PR #1119, opened 2014-12-29
Email templates for microsites

almathew:


default_settings.py
```diff
@@ -476,3 +476,77 @@
     'DIGITS': 1,
     'PUNCTUATION': 1
 }
+
+# email types
+ACTIVATION = 1
```
aspidites: Are we going to revisit this, or just keep the awkward gaps?

PR #1120, opened 2014-12-29
[PD-890] 404 when user isn't a member of current company on admin pages.

aspidites:
Decorator seemed like more work, since there'd be two function definitions just to get the same logic-- unless I'm missing a built-in decorator that gives me access to the user AND company.

It's also possible that I misunderstood what I should be guarding against.


aspidites:
This shouldn't have been closed

aspidites:
Tests are failing, but I'm reluctant to fix them until I know that my recent changes are what @almathew  was referring to. Especially since my original PR passed tests.

almathew:
Recent changes look ok, but hold off on this until Monday. I just came up with an edge case and so I want to ask @JLMcLaugh something.

aspidites:
@JLMcLaugh This is in "DO NOT MERGE" status because @almathew  wanted to ask you something about it.

aspidites:
So @almathew ... you're done with templates...right?

almathew:
Yep.

PR #1121, opened 2014-12-30
Updated registration email templates.

aspidites:
* Changes to helpers probably wasn't necessary. I was looking at it to find out what templates were used by the send_email function. I do think the change is clearer, though.
* for the password reset email, we could probably just pass in site as a context variable, but I did it this way so that it was explicit what I was after. Once we decide which context to pass in, we should be ok.


aspidites:
Will update when #1124 gets merged.

aspidites:
I'm done with this, just waiting on feedback.

almathew:
Merge conflicts. Not sure if you were aware of those.

aspidites:
Yeah. I've updated 3 times now. Every time something gets merged, I have to update. I also think I lost my changes in one of the merges.

PR #1127, opened 2015-01-02
Blocks, invitations, and forms [PD-910, 909, and 906]

almathew:


postajob/views.py
```diff
@@ -440,8 +440,10 @@ def get_context_data(self, **kwargs):
         kwargs['company'] = get_company(self.request)
         kwargs['prevent_delete'] = self.prevent_delete
         kwargs['on_admin_page'] = 'admin' in self.request.get_full_path()
-        # don't hide the company profile page
-        kwargs['on_admin_page'] = 'profile' not in self.request.get_full_path()
+        if kwargs['on_admin_page']:
```
almathew: @JLMcLaugh I'm not really sure how I feel about this section. I'd like to switch the admin part to at least be looking for `/posting/admin`, but at the same time we have a bit more control over what happens to the non-admin urls than we do where it's a problem on microsites.

I'm also wondering if we resolve the path or use the full string `profile`, since that's possible to mess up as well

seo/models.py
```diff
@@ -1292,7 +1292,7 @@ def save(self, *args, **kwargs):
         inviting_user = kwargs.pop('inviting_user', None)
         group = Group.objects.get(name=self.GROUP_NAME)
         self.user.groups.add(group)
-        if not self.pk:
+        if not self.pk and inviting_user:
```
almathew: @tdphillips what I did is a taking advantage of an odd situation, but I actually think the invitation emaills will turn out better for it. Do you have any objections to this?

almathew: @fezick I suppose I should ask you as well. Do you have objections to invitations only being sent if an inviting user can be determined? The invitee is added using the CompanyUser form, so that will always result in an invitation. This will prevent it from sending an invitation for postajob created CompanyUsers, though.

tdphillips: I have no objections. We assume a user exists anyway for this type of invitation.

One point of concern is ensuring that the invitee isn't also passed in as the inviting user.

myblocks/forms.py
```diff
@@ -73,4 +73,4 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super(RowForm, self).__init__(*args, **kwargs)
-        self.initial['template'] = models.raw_base_template(self.Meta.model)
```
aspidites: Did we move stuff to field sets? Otherwise, what motivated this change?

postajob/views.py
aspidites: I was wondering about that when I wrote it and went with the simplest thing that would work. If theres a more reliable way to check for admin pages in general than looking at paths, feel free to change it.

myblocks/forms.py
almathew: For some reason the old version was flat out overriding it no matter what. I haven't looked into it, as I know my fix works (and is used elsewhere), but if I had to guess I'd say that `self.initial` is what's loaded into the field no matter what wheres `self.fields[field].initial` is just initial data if some doesn't already exist.

aspidites: Works for me.

aspidites:
Code looks fine. Don't understand pd-906 well enough to merge.

aspidites:
Got things working locally and @almathew  explained pd-906 to me, which I've just confirmed is fixed. Unless there are objections, I'll be merging this once i confirm pd-910 and pd-909 are fixed as well.

aspidites:
PD-909 is a bit of a moving target, but this PR improves on the situation without introducing more issues, so I'm merging it. In a follow-up PR, i'll be reverting my changes until @fezick, @almathew ,andI can hash things out on Monday

PR #1129, opened 2015-01-02
Fixed time format.

aspidites:
Without this, all PR branches will fail.

### DIFF
--- a/mypartners/tests/test_views.py
+++ b/mypartners/tests/test_views.py
@@ -313,7 +313,9 @@ class PartnerOverviewTests(MyPartnersTestCase):
         for row in container('div', class_="product-card"):
             title = "Test Subject  - example-contact"
             today = date.today()
-            sub_title = today.strftime('%b.%e, %Y')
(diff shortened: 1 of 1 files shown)


almathew:
Not sure - won't this cause things to fail if the date is a two digit number?

aspidites:
What did you do differently, @almathew , since you fixed this in your branch?

almathew:
Probably not the best way to fix it, but:
https://github.com/DirectEmployers/MyJobs/pull/1127/files#diff-e337d3bf994a8853913c57a312419debR314

aspidites:
@almathew  and you're right. No go with 2 digits days.

aspidites:
Ah, I see what you did. Hmm.. that would work. Feels redundant. I'll use it for now though.

aspidites:
Investigated further, and unless we start using padded integers, manually constructing the date string as you did is the proper solution.

aspidites:
@almathew I prefer the format variety, but copied yours verbatim as not to introduce merge conflicts with your branch.

PR #1130, opened 2015-01-02
Invitation user experience [PD-847]

tdphillips:
invitation

All tests pass.

### DIFF
--- a/registration/tests/test_models.py
+++ b/registration/tests/test_models.py
@@ -335,13 +335,19 @@ class InvitationModelTests(MyJobsBase):
         self.assertTrue(self.admin.email in email.body)
         self.assertTrue(company.name in email.body)
 
-        return BeautifulSoup(email.body)
+        body = BeautifulSoup(email.body)
+
+        self.assertEqual(body.select('a')[0].attrs['href'],
+                         'https://secure.my.jobs' + reverse('home'))
+
+        return body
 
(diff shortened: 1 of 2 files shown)


aspidites:
Do we have to worry about invitations on microsites?

tdphillips:
What about them, specifically?

aspidites:
I was originally curious as to whether Company Name should appear in the signature some where, but perhaps that's redundant. 

Would highlighting the company name be an improvement or detraction? Something like this:
http://jsfiddle.net/yz4c16fa/

Also, does it make sense to use the inviting company rather than the inviting user? ("My.jobs on behalf of Company Name")

tdphillips:
Would that be a functional anchor or just a highlight?

Regarding inviting company vs. inviting user, this is the copy that I was provided with.

aspidites:
Functional if we have access to the company url, otherwise just a highlight (though in that case, I'm sure there are better ways to do so).

fezick:
The vast majority of invites will be for member tools on my.jobs. Lets deal with mircosite specific functionality when we get to it.

fezick:
The email looks good to me.

tdphillips:
@fezick I should mention that this doesn't contain the proposed subject change  that AJ asked for. I wasn't sure how desired that was.

fezick:
Right now it says invitation, which is generic enough for me.

PR #1131, opened 2015-01-02
Fix broken product form

aspidites:
Without this, the product form product count is broken.


almathew:
Does this have a ticket? Or was it just something you/someone else came across?

aspidites:
No ticket. It was a 2 minute fix while working on something else. I can make one though.

PR #1135, opened 2015-01-05
Invoice fix

almathew:


postajob/views.py
```diff
@@ -51,13 +51,15 @@ def jobs_overview(request):
 @company_has_access(None)
 def view_job(request, purchased_product, pk, admin):
     company = get_company_or_404(request)
-    product = PurchasedProduct.objects.get(pk=purchased_product)
-    if not product.owner == company:
+    purchased_product = PurchasedProduct.objects.get(pk=purchased_product)
+    if admin and not purchased_product.product.owner == company:
```
tdphillips: Would xor be too obscure?

`if bool(admin) ^ purchased_product.product.owner == company`

aspidites: Probably. Isn't `bool()` implied by `if`, or does the XOR operator mandate its use?

almathew: I made it a 2 line if statement specifically for readability already...

tdphillips: If admin were an instance of some model like I was originally thinking, `bool()` would be necessary. I think it's actually a bool already, however, so it's unnecessary.

PR #1136, opened 2015-01-05
Updated registration email templates.

aspidites:
Fixed the fact that my templates were being overridden.

Password Reset:
2015-01-05-111209_1465x304_scrot

Activation:
2015-01-05-111428_1432x239_scrot

### DIFF
--- a/registration/templatetags/password_reset_tags.py
+++ b/registration/templatetags/password_reset_tags.py
@@ -1,17 +1,37 @@
 from django.template import Library
 from django.conf import settings
 
+from seo.models import SeoSite
+
 register = Library()
 
 
 @register.simple_tag
-def get_current_seosite_domain():
+def get_current_seosite(attr=None, str_func=None):
     """
-    Gets the domain of the current SeoSite.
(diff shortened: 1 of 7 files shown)


aspidites:
For the record, this is what happens if I try to use underscores as an hr (since django password reset emails are plain text:
2015-01-05-111734_1482x349_scrot

registration/templatetags/password_reset_tags.py
```diff
@@ -1,17 +1,37 @@
 from django.template import Library
 from django.conf import settings
 
+from seo.models import SeoSite
+
 register = Library()
 
 
 @register.simple_tag
-def get_current_seosite_domain():
```
tdphillips: /templates/postajob/request_email.html still uses this name.

aspidites: Fixed

aspidites:
With dashes:
2015-01-05-130129_1447x340_scrot

@fezick ?

fezick:
:+1:

PR #1137, opened 2015-01-05
Prevent the deletion of already redeemed offline purchases via form [PD-911]

almathew:
All tests pass

### DIFF
--- a/postajob/tests/test_views.py
+++ b/postajob/tests/test_views.py
@@ -300,6 +300,35 @@ class ViewTests(PostajobTestBase):
         self.assertEqual(response.status_code, 404)
         self.assertEqual(PurchasedJob.objects.all().count(), 1)
 
+    def test_unexpire_expired_purchased_job(self):
+        product = PurchasedProductFactory(
+            product=self.product, owner=self.company)
+        job = PurchasedJobFactory(
+            owner=self.company, created_by=self.user,
+            purchased_product=product)
+        kwargs = {'pk': job.pk}
+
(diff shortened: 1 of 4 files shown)


postajob/views.py
```diff
@@ -767,6 +767,11 @@ class OfflinePurchaseFormView(PostajobModelFormMixin, RequestFormViewBase):
     update_name = 'offlinepurchase_update'
     delete_name = 'offlinepurchase_delete'
 
+    def delete(self):
```
tdphillips: Does this need to be decorated as well?

postajob/tests/test_views.py
```diff
@@ -300,6 +300,35 @@ def test_purchasedjob_delete(self):
...
+        data = dict(self.purchasedjob_form_data)
+        data['is_expired'] = False
+
+        response = self.client.post(reverse('purchasedjob_update',
+                                            kwargs=kwargs),
+                                    data=self.purchasedjob_form_data,
+                                    follow=True)
+        self.assertEqual(response.status_code, 200)
+
+        # Confirm that the job remains expired despite the attempt to
```
aspidites: So are we not expecting a user to ever check the expired box and uncheck it later?

almathew: This only happens if the job is past the max_expiration_date. Hence the changes to job on 313-316

aspidites: Ahhh, I misread the comment. You're enforcing the checkbox irrespective of the actual date. Derp. :+1:

PR #1138, opened 2015-01-05
Prevent creation of contact record and instead send error email when user is company user for multiple companies [PD-848]

almathew:
All tests pass

### DIFF
--- a/mypartners/views.py
+++ b/mypartners/views.py
@@ -1195,7 +1195,7 @@ def process_email(request):
         error = "Your account is setup as the admin for multiple companies. " \
                 "Because of this we cannot match this email with a " \
                 "specific partner on a specific company with 100% certainty. " \
-                "You will need to login to My.jobs and got to " \
+                "You will need to login to My.jobs and go to " \
                 "https://secure.my.jobs/prm to create your record manually."
         send_contact_record_email_response([], [], [], contact_emails,
                                              error, admin_email)
--- a/mypartners/views.py
+++ b/mypartners/views.py
@@ -1191,6 +1191,16 @@ def process_email(request):
     admin_user = User.objects.get_email_owner(admin_email)
     if admin_user is None:
         return HttpResponse(status=200)
+    if admin_user.company_set.count() > 1:
+        error = "Your account is setup as the admin for multiple companies. " \
+                "Because of this we cannot match this email with a " \
+                "specific partner on a specific company with 100% certainty. " \
+                "You will need to login to My.jobs and got to " \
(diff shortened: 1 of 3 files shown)


mypartners/views.py
```diff
@@ -1191,6 +1191,16 @@ def process_email(request):
     admin_user = User.objects.get_email_owner(admin_email)
     if admin_user is None:
         return HttpResponse(status=200)
+    if admin_user.company_set.count() > 1:
+        error = "Your account is setup as the admin for multiple companies. " \
+                "Because of this we cannot match this email with a " \
+                "specific partner on a specific company with 100% certainty. " \
+                "You will need to login to My.jobs and got to " \
```
tdphillips: go to

almathew: fixed

mypartners/views.py
```diff
@@ -1191,6 +1191,16 @@ def process_email(request):
     admin_user = User.objects.get_email_owner(admin_email)
     if admin_user is None:
         return HttpResponse(status=200)
+    if admin_user.company_set.count() > 1:
```
aspidites: For some reason, I thought during the meeting we discussed noting which companies a user belonged to as part of the response. Is that no longer (or was that never) the case?

aspidites: Also, the `\`s. PEP8. Parens would require less work should you decide to rewrap the lines, as well:

```python
 error = ("Your account is setup as the admin for multiple companies. " 
	      "Because of this we cannot match this email with a "
	      "specific partner on a specific company with 100% certainty. "
	      "You will need to login to My.jobs and go to "
	      "https://secure.my.jobs/prm to create your record manually.")
```

almathew: Perhaps? I took the wording directly from the ticket.

templates/mypartners/email/email_response.html
```diff
@@ -32,6 +32,7 @@
         {% endfor %}
     {% endif %}
 
+Please note: this inbox is not monitored and replies will not receive a response.
```
aspidites: :+1:

mypartners/views.py
aspidites: Ah. According to the ticket then, you're missing an anchor (as opposed to just a URL) for the secure link, and perhaps a signature (-My.jobs).

almathew: I left it as a link for all the clients that remove links. I have no idea how prevalent of an issue that is, but for how plain-text it is most browsers will auto-href it anyway.

almathew: Also, I added the signature to the footer with the do not reply text.

PR #1144, opened 2015-01-06
PRM Saved Search - force sort order to date (newest first)

galitskyd:
Creating a Partner Saved Search will now have sort_by = Date
Added a data migration to easily update our db on production. (sets all partner saved searches sort_by = Date)

### DIFF
--- a/mysearches/migrations/0003_partner_saved_search_default_sort_by_to_date.py
+++ b/mysearches/migrations/0003_partner_saved_search_default_sort_by_to_date.py
@@ -1,0 +1,414 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+    # This is only a data migration (no real migrations)
+    # so need to set no_dry_run to True
+    no_dry_run = True
+
+    def forwards(self, orm):
+        # Default sort_by set to 'Date' for Partner Saved Searches, see PD-912
+        for savedsearch in orm.SavedSearch.objects.filter(partnersavedsearch=True):
+            savedsearch.sort_by = 'Date'
+            savedsearch.save()
+
+    def backwards(self, orm):
+        pass
+
+    models = {
+        u'auth.group': {
+            'Meta': {'object_name': 'Group'},
+            u'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'name': ('django.db.models.fields.CharField', [], {'unique': 'True', 'max_length': '80'}),
+            'permissions': ('django.db.models.fields.related.ManyToManyField', [], {'to': u"orm['auth.Permission']", 'symmetrical': 'False', 'blank': 'True'})
+        },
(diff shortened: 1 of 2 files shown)


mysearches/migrations/0003_partner_saved_search_default_sort_by_to_date.py
```diff
@@ -0,0 +1,414 @@
...
+
+
+class Migration(SchemaMigration):
+    # This is only a data migration (no real migrations)
+    # so need to set no_dry_run to True
+    no_dry_run = True
+
+    def forwards(self, orm):
+        # Default sort_by set to 'Date' for Partner Saved Searches, see PD-912
+        for savedsearch in orm.SavedSearch.objects.filter(partnersavedsearch=True):
```
tdphillips: Does this work? I thought it would be something like `partnersavedsearch__isnull=False`

Also, you could probably update the queryset rather than iterating over it.

aspidites: `orm.SavedSearch.objects.filter(partnersavedsearch__isnull=False).update(sort_by='Date')` IIRC.

mysearches/migrations/0003_partner_saved_search_default_sort_by_to_date.py
```diff
@@ -0,0 +1,414 @@
...
+            savedsearch.sort_by = 'Date'
+            savedsearch.save()
+
+    def backwards(self, orm):
```
aspidites: Not that we should be backward migrating, but should we set it back to relevance here?

tdphillips: I haven't checked the data, but it's possible it wasn't relevance if the recipient changed it.

galitskyd: What @tdphillips said.

mysearches/migrations/0003_partner_saved_search_default_sort_by_to_date.py
almathew: `orm.PartnerSavedSearch.objects.update(sort_by='Date')`?

mysearches/migrations/0003_partner_saved_search_default_sort_by_to_date.py
```diff
@@ -0,0 +1,409 @@
...
+from django.db import models
+
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        # Default sort_by set to 'Date' for Partner Saved Searches, see PD-912
+        orm.SavedSearch.objects.filter(partnersavedsearch__isnull=False).update(sort_by='Date')
+
+    def backwards(self, orm):
```
aspidites: Again, are we not worried about setting this back to relevance in case of a backward migration? @JLMcLaugh?

tdphillips: Out of 1602 partner saved searches, zero of them are sort_by=Date. I retract my previous comment.

PR #1145, opened 2015-01-07
[PD-890] Change permissions on admin pages for postajob

aspidites:
## Description
Specifically, this ensures that a user can only access the admin page for the current site they are visiting *if* that site belongs to a site package owned by a company the logged in user is a part of.'

## Remaining Tasks
- [x] Finish wrapping appropriate views
- [x] Add comments
- [x] Write tests

## Notes for Reviewers
* In implementing this, the query I used is as obtuse as the description I gave above. It was suggested that I use `settings.SITE.canonical_company`, but I'm weary of doing so because of the fact that it could very well not be set, where as the reverse relationships seem a safer bet. I'm willing to use `canonical_company` so long as I have access to a future-dated "I told you so". 

* Does anyone else have a better name for this decorator? I wanted something long enough to be descriptive but short enough to avoid being annoying to type, and I think I've failed in both respects. 
* Are there any views which I forgot to wrap, or which shouldn't have been wrapped to begin with? In particular, I'm curious about the form views, `view_job`, `view_invoice`, and `view_request`, `resend_invoice`.


universal/decorators.py
```diff
@@ -40,6 +41,27 @@ def wrap(request, *args, **kwargs):
...
+    That is, if John is visiting testing.jobs, which is in a site package owned
+    by DirectEmployers, but John isn't a company user for DirectEmployers, he
+    will see a 404 page.
+    """
+
+    @wraps(view_func)
+    def wrap(request, *args, **kwargs):
+        import ipdb; ipdb.set_trace()
+        if not settings.SITE.sitepackage_set.filter(
+                owner__in=request.user.company_set.all()).exists():
```
JLMcLaugh: Query logic like this should be in the models if possible. postajob BaseModel has a user_has_access method. You could build a model manager, and use a similar name. @almathew or @galitskyd should weigh in on whether you've caught all the views we need. This is probably a good example of why we've started using class-based views.

universal/decorators.py
```diff
@@ -40,6 +41,27 @@ def wrap(request, *args, **kwargs):
...
+    current seo site.
+
+    That is, if John is visiting testing.jobs, which is in a site package owned
+    by DirectEmployers, but John isn't a company user for DirectEmployers, he
+    will see a 404 page.
+    """
+
+    @wraps(view_func)
+    def wrap(request, *args, **kwargs):
+        import ipdb; ipdb.set_trace()
```
JLMcLaugh: ipdb

universal/decorators.py
almathew: The decorator needs moved to urls at the very least, as we `view_request` is both an admin and non-admin view.

aspidites:
Moved the query and changed where things were being decorated at.

aspidites:
Might be an issue with the `user.can_access_site` logic.

aspidites:
Was getting unexpected 404's, so tried relying on canonical_company, which wasn't reliable. Turns out it was error in the logic of a different branch. This one should be fine.

tdphillips:
:+1:

Anyone else?

PR #1152, opened 2015-01-09
Log saved search sends/failures [PD-819]

tdphillips:
Also adds categories to all emails.


tdphillips:
Adding a category to password reset links caused them to be interpreted as html instead of plain text. Here's my fix:
2015-01-09-155148_705x201_scrot

Still not perfect. I'm open to suggestions.

All tests pass.

tdphillips:
Considering each different category is basically a special snowflake as far as formatting is concerned, I don't think it would make anything cleaner if I were to move them to settings as I had mentioned.

Removing the DO_NOT_MERGE label. If anyone has strong feelings on the matter, I'll revisit it.

tdphillips:
The most recent commit logs the reason that @almathew asked for in the ticket.

All tests pass.

JLMcLaugh:
Failed test.

tdphillips:
Tests pass.

PR #1153, opened 2015-01-09
[PD-920] Disable opted-out emails on saved search form

aspidites:
This disables email addresses in the saved search form if the user who owns it is not opted in. For the disabled ones, a tool tip is displayed:

2015-01-09-153815_3840x1080_scrot


aspidites:
Rewriting this.

PR #1154, opened 2015-01-12
[PD-920] Disable opted-out emails on saved search form

aspidites:
## Changes
- A partner saved search that hasn't been sent says "Never" for "Last sent" instead of "None"
- opted out emails are disabled with help text provided:
2015-01-12-091137_827x385_scrot

- Unlike the last PR I submitted with the same title, you can save the form.


aspidites:
**note:** "communication" is now "communications". Didn't feel like taking another screenshot, since I had to crop the last one.

PR #1157, opened 2015-01-13
[PD-788] Opt-out notifications now display properly

aspidites:
**NOTE: ~~I'm holding off writing unit tests for this until I take care of *some* of the circular imports that we have within our model modules~~ I *think* I've taken care of circular imports for the affected apps. It's on a separate branch (which I'll create a PR for once I've finished the other apps)**

## Remaining Tasks
- [ ] unit tests

No visual changes. Basically makes it so that this feature works as intended. If you are logged in as a user and add another user to a partner saved search, should that user opt out of emails, the original user you logged in as should receive an email and in-app notification that the other user has opted out.


myjobs/models.py
```diff
@@ -536,7 +537,11 @@ def send_opt_out_notification(self):
         Notify saved search creators that a user has opted out of their emails.
         """
         subject = "My.jobs Partner Saved Search Update"
-        saved_searches = self.partnersavedsearch_set.distinct()
+
+        # partnersavedsearch_set doesn't account for regular saved searches
+        # with a pss field
```
tdphillips: What do you mean pss field?

galitskyd: I think he means savedsearch.partnersavedsearch

aspidites: What @galitskyd said. I'll rewrite.

tdphillips: There's not actually a list for a user's parter nsaved searches (yay for not using related_name). It looks like partnersavedsearch_set is the set of saved searches that you created, not the ones created for you.

almathew:
Tests?

PR #1158, opened 2015-01-14
Fix sendgrid header [PD-931]

tdphillips:
I don't know how I missed this.

Tests pass.


tdphillips:
The latest commit also makes the admin usable.

aspidites:
Checked locally and compared against master. :+1:

PR #1159, opened 2015-01-14
Multiple standard facets

almathew:


aspidites:
tests :-P

seo/breadbox.py
```diff
@@ -0,0 +1,260 @@
+from django.core.urlresolvers import reverse, reverse_lazy
+from django.conf import settings
+from django.template.loader import render_to_string
+
+from seo import helpers
+from seo.models import CustomFacet
+
+
+class Breadbox(object):
```
tdphillips: :-1:

almathew: Blame @JLMcLaugh

almathew:
So far as I can tell, everything that is not related to tests is complete. Testing may uncover some new things, but you can start assuming it should be completely working.

seo/breadbox.py
```diff
@@ -0,0 +1,260 @@
...
+    class Breadcrumb(object):
```
aspidites: Why the inner-class? One function which is __init__...

almathew: prettier than a named tuple

aspidites: I disagree with both of those points (more details tomorrow), but I was asking because it seems wrong to have the inner class all together.

aspidites: ### Beauty
1. Given the breadcrumb's attributes would accessed in the same way (eg. `breadcrumb.url`), I fail to see how one could be prettier than the other. 
2. If you are referring to how they are declared, I suspect that is a subjective opinion, as I prefer the lightweight look of a namedtuple over a nearly worthless class with a single `__init__` method:

```python

# class
class Breadcrumb(object):
    def __init__(self, display_title='', url_reverse_lazy('view_jobs'),
                        breadcrumb_class='breadcrumb_bread_box'):
        self.url = url
        self.display_title = display_title
        self.breadcrumb_class = breadcrumb_class

# namedtuple
Breadcrumb = namedtuple('Breadcrumb', 'url display_title breadcrumb_class')

# the largest omission seems to be that of default values, for which I'd suggest what in Haskell might be referred to as smart constructors:
DEFAULT_BREADCRUMB = Breadcrumb('', reverse_lazy('view_jobs'), 'breadcrumb_bread_box')
make_breadcrumb = lambda **kwargs: DEFAULT_BREADCRUMB._replace(**kwargs)
```

### Mutability
1. I don't see where any instance of breadcrumb is being mutated. Cold you show me where mutability is vital? The closest you seem to get is replacing Breadbox's individual breadcrumb instances entirely, which wouldn't pose a problem for namedtuples. 
2. While namedtuples are immutable, there is always the `_replace` method. As mutations are infrequent at best (though I haven't seen an instance of mutation at all), I doubt `breadcrumb._replace(url="test.jobs")` will be very burdensome compard with `breadcrumb.url = "test.jobs"`, especially since I suspect that mutation would happen within the context of `Breadbox`'s interface.

JLMcLaugh: I'm not going to argue one way or the other about aesthetics unless the class was a mess of getters and setters. Mutability is probably the main point. If we're counting on using _replace, then a class would have been the right answer (it would be breadcrumb = breadcrumb._replace(url="test.jobs"), and tuples imply an intended immutability anyways. If you want to hint that the logic for deciding the breadcrumb values belongs in Breadbox, and Breadcrumb is just a container for returning results, namedtuple works. If you can see the values getting edited by different functions, template tags, or views, you'd want a class. Just be careful we don't end up with the same mess that led to this class getting created. If more logic might end up in Breadcrumb itself (something based on breadcrumb_class probably), a class is the way to go.

aspidites: "If more logic may" - I'll just restate my YAGNI argument here... It isn't
that hard to promote a data type.

almathew:
Old tests are now passing. I'm on to writing new ones.

almathew:
Tests for the new class are done.

seo/breadbox.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def _remove_param_from_query_dict(self, param):
+        query_dict = self.query_dict.copy()
+        if param in query_dict:
+            del query_dict[param]
+        return query_dict
+
+    def all_breadcrumbs(self):
+        breadcrumbs = (self.custom_facet_breadcrumbs +
+                       self.location_breadcrumbs + self.moc_breadcrumbs)
```
JLMcLaugh: Why is moc not a single_breadcrumb?

almathew:
Added comments where necessary.

PR #1160, opened 2015-01-14
Add location dropdowns to company profile [PD-891]

tdphillips:
The modified postajob javascript should have no ill effects on company profile. I'm not super pleased with the id prefixes, but I would like having to duplicate a good portion of javascript even less.

This also completes the removal of accordions on job forms.

All tests pass.


aspidites:
Default country is Afghanistan instead of the United States

tdphillips:
When you're looking at a company profile, you're editing one. I guess I can inspect the current data and, if blank, set it to United States.

templates/postajob/myjobs/job_form.html
```diff
@@ -113,8 +113,6 @@
         var form_count = {{ formset.total_form_count }};
         var form = '{{ formset.empty_form|escapejs }}';
         $(document).ready(function() {
-            accordion('.header', '.contents');
```
aspidites: Ugh. Thought I got rid of these forever ago. Thanks

aspidites:
Did `templates/dseo/purchasedproduct.html` not have to be updated?

tdphillips:
That and the template asked about in irc don't reference javascript, except through the form's media.

tdphillips:
There are a grand total of zero company profiles in production. That being the case, I went with setting a default on the model rather than trying to detect that a company profile already exists with an empty company. If you want to test locally, you'll need to delete your profile.

All tests pass.

aspidites:
@tdphillips you'll probably want to delete company profiles on QC when this gets merged so that it doesn't get rejected. That, or make a note of it in the ticket so testers know?

tdphillips:
Whatever is easiest for @fezick. I can delete them all (they'll get recreated when the company profile is visited), I can update them all to have country=United States, or we can let them all show Afghanistan.

aspidites:
Tested locally. :+1: , after surmounting my own stupidity.

tdphillips:
Javascript issue. The line of code in my most recent commit did not cause any issues for job locations because they were never edited, only created/removed.

aspidites:
After a `git pull` and `CompanyProfile.objects.all().delete()`, I'm still seeing Alabama when I try and edit a Company Profile.

tdphillips:
Let's see if this works.

PR #1161, opened 2015-01-14
topbar - reload when a user changes companies [PD-935]

galitskyd:
Added redirect/reload when changing company in topbar.

Removed some things that weren't needed now that the page will reload.

### DIFF
--- a/templates/includes/topbar.html
+++ b/templates/includes/topbar.html
@@ -273,11 +273,6 @@
         });
 
         {% if group_member and company_name|length > 1 %}
-        // See if a cookie is already set. If so update employer links.
-        var cid = read_cookie();
-        if(cid)
-            update_employer_tools(cid);
-
         // Build multiple company selector
         get_companies();
     });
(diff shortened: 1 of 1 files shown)


aspidites:
:+1:  on initial inspection.

templates/includes/topbar.html
```diff
@@ -425,19 +418,42 @@ <h3 id="topbar-logo">My.jobs</h3>
...
-        partner_link.search = "company=" + company_id;
+    function process_reload() {
+        /*
+        Looks at what path the user is currently in.  If the path is showing
+        an app that changing company matters redirect the user to a desired
+        location instead of causing a 404.
+         */
+        var path = window.location.pathname,
+            // Apps that changing company IDs would be a problem.
+            app_names = ["candidates", "prm", "posting"],
```
fezick: Are there other places where this list might be used? Should we consider making this an entry in the settings file and passing it as a context variable? This seems like it would be easy to miss if the list ever changes.

almathew: This would best be classified as "pages that would change if the company were to change". Do we foresee any other way of the company changing in the future? Because I can't come up with any other reason we would care about these pages.

Putting it in the settings file isn't that big of a deal and doesn't seem like a horrible idea. The worst point about that is it puts us one setting closer to the point in which we have a settings file so big that we don't bother to look there for that information either.

PR #1163, opened 2015-01-14
Fix for task update solr

mklauber:
If no jobs from the buid exist in solr, that trying to index from SEOXML will fail.


aspidites:
I know tests are still running, but is this waiting on any particular reviewers? Change is small enough that I'd suspect not, but considering what is being touched, I though I'd ask.

JLMcLaugh:
It looks fine to me, just waiting for tests to finish.

PR #1166, opened 2015-01-16
[PD-930, PD-946] Updated disable saved search and saved search overview pages

aspidites:
## Saved Search Overview
I added information about the partner who created the saved search since those can't be deleted by the use. I also linked to the partner page.
Screenshot:
2015-01-16-105115_1052x1056_scrot

## Disable Saved Search
Before, when clicking the delete all saved searches link in the confirmation email, you'd still see a delete all link, which would delete partner saved searches as well. I originally hid this, but realized it would prevent users from deleting all saved searches. After first trying to separate the content on this page by saved search type as suggested, I went with a delete modal instead, a la partner library. I didn't bother listing the partner saved searches that wouldn't be deleted explicitly since clicking delete redirects to the saved search overview anyway.

In short, disabling a single partner saved search hides the delete button (that's already in master), but deleting all searches reenables that button, which now has modal behavior (in all cases, not just this one).

Screenshot:
2015-01-16-105450_1052x1056_scrot

## Remaining Tasks
- [ ] Unit tests

### DIFF
--- a/templates/mysearches/saved_search_disable.html
+++ b/templates/mysearches/saved_search_disable.html
@@ -56,7 +56,7 @@
                                     {% plural %}
                                     Delete them instead
                                     {% endblocktrans %}
-                                </button>
+                                </a>
                             </td>
                         </tr>
                     {% endif %}
--- a/templates/mysearches/saved_search_disable.html
+++ b/templates/mysearches/saved_search_disable.html
@@ -50,13 +50,13 @@
                     {% if searches|length and not contains_pss %}
                         <tr class="profile-section">
                             <td>
-                                <a href="{%url 'delete_saved_search'%}?id={{search_id}}&verify={{email_user.user_guid}}">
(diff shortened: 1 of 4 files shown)


templates/mysearches/saved_search_disable.html
```diff
@@ -50,13 +50,13 @@
                     {% if searches|length and not contains_pss %}
                         <tr class="profile-section">
                             <td>
-                                <a href="{%url 'delete_saved_search'%}?id={{search_id}}&verify={{email_user.user_guid}}">
+                                <a id="delete" type="button" data-toggle="modal" data-target="#delete-confirm">
```
tdphillips: Should this be a button instead of an anchor?

aspidites: No.  A button has the wrong styling.

tdphillips: In that case, line 59 should probably be reverted.

aspidites: Fixed

PR #1168, opened 2015-01-16
Ajax jobs by location [PD-924]

almathew:
Stacked on top of https://github.com/DirectEmployers/MyJobs/pull/1159 since I was working on some of the same areas. Commits specific to these tickets are 6421204 through 85355d5.

All tests pass.


tdphillips:
For anyone as lazy as me:
https://github.com/DirectEmployers/MyJobs/compare/multiple-std-facets...ajax-jobs-by-location

seo/views/search_views.py
```diff
@@ -74,6 +75,68 @@
...
+                                    facet_limit=num_items,
+                                    sort_order=sort_order)
+    featured_jobs = helpers.get_featured_jobs(default_sqs=sqs,
+                                              jsids=settings.SITE_BUIDS,
+                                              filters=filters,
+                                              facet_limit=num_items,
+                                              sort_order=sort_order)
+
+    facet_counts = default_jobs.add_facet_count(featured_jobs).get('fields')
+    facet_counts = facet_counts['lat_long_%s_slab' % facet_field_type]
```
tdphillips: lat_long_buid_slab is the only field that looks like this. Is that going to change?

almathew: It was probably stupid of me to prematurely set up this part knowing that I'll have to add the fields later anyway, but I do believe it's the plan to add more later.

seo/views/search_views.py
```diff
@@ -74,6 +75,68 @@
 LOG = logging.getLogger('views')
 
 
+def ajax_geolocation_facet(request):
```
aspidites: Do we not need to guard this request to ensure it is only called via ajax (`if request.is_ajax()`)?

almathew: This isn't really protected data, and it's behind a Disallow for bots. It doesn't gain much by being behind one.

aspidites: We gain the comfort of knowing that we won't get 500 errors in production:
2015-01-20-094034_1916x1056_scrot

almathew: Your error is in `home_page()`. While worthwhile to not have 500 errors, it's not related to the ajax-ness of this page.

aspidites: Accessing a view expecting an AJAX response will result in a 500 error
though -we've experienced this already with mymessages a few months ago.

aspidites:
My error was caused by the wrong version of solr running.

seo/views/search_views.py
almathew: The mymessages error we were having was because something didn't return a response if the request wasn't ajax. This returns a response regardless of ajax status.

PR #1174, opened 2015-01-21
[PD-960] Honor date filters on export

aspidites:
* Makes sure that the current date filter is honored when exporting to various formats in contact records.


static/record_overview.160-21.js
```diff
@@ -406,6 +407,26 @@ function update_search_url(data) {
...
+/*
+Updates download links so that date filters match what is being displayed
+*/
+function update_download_links(data){
+    var csv_link = $(".records-csv-export-link"),
+        xml_link = $(".records-xml-export-link"),
+        // printer  friendly
+        html_link = $(".records-printer-friendly-export-link");
+
+    csv_link.attr('href', csv_link.attr('href').split('?')[0]
```
JLMcLaugh: Is this dropping every other query parameter?

tdphillips: Isn't location.search everything after (and including) the "?" anyway?

tdphillips:
Might want to hold off deploying. I had some test failures locally.

PR #1177, opened 2015-01-21
Fix

almathew:


seo/helpers.py
```diff
@@ -361,14 +361,14 @@ def bread_box_company_heading(company_slug_value):
     if not company_slug_value:
         return None
 
+    kwargs = {'title_slug': company_slug_value}
+    business_unit = BusinessUnit.objects.filter(**kwargs)
```
tdphillips: If you don't want to try/except, appending a .first() will return a BusinessUnit or None.

PR #1182, opened 2015-01-21
Myreports Init [PD-961] (maybe [PD-962])

galitskyd:
New app called Myreports

New url path /reports/

only staff can see links and not get 404's from hitting views


tdphillips:
Should this be merged or is it effectively a feature branch that everything will be built on top of?

aspidites:
@tdphillips dunno. My work plays on top of this (so probably needs to be merged first if nothing else). It also includes a bug fix, so perhaps it should be merged?

templates/includes/topbar.html
```diff
@@ -108,6 +111,9 @@ <h3 id="topbar-logo">My.jobs</h3>
                                             <li><a id="partner-link" class="no-show sub-nav-item employer-menu" href="{{ ABSOLUTE_URL }}prm/view?company={{ company.id }}">{% trans "PRM" %}</a></li>
 
                                         {% endif %}
+                                        {% if user.is_staff %}
+                                            <li><a id="reports-link" class="no-show sub-nav-item employer-menu" href="{{ ABSOLUTE_URL }}reports/view">{% trans "Reports" %}</a></li>
```
tdphillips: I thought one of the tickets said reporting will only be on secure. Is that not the case?

aspidites: That's correct. I don't think I understand why you're asking, though. Top bar only shows up on secure, and the link should only show up for staff members.

almathew: Topbar shows on network sites, but it will link back to secure so it's pretty much just like PRM in that regard.

tdphillips: This template is also rendered on certain microsites. Check python.jobs for instance.

tdphillips: Never mind. There's an ajax request for the topbar, meaning ABSOLUTE_URL should always be secure.

PR #1186, opened 2015-01-22
[PD-985] Made sure logged out users are still being redirected to login.

aspidites:
Makes sure that anonymous users are redirected to login with a next url instead of seeing a warning.


aspidites:
While working on reports, I found a better solution to this.

PR #1189, opened 2015-01-23
Models and admin for 2 additional custom facet groups [PD-846]

almathew:
I'm going to make everyone read really small pull requests instead of one big one this time.

This is the admin + models for the new custom facet groups.

All tests pass.


tdphillips:
TL;DR.

PR #1192, opened 2015-01-23
Improved logging of received SNS messages and sent tasks.

mklauber:


seo/views/search_views.py
```diff
@@ -1866,12 +1869,7 @@ def confirm_load_jobs_from_etl(response):
                      'ad875783-c49e-49ff-b82a-b0538026e089',
                      '0ab41358-8323-4863-9f19-fdb344a75a35',)
 
-    LOG.info("sns received for ETL", extra={
-        'view': 'send_sns_confirm',
-        'data': {
-            'json message': response
-        }
-    })
```
mklauber: Removed all this because it wasn't being used to actually update the logs.  Sometimes simpler is better.

aspidites:
Looks good. @mklauber  have you looked for any places where it might be beneficial to log a warning or an error? Cursory glace at where we catch exceptions seems to indicate a bunch of false positives (we're just asking for forgiveness, not handling legit exceptions), but I wouldn't know since I only got to like 400 or so before I stopped looking out of frustration.

PR #1194, opened 2015-01-23
Filters to individual functions

almathew:
Probably broken down a bit further than I absolutely had to go, but in general it's better for inheritance than it was. Mostly just massive amounts of moving code.

All tests pass.

### DIFF
--- a/seo/filters.py
+++ b/seo/filters.py
@@ -123,7 +123,7 @@ class FacetListWidget(object):
         if not self._show_widget(self.items):
             return
 
-        output = [self._render_list()]
+        output = [self._as_ul()]
 
         if self._has_hidden_items or self._show_more(self.items):
             more_less = self._render_more_less()
@@ -131,7 +131,7 @@ class FacetListWidget(object):
 
         return mark_safe('\n'.join(output))
 
-    def _render_list_item(self, item):
+    def _render_li(self, item):
         """
         Turns a facet tuple into an li containing the correct link
         to the item.
@@ -193,7 +193,7 @@ class FacetListWidget(object):
 
         return href
 
-    def _render_list_items(self):
+    def _render_lis(self):
         """
         Renders all the items in self.items.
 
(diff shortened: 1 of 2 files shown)


almathew:
I thought there was a problem with my code, but as it turns out some of the jobs (on staging) actually just have `</br>` in the location. Who knew?

seo/filters.py
```diff
@@ -47,25 +47,35 @@ def join_paths_of_same_type(item_type, path1, path2):
...
+        self.selector_type = self.widget_type.replace('_', '')
+
+        self.items = filter(lambda x: x[0], items)
         self.path_dict = self.filters_to_paths(filters.copy())
-        self.num_items = self.site_config.num_filter_items_to_show
-        self.offset = offset or self.num_items * 2
+        self.num_to_show = self.site_config.num_filter_items_to_show
+        self.offset = offset or self.num_to_show * 2
 
         if hasattr(request, 'META') and not query_string:
```
aspidites: don't request objects always have a META attribute? If so, the only way this would fail when there isn't a query_string is if the wrong type passed to the constructor, which I would deem a programmer error.

seo/filters.py
```diff
@@ -78,107 +88,158 @@ def filters_to_paths(filters):
...
+        :return: A string containing the facets in self.items rendered as a ul.
+
+        """
+        self._has_hidden_items = False
+        self._num_items_rendered = 1
+
+        if not self._show_widget(self.items):
+            return
+
+        output = [self._render_list()]
```
aspidites: `_render_list()` is named as though it should produce a list. Are we wanting a nested list? If it isn't returning a list, why is list in the name?

almathew: render list as in `<ul>`

aspidites: Maybe we can take a chapter out of Django's Form class -> `as_ul()`?

seo/filters.py
almathew: There are dummy requests we use in some areas (the tests being one) that won't always have META. 

I'm trying really hard not to change existing functionality here, just to move it around so I can subclass and override some of the smaller functions in my next pull request.

seo/filters.py
almathew: Sure. I like it.

PR #1196, opened 2015-01-26
[PD-955, PD-977, PD-969] PRM Reporting Output

aspidites:
## Description
This **WIP** branch is about implementing the django AJAX view responsible for delivering data to the MyReports app. It isn't complete, but I wanted to put it up so that:
* I could get early feedback in the event that I am heading the wrong direction
* Keep everyone up to speed on the progress I've made

## Changes
* Updated unit test to account for the lack of an explicit overview page
* Wrote the `search_records` function which takes column names as parameters and intelligently* filters based on those values
* Added a unit test for the `search_records` function (partially incomplete)

## Remaining Tasks and Known Issues
- [ ] Take into account date rangers, rather than simply a start date
- [ ] Do something more intelligent with `ForeignKey`s
- [ ] Allow for other result types (XML, PDF, CSV)
- [ ] Either derive a version of the `search_records` view function for `Contact`s, or enhance it to allow results to be queried on `Contact`s (rather than `ContactRecord`s) 

* Intelligently meaning I naively assume that for particular field types, we will be doing the same kind of queries.


aspidites:
played against the wrong branch...

PR #1197, opened 2015-01-26
 [PD-955] PRM Reporting Output

aspidites:
### Description
This  branch is about implementing the django AJAX view responsible for delivering data to the MyReports app. It isn't complete, but I wanted to put it up so that:
* I could get early feedback in the event that I am heading the wrong direction
* Keep everyone up to speed on the progress I've made

### Changes
* Updated unit test to account for the lack of an explicit overview page
* Wrote the `search_records` function which takes column names as parameters and intelligently* filters based on those values
* Added unit tests for the `search_records` function
* Select models now have a custom manager which provides `from_search` method used to filter based on search terms in a request. This manager intelligently filters the `Contact`, `ContactRecord`, and `Partner` models. Parameters can be further processed by adding a `parse_parameters` class method to a model. I didn't put it in the model because I didn't want to have to create a bunch of sub-classes. I may experiment with that once other things are in play to see how the aesthetics of the two approaches feel.

### Remaining Tasks
- [x] Take into account date rangers, rather than simply a start date
- [x] Do something more intelligent with `ForeignKey`s
- [x] Either derive a version of the `search_records` view function for `Contact`s, or enhance it to allow results to be queried on `Contact`s (rather than `ContactRecord`s) 
- [x] Write model-specific filter helpers
- [x] Determine if these helpers are a good fit for custom managers on the model
- [x ] Take care of todo items, such as moving code to appropriate modules
* Intelligently meaning I naively assume that for particular field types, we will be doing the same kind of queries.

#### Note: I merged with master, so this will bring the FEATURE-myreports branch up-to-date


myjobs/tests/factories.py
```diff
@@ -3,20 +3,16 @@
 
 
 class UserFactory(factory.django.DjangoModelFactory):
-    FACTORY_FOR = User
-    email = 'alice@example.com'
+    class Meta:
+        model = 'myjobs.User'
+
+    email = factory.Sequence(lambda n: 'alice%n@example.com' % n)
```
almathew: %s?

aspidites: Good catch. Was supposed to be %i, actually (Since I don't have to worry about floating point precision).

aspidites:
Going to see if I can make the PR cleaner (not showing every commit since the beginning of time).

aspidites:
This PR will look less intimidating once #1212 is merged.

tdphillips:
Conflicts.

PR #1198, opened 2015-01-26
A widget for facets

almathew:
Remaining steps for feature branch: Javascript.

Allows multiple facets to show up on the right hand column thing.

All tests pass.

### DIFF
--- a/middleware.py
+++ b/middleware.py
@@ -245,19 +245,25 @@ class MultiHostMiddleware:
             else:
                 setattr(settings, v, '')
 
-        settings.CACHE_MIDDLEWARE_KEY_PREFIX = "%s" % (my_site.domain,)
-        default_site_facets = SeoSiteFacet.objects.filter(
-            seosite=my_site).filter(
-                facet_type=SeoSiteFacet.DEFAULT)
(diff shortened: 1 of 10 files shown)


middleware.py
```diff
@@ -245,22 +245,29 @@ def process_request(self, request):
...
+
+        default = SeoSiteFacet.DEFAULT
+        default_site_facets = site_facets.filter(facet_type=default)
+        settings.DEFAULT_FACET = custom_facets_ops_groups(default_site_facets)
+
+        featured = SeoSiteFacet.FEATURED
+        featured_site_facets = site_facets.filter(facet_type=featured)
+        settings.FEATURED_FACET = custom_facets_ops_groups(featured_site_facets)
+
+        standard = SeoSiteFacet.STANDARD
```
JLMcLaugh: Are we querying these for every request and re-querying them in views with get_custom_facets?

almathew: I was actually just looking at that. We do it again in `get_solr_facet()` but it's for a slightly different (but more accurate) queryset, so I'll probably end up switching those around.

PR #1199, opened 2015-01-26
Made it so that editing a partner saved search uses the correct email.

aspidites:
* When a previous user has unsubscribed, goes back to no email address being selected
* Otherwise, selects the previously selected email
2015-01-26-164137_1916x1056_scrot


aspidites:
@almathew 's branch is addressing the issue I was meaning to address with this PR (I thought she was only disabling the field).

PR #1200, opened 2015-01-26
Disable the saved search recipient after it's already been set

almathew:
All tests pass.

### DIFF
--- a/templates/includes/error-highlight.html
+++ b/templates/includes/error-highlight.html
@@ -26,11 +26,11 @@
     </div>
     <div class="profile-form-input {{ field.label|slugify }}-field">
         {% if field.label == 'Send Results to *' %}
-            <select id="id_email" name="email" style="display: inline-block;">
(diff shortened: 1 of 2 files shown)


templates/includes/error-highlight.html
```diff
@@ -26,11 +26,11 @@
     </div>
     <div class="profile-form-input {{ field.label|slugify }}-field">
         {% if field.label == 'Send Results to *' %}
-            <select id="id_email" name="email" style="display: inline-block;">
+            <select id="id_email" name="email" style="display: inline-block;" {% if form.fields.email.widget.attrs.disabled %}disabled{% endif %}>
```
aspidites: How is this any better than what I had? Rather than checking for one attribute, you're just checking for a different one, which you had to add logic for, so the code is duplicated anyway.

almathew: When you're looking for a disabled field people will tend to look in forms.py, but really it's not that big of a deal either way.

PR #1201, opened 2015-01-27
Updated settings files to use the new solr URL.

mklauber:


mklauber:
This is staying like this b/c it only affects production, and is dangerous needs careful deployment and monitoring.

PR #1205, opened 2015-01-27
[PD-1000] Stop the postajob listing page from redirecting on login

aspidites:
More fixes to my decorator...this makes it so that you aren't redirected to login on /posting/list


tdphillips:
This is based off of reporting.

aspidites:
Not any more. Fixed. Thanks for the catch.

PR #1203, opened 2015-01-27
More button for multiple facets

almathew:
This is the last piece of the feature branch. Merge after https://github.com/DirectEmployers/MyJobs/pull/1198
All tests pass.

PR #1207, opened 2015-01-28
Feature facet groups

almathew:
Merge after https://github.com/DirectEmployers/MyJobs/pull/1203 and after you've actually looked at the output locally.


almathew:
You can now feel safer merging this knowing that it won't be deployed for a few days :+1:

PR #1209, opened 2015-01-30
Myreports - Front-end Report creation, management, send

galitskyd:
Just up for preview at the moment.

### DIFF
--- a/myreports/views.py
+++ b/myreports/views.py
@@ -7,5 +7,13 @@ from myreports.decorators import restrict_to_staff
 
 @restrict_to_staff()
 def reports(request):
+    if request.is_ajax():
+        response = HttpResponse()
+        template = 'myreports/prm/page{page}.html'
+        html = render_to_response(template.format(page=request.GET['page']),
+                                  {}, RequestContext(request))
+        response.content = html.content
+        return response
+
     return render_to_response('myreports/reports.html', {},
                               RequestContext(request))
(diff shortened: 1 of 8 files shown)


myreports/views.py
```diff
@@ -1,11 +1,52 @@
...
+        html = render_to_response(template.format(page=request.GET['page']),
+                                  {}, RequestContext(request))
+        response.content = html.content
+        return response
+
     return render_to_response('myreports/reports.html', {},
                               RequestContext(request))
+
+
+def prm_filter_partners(request):
```
galitskyd: Crappy place holder view that is suppose to return all.

myreports/views.py
```diff
@@ -1,11 +1,52 @@
...
+            'partners': partners
+        }
+        response = HttpResponse()
+        template = 'myreports/includes/prm/partners.html'
+        html = render_to_response(template, data, RequestContext(request))
+        response.content = html.content
+        return response
+
+
+def prm_filter_contacts(request):
```
galitskyd: Crappy place holder view that is suppose to return all.

static/reporting.160-26.js
```diff
@@ -0,0 +1,353 @@
...
+            $(all_checkboxes).each(function(element) {
+                $(all_checkboxes[element]).prop("checked", true);
+            });
+
+            // remove disabled on next button if it is disabled
+            if($(next_btn).hasClass("disabled")) {
+                $(next_btn).removeClass("disabled");
+            }
+        } else {
+            // turn all chekcs off
```
tdphillips: checks

PR #1216, opened 2015-02-02
Corrected log location

mklauber:


aspidites:
I can't make tests pass for this since it requires changing server settings.

PR #1218, opened 2015-02-02
[PD-921] Ignore includeinindex flag on certain BusinessUnits.

mklauber:
Created the ignore_includeinindex boolean, and updated
tasks.import_job_source to check it.


import_jobs.py
```diff
@@ -44,11 +44,9 @@ def update_job_source(guid, buid, name):
...
     # Lookup the jobs, filter then, transform them, and then load the jobs
-    jobs = get_current_jobs(guid)
-    # Hack to ignore the includeinindex bit for pearsoninternal.jobs
-    # Should be temporary.
-    if buid not in ["31578", "31094", "23525", "23500"]:
-        jobs = filter_current_jobs(jobs)
+    zf = get_jobsfs_zipfile(guid)
+    jobs = get_jobs_from_zipfile(zf, guid)
+    jobs = filter_current_jobs(jobs, bu)
     jobs = list(hr_xml_to_json(job, bu) for job in jobs)
```
aspidites: Probably better to use a list comprehension directly, rather than converting a generator expression to a list:
`jobs = [hr_xml_to_jason(job, bu) for job in jobs]`

import_jobs.py
```diff
@@ -92,15 +95,27 @@ def get_current_jobs(guid):
...
+    """Get a list of xml documents representing all the current jobs.
+
+    Input:
+        :guid: A guid used to access the jobsfs server.
+    :return: [lxml.eTree, lxml.eTree,...]"""
+    logger.debug("Getting current Jobs for guid: %s", guid)
+
+    # Delete any existing data and use the guid to create a unique folder.
+    directory = "/tmp/%s" % guid
     if os.path.exists(directory):
```
aspidites: Do we want to log when an existing directory is removed?

seo/migrations/0010_auto__add_field_businessunit_ignore_includeinindex.py
```diff
@@ -0,0 +1,406 @@
+# -*- coding: utf-8 -*-
```
aspidites: Does this *have* to be two migrations, or could it conceivably be one since we are introducing them at the same time?

seo/tests/test_import_jobs.py
```diff
@@ -4,7 +4,9 @@
 from django.conf import settings
 
 from seo_pysolr import Solr
-from import_jobs import DATA_DIR, add_company, remove_expired_jobs, update_solr
+from import_jobs import DATA_DIR, add_company, remove_expired_jobs, update_solr, get_jobs_from_zipfile,\
```
aspidites: PEP8:
```python
from import_jobs import (DATA_DIR, add_company, remove_expired_jobs, update_solr, 
                                        get_jobs_from_zipfile)
```

seo/tests/test_import_jobs.py
```diff
@@ -145,3 +147,48 @@ def test_remove_expired_jobs(self):
...
+        
+        with open(self.zipfile) as zf:
+            self.jobs = list(get_jobs_from_zipfile(zf, "ce2ca701-eeca-4c13-96ba-e6bde9cb7060"))
+            
+        self.businessunit = BusinessUnitFactory(id=0)
+    
+    def tearDown(self):
+        pass
+    
+    def test_filtering_on_includeinindex_bit(self):
```
aspidites: Docstrings?

seo/migrations/0010_auto__add_field_businessunit_ignore_includeinindex.py
mklauber: I'd prefer to make it two migrations, because combining data and schema changes would be expected.  Also, the data migration can easily be removed or modified, if the assumptions (those business units need to be set True) stops being valid.

import_jobs.py
mklauber: Frankly, we need a better way to handle this.  There's two possible reasons why a directory would exist.  
1) The previous time that guid was run on this server, it failed.  In this case, it's safe to remove the directory.
2) Two jobs for the same guid are running on different workers on this server.  In this case it's not safe to remove the directory.  But this case has never arisen, and shouldn't occur since tasks are not usually queue more than 4 times a day.  

For now, I'd prefer to ignore this, and maybe create an enhancement ticket.  Thoughts?

seo/tests/test_import_jobs.py
mklauber: Added.

tdphillips:
:+1: pending @aspidites thoughts on the remaining point.

import_jobs.py
aspidites: Enhancement ticket is fine by me, as it seems as we may want to think these cases through. Important for me is to keep eyes on it.

seo/migrations/0010_auto__add_field_businessunit_ignore_includeinindex.py
aspidites: Fair point.

aspidites:
:+1:

PR #1225, opened 2015-02-03
Default saved search sorting to date [PD-1014]

almathew:
All tests pass


mysearches/migrations/0007_partner_saved_search_default_sort.py
```diff
@@ -0,0 +1,452 @@
...
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        # Default sort_by set to 'Date' for Partner Saved Searches, see PD-912
+        orm.SavedSearch.objects.filter(partnersavedsearch__isnull=False).update(sort_by='Date')
```
almathew: Assuming the fix actually works this time it shouldn't be needed again.

mysearches/forms.py
```diff
@@ -219,6 +219,8 @@ def clean(self):
     def save(self, commit=True):
         self.instance.feed = self.cleaned_data.get('feed')
         is_new_or_change = CHANGE if self.instance.pk else ADDITION
+        if not self.instance.pk:
```
aspidites: not having a pk implies a new form and thus not user specified?

mysearches/forms.py
```diff
@@ -232,7 +234,6 @@ def save(self, commit=True):
             }
             Invitation(**invite_args).save()
             # Default sort_by for new Partner Saved Searches, see PD-912
```
aspidites: If you're removing the code, the comment can probably be removed or updated and moved to your changes above, right?

mysearches/forms.py
almathew: Implies new form and therefore default should be date. I don't really care about the user.

mysearches/forms.py
almathew: I've updated my comment. The datamigration is purely historical, so I'm just going to leave it as-is.

PR #1235, opened 2015-02-05
New notification architecture [PD-956]

tdphillips:
A lot of this is ajax and javascript, which isn't really testable at the moment. Let me know of any omissions.

Also adds a landscape config, because gamification.

That being said, all tests pass.


mymessages/models.py
```diff
@@ -165,8 +169,11 @@ def get_messages(user):
...
     """
     now = timezone.now().date()
-    messages = Message.objects.prefetch_related('messageinfo_set').filter(
-        Q(group__in=user.groups.all()) | Q(users=user),
-        Q(expire_at__isnull=True) | Q(expire_at__gte=now)).distinct()
+    try:
+        messages = Message.objects.prefetch_related('messageinfo_set').filter(
+            Q(group__in=user.groups.all()) | Q(users=user),
+            Q(expire_at__isnull=True) | Q(expire_at__gte=now)).distinct()
+    except ValueError:
```
tdphillips: I'm not super happy with this, but the page that is loaded after user deletion tries to get the (now deleted) user's messages. This prevents that from turning into a 500.

tdphillips:
Adding comments.

tdphillips:
Shamelessly nicked @aspidites solution.

aspidites:
@tdphillips: I see you also nicked my lack of comments :-P. Could you toss on a one liner explaining why that line is even necessary? Similar to what you had with the try/except would do, if not shorter.

galitskyd:
I have a few suggestions.
screen shot 2015-02-06 at 9 14 17 am
* If I didn't know any better I wouldn't have realized I could click on the tr's.  Something that can help with this is on hover of a tr change the cursor to "pointer".  
* I'd recommend change "No new messages" to "No new unread messages" and add some padding on the left. (The left part of the N blends with border).  I thought I didn't have any messages but in reality I didn't have any unread messages.

@fezick should there be a new ticket for figuring out what we want to do with current message functionality?  Currently messages still are shown at the top of every page when a message is unread.

tdphillips:
I thought I took care of the left padding already. Oops.

.landscape.yaml
```diff
@@ -0,0 +1,2 @@
+ignore-patterns:
+    - (^|/)migrations(/|$)
```
almathew: I don't know how important it is to your life, but putting things like fixtures in here would probably make our code look better too.

tdphillips:
The most recent commits address @galitskyd and @almathew's comments. I'm waiting for tests to pass before I remove the label.

almathew:
Why is only "See All" highlighted when you hover over that line? 

hover_see_all


I'm not sure why this bothers me, but I'm pretty sure there's some UI/usability thing I have to learn from this.

tdphillips:
I don't know, but it made sense at the time. I'll see if I can't fix the weird height of the background while I'm at it.

almathew:
It's cool, and I'm sure @galitskyd has something to say about it. I just wasn't sure.

tdphillips:
Too late.

tdphillips:
I had to clear my cache to get the full inbox link clickable.

Running tests now.

tdphillips:
Tests passed locally. Let's see if Jenkins concurs (#2).

tdphillips:
Since none of this is on master yet, I deleted and recreated my one migration. If you're testing, you'll need to roll back before pulling.

tdphillips:
Tests are failing locally but I didn't touch that stuff. Let's see what happens.

PR #1246, opened 2015-02-06
Update .mailmap

tdphillips:
Did I miss anyone?


tdphillips:
This will not fix your commit history, so we are all still lazy bums.

PR #1252, opened 2015-02-09
[PD-1030] Proper jobs-per-email count when viewing a saved search

aspidites:
* made sure the jobs_per_email field was passed to the function that parses feeds
* made sure that said value isn't ignored in that function

I have a unit test written for this, but I am apparently passing the wrong arguments to parse_feed. Will push when I've figured it out.

### DIFF
--- a/mydashboard/tests/test_views.py
+++ b/mydashboard/tests/test_views.py
@@ -124,7 +124,7 @@ class MyDashboardViewsTests(MyJobsBase):
             solr = pysolr.Solr(location)
             solr.add(dicts)
 
-    @unittest.skip("Need to decide whether or not we want duplicate candidates" 
+    @unittest.skip("Correct behavior undefined with respect to duplicates.")
     def test_number_of_searches_and_users_is_correct(self):
         response = self.client.post(
             reverse('dashboard')+'?company='+str(self.company.id))
--- a/mydashboard/tests/test_views.py
+++ b/mydashboard/tests/test_views.py
@@ -126,11 +126,8 @@ class MyDashboardViewsTests(MyJobsBase):
     def test_number_of_searches_and_users_is_correct(self):
         response = self.client.post(
             reverse('dashboard')+'?company='+str(self.company.id))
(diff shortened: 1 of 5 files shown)


mydashboard/tests/test_views.py
```diff
@@ -123,14 +124,12 @@ def add_analytics_data(self, type_, num_to_add=2):
             solr = pysolr.Solr(location)
             solr.add(dicts)
 
+    @unittest.skip("Need to decide whether or not we want duplicate candidates"
```
tdphillips: Missing a bit of punctuation.

mysearches/helpers.py
```diff
@@ -133,7 +133,7 @@ def get_json(json_url):
 
 
 def parse_feed(feed_url, frequency='W', num_items=100, offset=0,
-               return_items=5, use_json=True, last_sent=None,
+               return_items=None, use_json=True, last_sent=None,
```
almathew: Is there a specific reason we actually need both return_items and num_items? @tdphillips, do you know what the goal ever was of having both?

tdphillips: Something something getting a very rough count of jobs.

mydashboard/tests/test_views.py
aspidites: That bit of punctuation would have made the line wrap. I'm not bothering with it. Especially when the output isn't even printed.

mysearches/helpers.py
aspidites: Because at one point they both mattered. It was the number of jobs fetched vs the number of jobs returned by the function. That said, I introduced neither, and they existed long before this change.

mydashboard/tests/test_views.py
aspidites: Oh crap. You meant the ')'.

tdphillips: That is not the bit of punctuation I'm referring to. Tests are failing because `.skip(` doesn't have an end parenthesis.

PR #1258, opened 2015-02-10
Feature: MyReports

aspidites:
Moves the work done over the last week and a half into the quality control branch. Everything here has already been reviewed at least once.

## Brief Overview of Added Features:
- `SearchParameterQuerySet` (and manager) which provides the `from_search` method, which intelligently parses query parameters into sensible model filter arguments
- `filter_records` view function which can be used to filter any app's model, so long as they have registered the appropriate `QuerySetManager`. This function can also be used to sort results, and provides a cache which may optionally be ignored.
- Javascript foundation for describing a report with wizard-style pages (@galitskyd , please update this description if it's inaccurate)
- Rudimentary reporting UI which currently filters by date, state, partner, and contact
- JSON to be used to format the report is generated, though not currently utilized


mypartners/models.py
```diff
@@ -170,6 +330,38 @@ class Partner(models.Model):
...
     owner = models.ForeignKey('seo.Company')
 
+    company_ref = 'owner'
+    objects = SearchParameterManager()
+
+    @classmethod
+    def _parse_parameters(cls, parameters, records):
+        """Used to parse state during `from_search()`."""
+
+        start_date = parameters.pop('start_date', None)
```
almathew: Are we really sure `start_date` and `end_date` are never needed again before popping them?

aspidites: Yes. If we need them again, its because we are querying a different model, in which case it would be a separate request anyway.

PR #1263, opened 2015-02-11
[ST-862] Sites now populate when choosing a group

aspidites:
I technically only changed the first line. The others are showing as changed because I removed a level of indentation.


galitskyd:
:+1: after tests pass I'll merge

PR #1264, opened 2015-02-12
Add count to non-mobile topbar [PD-956]

tdphillips:
2015-02-12-094454_268x191_scrot

### DIFF
--- a/myjobs/tests/test_views.py
+++ b/myjobs/tests/test_views.py
@@ -2,6 +2,7 @@ import base64
 from bs4 import BeautifulSoup
 from datetime import timedelta, date
 from importlib import import_module
+import re
 import time
 import uuid
 from urllib import urlencode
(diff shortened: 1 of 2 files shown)


myjobs/tests/test_views.py
```diff
@@ -881,6 +882,10 @@ def test_messages_in_topbar(self):
             infos[0].mark_read()
 
             response = self.client.get(reverse('home'))
+            search = re.search('id="menu-inbox">%s<' % (num_messages-1, ),
```
aspidites: Do we need a regular expression to test whats already available by inspecting the response context?

tdphillips: Nope, I originally thought there would be whitespace, which has since been removed from the regex, and just left it in.

tdphillips: I was running tests again after removing the regex. I guess I don't have to now.

aspidites: I figured you'd be making other improvements, so addressing that unit test could be done in another iteration. Feel bad about the initial PR, so didn't want to hold up a new feature on a nitpick.

PR #1265, opened 2015-02-12
Add margin between message count and envelope icon [PD-956]

tdphillips:
I didn't bump the css version number because it's not yet on production and qc should still be loading files locally. Is that okay with everyone?

All spaces at the end of lines seem to have been removed when I saved the file in vim. Not complaining, just didn't expect it.

Hopefully this is one of my last changes.

All tests pass.


aspidites:
@tdphillips that's fine by me as far as not bumping the version number @galitskyd  and I took the same strategy with respect to reporting.

PR #1267, opened 2015-02-13
Sendgrid to celery [PD-1009]

almathew:
I think I got it all.

Requires `SENDGRID_BATCH_POST_USER` and `SENDGRID_BATCH_POST_PASSWORD` in secrets.py. Both are strings. I don't really care what you make the strings.

All tests pass.

### DIFF
--- a/default_settings.py
+++ b/default_settings.py
@@ -125,6 +125,9 @@ CELERY_QUEUES = {
     },
     'myjobs': {
         'binding_key': 'myjobs.#'
+    },
+    'sendgrid': {
+        'binding_key': 'sendgrid.#'
     }
 }
 CELERY_ROUTES = {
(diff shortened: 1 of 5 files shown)


tdphillips:
Regarding `SENDGRID_BATCH_POST_USER` and `SENDGRID_BATCH_POST_PASSWORD`:

If you change them from what SendGrid currently uses, be sure to also change them on SendGrid. That likely won't be immediately important as I'm betting event posting will still be off when this is deployed, but it should be rectified soon after.

almathew:
@tdphillips I've shared what should be there with @mklauber as well. I think the only time it's really important is when deploying.

aspidites:
@almathew what remains for this to be considered ready to merge?

almathew:
Deployment issues.

PR #1269, opened 2015-02-16
QC -> Staging

almathew:
Is this right?


aspidites:
Maybe? I haven't the slightest cue.

PR #1272, opened 2015-02-17
Event models

almathew:
I'm mostly guessing about what might work and if I'm even doing it correctly at this point, so I'm opening it for suggestions.

In the spirit of the ticket, I feel like I'm done plus some. I'd still like to get some tests and maybe triggers, though.

### DIFF
--- a/myemails/models.py
+++ b/myemails/models.py
@@ -3,43 +3,61 @@ from django.db import models
 
 
 class EmailTemplate(models.Model):
+    # header = models.ForeignKey()
+    # footers = models.ForeignKey()
+
     pass
 
 
 class Event(models.Model):
+    COMPARISON_CHOICES = (
+        ('', 'is equal to'),
+        ('__gte', 'is greater than or equal to'),
+        ('__lte', 'is less than or equal to'),
+    )
     email_template = models.ForeignKey('EmailTemplate', null=True)
     is_active = models.BooleanField(default=True)
     owner = models.ForeignKey('seo.Company')
 
+    compare_using = models.CharField(choices=COMPARISON_CHOICES)
+    model = models.ForeignKey(ContentType)
+    field = models.CharField(max_length=255)
+
     class Meta:
         abstract = True
 
-    def send_email(self):
-        return
-
-
-class ActionEvent(Event):
-    pass
-
-
-class CronEvent(Event):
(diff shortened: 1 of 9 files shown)


myemails/models.py
```diff
@@ -0,0 +1,110 @@
+from datetime import datetime, timedelta
+
+from django.contrib.auth.models import ContentType
+from django.db import models
+from django.template import Template
+
+
+class EmailSection(models.Model):
+    SECTION_TYPES = (
```
aspidites: :+1:  to this.

myemails/models.py
```diff
@@ -0,0 +1,110 @@
...
+        :param for_object: The object the email is being rendered for.
+        :return: The rendered email.
+        """
+        context = self.build_context(for_object)
+        template = Template(self.header + self.body + self.footer)
+        return template.render(context)
+
+
+class Event(models.Model):
+    email_template = models.ForeignKey('EmailTemplate')
```
aspidites: Do we need some sort of time stamp field? Also, is an explicit creatd_by needed, or are we just relying on reverse relationships for that?

myemails/models.py
```diff
@@ -0,0 +1,110 @@
...
+    is_active = models.BooleanField(default=True)
+    owner = models.ForeignKey('seo.Company')
+
+    model = models.ForeignKey(ContentType)
+    field = models.CharField(max_length=255)
+
+    class Meta:
+        abstract = True
+
+    def send_email(self, for_object):
```
aspidites: Is this a django api thing? I feel like i've seen like 20 send_email functions in spurious places.

aspidites:
Tests would be a good way to spec out what the API is missing. That whole test-driven development thing. Doing so in reports helped me out in that I was able to hone in rather quickly on what would be easiest/most flexible, both in implementing, and in interfacing with.

myemails/models.py
almathew: It will probably need a `created_by` field in the future, but without the forms there's no useful way of populating the information.

myemails/models.py
almathew: We send a lot of emails.

aspidites: I don't think my question was clear. Coming over.

PR #1273, opened 2015-02-17
[PD-915] Better representation for library data source

aspidites:
This changes the way that data source for a partner and contact are presented as well as adds the contact's location to the contact card. I also fixed a regression that caused contact locations not to show up in partner details.

# Remaining Tasks
- [x] screenshots
- [x] finish local manual testing (since it includes some template tasks)
- [x] update unit tests so that they are ensuring that the library association happens, rather than that tags are created

# Screenshots

## Before
2015-02-17-132057_627x126_scrot
2015-02-17-132110_624x203_scrot

## After
2015-02-17-141512_624x136_scrot
2015-02-17-142124_625x222_scrot


aspidites:
For some reason the data migration isn't performing consistently (contact library isn't being set).

aspidites:
Was an issue with swapping between branches that aren't aware of the data migration. Creating a backward method for that migration, then will remove DNM tag.

aspidites:
Should now be possible to do a backward migration do get back the old data_source named tags.

templates/mypartners/includes/contact_card.html
```diff
@@ -2,9 +2,21 @@
...
+        {% endfor %}
+    </div>
+    {% endif %}
+    <div class="product-details">
+        {% if contact.email %} {{contact.email}} {% endif %}
+        {% if contact.phone %} {% if contact.email %} &#8226; {% endif %} &#9742;{{contact.phone}} {% endif %}
+    </div>
+    {% if contact.library %}
+    <div class="product-details">
+        <i class="icon icon-list-alt icon-black" alt="original import source"></i>
```
tdphillips: I'll pull this down when my tests finish locally, but I'm not sure if `icon-black` exists. Black is the default state for icons - one of my PRs re-enabled the option to use white icons.

tdphillips:
:+1: but I have a bad track record with PRs. Confirmed my comment regarding icon-black, but it's not hurting anything having a class that doesn't add styles.

PR #1276, opened 2015-02-17
List of fields used [PD-498]

almathew:
Open so @JLMcLaugh  can take a look, but really all tests pass so other than continuing to manually test I'm not sure what else I want to do.

### DIFF
--- a/seo/tests/test_views.py
+++ b/seo/tests/test_views.py
@@ -737,8 +737,10 @@ class SeoSiteTestCase(DirectSEOTestCase):
         null search results - Ticket MS-378
 
         """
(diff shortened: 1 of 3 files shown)


seo/tests/test_views.py
```diff
@@ -737,6 +737,9 @@ def test_custom_moc(self):
         null search results - Ticket MS-378
 
         """
+        base_search_fields = deepcopy(helpers.search_fields)
+        helpers.search_fields = helpers.search_fields + ['onet', 'moc']
```
aspidites: helpers.search_fields += ['onet', 'moc']

PR #1279, opened 2015-02-19
[PD-1055] Sanitize State and City input

aspidites:
The regex idea didn't pan out, and a custom query seems like more work for a problem that could simply be avoided in the first place.
* Introduced a new form type which sanitizes all of its inputs on save.
* added data migration to sanitize existing locations to take care of the original bug report

I didn't bother sanitizing everything, as that would be time consuming, and it wasn't determined that other fields were an issue. That said, going forward, fields will be sanitized. The introduced class can be used in other places, but I didn't do so, as there may be legitimate places where white space counts (configurations and custom pages for example)

~~This is marked as do not merge because it introduces a migration which needs to happen after #1273~~


aspidites:
Didn't use F objects as I need to modify the value, rather than simply reference it, and I couldn't find a way to get the string representation of said value. For instance, str(F('state')) returns "DEFAULT", rather than something like "New York".

mypartners/migrations/0008_normalize_location_data.py
```diff
@@ -0,0 +1,434 @@
...
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        """Normalizes city and state data."""
+        locations = Location.objects.filter(
```
tdphillips: Would Q objects make a difference? (`Q(state__...) | Q(city__...)`)

aspidites: Not sure. I assumed they'd be equivalent considering they are still getting or'd together. I'll take a peak at the generated SQL and get back to you

aspidites: @tdphillips Identical results:
```python
 str(Location.objects.filter(Q(city__icontains=' ') | Q(state__icontains=' ')).query)
Out[9]: 'SELECT `mypartners_location`.`id`, `mypartners_location`.`label`, `mypartners_location`.`address_line_one`, `mypartners_location`.`address_line_two`, `mypartners_location`.`city`, `mypartners_location`.`state`, `mypartners_location`.`country_code`, `mypartners_location`.`postal_code` FROM `mypartners_location` WHERE (`mypartners_location`.`city` LIKE % %  OR `mypartners_location`.`state` LIKE % % )'

In [10]: str((Location.objects.filter(city__icontains=' ') | Location.objects.filter(state__icontains=' ')).query)
Out[10]: 'SELECT `mypartners_location`.`id`, `mypartners_location`.`label`, `mypartners_location`.`address_line_one`, `mypartners_location`.`address_line_two`, `mypartners_location`.`city`, `mypartners_location`.`state`, `mypartners_location`.`country_code`, `mypartners_location`.`postal_code` FROM `mypartners_location` WHERE (`mypartners_location`.`city` LIKE % %  OR `mypartners_location`.`state` LIKE % % )'
```

I also didn't use Q because I didn't realize migrations import django.db.models by default, and was feeling lazy. Q objects would be less repetitious, but as this is just a data migraiton, I'm not sure I  care.

tdphillips: Good enough for me.

tdphillips:
No tests?

Otherwise, :+1:

aspidites:
@tdphillips added unit test. Also decided to split on white space as catching "legit" whitespace was going to make things clunky (eg. consuming tabs, but not new lines, but also spaces). I figure we can expand the api to allow for the changing of sanitizing character set if need be.

PR #1280, opened 2015-02-19
Allow saved searches with old domains to be viewed [PD-994]

tdphillips:
This is mainly being opened to confirm that Jenkins agrees when I say all tests pass.


tdphillips:
And Jenkins agrees. Feel free to rip this apart. I did it as simply as I could.

almathew:
:octopus:

mysearches/tests/helpers.py
```diff
@@ -26,7 +27,9 @@ def _inner(url, *args, **kwargs):
         :file: File-like object
         """
         feed = False
-        if 'feed/rss' in url:
+        if '404.com' in url:
+            raise HTTPError(url=None, code=404, msg=None, hdrs=None, fp=None)
```
tdphillips: Docs aren't helpful regarding HTTPError. They imply that there are only two arguments, code and body. There are actually six, which doesn't include body, and all are required.

aspidites: Any reason we aren't using [`django.http`](https://docs.djangoproject.com/en/1.7/topics/http/views/#returning-errors) for this?

tdphillips: We use urllib2 to retrieve most things, which throws an HTTPError. These two should be functionally equivalent (disclaimer: I haven't actually tried HttpResponseNotFound so this is a semi-educated guess).

TL;DR: We explicitly catch HTTPError so I was going for a level of consistency.

I'll check out django.http.

aspidites: AH, ok. :+1:  then.
TL;DR: :+1:

templates/mysearches/view_full_feed.html
```diff
@@ -47,6 +47,8 @@
                             </td>
                         </tr>
                     {% endfor %}
+                {% elif items == None %}
+                    <tr><td>This site is no longer in operation. You will need to delete or modify this saved search.</td></tr>
```
aspidites: Is this wording sanctioned, or just an ad lib. It reads fine, but I wasn't sure if:
* we were worried about "no longer in operation" sounding too close to "broken" or "we lost this company's business"
* "You will need to" sounded to authoritative (even if accurate) and might need to be replaced with "Please".

tdphillips: Just ad lib. @fezick probably knows better, but I can't currently think of something that says "this url is bad" that doesn't also imply "our stuff is broken" or "this company doesn't like us anymore."

fezick: The domain for this saved search is no longer valid. Please edit the domain or delete this saved search.

aspidites: :+1:  for conveying *what* doesn't work. Only question is, does the error message only show when the feed is broken, or does it additionally show if something else is wrong with the saved search?

aspidites:
Other than the messaging conversation currently in progress, things seem to function as expected, so I'm good with this.

templates/mysearches/view_full_feed.html
tdphillips: If we can't download the feed for any reason, this would get shown.

tdphillips: Text updated.

PR #1286, opened 2015-02-23
Replaced bleach with strip_tags.

aspidites:
Tests take forever to run as-is, do we really want to generate 500+ contact records just to test this, or is it fine as is?

### DIFF
--- a/mypartners/views.py
+++ b/mypartners/views.py
@@ -1,27 +1,23 @@
-import bleach
-from collections import OrderedDict
-from datetime import date, datetime, timedelta
+from datetime import date, timedelta
 from email.parser import HeaderParser
 from email.utils import getaddresses
 from itertools import chain
 import json
 from lxml import etree
-import newrelic.agent
 import pytz
 import re
-import sys
 import unicodecsv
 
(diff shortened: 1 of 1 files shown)


tdphillips:
What is the likelihood that a contact record contains something that resembles but is not an HTML tag? The docs for strip_tags claim it just looks for something enclosed in angle brackets.

aspidites:
I assume rare and that such a use case is something that would be turned up during the week of qc.

tdphillips:
Good enough for me.

:+1:, anyone else?

mypartners/views.py
```diff
@@ -1157,7 +1152,7 @@ def prm_export(request):
                 ', '.join([val.name for val in v.all() if val]) for v in values
             ]
             # Remove the HTML and reformat.
-            values = [bleach.clean(v, [], strip=True) for v in values]
+            values = [strip_tags(v) for v in values]
```
JLMcLaugh: Is this marked safe on the front end? strip_tags isn't secure, but if we're escaping, then it should be fine.

PR #1287, opened 2015-02-24
[PD-791] Reduce number of queries topbar makes

aspidites:
Before, the number of queries made by the top bar grew linearly proportional to the number of companies the logged in user belonged to. This is now constant. For example, with 23 companies, 50 queries were being made, where as with this code, that was reduced to 27. Other queries seem to be tied to middleware and a few other things, which I expect can be tackled as a separate issue.

### DIFF
--- a/myjobs/templatetags/common_tags.py
+++ b/myjobs/templatetags/common_tags.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from myjobs import version
 from myjobs.models import User
 from myjobs.helpers import get_completion, make_fake_gravatar
-from seo.models import CompanyUser
+from seo.models import Company
 from universal.helpers import get_company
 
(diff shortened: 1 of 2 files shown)


myjobs/templatetags/common_tags.py
```diff
@@ -85,16 +85,11 @@ def get_company_name(user):
...
     """
 
-    try:
-        companies = CompanyUser.objects.filter(user=user)
-        company_list = [company.company for company in companies]
-        return company_list
-    except CompanyUser.DoesNotExist:
-        return {}
+    # Only return companies for which the user is a company user
+    return user.company_set.filter(companyuser__user=user)
```
tdphillips: Since CompanyUser is a through table between Company and User, shouldn't this be the same without the `companyuser__user=user`?

PR #1289, opened 2015-02-25
Add fields to partner saved search detail page [PD-1027]

tdphillips:
The number of new jobs is shown under "Jobs per Email" as long as it is different from what the member chose. In this picture, we tried to do 100 jobs per email but only 44 were new. In the event that there were 100 new jobs, the 100 would still be displayed but there would be no parenthesized number next to it.

Right now the number in the header and the number of new jobs are the same. I don't think that's a given, but I'm not sure.

2015-02-25-085246_975x647_scrot

All tests pass.

### DIFF
--- a/mypartners/tests/test_views.py
+++ b/mypartners/tests/test_views.py
@@ -719,10 +719,12 @@ class SearchFeedTests(MyPartnersTestCase):
         details = soup.find(class_="sidebar")
         self.assertIn('Active', details.find('h2').get_text())
         texts = ['http://www.my.jobs/jobs',
+                 'None',
                  'Weekly on Monday',
                  'Relevance',
                  'Never',
                  'alice@example.com',
+                 str(self.search.jobs_per_email),
                  'All jobs from www.my.jobs']
         details = details('span', recursive=False)
         for i, text in enumerate(texts):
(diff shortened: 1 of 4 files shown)


tdphillips:
This should probably wait for Seth to sign off on the change.

aspidites:
@tdphillips I'm pretty sure AJ requested this change, not Seth.

tdphillips:
You saw nothing.

mypartners/tests/test_views.py
```diff
@@ -719,10 +719,12 @@ def test_details(self):
         details = soup.find(class_="sidebar")
         self.assertIn('Active', details.find('h2').get_text())
         texts = ['http://www.my.jobs/jobs',
+                 'None',
```
aspidites: When is this used?

aspidites:
Is it intended that Notes says "Notes"? Also, what does it look like with multiple source codes?

mypartners/tests/test_views.py
tdphillips: The values in this list correspond to the <span> tags in view_full_feed.html. That one in particular is displayed when a partner saved search doesn't include source codes or campaigns.

tdphillips:
I could have picked a better value, but the string "Notes" is what I typed into the notes field.

The source code/campaign spot is basically a query string. `code_1=foo&code_2=bar`

aspidites:
It's been a couple of days. If AJ doesn't like the way it looks, we can just change it later.

PR #1292, opened 2015-02-25
Custom search box text [PD-712]

almathew:
All tests pass

### DIFF
--- a/templates/search_box.html
+++ b/templates/search_box.html
@@ -9,12 +9,12 @@
     <div class="row standard">
       <div class="searchHeader" role="search">
           <div class="searchTab">
-              <label for="location">{% if site_config.where_label %}{{ site_config.where_label }}{% else %}{% trans "Where" %}{% endif %}</label>
+              <label for="location">{{ site_config.where_label|default:_("Where") }}</label>
               <input title="Search Location" type="text" name="location" id="location" class="micrositeLocationField" {% if location_term != "\*" %}value="{{ location_term }}"{% endif %} {% if site_config.where_placeholder %}placeholder="{{ site_config.where_placeholder }}"{% endif %}/>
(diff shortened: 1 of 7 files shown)


templates/search_box.html
```diff
@@ -8,18 +8,16 @@
...
-
-        </div>
-        <div class="searchTab">
-          <label for="q">{% trans 'What' %}</label>
-          <input title="Search Phrase" type="text" name="q" id="q" class="micrositeTitleField" value={% if title_term != "\*" %}"{{ title_term }}"{% else %}""{% endif %} />
-          <span class="helptext" id="what">{% blocktrans %}job title, keywords{% endblocktrans %}</span>
-        </div>
-        
+          <div class="searchTab">
+              <label for="location">{% if site_config.where_label %}{{ site_config.where_label }}{% else %}{% trans "Where" %}{% endif %}</label>
```
aspidites: Will a default filter not work? Something like `{{ site_config.where_label|default:"Where" }}`

aspidites: Oh. Trans. Ugh.

almathew: Yeah. I think they actually use it too. I'm afraid to find out.

aspidites: http://stackoverflow.com/questions/3595099/django-trans-tag-within-a-default-filter

almathew:
I've addressed @aspidites's comments a bit, but the place that probably needs it the most (help text) will need to be a multi-line change in order to continue marking it as safe. Opinions?

aspidites:
@almathew : Out of curiosity, is this where bleach would come in? That is, if we are concerned about marking things as safe, would bleach resolve those concerns, or be over kill? (Yay, super slow rendering because of bleach!).

almathew:
I'm marking it as safe to contain html.

aspidites:
I thought bleach did that, but removed unsafe html. That whole cake and eating it too, thing?

PR #1296, opened 2015-02-27
[PD-922] Made templates UTF-8 by default and changed Configuration Table to UTF-8

aspidites:
Changed the Configuration table in seo to use UTF-8 and added a chareset line to each of the base templates.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -10,7 +10,7 @@
     {% block site-title %}
     <title>My.jobs</title>
     <meta name="title" content="My.jobs">
-    <meta charset= “utf-8”>
+    <meta charset= "utf-8">
     {% endblock %}
 
     {% block meta %}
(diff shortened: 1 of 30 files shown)


galitskyd:
:+1: @JLMcLaugh

templates/base.html
```diff
@@ -10,6 +10,7 @@
     {% block site-title %}
     <title>My.jobs</title>
     <meta name="title" content="My.jobs">
+    <meta charset= “utf-8”>
```
tdphillips: Fancy quotes intentional?

aspidites: In so much as it was a copy-paste, yes. Otherwise no. I wonder if i fix that if i can get rid of the unit test changes...

aspidites:
Much better. Thanks @tdphillips

PR #1299, opened 2015-03-03
[PD-1083] Standardize Locations in Forms

aspidites:
*Note: This isn't finished yet, but I wanted a way to run tests while still working on it

# What's been done
- deleted json(p) files that we weren't using
-  moved location_data module from postajob to myjobs (the alternative was universal)
- added missing regions to location data (AA, AE, AP, etc)
- started using location_data in reports app and prm portion of mypartners app

# What's left to do
- [ ] use location_data instead of countries.json
- [ ] completely remove jsondata directory
- [x] replace all mypartners forms state and city fields with somethin similar to what's used in postajob
- [ ] investigate whether or not it's useful to keep the state dropdown used in PRM (eg. there's no intelligent behavior happening since regions are not considered)
- [ ] find a clean way to factor out what's in the Location form's __init__


static/location.162-3.js
```diff
@@ -0,0 +1,35 @@
...
+        if($(this).val() === "USA") {
+            $('#id_state_0').show();
+            $('#id_state_1').hide();
+            $('#id_state_2').hide();
+        } else if($(this).val() === "CAN") {
+            $('#id_state_1').show();
+            $('#id_state_0').hide();
+            $('#id_state_2').hide();
+        } else {
+            console.log('here');
```
galitskyd: remove this please

aspidites:
Generally works but django's api is just quirky enough that I don't want to be responsible for maintaining this. Going back to retrofitting @tdphillips' work.+

PR #1302, opened 2015-03-04
[PD-1022] Removed Country selection from PRM and changed state input to drop down.

aspidites:
Good bye country selection.


tdphillips:
It should be noted, IIRC, that no entries currently exist for countries other than the US.

PR #1307, opened 2015-03-06
Blocks that search

almathew:
Opinions?


myblocks/context_tools.py
```diff
@@ -0,0 +1,109 @@
+from itertools import chain
+
+from django.core.urlresolvers import resolve, reverse
+
+from seo import cache, helpers
+from seo.breadbox import Breadbox
+
+
+def get_breadbox(request):
```
almathew: The plan is to memoize each of these so that even if they're called multiple times per request, they're still only actually executed once per request.

aspidites:
Seems fine. Nothing jumps out at me, anyway.

almathew:
This is now ready for full review.

PR #1308, opened 2015-03-06
Update css and links for inbox in topbar [PD-1111]

tdphillips:
All tests pass.

### DIFF
--- a/static/topbar.162-6.css
+++ b/static/topbar.162-6.css
@@ -543,3 +543,16 @@ ul.submenu > li > a:hover {
         text-align: left;
     }
 }
+
+#menu-inbox ~ ul li > p {
+    color: white;
+    padding-left: 4px;
+}
+
+#menu-inbox > div.icon-envelope {
+    margin-left: 3px;
+}
+
+.icon-white {
+    background-image: url("//d2e48ltfsb5exy.cloudfront.net/content_mj/files/images/glyphicons-halflings-white.png");
+}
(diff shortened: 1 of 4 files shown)


static/topbar.162-6.css
```diff
@@ -542,4 +542,17 @@ ul.submenu > li > a:hover {
...
+#menu-inbox ~ ul li > p {
+    color: white;
+    padding-left: 4px;
+}
+
+#menu-inbox > div.icon-envelope {
+    margin-left: 3px;
+}
+
+.icon-white {
```
tdphillips: This was copied from my.jobs.css. Removing it from there feels wrong, so it's duplicated. Suggestions?

aspidites: If the inbox is only visible in the topbar, what reason would there be to keep it in my.jobs.css?

tdphillips: This isn't inbox-specific - it allows any white icon to be used.

aspidites: Then perhaps the selector shouldn't be so specific (#menu-inbox ~/>)? Maybe I just don't get what the the selectors are doing. If it's only meant to enable white icons, I'd probably want some aptly named class rather than an id selector.

tdphillips: This one specific rule enables white icons. Everything above it was in the My.jobs css and has been removed.

aspidites: OK

aspidites: Oooooooooh, I was looking at lines 545-555, not 556. I'm :+1: now.

PR #1321, opened 2015-03-12
My Report Additions

galitskyd:
Lots of refactoring.
Added state dropdown.
Added a required param on Field objects.
Added a String prototype called capitalize which capitalizes the first Char of a string using syntax such as String.capitalize()

Fixed bugs:
- Fixed animation queue on lists. 
  - Now, starting a new animation will cause the current animation to end immediately and start the new animation.
- Fixed checkbox logic.
  - The all checkbox was not causing data to save.
  - The first time someone changed a state of a checkbox that state wasn't changing during checks so it wasn't saving properly.
- Fixed an issue where in readable data keys with value of 0 or '' would still show.
- Fixed an issue on a redirect that would trigger the beforeunload listener.


static/reporting.162-10.js
```diff
@@ -41,9 +32,19 @@ Report.prototype.create_fields = function(types) {
...
 
+  // Send the user a message when they try to unload the page saying their
+  // progress will not be saved and will be lost.
+  /*
+  window.addEventListener("beforeunload", function(e) {
+    if (!reload) {
+      e.returnValue = "You have unsaved changes!";
+    }
+  });
+  */
```
galitskyd: This is commented out at the moment because this is annoying during development.

PR #1322, opened 2015-03-12
[PD-1131, PD-1132] Report Display

aspidites:
## Summary
The code is really rough at the moment, and some of it is dead code, but the plan is to go back and prune once we see what bits we don't need. 

## Useful Parts
- refactored views:
    - `filter_records` now simply returns a query set
    - `view records` acts like the old filter_records except that it only recognizes json output and doesn't handle sorting. The count option does still work
    - `create_report` actually creates a report object
- added myreports.helpers
    - `serialize` acts like `django.core.serializers.serialize` except that it accounts for annotated records and uses a flatter representation
    - `parse_params` converts a querydict into a regular dict, intelligently converting list-like params into tuples (lists aren't hashable)

- there is a Report model which essentially is a representation of the cache_key used for the `filter_records` function used above. We may decide (hopefully before this hits production) that some of the fields are unnecessary. I did it this way as I'm contemplating using report pk's as cache keys rather than python objects, but am not quite sure yet.

## Useless Parts (That could be useful later)
- `myreports.reports` has a PRMReport object that reproduces the charts used in the old reports app
- `templates.myreports.prm_report.html` - makes use of the above report object, demonstrating how easy it is to add charts to a view

## Broken Parts (aka future improvements)
I'd typically do these kinds of things in the PR itself, but given the deadline and current size of the PR, I wanted to get this into @galitskyd's hands as soon as possible. That said, these are deficiencies that I'm already aware of:

- should be using a class-based view, dispatching on get/post accordingly
- update urls to take advantage of this so that we aren't being silly and just appending get/generate to url paths
- update unit tests: this week was a big sandbox for me, so I didn't feel like writing unit tests only to have to rewrite them. Now that the dust has settled, I can start to implement tests for this.


myreports/helpers.py
```diff
@@ -0,0 +1,38 @@
...
+    params = {}
+    for key in querydict.keys():
+        value = querydict.get(key)
+        value_list = querydict.getlist(key)
+
+        # parsing a list parameter as a regular parameter only captures the
+        # last item, so if trying both ways returns the same value, we can
+        # be sure that it's not a list
+        if value:
+            if value == value_list[0]:
```
tdphillips: Could this also be done by checking the length of `value_list`?

I'm not entirely sure what can be parsed, but is it possible for `value_list` to contain the same value multiple times?

aspidites: If by "this" you mean the entire block, no. Reason being that both strings and lists have a length, so I'm not gaining anything. If you just mean the bit where I check for the 0th item, I'm not sure. This code is already in production so was written weeks ago. I think there was a specific reason I didn't just check length, but I don't remember, sorry.

I don't think its possible to have the same value multiple times. "?foo=bar&foo=baz" would be parsed as foo = [bar, baz]. We played with something similar when prm is being built and I think both foos being bar still parsed correctly.

PR #1324, opened 2015-03-12
Fixed utf-8 migration.

aspidites:
- renamed migration so it wasn't a conflict
- added guard for sqlite so that we don't have to fake during tests or local development


galitskyd:
Works on sqlite3 :+1:

PR #1330, opened 2015-03-13
Fall back to solr query and title casing for locations so that we can add the breadbox to the no jobs page [PD-1141]

almathew:
Tests pass


almathew:
I think I got them all:
- Custom facets depend on the CustomFacet table
- Company depends on the Company table
- Military depends on the Moc table
- Title already had a fallback


So that left location, I think? This felt too easy.

PR #1335, opened 2015-03-16
Fix numbers [PD-1154]

tdphillips:
I've made the number look like the other number.

All tests pass.


aspidites:
@tdphillips  did you verify that the number we decided to display matches whatever is showing up in emails?

tdphillips:
Not yet.

tdphillips:
Just checked. The number equals the other number, as it apparently should.

tdphillips:
Except I forgot to push...

PR #1340, opened 2015-03-17
Reporting Refactor

aspidites:
Some minor refactoring before I begin working on generating chart data for the new reports template. I'm putting this up now because I'd like to get the model changes in before they hit production (so that we don't have two migrations). Speaking of which, @JLMcLaugh  would you be opposed to me working these changes as a hotfix against staging? Alternatively, since (as far as I understand) myreports isn't being released until the end of this sprint, we could do a zero migration before moving forward for myreports.

## Model Changes
- added a new manager to `ContactRecord` which will allow us to get at numbers more easily: `emails`, `phone_calls`, `meetings`, `applications`, `interviews`, `hires`, and the not so aptly labeled `records`.\
- added new properties to `Contact` instances in a similar vein: `communications`, 'referrals`, `contact_records`.
- `path` has been replaced by `app` and `model`
- It is now possible to get the records queryset by accessing a `Report` instance's `queryset` property
- changed params to text field with unlimited length to prevent truncation

## View Changes
- renamed `get_records` to `view_records`

### DIFF
--- a/default_settings.py
+++ b/default_settings.py
@@ -260,7 +260,6 @@ INSTALLED_APPS = (
     'saved_search',
     'taggit',
     'fsm',
-    'report_tools',
 )
 
 # Captcha SSL
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -284,18 +284,6 @@ class Contact(models.Model):
         query_string = urlencode(params)
         return "%s?%s" % (base_urls[self.content_type.name], query_string)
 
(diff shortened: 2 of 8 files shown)


aspidites:
Been doing work on top of this in another branch, and this hasn't been merged yet, so I'm just going to put that one up when it's ready. Also, since I haven't gotten a response about migrations, I'm just going to make this one 0002.

PR #1344, opened 2015-03-17
Search and job headers

almathew:


almathew:
Merge after #1342 for less changes.

PR #1348, opened 2015-03-18
Multiple sites per page [PD-1174] and some accidental caching [PD-1146]

almathew:


aspidites:
"and some accidental caching"?

almathew:
I meant for it to be a separate branch.

myblocks/context_tools.py
```diff
@@ -234,6 +234,12 @@ def get_query_string(request):
 
 
 @Memoized
```
aspidites: Can we use the built in [memoize decorator](https://docs.djangoproject.com/en/1.6/_modules/django/utils/functional/)? (that I mentioned about a week ago)

almathew: I tried messing around with it when I first started memoizing after you brought it up. There's an issue with the way the django's version handled hashing that was either causing requests to always match or be unique (I honestly forget which, but I think it was be unique).

myblocks/models.py
```diff
@@ -502,9 +528,33 @@ def pixel_template(self):
...
+            return rendered_template
         context = self.context(request, **kwargs)
         template = Template(self.get_template(request))
-        return template.render(RequestContext(request, context))
+        rendered_template = template.render(RequestContext(request, context))
+        blocks_cache.set(key, rendered_template, settings.MINUTES_TO_CACHE*60)
+        return rendered_template
+
+    def render_cache_prefix(self, request):
+        page = '::'.join([str(self.pk), str(self.updated)])
```
aspidites: modulo string formatting is faster (in python 2), and arguably clearer -> `'%s::%s' % (self.pk, self.updated)`. Also doesn't require multiple calls to `str`.

myblocks/context_tools.py
aspidites: OK. Decorator class may be overkill. That you are overriding `__call__` seems like a good indicator that what you really want is a decorator function.

almathew: Copy + paste from here: https://wiki.python.org/moin/PythonDecoratorLibrary#Memoize

aspidites: [From that same page](https://wiki.python.org/moin/PythonDecoratorLibrary#Alternate_memoize_as_nested_functions):
```python
def memoize(obj):
     cache = obj.cache = {}
 
     @functools.wraps(obj)
     def memoizer(*args, **kwargs):
         key = str(args) + str(kwargs)
         if key not in cache:
             cache[key] = obj(*args, **kwargs)
         return cache[key]
     return memoizer
```

PR #1350, opened 2015-03-19
Add source to Partner model [PD-1162]

tdphillips:
The source field appears on the new/edit partner forms. In the event that a partner was added from the library, the source is automatically populated with the data source for that entry in the library. It can later be edited if, as Hal suggested, the partner was met at a con of some type and was later located in our library.

Updating the form to be more pleasing will be done in PD-1163.

All tests pass.

### DIFF
--- a/mypartners/forms.py
+++ b/mypartners/forms.py
@@ -17,7 +17,6 @@ from mypartners.widgets import (MultipleFileField,
 from universal.forms import NormalizedModelForm
 
 
-
 def init_tags(self):
     if self.instance.id and self.instance.tags:
         tag_names = ",".join([tag.name for tag in self.instance.tags.all()])
@@ -193,6 +192,11 @@ class NewPartnerForm(NormalizedModelForm):
                 widget=forms.TextInput(
                     attrs={'placeholder': 'Partner Organization',
                            'id': 'id_partner-partnername'})),
+            'partnersource': forms.CharField(
+                label="Partner Source", max_length=255, required=False,
+                widget=forms.TextInput(
+                    attrs={'placeholder': 'Partner Source',
+                           'id': 'id_partner-partnersource'})),
             'partnerurl': forms.URLField(
                 label="Partner URL", max_length=255, required=False,
                 widget=forms.TextInput(attrs={'placeholder': 'Partner URL',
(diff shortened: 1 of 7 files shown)


templates/mypartners/includes/partner_card.html
```diff
@@ -75,7 +75,7 @@
         <div class="product-details">
             {% if on_page == 'prm' %}
                 <i class="icon icon-list-alt" alt="original import source"></i>
-                {{ partner.library.data_source }}
```
tdphillips: When a partner is initially created from the library, partner.library.data_source and partner.source will be identical. It can later be modified. Rather than displaying both or doing calculations to determine which should be shown, I opted for just including the source.

mypartners/migrations/0009_auto__add_field_partner_source__add_field_contact_library.py
```diff
@@ -0,0 +1,441 @@
...
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding field 'Partner.source'
+        db.add_column(u'mypartners_partner', 'source',
+                      self.gf('django.db.models.fields.CharField')(default='', max_length=255, blank=True),
+                      keep_default=False)
+
+        # Adding field 'Contact.library'
+        db.add_column(u'mypartners_contact', 'library',
```
tdphillips: Marked as don't merge until I find out what's happening here.

mypartners/forms.py
```diff
@@ -193,6 +192,11 @@ def __init__(self, *args, **kwargs):
                 widget=forms.TextInput(
                     attrs={'placeholder': 'Partner Organization',
                            'id': 'id_partner-partnername'})),
+            'partnersource': forms.CharField(
```
aspidites: Any reason to not just be consistent and go with "data_source"? In the template you'd then not have to distinguish between the two.

aspidites:
I assume you'll also need to change mypartners/helpers.new_partner_from_library

mypartners/forms.py
tdphillips: True.

tdphillips:
new_partner_from_library already adds the source. I'll just need to update it to use data_source.

tdphillips:
I suppose this should also include a data migration to update everything already created from a library item.

tdphillips:
Warning: This PR now includes yet another migration that modifies most of the entries of the Partner table.

mypartners/migrations/0010_update_partner_data_source.py
```diff
@@ -0,0 +1,435 @@
...
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        "Write your forwards methods here."
+        # Note: Don't use "from appname.models import ModelName". 
+        # Use orm.ModelName to refer to models in this application,
+        # and orm['appname.ModelName'] for models in other applications.
+        partners = orm.Partner.objects.filter(library__isnull=False)
```
aspidites: Is this where F objects would be handy? update instead of a for loop?

tdphillips: Tried that. `library` is a foreign key.

`django.core.exceptions.FieldError: Joined field references are not permitted in this query`

aspidites: Ok. Looks good then.

PR #1353, opened 2015-03-19
[PD-1177] MyReports Cleanup

aspidites:
# Description
Does a bunch of stuff, really. 

* Removes django-report-tools dependency
* re-enables myreports unit tests that were disabled about a week ago
* gets rid of unused functionality (can be re-introduced if we decide at a later date that we have a use for it)
* simplifies the signature of a few functions in myreports.views
* adds properties to mypartners.models.ContactRecord
* exposed the numbers needed to create the charts available in current PRM for use in new reports
* documented everything that I wrote (undocumented functions weren't created by me)
* changed the structure of the reports model slightly
* CSV export looks much better than the current reporting now

~~The only thing left to do before I consider this ready is to write unit tests for new functionality and properly decorate views~~. That said, don't let that stop you from reviewing the code. I'd much rather find out as I'm working on this that other things need to be updated rather than after I'm done.

# Shortcut
[Here](https://github.com/DirectEmployers/MyJobs/compare/myreports-name-clone...reports-cleanup) are the relevant changes. What's in the PR technically has a lot of @galitskyd's changes. If that's too overwhelming, visit the link above to see what I worked on specifically.


myreports/helpers.py
```diff
@@ -6,25 +6,108 @@
...
+from django.utils.html import strip_tags
 from mypartners.models import CONTACT_TYPE_CHOICES
 
 
-def serialize(fmt, records, output=None, counts=None, as_is=False):
-    # TODO: see if i can preserve annotations using `values`
-    if as_is:
-        data = records
-    else:
+# TODO:
```
tdphillips: Is this TODO going to be part of later tickets or this one?

tdphillips:
Looks good to me. Is the intention to still add unit tests?

myreports/helpers.py
aspidites: The humanizing TODO will be part of another branch, I think. The one about values has already been addressed locally.

aspidites:
@tdphillips: I'm adding tests as we speak I marked this as 'merge at your own risk' rather than do not merge', because I don't mind having those as  part of a separate pr, but understand if they are a requisite as well. The old tests are already running again, so only new functionality needs to be tested. I don't have intentions of writing tests for the views that @galitskyd created, though.

aspidites:
@JLMcLaugh onlything left to do is write tests for the helper functions. The views all have tests, and I haven't written any model tests because there is nothing extraordinary about the Report model.

aspidites:
Done messing with stuff now. I decided to finish CSV exporting. This hadn't been merged yet, so I decided to merge that work with this one.

PR #1354, opened 2015-03-20
Tests for blocks

almathew:


aspidites:
@almathew : This looks like it's more than tests. Mind expanding the description to be more useful?

almathew:
It's mostly tests, or helper functions to help with tests.

Looking through the only non-test changes are:
- Added the Page status column to the admin search view (I was getting frustrated with it not being there)
- Expanded get location term so it worked with job detail and search results pages (an issue identified by tests)
- Updated the Page cache key so it doesn't fail when there's unicode in the query string or path (problem identified by testing) 
- Removed handle_postprocessing since it's completely unused and didn't do anything anyway
- A boatload of pep8 to figure out what's going on with mystery test failures

Changes made to help with testing that aren't in the tests file:
- Removed duplicate javascript entries for ColumnBlock and Rows (this was originally handled by Page, but again, doing it on this level made testing easier)
- Added context, required_js to Rows

aspidites:
Thanks. Will review again once jenkins is behaving.

aspidites:
Everything seems in order, though I wouldn't begin to know how to test this thing locally. That said, you may want to review the code for `myjobs.tests.test_views.TestClient`, as I've updated how it works a bit which would help you remove a lot of the client boilerplate you have in your tests (take a look [here](https://github.com/DirectEmployers/MyJobs/blob/reports-cleanup/myreports/tests/test_views.py#L57) to see what I mean).

PR #1358, opened 2015-03-23
not

almathew:


seo/templatetags/job_setup.py
```diff
@@ -66,7 +66,7 @@ def create_arranged_jobs(request, featured_jobs, default_jobs, site_config):
             'class': 'default_jobListing direct_hiddenOption'
         })
 
-    if not jobs or jobs[0]['jobs'] or not jobs[1]['jobs']:
+    if not jobs or (not jobs[0]['jobs'] and not jobs[1]['jobs']):
```
aspidites: Probably a good indication that we should be documenting intent? In fact, I see exactly two comments in the entire file.

aspidites: Update: specifically, the fact that we had deployment issues directly related to misunderstanding boolean logic seems to indicate that we should document what the logic should be asserting. Especially if there is a chance this will ever be updated in the future.

PR #1361, opened 2015-03-23
Segment new partner form [PD-1163]

tdphillips:
2015-03-23-164455_648x821_scrot

All tests pass.

### DIFF
--- a/mypartners/forms.py
+++ b/mypartners/forms.py
@@ -181,7 +181,6 @@ class NewPartnerForm(NormalizedModelForm):
             self.fields['notes'] = notes
 
         for field in self.fields.itervalues():
-            field.label = "Primary Contact " + field.label
             # primary contact information isn't required to create a partner
             field.required = False
         model_fields = OrderedDict(self.fields)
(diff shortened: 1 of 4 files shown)


tdphillips:
@fezick I don't know if you wanted all of these labels to be changed, but I like how it turned out.

Everything in the first section would normally be prefixed with "Partner" while everything in the second would be prefixed with "Primary Contact."

aspidites:
Being pedantic, but they are legitimate concerns:
- Technically, locations are optional, and you may have multiple. As such, I wonder if there should be a "Location" section
- Regardless, I think the partner information should be marked with a legend

I'm conflicted as to whether I should be impressed that you were able to add what needed to be added without rewriting the form, or sad that you weren't motivated to rewrite the form and do things "correctly". 

Nice work overall.

mypartners/forms.py
```diff
@@ -273,6 +272,18 @@ def save(self, commit=True):
...
         return None
 
+    def get_field_sets(self):
+        """
+        NewPartnerForm is a combination Partner and Contact form. As
+        self.fields has already been turned into an OrderedDict in __init__,
+        we can easily segment our form into fieldsets.
+        """
+        sections = self.fields.keys()[:4], self.fields.keys()[4:]
+        field_sets = [
```
aspidites: TIL a form's fields can be accessed using subscript/`dict`/`__getitem__` syntax.

aspidites:
I suspect that "Source" is only obvious to the individual who requested the feature. I, as a new user (who probably didn't pay attention to the live training session), would have no clue what source meant. Perhaps help text?

tdphillips:
You can't have multiple locations on a contact that I can see. It's also entirely optional.

tdphillips:
@fezick Suggestions?

aspidites:
Spoke with @fezick  directly, who addressed all of my questions. This is good to go, and I've checked it locally.

fezick:
Yup, sorry. I didn't see this until @aspidites brought it my attention.

PR #1363, opened 2015-03-24
fix staging pages

almathew:


aspidites:
Fix staging pages....what's broken?

seo/views/search_views.py
```diff
@@ -97,6 +97,7 @@ def post(self, request, *args, **kwargs):
         return super(FallbackBlockView, self).post(request, *args, **kwargs)
 
     def set_page(self, request):
```
aspidites: This entire method (set_page) feels wrong. Nested exceptions, which aren't consistent with each other. In some cases, page is set to None, but in others we just pass, so page doesn't even exist. How is this method meant to operate in plain English?

almathew:
The problem: If a user is logged in and staff they should get the staging page (if there is one). Instead it's being overwritten by the section that attempts to set a production page.

This tries to:
1. Set a staging page if the user is logged in and staff.
2. If a staging page wasn't set, try to set a production page.
3. If a production page can't be found, try to fall back to a default page (this would be important should we decide to switch all of the network sites over to Pages)
4. If none of those can be done, set the page to none. The `handle_request` function then appropriately falls back to the existing view if a page can't be set.

This is similar logic to how we set configurations, except the configurations don't fall back because we always know that there's one for sites__pk=1 (since it's in the default fixtures).

aspidites:
So assuming a staff user, page should be:
- staging or production or default or None

(Assuming boolean logic where the left alternative is considered before the right one). Is this correct, @almathew?

almathew:
Correct. In that order. And if one is set, it should not be overwritten by one of the later ones.

aspidites:
Pulled the branch locally. Less concerned about page not being set during `IndexError` now since I see that page is being initialized as None to begin with.

aspidites:
@almathew, I think [this gist](https://gist.github.com/aspidites/739ababef910e8413700) accomplishes the same thing without the nested try/excepts. It looks longer than it really is because I over-explained things.

aspidites:
The key points are to take advantage of the ORM's features (first() returning a None on failure, and incrementally building up the correct query to begin with, rather than going the trial/error route.

aspidites:
@almathew indentation was wrong in my previous gist -- updated. Also removed comments, as they were a bit noisy.

PR #1366, opened 2015-03-25
Normalized context and added domain to the cache key

almathew:


myblocks/context_tools.py
```diff
@@ -72,12 +72,6 @@ def get_breadbox(request):
 
 
 @Memoized
-def get_count_heading(request):
```
almathew: This was a duplicate of get_results_heading

myblocks/models.py
```diff
@@ -558,6 +558,12 @@ def render(self, request, **kwargs):
         return rendered_template
 
     def render_cache_prefix(self, request):
+        domain = ''
+        if request.user.is_authenticated() and request.user.is_staff:
```
aspidites: Doesn't `get()` return None if the value isn't  found? As such, there is a possibility that line 564 won't be executed. 

I think you'll either need to pass a default  value of '', or change your if to `if not domain`.

almathew:
Added comments while addressing @aspidites's concern

PR #1379, opened 2015-03-31
Mess with the interaction between the More and back buttons [PD-322]

tdphillips:
Mainly opening for tests. There are also some issues with a weird merge.


aspidites:
Easiest way to test this locally?

aspidites:
I assume you know your merge files (*.orig) are still being committed?

tdphillips:
I'm not convinced this is ready to be merged, but you can test if you want. The .orig files exist on staging, which I built this off of.

The easiest way to test:
* Point to staging database and solr
* Override your chosen domain in /etc/hosts
* Do a search and click more a bunch
* View a job
* Click back
* ???
* Profit (Also you should have a bunch of jobs loaded instead of only the default number)

aspidites:
I guess the follow question in that case would be why is this being built off of staging instead of quality control?

tdphillips:
Because staging solr was in the state I wanted and this was pre-qc->staging deployment.

aspidites:
I see. Two things then:
1. Can you git rm --cached the orig files. At least then, when this hits staging it'll remove the orig files without us having to bother on staging.
2. Can you sync with qc at some point:
2015-03-31-145635_377x39_scrot
Doing so should get rid of extra line noise in the diff, even if this isn't ready to be merged.

tdphillips:
I was planning on just redoing the commit on a new branch. I don't want to take a chance on messing things up further.

static/pager.162-31.js
```diff
@@ -319,10 +319,12 @@ $(document).ready(function(){
 		var parent = $(this).parent();
 		var num_items = parseInt(parent.attr('data-num-items'));
 		var offset = parseInt(parent.attr('offset'));
+        console.log(offset);
```
aspidites: Is this log supposed to be here?

tdphillips: Nope.

PR #1380, opened 2015-03-31
Make back button work from job detail view [PD-322]

tdphillips:
The easiest way to test:

* Point to staging database and solr
* Override your chosen domain in /etc/hosts
* Do a search and click more a bunch
* View a job
* Click back
* ???
* Profit (Also you should have a bunch of jobs loaded instead of only the default number)


tdphillips:
Feel free to rip this apart. Browsers that lack history.replaceState are no longer supported.

tdphillips:
Someone who is better at seo than I should take a look at what I did. I saw no way of resolving my ticket without discarding a bit of code and I'm not sure if that's acceptable.

aspidites:
Results are re-ordered on back. Before viewing a job:
2015-04-01-124934_1916x1056_scrot
After clicking back:
2015-04-01-124957_1916x1056_scrot

almathew:
We use the default ordering unless you specifically order by date. That shouldn't change.

aspidites:
@almathew  I didn't specify anything. I simply clicked the back button. Did no sorting at all.

tdphillips:
Destroying all of my work.

PR #1381, opened 2015-04-01
PRM Changelog

almathew:
Adds the delta field to ContactLogEntry. This is a json dump of the actual change that was made. The content for the delta is as human readable as I could possibly get it.


mypartners/helpers.py
```diff
@@ -125,9 +187,11 @@ def get_change_message(form):
     change_message = []
     if not form:
         return ''
+
     if form.changed_data:
         change_message = (ugettext('Changed %s.') %
                           get_text_list(form.changed_data, ugettext('and')))
+    get_form_delta(form)
```
tdphillips: I could be mistaken, but it looks like get_form_delta doesn't mutate the form. If that's the case, this line doesn't do anything.

almathew: It was left over from the remnants of how I was going to do it originally. I've removed it.

aspidites:
Yay, no description

mypartners/forms.py
```diff
@@ -307,14 +308,17 @@ def __init__(self, *args, **kwargs):
...
                 choices.insert(0, ('', "No Primary Contact"))
             else:
                 choices.append(('', "No Primary Contact"))
         else:
             choices.insert(0, ('', "No Primary Contact"))
+
         self.fields['primary_contact'] = forms.ChoiceField(
-            label="Primary Contact", required=False, initial=choices[0][0],
+            label="Primary Contact", required=False,
+            initial=unicode(choices[0][0]),
```
aspidites: Since we define choices, is there any reason to not just make sure that choices are unicode to begin with? That is, instead of ('', "No Primary Contact")`, why not `(u'', u'No Primary Contact')` and forgo the explicit conversion? Also, any reason for mixed quotations?

mypartners/helpers.py
```diff
@@ -117,6 +122,63 @@ def log_change(obj, form, user, partner, contact_identifier,
     )
 
 
+def get_form_delta(form):
+    """
+    Determines the changes made by a newly submitted form. Provided delta
+    values should be at least semi-human readable.
+
+    :param form: The form that is being checked for changes.
+    :return: A dictionary of changes. The format is:
```
aspidites: What's the point of the outer `dict`, given it doesn't have any more keys?

mypartners/helpers.py
```diff
@@ -117,6 +122,63 @@ def log_change(obj, form, user, partner, contact_identifier,
...
+            # form.initial or form.fields[field].initial. Django
+            # favors form.initial in their code, so this code prefers
+            # form.initial first as well.
+            initial_val = form.initial.get(field, form.fields[field].initial)
+            initial_val = form.fields[field].to_python(initial_val)
+
+            new_val = form.data.get(field, '')
+            new_val = form.fields[field].to_python(new_val)
+
+            if type(form.fields[field]) == MultipleFileField:
```
aspidites: `isinstance` > `type`

mypartners/forms.py
almathew: They're mostly longs (the ids). No Primary Contact would be one of the few cases where the casting isn't necessary.

mypartners/helpers.py
```diff
@@ -117,6 +122,63 @@ def log_change(obj, form, user, partner, contact_identifier,
...
+                # for all the added attachments.
+                if type(form.files) == MultiValueDict:
+                    initial_val = None
+                    new_val = [f.name for f in form.files.getlist(field)]
+
+            elif hasattr(form.fields[field], 'choices'):
+                # Coerce the keys to unicode, since the data values
+                # we will be getting back should be unicode.
+                choices = form.fields[field].choices
+                choices_dict = {unicode(x[0]): x[1] for x in choices}
```
aspidites: Same comment as above. We have control over choices, so no need to coerce to unicode when we can make sure it's unicode to begin with.

mypartners/helpers.py
almathew: I don't think I understand your question. It's a dictionary of dictionaries. `form_field_name` is just a stand in for any form field that has changed.

mypartners/helpers.py
```diff
@@ -117,6 +122,63 @@ def log_change(obj, form, user, partner, contact_identifier,
...
+
+                initial_val = choices_dict.get(initial_val, initial_val)
+                new_val = choices_dict.get(new_val, new_val)
+
+            if ((initial_val != new_val)
```
aspidites: I assume this is to be read as "if initial_val and new_val are different and at least one of them is truthy" ?

Also, the `and` and `or` imply `bool`: `if (inital_val != new_val) and initial_val or new_val:`

mypartners/forms.py
aspidites: Oh, ok.

mypartners/helpers.py
aspidites: Ah. It wasn't apparent to me that `form_field_name` was a placeholder value. Now that I do, I see where the outer dict is necessary.

PR #1384, opened 2015-04-01
[PD-1212] Allow fields to be specified for a report

aspidites:
Check the ticket for a full description. In effect, this allows `values` to be specified, which allows the queryset returned by the API to have only those columns necessary for the report and/or view using them. For instance, the partner list in the current implementation of myreports only relies on partner name and contact record count, so this ticket trims the fat, so to speak, making the resulting response 1/3 the size of what it was before this was implemented. 

A bunch of crazy stuff happened in implementing this, so while these things are documented, I've also added notes throughout the PR to explain the rational for such quirky decisions. Hopefully I've missed something obvious, and one of you has a better alternative to the problems I faced in writing this.

### DIFF
--- a/myreports/helpers.py
+++ b/myreports/helpers.py
@@ -31,22 +31,27 @@ def humanize(records):
 
     for record in records:
         # make tag lists look pretty
-        record['tags'] = ', '.join(record['tags'])
+        if 'tags' in record:
+            record['tags'] = ', '.join(record['tags'])
         # get rid of pks
         record.pop('pk', None)
         # human readable contact types
-        record['contact_type'] = CONTACT_TYPES[record['contact_type']]
+        if 'contact_type' in record:
+            record['contact_type'] = CONTACT_TYPES[record['contact_type']]
         # strip html and extra whitespace from notes
-        record['notes'] = parser.unescape('\n'.join(
-            ' '.join(line.split())
-            for line in record['notes'].split('\n') if line))
-        # second pass to take care of extra new lines
-        record['notes'] = '\n'.join(
-            filter(bool, record['notes'].split('\n\n')))
+        if 'notes' in record:
+            record['notes'] = parser.unescape('\n'.join(
+                ' '.join(line.split())
+                for line in record['notes'].split('\n') if line))
+            # second pass to take care of extra new lines
+            record['notes'] = '\n'.join(
+                filter(bool, record['notes'].split('\n\n')))
 
         # get rid of nones
-        record['created_by'] = record['created_by'] or ''
-        record['length'] = record['length'] or ''
+        if 'created_by' in record:
+            record['created_by'] = record['created_by'] or ''
+        if 'length' in record:
+            record['length'] = record['length'] or ''
 
     return records
 
--- a/myreports/helpers.py
+++ b/myreports/helpers.py
@@ -117,18 +117,18 @@ def serialize(fmt, data, counts=None, values=None):
         if values:
             records = []
             haystack = []
+
             for record in data:
-                needle = [record[value] for value in values]
+                needle = [record[value] for value in values] or record['pk']
+
                 if needle not in haystack:
                     haystack.append(needle)
-                    records.append(record)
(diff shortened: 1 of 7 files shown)


myreports/helpers.py
```diff
@@ -33,22 +31,27 @@ def humanize(records):
 
     for record in records:
```
aspidites: It is not only possible that contact record isn't the model being traversed, but that not all columns are being used in that model. I have plans to refactor this method (and perhaps even obsolete it), but wanted to simply get things working this iteration.

myreports/helpers.py
```diff
@@ -77,8 +80,11 @@ def parse_params(querydict):
     return params
 
 
-# TODO: Find a better way to handle counts
-def serialize(fmt, data, counts=None):
+# TODO:
+#   * find a better way to handle counts
+#   * do something other than isinstance checks (duck typing anyone?)
+
+def serialize(fmt, data, counts=None, values=None):
```
aspidites: Manually threading values through so that I can get the natural keys for foreign keys. If I simply ran `values()` before or after, such keys wouldn't be translated, and thus we'd have integers for those values instead.

myreports/helpers.py
```diff
@@ -97,15 +103,27 @@ def serialize(fmt, data, counts=None):
...
-                    'python', data, use_natural_keys=True)]
+                    'python', data, use_natural_keys=True, fields=values)]
 
         if counts:
             data = [dict({'count': counts[record['pk']]}, **record)
                     for record in data]
 
+        # Cant' use a ValueQuerSet for serialize, which means we lose the
+        # ability to use distinct. As such, we fake it by doing so manually
+        records = []
```
aspidites: Unfortunately, because I don't run `values` before hand (`ValuesQuerySet`s aren't serializable using django's serialize function), I don't have distinct values, so I do so manually here, creating a unique key (needle) from the  values passed in and only keeping one instance of each. I couldn't simply convert the dicts into a set because of the fact that pk's are included, which are by their nature unique.

myreports/models.py
```diff
@@ -31,5 +33,22 @@ def python(self):
 
     @property
     def queryset(self):
-        pks = [record['pk'] for record in self.python]
-        return get_model(self.app, self.model).objects.filter(pk__in=pks)
+        model = get_model(self.app, self.model)
```
aspidites: Simply returning a `ValuesQuerySet` doesn't have the attributes we need (for instance, for ContactRecord, emails, referrals, etc), which renders them useless as a return value. I can't simply query on values because they aren't distinct, so I would get duplicates again. I can't pass a column to distinct() since we aren't using postgresql, and I can't use values for the reason listed above (and the fact that it excludes all other columns, which is not the behavior we want for this use case). Ideally, I'd be able to do a `SELECT * from {{ model } ORDER BY ( {{ values }})`, but I haven't found a way to do so using Django's ORM. There is `raw`, but that doesn't seem to be chainable, and I'd rather not have to do SQL mangling to get the correct query for the correct model.

myreports/views.py
```diff
@@ -92,33 +92,17 @@ def view_records(request, app, model):
         # remove non-query related params
         params.pop('csrfmiddlewaretoken', None)
         count = params.pop('count', None)
+        values = params.pop('values', [])
 
-        # TODO: REMOVE THIS ---------------------------------------------------
```
aspidites: This was a hack used to get contact_names for the contact list in report generation. Now that the api natively supports values, this is unnecessary.

static/reporting.162-26.js
```diff
@@ -488,7 +488,9 @@ List.prototype.filter = function(filter) {
   "use strict";
   var url = location.protocol + "//" + location.host, // https://secure.my.jobs
       data = {},
-      list = this;
+      list = this,
+      name = "name",
```
aspidites: Used to dynamically choose the right property. For partners, name is name, but for contact records, name is contact_name, for instance.

PR #1385, opened 2015-04-01
Added is_site_family boolean to site tags to allow for site families [PD-1155]

almathew:
Migration 0014 was a little messed up, so I had to fix that as well.


aspidites:
Looks good other than that pesky duplicate transaction thing. I had to do the same in my other branch.

aspidites:
Seems fine. Are we concerned about the implications of making a site tag a site family, tagging sites as such, then making that tag a regular one again? That is, what does it mean operationally to convert a site tag to a site family and back again? Anything? Nothing?

almathew:
Demoting via checkbox is fine. It will be the fastest way to handle the people who don't renew and then renew three weeks later.

tdphillips:
This doesn't differentiate which site is the parent, does it?

almathew:
"Groups" were never going to be a good way to do that. @JLMcLaugh and I were discussing namespacing which would be the best way to denote the parent. Unfortunately this is something that I can explain what my intentions were but until it has an actual use case it will be a complete guess as to whether or not it works.

I'm open to other suggestions for how to do this, but the only other alternative I see is adding another model/admin, which was something that we were trying to avoid.

almathew:
Alternatively, I can add a relationship with SeoSite that will allow you to explicitly denote parent without any fancy namespacing. Because everything needs a relationship with SeoSite.

aspidites:
Or you could do something horrible and add yet another field to this model "is_parent".

almathew:
Now with more foreign key.

PR #1387, opened 2015-04-01
Pager for blocks

almathew:


almathew:
I'm going to leave this open for a little while so I can bother people (specifically @galitskyd and @JLMcLaugh) about exactly what I've just done to pager and whether or not it's a good idea.

PR #1388, opened 2015-04-01
Improve companyuser assignment, PD-1200

fezick:


seo/models.py
```diff
@@ -620,7 +620,11 @@ class Company(models.Model):
 
     """
     def __unicode__(self):
-        return "%s" % self.name
+        if self.company_user_count == 0:
+            return "%s (%s users) **Might be a duplicate**" % (
```
tdphillips: If companies that have errantly been duplicated also have prm access (field on Company), could we check that field and potentially reduce false positives?

fezick: Probably, but that is dependent on the field being set properly. Also, company lists are retrieved for  than just the autocomplete on companyuser (seo.seosite, for example). I didn't want to do anything that fundamentally changed how company worked, so I opted for visual cues so that member services would be able to differentiate them. That said, I am open to any suggestions for improving this process.

fezick:
I'll check why this is failing tomorrow. My guess is that the company __init__ change is breaking a string comparison.

almathew:
Tests are failing because of an issue with the Authorize.net API. I don't think there's anything we can really do about that.

fezick:
In that case, I shall leave it as is.

templates/admin/seo/companyuser/change_form.html
```diff
@@ -0,0 +1,90 @@
...
+     {% if has_change_permission %}<a href="../">{{ opts.verbose_name_plural|capfirst }}</a>{% else %}{{ opts.verbose_name_plural|capfirst }}{% endif %} &rsaquo;
+     {% if add %}{% trans "Add" %} {{ opts.verbose_name }}{% else %}{{ original|truncatewords:"18" }}{% endif %}
+</div>
+{% endif %}{% endblock %}
+
+{% block content %}<div id="content-main">
+<div id='checkCompanyWarning'>
+<h2>Note</h2>
+Please verify that the company you are assigning this user to is the correct company. 
+Companies with zero users are usually dupicates unless this is the first user for
```
tdphillips: dupicates

seo/models.py
```diff
@@ -655,7 +659,19 @@ def associated_jobs(self):
...
+        Counts how many users are mapped to this company. This is useful for
+        determining which company to map companyusers to when two company
+        instances have very similar names.
+        
+        It is treated as a property of the model.
+        
+        """
+        return len(CompanyUser.objects.filter(company=self))
+        
+    company_user_count = property(get_company_user_count)
```
aspidites: Few things:
1. If this is going to be a property, it's probably better to not have the `get_` prefix.
2. `@property` above the method definition is the canonical way to create a read only property, rather than running the decorator manually
3. instead of `len(CompanyUser.objects.filter(company=self))`, you can do `self.companyuser_set.count()`, which I suspect will perform better.

templates/admin/seo/companyuser/change_form.html
```diff
@@ -0,0 +1,90 @@
+{% extends "admin/base_site.html" %}
+{% load i18n admin_modify staticfiles %}
+{% load url from future %}
```
aspidites: Load url from future?

templates/admin/seo/companyuser/change_form.html
fezick: what is duplicated?

tdphillips: Sorry, "duplicates" is spelled "dupicates"

templates/admin/seo/companyuser/change_form.html
fezick: I copied the change form template directly from django.contrib. I didn't change anything that wasn't directly related to my display changes.

aspidites: Ah. Is the whole template copied verbatim, then?

PR #1392, opened 2015-04-02
Pager for blocks

almathew:
Makes pager usable for blocks. This took a few large changes:
* Removal of a ton of unused code.
* `ajax_get_jobs()` now allows for the first SearchResult block on a SearchResults page to render the response.
* The more/less button clicks on filters are now contained by id within the parent div for all the filters in that block. This means you can have multiple SearchFilter blocks per Page now. 
* The more button focus works assuming a specific structure. It's not ideal that the structure is required, but the structure is open enough that the types of changes Digital Strategy will want to make can still be made.
* I'd also like to think that pager is more readable now, but maybe it's just me.

I'm working on some additional google docs documentation that will explain the requirements that pager has, but those are actually going to be more relaxed then they are now, so don't let the documentation hold you up on this.


aspidites:
This makes me nervous, in particular because we don't test any of our javascript, so I have no way of being assured that what's been deleted doesn't break something. The fact that this messes with microsites makes me that much more unsure. 

The python seemed fine at a cursory glance.

aspidites:
Well, it still has to go through staging, so hopefully if something does break it'll break in obvious ways. Maybe ping Jim once this goes out to make sure, @almathew ?

PR #1393, opened 2015-04-03
[PD-1210] Myreports navigation

aspidites:
From what I've gathered by playing with this branch over the last couple of days (and basing my changes on top of it), this branch seems to properly implement browser navigation (forward and back) between the overview, generation, and view pages. The one caveat is IE 8 and 9, which don't properly support the history API. 

I'd strongly suggest everyone pull this branch and play with it, paying close attention to the console.


tdphillips:
If I initiate report creation, press the back button to go back to selecting report type, and repeat several times before finally creating a report, one report is created for each aborted attempt in addition to the final, intended report.

The view, clone, and download icons on the report archive do nothing. @aspidites informed me that this was a conscious decision and the archive was an afterthought, but it feels like they should do something.

tdphillips:
If you name a report the same as an already-existing report, the existing report's json file gets replaced with the new one.

PR #1394, opened 2015-04-03
Decrease number of configuration saves on cache clear for custom facets [ST-991]

almathew:
This reduces the number of saves on configuration by making the actual number of configurations cleared distinct.

This probably is not a long term fix for the problem - save (and therefore this signal) is called multiple times per admin form save. Therefore, adding more configurations associated with a facet will still increase the amount of total configuration saves by more than 1. The worse offender for this is in `change_view` for `RowPermissionAdmin`, which is used by most of the forms on admin. I'm not ready to mess with that just to decrease the time a little bit more. This is already going from ~500 saves per Providence CustomFacet to ~60 saves, which is enough of an improvement for now.


seo/models.py
```diff
@@ -356,6 +355,27 @@ def _make_qs(self, field, params):
...
+    # circular import condition.
+    obj = kwargs['instance']
+
+    if not obj.seosite_set.all():
+        return
+
+    configs = Configuration.objects.filter(seosite__facets=obj)
+
+    for config in configs:
+        config.save()
```
almathew: Original code was in `seo/urls/search_urls.py`

aspidites: I think `object.seosite_set.exists()` is both more contextually clear, and faster:
```python
In [4]: timeit("from seo.models import Configuration; bool(Configuration.objects.first().seosite_set.all())", number=1000)
Out[4]: 2.48301100730896

In [5]: timeit("from seo.models import Configuration; Configuration.objects.first().seosite_set.exists()", number=1000)
Out[5]: 1.855492115020752
```

mklauber:
:thumbsup:

PR #1399, opened 2015-04-03
Fixed possible race condition between workers.

mklauber:


import_jobs.py
```diff
@@ -14,6 +14,7 @@
 from lxml import etree
 from django.conf import settings
 from django.db import IntegrityError
+from billiard import current_process
```
tdphillips: Is this a dependency of something already installed? It's not in requirements.

JLMcLaugh: It's included with Celery, a fork of Python multiprocessing.

aspidites: What does the fork offer that the standard library doesn't?

import_jobs.py
```diff
@@ -110,11 +111,20 @@ def get_jobs_from_zipfile(zipfileobject, guid):
     :return: [lxml.eTree, lxml.eTree,...]"""
     logger.debug("Getting current Jobs for guid: %s", guid)
 
+
+    # Get current worker process id, to prevent race conditions.
+    try:
+        p = current_process()
+        process_id =  p.index
+    except:
```
aspidites: This seems like a bad idea. Silent failures? I suspect we should be more specific about what we expect to fail here.

import_jobs.py
```diff
@@ -110,11 +111,20 @@ def get_jobs_from_zipfile(zipfileobject, guid):
...
+        process_id = 0
+
+
     # Delete any existing data and use the guid to create a unique folder.
-    directory = "/tmp/%s" % guid
+    directory = "/tmp/%s/%s" % (process_id, guid)
     if os.path.exists(directory):
         shutil.rmtree(directory)
-    os.mkdir(directory)
+    os.makedirs(directory)
```
aspidites: Are we worried about the fact that this could fail? (eg, permission errors).

import_jobs.py
mklauber: Frankly, we don't care.  Celery requires it, so it's not like we can avoid it.  But to answer the question, I think there's some performance improvements, and also the ability to name and track subprocesses.

import_jobs.py
mklauber: The directory **should** always be a subdir in `/tmp` so it should be accessible.  But if it's not, there's nothing we can do about it, so letting the task fail, and bubble up to a celery retry is the desired behavior.  That'll kick it off again on another server, likely letting the task finish if this is a localized server problem.  Otherwise, we will still have a log of the failures, and can retry them once the promblem is resolved.

import_jobs.py
mklauber: This is a special case for handling when we're not running in multiprocess mode.

mklauber:
Finally got all tests passing.

import_jobs.py
aspidites: Such a case doesn't throw a specific error? I'd assume IOError?

import_jobs.py
```diff
@@ -110,11 +111,20 @@ def get_jobs_from_zipfile(zipfileobject, guid):
...
+        process_id =  p.index
+    except:
+        process_id = 0
+
+
     # Delete any existing data and use the guid to create a unique folder.
-    directory = "/tmp/%s" % guid
+    directory = "/tmp/%s/%s" % (process_id, guid)
     if os.path.exists(directory):
         shutil.rmtree(directory)
```
JLMcLaugh: I know this was a problem before this pull request, but we need to make sure this is a clean guid before running a command like rmtree on it.

import_jobs.py
```diff
@@ -110,11 +111,21 @@ def get_jobs_from_zipfile(zipfileobject, guid):
...
+    try:
+        p = current_process()
+        process_id =  p.index
+    except AttributeError:
+        process_id = 0
+
+
     # Delete any existing data and use the guid to create a unique folder.
-    directory = "/tmp/%s" % guid
+    assert "/" not in guid, "GUID: %s should not contain slashes." % guid
```
JLMcLaugh: I don't think this is enough. At a minimum they could still clear out /tmp. Try something like os.path.commonprefix(['/tmp/MY_PID', os.path.abspath(directory)] == '/tmp/MY_PID'

mklauber:
Woo!  Thanks @tdphillips

PR #1403, opened 2015-04-06
Case insensative checks for partner and contact

aspidites:
Disabled forms when partner name or contact name is empty or a duplicate. Also made sure check was case insensitive.


mklauber:
:thumbsup:

PR #1404, opened 2015-04-07
Don't allow prm users to activate partner saved searches if the recipient unsubscribes [ST-1047]

tdphillips:
The necessary tests were less verbose than I thought they would be. If you want more proof of uneditability than the field being disabled, I can come up with something.


tdphillips:
Adding some comments.

tdphillips:
Added some comments.

tdphillips:
I'm working on another ticket at the moment and I don't want to introduce another heisenbug, so let's not merge this until I find out what's going on.

tdphillips:
I ran that test case 100 times and all runs passed. Let's see what happens.

PR #1405, opened 2015-04-07
Update requirements.txt

mklauber:
Fixed jira-python being renamed.  This was merged to QC, but we'll need it in staging for the next deployment.


aspidites:
Any reason to not use the latest version ([0.46])(https://pypi.python.org/pypi/jira)?

PR #1416, opened 2015-04-08
[PD-1235] Rudimentary UI for Column Selection Before Download

aspidites:
The draggable items could still use some styling, but I wanted to get the basic functionality in there first. 
2015-04-08-155910_1916x1056_scrot

- while you can visually re-order columns, they don't persist between downloads (back-end change required which is part of a different ticket
- column selection *is* persisted between downloads
- pop state implemented (though @galitskyd  has informed me that other changes are required for IE, which we plan to tackle separately, as it requires some of his changes).

As a result of these requested changes, it should be noted that:
- the json saved to disk is larger, as I was asked not to prune unused data (deselected columns) from the results
- as such, that also means that downloading of complex reports takes longer
- deselecting columns does mean less data needs to be transferred and thus has a positive impact on resulting CSV size.

I'm intentionally ignoring the possibility of other export formats for now, as I have a lot of other things on my plate.

### DIFF
--- a/myreports/helpers.py
+++ b/myreports/helpers.py
@@ -112,23 +112,26 @@ def serialize(fmt, data, counts=None, values=None):
             data = [dict({'count': counts[record['pk']]}, **record)
                     for record in data]
 
-        # Cant' use a ValueQuerSet for serialize, which means we lose the
-        # ability to use distinct. As such, we fake it by doing so manually
-        if values:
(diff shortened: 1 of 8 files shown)


aspidites:
Closing since we decided to do something slightly different on the UI. My other branch has this and some other changes anyway.

PR #1420, opened 2015-04-10
Track and replay clicks on the more button [PD-322]

tdphillips:
Waiting on input from the product owner on whether I need to do the same for blocks. In the meantime, let's see if we actually want to go through with this.


static/pager.160-29.js
```diff
@@ -253,11 +255,27 @@ $(document).ready(function(){
...
+                    if (!isNaN(total_clicks) && num_clicks < total_clicks) {
+                        $('#button_moreJobs').click();
+                    }
+                }
+            }
+        });
+    });
+
+    if (!isNaN(total_clicks)) {
+        $('#button_moreJobs').trigger('click');
```
tdphillips: I know this could be `.click()` but I was fighting with some issues and changed this in an attempt to fix them. It didn't and I just haven't changed it back as it's not hurting anything.

aspidites:
Confirmed that refreshing a page remembers how many times I clicked the more button. I assume that any history concerns are outside of the scope of this ticket? That is, is the end result of this ticket meant to address the fact that currently in production, refreshing a page loses how many results are on the more page?

tdphillips:
I think so, but my understanding has never survived impact with product owners so YMMV.

aspidites:
OK. I'd consider this done then. Don't quite understand the hash voodoo magic, but the page functions as expected.

aspidites:
Don't we need to bump pager's version number? Or is what ashley changed not on production yet/won't hit production before this?

tdphillips:
I forgot.

PR #1421, opened 2015-04-10
[PD-1213][PD-1235][PD-1243]  Reporting Download Screen

aspidites:
# Overview
This implements the new download configuration screen. Instead of immediately downloading a CSV file, the user is taken to a screen which affords them the ability to change the column and sort order.

# Screenshot
2015-04-13-072306_1916x1056_scrot

# Features
- columns may be reordered by dragging and dropping them
- columns may be included or excluded by clicking on the checkbox or div containing the checkbox
- sort by can be specified for the resulting output file
- after export, if a user re-visits that same report, the column ordering is persisted

# Remaining Tasks
While this is rather functional already, there are a couple of things that I would like to do (either before this is merged, or as a follow-on ticket.
- [x] currently order_by isn't persisted in the database, which also means that they aren't persisted in the interface
- [x] currently, the backend only does the order-by on export, as that was what was required for this ticket, but I'd like to make that possible on the initial query so that, for example, the contact list can be sorted by name.
- [ ] Unit tests for the new functionality that is provided by the backend
- [x] ~~I'd really like to refactor some of the code. At the moment, things are done dirty, as we had deadlines to meet. As an example, I'm traversing the record set multiple times for certain transformations, when I could probably do so in a single pass after a bit of refactoring.~~ create a task to clean up some of the code which is functional, but could be cleaner

See [PD-1259](https://directemployersfoundation.atlassian.net/browse/PD-1259) for the ticket mentioned above


galitskyd:
The select all row doesn't line up how I thought we wanted.

Also, on Mac's dropdown cuts off 'descending'
screen shot 2015-04-10 at 8 42 07 am

aspidites:
I've addressed @galitskyd's complaints and updated the screenshot.

aspidites:
To determine whether this should be merged now or after unit tests are written (After all, how much stuff this sprint has been merged without new tests? Surely I'm not a special case.), I thought I'd note what currently isn't covered by tests:

- passing a values parameter to a view
    - determines which columns should appear in resulting output
    - also determines which order those columns should appear in the resulting output
- passing the order_by parameter to a view
    -  determines the sort order for the resulting data
- the new downloads view
    - renders a template which passes as context the columns to include, the sort order, and which 
      ones are enabled
- parse params/serialize
    - I probably just need to update existing test to ensure that stripping of whitespace and html works 
      as expected

PR #1422, opened 2015-04-10
[PD-1205] Fixed off by one on readable data

aspidites:
Because you know, I've not been programming for years in languages where array-like structures are 0-indexed...

### DIFF
--- a/static/reporting.163-8.js
+++ b/static/reporting.163-8.js
@@ -375,10 +375,11 @@ Report.prototype.readable_data = function() {
 
         if (key === "contact type") {
           value = $.map(checklists.contact_type, function(item, index) {
-            if (value.indexOf(item.value) > 0) {
+            if (value.indexOf(item.value) !== -1) {
               return item.label;
             }
           });
+          console.log(value);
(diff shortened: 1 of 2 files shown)


static/reporting.163-8.js
```diff
@@ -375,10 +375,11 @@ Report.prototype.readable_data = function() {
 
         if (key === "contact type") {
           value = $.map(checklists.contact_type, function(item, index) {
-            if (value.indexOf(item.value) > 0) {
+            if (value.indexOf(item.value) !== -1) {
               return item.label;
             }
           });
+          console.log(value);
```
galitskyd: console.log

almathew:
Do the version numbers of these static files need updated?

aspidites:
@almathew def.jobs probably, reporting, probably not.

aspidites:
Actually, i just removed a console log statement from def.ui.js so I really don't think it matters either if that's cached. I remember why I didn't bother updating the name now.

almathew:
Works for me.

PR #1434, opened 2015-04-14
[PD-1223] Reports Redirects on Failed AJAX Requests

aspidites:
- temporary work around, since apparently this is supposed to be default behavior


static/reporting.163-14.js
```diff
@@ -728,6 +728,12 @@ List.prototype.filter = function(filter) {
       updateShowModal();
     },
     error: function(e) {
+      // work-around; this should be handled by my.jobs.xxx-x.js:31, but isn't
+      // for some reason
+      if (e.status === 403) {
```
tdphillips: I don't know if you caught it in IRC, but the reason this might not be working anymore is that I'm unsure if we're using 403s anymore. I was under the impression that everything was now either a 404 or a 30x.

aspidites: Run server is definitely returning 403s. The code I point to is also supposed to be checking for 403s but doesn't ever get called (that is, the error handler isn't called, so 403 or not doesn't really matter).

tdphillips: Got it.

PR #1435, opened 2015-04-14
The beginnings of merging the api

mklauber:


api/decorators.py
```diff
@@ -0,0 +1,44 @@
+import json
+import urllib2
+
+from django.http import HttpResponse, HttpResponseForbidden
+
+from api.models import APIUser
+
+
+def sns_json_message(f):
```
mklauber: I think we already have this at https://github.com/DirectEmployers/MyJobs/blob/quality-control/seo/decorators.py#L134

api/models.py
```diff
@@ -1,7 +1,208 @@
...
+        u.is_active = True
+        u.is_superuser = True
+        u.gravatar = u.email
+        u.set_password(password)
+        u.save(using=self._db)
+        u.make_guid()
+        return u
+
+
+class User(AbstractBaseUser, PermissionsMixin):
```
mklauber: Probably need to remove this too.

api/tasks.py
```diff
@@ -0,0 +1,51 @@
+from celery import Celery
```
mklauber: Don't think we need this file.

api/views.py
```diff
@@ -1 +1,146 @@
...
+
+from api.decorators import authorize_user, sns_json_message
+from api.helpers import (get_cntl_query, get_rc_query,
+                         get_query, get_query_as_string,
+                         grouper)
+from api.xmlparse import JobFeed
+from tasks import task_update_solr
+
+
+@csrf_exempt
```
mklauber: Probably don't need this view.

api/models.py
almathew: We don't actually share a database on this one. There are a few approaches we can take to this. I'm curious if anyone has any suggestions first, though.

mklauber:
Just wanted to make notes on the things we need to remove early on here.

almathew:
I've addressed a number of @mklauber's points. There is still the point to consider of the not shared database this time around. I have an idea for how I planed to solve it, but I'm sure someone else has an opinion as well.

almathew:
Unfortunately this pull request is getting borderline un-reviewable, so I'm going to go ahead and start commenting on it now.

This pull request is for merging the usjobs-api project into the MyJobs project. There are two reasons we're doing this:
1. It means that solr is managed all in one place, which makes sense since there will be only one solr instance.
2. It makes deployment easier. This is particularly important because we don't deploy the api very often.

Besides the new solr fields (and all the copied and pasted code from the api project) the biggest new change is the introduction of a router to support multiple databases. You will need to add this new database (the key is api) to your settings file. I highly recommend pointing it to a different database (or file if you're using sqlite) than "default" to get the full experience, as there are some errors that you will only find if the databases are actually in different locations.

The api also comes with its own urls, just as myjobs and dseo do. There aren't really any other special settings.

api/migrations/0001_initial.py
```diff
@@ -0,0 +1,168 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db, dbs
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+db = dbs['api']
```
almathew: This and the `no_dry_run = True` a few lines down are an unfortunate hack.

Setting `db` here allows these migrations to be run on the api database and not the default database. 
The south implemenation of dry run for mysql is done as a bit of a hack - basically they run the migration, roll it back, and then "actually" run the migration. It looks like the dry run is done on the api database, the rollback (which is completely out of our control) is done on the default database, and then the actual migration is applied again to api. Since the api changes never got rolled back, during the second migration south freaks out and runs the backwards migration instead, resulting in hours of confused frustration. So instead `no_dry_run = True` means the migration gets run once and only once.

All migrations are recorded on the default database. This is both out of our control and a really good thing.

almathew:
This pull request should not be complete. I'm going to leave it as "do not merge" until solr has been migrated.

Keep in mind that most of this is not new code.

aspidites:
@almathew any way of getting a diff of what is actually new code, or noting explicitly what is new?

almathew:
Magic. I've  been noting most of it, but there were a few more recent changes that I probably should point out.

apache/django.wsgi
```diff
@@ -11,8 +11,13 @@ if PROJECT_DIR_PARENT not in sys.path:
     sys.path.append(PROJECT_DIR_PARENT)
 
 os.environ['CELERY_LOADER'] = 'django'
-os.environ['DJANGO_SETTINGS_MODULE'] = 'settings'
 
 import django.core.handlers.wsgi
-application = django.core.handlers.wsgi.WSGIHandler()
-application = newrelic.agent.wsgi_application()(application)
+
+def application(environ, start_response):
```
almathew: This is the only way I was able to get variables from the apache conf loaded by the wsgi file.

routers.py
```diff
@@ -0,0 +1,53 @@
+def is_api_model(model):
+    return model.__module__ == 'api.models'
+
+
+class APIRouter(object):
```
almathew: Router that correctly routes things from the api app to the api database.

seo/search_indexes.py
```diff
@@ -42,34 +42,30 @@ def __init__(self, *args, **kwargs):
...
+        # blank, the type gets overridden from string to text_en
+        self.trust_field_type = True
 
 
 class JobIndex(indexes.SearchIndex, indexes.Indexable):
     """
     All fields that you want stored will be put here, in django model form
     """
-    job_source_name = indexes.CharField()
+    all_locations = MultiValueSpecialField(field_type='location_char')
```
almathew: We added this and a few other fields (below) for the api to use.

seo/search_indexes.py
```diff
@@ -42,34 +42,30 @@ def __init__(self, *args, **kwargs):
...
-
-class MultiValueIntegerField(indexes.MultiValueField):
-        field_type = 'int'
+class MultiValueSpecialField(indexes.MultiValueField):
+    def __init__(self, *args, **kwargs):
+        self.field_type = kwargs.pop('field_type', 'text_en')
+        super(MultiValueSpecialField, self).__init__(*args, **kwargs)
+        # This attribute is checked in the search backend. If facet_for is
+        # blank, the type gets overridden from string to text_en
+        self.trust_field_type = True
```
almathew: I switched from the hacky `facet_for` which required a match to the model (which we didn't always have) to `trust_field_type`. Basically this is just saying 'I know what I'm talking about so use the field type I'm passing you rather than changing it to `text_en` like you do everything else` when creating the schema. The actual change that trusts the field type is in `search_backends.py`.

tasks.py
```diff
@@ -700,9 +700,9 @@ def task_update_solr(jsid, **kwargs):
 
 
 @task(name='tasks.etl_to_solr', ignore_result=True, send_error_emails=True)
-def task_etl_to_solr(guid, buid, name):
+def task_etl_to_solr(guid, buid, name, fc, staffing_firm, industries):
```
almathew: Chen added the new required fields for us.

transform.py
```diff
@@ -272,7 +275,8 @@ def hr_xml_to_json(xml, business_unit):
     # This has to be seo.joblisting, otherwise the jobs won't be included
     # in the search results.
     job['id'] = 'seo.joblisting.%s' % guid.replace('-', '')
-
+    job['all_locations'] = [zipcode, city, state, state_short,
+                                "%s, %s" % (city, state), country]
```
almathew: And we add the new required fields to solr. These are for the hrxml files. `xmlparse.py` also has similar changes for seoxml.

almathew:
We also renamed the settings files so they could be used by apache and added the new 'api' entry to databases, which I mentioned in a previous comment.

There's some updates to test data (adding the new fields) but other than that nothing particularly remarkable.

PR #1436, opened 2015-04-14
[PD-1225] Regenerate report contents on init if it is supposed to have results.

aspidites:
If a report has results, but the file associated with that result doesn't exist, I rerun the database query and store the results. 

Not sure if I should be checking that the location is writable in the first place, or if there is a canonical django way to do that. I assign contents to _results because it didn't make sense to have to re-read a file just to get at hte contents immediately after (via Report.python or Report.json)


aspidites:
Totally misunderstood what I was supposed to be doing.

PR #1437, opened 2015-04-14
[PD-1225] Regenerate report contents on init if it is supposed to have results.

aspidites:
- added regenerate method to report model which allows a models results to be regenerated (json file)
- if file can't be found, download icon changes to regenerate icon until pressed; once regenerated, it turns back into a download icon
- updated archive page (both this, and added access to the new download customization functionality)


almathew:
merge conflicts

PR #1440, opened 2015-04-14
backported files from pd-1082 to staging

aspidites:
Was able to back port changes in pd-1082 by simply checking out those files instead of cherry-picking.

### DIFF
--- a/templates/mypartners/edit_item.html
+++ b/templates/mypartners/edit_item.html
@@ -44,7 +44,7 @@
                 <div class="actions pull-right">
                     {% if contact %}<button id="delete" type="button" data-toggle="modal" data-target="#delete-confirm" class="btn">{% trans 'Delete' %}</button>{% endif %}
                     <a class="btn" href="{% if partner %}{% url 'partner_details' %}?partner={{ partner.id }}{% else %}{% url 'prm' %}{% endif %}">{% trans 'Cancel' %}</a>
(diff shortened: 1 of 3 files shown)


templates/mypartners/edit_item.html
```diff
@@ -153,75 +153,13 @@ <h2 class="top">Tips</h2>
...
-                isFormValid = $(this).val().length;
-            }
-
-            $("#item-save").prop("disabled", !isFormValid);
-
-        });
-    });
-
+    var partner = {{ partner.pk|default:"0" }},
+        partners = {{ partners|safe }},
```
almathew: Are we sure these are actually safe?

I feel like this is the most ridiculous question on the planet, but it has been a point of discussion in the past.

almathew: I just noticed that it was a copy/paste or whatever. Point still stands, but this may be a point for later discussion on security rather than a right now thing.

aspidites: Given:
- https://github.com/DirectEmployers/MyJobs/blob/pd-1082-staging/mypartners/views.py#L201
- https://github.com/DirectEmployers/MyJobs/blob/pd-1082-staging/mypartners/views.py#L211

Yes, I'm sure they're safe. The only real counter argument might  be that there isn't any guarantee that users won't use arbitrary javascript as a partner name when creating them, but that is both unlikely, and a security concern for partner creation, not this PR.

PR #1441, opened 2015-04-15
[PD-1276] Update robots.txt files

fezick:
Strips most of the robots.txt disallow rules from robots.txt for my.jobs and S3. Once merged, the S3 robots file will be pushed to the root of src-nlx-org folder. It previously wasn't in any version control.

### DIFF
--- a/cdn_files/src-nlx-org/robots.txt
+++ b/cdn_files/src-nlx-org/robots.txt
@@ -1,0 +1,3 @@
+User-agent: *
+Disallow: /
+
--- a/cdn_files/src-nlx-org/robots.txt
+++ b/cdn_files/src-nlx-org/robots.txt
@@ -1,3 +1,1 @@
 User-agent: *
-Disallow: /
-
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,16 +1,10 @@
-Sitemap: http://find.ibm.jobs/sitemap.xml
 User-agent: *
-Disallow: /login/
-Disallow: /logout/
 Disallow: /associate/
-Disallow: /done/
 Disallow: /error/
 Disallow: /admin/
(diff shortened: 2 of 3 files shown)


aspidites:
Just need to re-add the disallow for feed files that Seth requested a few months back.

aspidites:
Speaking of, if we are only concerned with allowing Google (or whoever else) to browse static files (js and css in particular), do we really want to remove the other disallow lines? In particular, removing the ones for login and logout seems suspect.

fezick:
It would if there was anywhere in the system that really used them. They only come into play in the case that post-a-job is enabled. A lot of these go back to defaults that were added with the initial commits - in Feb 2011. We can always add them back if we encounter issues. @haliscooper wanted to be aggressive in trimming the file.

aspidites:
Ah, ok. Sounds fair. Perhaps double-check Jira to make sure there aren't
explicitly requested lines asked for in (recent) tickets, similar to the
one I pointed out? Other than that, I'm :+1:. How do I run the google
mobile check to see if this does what it's supposed to?

fezick:
I doubt we can without an externally accessible URL.

aspidites:
Oh. Never mind, then. I'm done  being pedantic, I promise :-)

fezick:
Only other ticket was one where we added the AJAX calls. It was iffy then, and I think with some of the new functionality, removing it is better than leaving it.

aspidites:
I'm sold.

fezick:
Why is this marked do not merge?

aspidites:
Because at the time that I marked it, you hadn't added back the like for
ignoring feed files.

fezick:
OK, I'm going to remove it then.

PR #1447, opened 2015-04-17
[PD-1277] More efficient Record Traversal

aspidites:
# Optimizations
- rather than traverse records 5 times, I do so once
- use lxml (c based) instead of django (pure python) to strip HTML
- the presence of the downloads page deprecates the need to remove pk in `humanize`
- on the download page, I replaced divs with labels, which allowed me to reduce the number of bound events on that page

# Bug Fixes
While I was at it, I 
- fixed regression that caused field checkboxes on the downloads page to not be clickable
- protected the download page from regular GET requests, as doing so meant events werent bound and styling wasn't applied

# Tests
Before making each of the above optimization,  I created a unit test to ensure integrity of previous behavior:
- ensure order_by is respected
- ensure column order and presence is respected (python serialization only)
- ensured html is stripped from exported values


aspidites:
I based this on #1442, which is why it's marked as do not merge. Once that one is merged, the tag can be removed.

PR #1450, opened 2015-04-17
[PD-1282] Wrote the CheckBoxField and CheckListField prototypes

aspidites:
- wrote prototypes mentioned in the title
- fixed report.save so that all (and not just the first) fields validate and populate data to send to the server
- synced with quality control (which is why so many files seem to have been changed. All my work was exclusively in static/reporting.163-14.js though)


aspidites:
@galitskyd this should finally be ready to go. Apparently I was too tired. I thought I moved the label end tag to the right spot, but I guess I didn't. In any case, it's correct now. Pressing generate report fails because of the contact and partner lists (which I confirmed by first removing the checklist from the report, then removing those lists). 

tl;dr - This is safe to merge.

aspidites:
@galitskyd made the review easier by moving my file back to -13 instead of -14. once done, do you mind updating your file, as trying to merge with quality control will cause a merge conflict.

PR #1456, opened 2015-04-21
Deactivate and mark unsubscribed a user's prm searches upon account deletion [PD-560]

tdphillips:
All tests pass.

### DIFF
--- a/myjobs/models.py
+++ b/myjobs/models.py
@@ -646,8 +646,10 @@ def delete_user(sender, instance, using, **kwargs):
     saved searches will be deleted and partner saved searches will have the
     user nullified.
     """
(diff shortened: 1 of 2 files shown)


myjobs/models.py
```diff
@@ -646,8 +646,10 @@ def delete_user(sender, instance, using, **kwargs):
     saved searches will be deleted and partner saved searches will have the
     user nullified.
     """
-    instance.savedsearch_set.filter(partnersavedsearch__isnull=False).update(
-        user=None)
+    from mysearches.models import PartnerSavedSearch
```
tdphillips: This is an instance where having the import at the top of the file breaks everything.

PR #1457, opened 2015-04-21
Made __unicode__ function not make queries [ST-1181]

almathew:
This prevents any Select/Select.Multiples for Company from doing 30,000 queries.

I feel like this is a good place to complain about the lack of documentation for django_extensions.


almathew:
This will be ready to merge again in like 5 minutes.

almathew:
Added an ssl change. It can be merged now.

PR #1459, opened 2015-04-21
[PD-1289] Dynamic Filtering Lists

aspidites:
- filtered list now works
- introduced an event that fires when report data is changed
- introduced extendData to be used to ensure that the event is fired
- added an ignore property to filtered lists which is used to determine whether or not a field changing should cause that list to update
- refactored validate so that we don't have to keep constructing anonymous validate functions 
- I didn't bother refactoring filtered list into common parts yet, though I have a few ideas

# TODO
- [x] rearrange method declaration order
- [x] implement validation for state drop-down and tags
- [x] fix order of fields in the human readable report overview
- [x] reduce number of calls made to backend on report submission


static/reporting.163-20.js
```diff
@@ -99,9 +106,10 @@ Report.prototype.save = function() {
     return false;
   } else {
     this.fields.forEach(function(field) {
-      $.extend(report.data, field.onSave());
+      report.extendData(field.onSave());
```
galitskyd: Changing this back to $.extend would fix the issue of where all fields attempt to filter.

PR #1471, opened 2015-04-23
[PD-964] Links to communication records

aspidites:
This makes each name in the report view clickable, so that you can view specific records. I based it off of work from earlier in the sprint, so once those other tickets are merged, this one should look less intimidating.


aspidites:
Apparently, I folded these changes into #1470 , which is then part of (what will be) @galitskyd's PR.

PR #1472, opened 2015-04-24
Make deactivation of saved searches and deletion of the recipient set unsubscriber and send notifications [ST-1047][PD-560]

tdphillips:
I've still got some issues with seo tests, but everything else passed. Let's see what Jenkins says.

### DIFF
--- a/mysearches/forms.py
+++ b/mysearches/forms.py
@@ -127,6 +127,8 @@ class PartnerSavedSearchForm(RequestForm):
     def __init__(self, *args, **kwargs):
         choices = PartnerEmailChoices(kwargs.pop('partner', None))
         super(PartnerSavedSearchForm, self).__init__(*args, **kwargs)
+        setattr(self.instance, 'changed_data', self.changed_data)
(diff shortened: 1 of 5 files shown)


mysearches/forms.py
```diff
@@ -127,6 +127,8 @@ class PartnerSavedSearchForm(RequestForm):
     def __init__(self, *args, **kwargs):
         choices = PartnerEmailChoices(kwargs.pop('partner', None))
         super(PartnerSavedSearchForm, self).__init__(*args, **kwargs)
+        setattr(self.instance, 'changed_data', self.changed_data)
```
tdphillips: I opted to duplicate these two lines exactly once rather than create an annoying new subclass of RequestForm that did nothing but this and was hard to name to boot..

tdphillips:
The UI is unchanged, so screenshots would be fairly useless.

aspidites:
Code seems fine. Didn't test functionality locally, though.

aspidites:
Any objections to me merging this? @mklauber @almathew @galitskyd ?

PR #1477, opened 2015-04-27
[PD-1281][PD-1289][PD-1316] Preliminary changes for PD-1204

galitskyd:
This PR/branch is creating a better framework for Reporting.  Now, tickets such as PD-1204 will be less of a hack-fest.  

Before when creating new fields you'd have to specify a type in the parameters.  Then there would be one render function that would have a bunch of if's. If "date" do x, if "state" do y, and ect.  When adding new fields or reports one would have to be wary of other parts of the code, hundreds of lines apart.  Also, on the report level there was a single change event listener on all fields that had a bunch of ifs that would update ```Report.data``` which filteredLists relied on for filtering Partner and Contact and sometimes would actually modify the data causing risk of data lose. :-1: 

Now each field is its own type.  A field with type "text" is now called TextField, a field with type of "date" is now called DateField, you see the pattern.  Also, each field type can inherit off of a Field object and overwrite specific functions such as renderLabel, render, bindEvent, validation, showErrors, removeErrors, onSave, ect as needed.  Overwriting bindEvent will allow a developer make their own event listeners on that specific field and field type.  FilteredLists now don't require ```Report.data``` because ```Field.onSave``` can be used to retrieve data from that Field and doesn't modify any data.  ```Report.data``` is only set once the user hits "Generate Report" which one visual difference is that previewing the report information will always be in the same order as the fields on the form.  Below, is a nicer list of what is new and what other improves have been made to reporting. :+1: 

### What is new?

 * Validation
required
  * On change, a field will run validation immediately.
  * Can log and show multiple errors separated by commas.
  * Having a field that is required actually means something now.

 * Tag Filtering 
tags
  * For filtering multiple tags comma separation is needed.
  * Autocomplete based off of that company's tag list.
  * Selecting an item from autocomplete will automatically add a comma and space proceeding the item for convenience. 

 * Help text for fields. (Can be seen in Tag Field picture)

### What has been improved?
 * The API to build new fields.
  * Better use of Prototype pattern.  
    * Now when someone wants to add a new Field type they can just inherit off of Field object and overwrite functions as needed as opposed to adding ifs to a bunch of functions.
    * No longer needs to worry about a Field of one type possibly messing with another Field of a different type.  Abstraction!
    * Flexibility
 * FilteredLists (Partners and Contacts)
  * Query speeds have been improved.
  * Contact FilteredList waits for Partner to finish to accurately get results when data changes. 
filtering
  * Now you can watch spinners while you wait.
  * Rendering has been improved.
    * The state of the accordion persists so if you have a list showing/open it will stay open when the list updates. (before they would constantly close)
    * Switched to using labels instead of divs which saves writing code for clicking divs to check checkboxes.
 * DateField Improvements
datefield
  * Default start and end date
    * Start Date set to: 01/01/2014 (start of PRM)
    * End Date default to: Yesterday's Date.
  * Start date can't be later than the end date and vice versa.
  * Currently only field that has multiple errors, mentioned above.
 * Preview information before generating a report will always be in the same order as the fields show up on the form.
 * Filtered lists are aware of when other fields change, and can choose to ignore such updates. Otherwise, they are re-filtered when such a change occurs.
 * The large ajax spinner only shows up on report creation, and even that will be going away soonish (I'm allowing for reports to be created in parallel, instead of having to do so sequentially)
 * When all choices of a particular field are selected (contact type, partners, contacts), those choices are explicitly sent, which will ensure that the numbers seen in the generated report are consistent (not the case currently in production). This also means that deselecting and reselecting all choices no longer produces different partner and contact counts.


aspidites:
@galitskyd: feel free to move these to your message and delete this comment if you wish) 

# More Improvements
- Filtered lists are aware of when other fields change, and can choose to ignore such updates. Otherwise, they are re-filtered when such a change occurs.
- The large ajax spinner only shows up on report creation, and even that will be going away soonish (I'm allowing for reports to be created in parallel, instead of having to do so sequentially)
- When all choices of a particular field are selected (contact type, partners, contacts), those choices are explicitly sent, which will ensure that the numbers seen in the generated report are consistent (not the case currently in production). This also means that deselecting and reselecting all choices no longer produces different partner and contact counts.

PR #1478, opened 2015-04-27
[PD-1318][PD-1319] State/City filtering works properly in reports

aspidites:
- Code is a bit redundant, but I suspect the model work on contact/contactrecord will make a lot of this redundant, which is why I didn't bother thinking too hard about something more DRY. 
- went ahead and fixed the indentation issue in helpers. This code doesn't exist in QC as I rewrote that entire section, but rather than propagate those changes forward, I thought it best to do something simpler here until those changes naturally propagated.


aspidites:
Pardon the size of the request-- really, the only files I touched were those that are those related to reporting. The others are from @almathew's work which fixes seoxml unit tests.

PR #1479, opened 2015-04-28
[PD-1317] Tags Improvements for Reporting

aspidites:
- Changed to AND's for tags instead of OR's. 
- for Partner filtered list, check for contactrecord tags instead of partner tags
- ignore tags for contact list

Waiting on #1477 , which will simplify the diff for this one.


aspidites:
Also, I know there are merge conflicts. Will fix after @galitskyd's stuff is merged.

PR #1483, opened 2015-04-28
[PD-1322] Polyfill missing ECMAScript 5 functions

galitskyd:
Until we drop support of IE8 es5-shims will polyfill most of the missing functions introduced in ES5.

### DIFF
--- a/templates/mypartners/edit_location.html
+++ b/templates/mypartners/edit_location.html
@@ -59,5 +59,5 @@
 
 {% block extra-js %}
 <script src="{{ STATIC_URL }}tag-it.min.js{{ gz }}"></script>
-<script src="{{ STATIC_URL }}prm.163-28.js{{ gz }}"></script>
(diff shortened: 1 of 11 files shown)


aspidites:
Since this is going to quality control anyway, do you mind going through the static files and removing the definitions that are deprecated by the existence of these shims? For example, the indexOf definition in PRM?

templates/mypartners/edit_location.html
```diff
@@ -59,5 +59,5 @@ <h2 class="top">Tips</h2>
 
 {% block extra-js %}
 <script src="{{ STATIC_URL }}tag-it.min.js{{ gz }}"></script>
-<script src="{{ STATIC_URL }}prm.163-8.js{{ gz }}"></script>
+<script src="{{ STATIC_URL }}prm.163-28.js{{ gz }}"></script>
```
galitskyd: this is wrong version #

PR #1495, opened 2015-05-05
Add admin pages for event->site mapping [PD-805][PD-778]

tdphillips:
The ticket specifies tying templates to site families. I'm unsure how that can be done with the schema we currently have - filter sites by sitetag__is_site_family=True, possibly.

In any case, this adds a boat load of admin screens.

I want to see what Jenkins thinks.


myemails/tests/test_admin.py
```diff
@@ -0,0 +1,52 @@
...
+            self.user.is_superuser = superuser
+            self.user.save()
+            form = admin.get_form(self.request)()
+            email_template = form.fields['email_template']
+            query = str(email_template.queryset.query)
+            if superuser:
+                self.assertFalse('WHERE' in query)
+            else:
+                self.assertTrue(
+                    'WHERE ("myemails_emailtemplate"."owner_id" = %s' % (
```
tdphillips: I'm sorry.

postajob/models.py
```diff
@@ -765,6 +770,10 @@ class Request(BaseModel):
     deny_reason = models.TextField(_('Reason for denying this request'),
                                    blank=True)
 
+    @property
```
tdphillips: Unused for now, but I planned on using this for certain `ValueEvent`s before it was determined that that functionality is out of the scope for this.

myemails/tests/test_admin.py
```diff
@@ -0,0 +1,55 @@
...
+            form = admin.get_form(self.request)()
+            email_template = form.fields['email_template']
+            query = str(email_template.queryset.query)
+            if superuser:
+                self.assertFalse('WHERE' in query)
+            else:
+                if connection.vendor == 'sqlite':
+                    test = 'WHERE ("myemails_emailtemplate"."owner_id" = %s'
+                else:
+                    test = 'WHERE (`myemails_emailtemplate`.`owner_id` = %s'
```
tdphillips: This annoys me more than it should.

aspidites: @tdphillips  did you mean to use backticks in the second WHERE clause for the connection.vendor if statement? Is mysql that finnicky?

tdphillips: Yes and yes. I'm comparing it to the generated query, so I'm subject to the whims of the DB backend.

static/emails.js
```diff
@@ -0,0 +1,55 @@
+(function($) {
```
aspidites: Yay, IIFE!

tdphillips: Yay, @galitskyd!

static/emails.js
```diff
@@ -0,0 +1,55 @@
...
+                    // with that model.
+                    if (model_fields.hasOwnProperty(value)) {
+                        // We've already retrieved the fields for this model;
+                        // pull it from our cache and select its first value.
+                        $select = model_fields[value];
+                    } else {
+                        // We have not retrieved this model's field list yet.
+                        $.getJSON('/emails/get-fields/', 'model=' + value,
+                            function (data) {
+                                // TODO: Can I get by without an ever-expanding if chain?
```
galitskyd: Yes, with a ternary operator.

``` fields = window.location.href.indexOf('valueevent') > -1 ? data['value'] : data['time']; ```

Comment (tdphillips) [inline static/emails.js]: :+1:

For some reason, I was thinking there might be more than value and time. There shouldn't be.

File: static/emails.js [hunk 6]
```diff
@@ -0,0 +1,55 @@
...
+        $model.change(
+            function () {
+                var fields,
+                    field,
+                    $input = $('#id_field'),
+                    selected = $input.val(),
+                    $select = $('<select id="id_field" name="field"></select>'),
+                    options = ['<option value="">---------</option>'],
+                    value = $model.val();
+                if (value.length > 0) {
```
galitskyd: value.length == 0 is a falsy value and anything > 0 is truthy. Doesn't have to be changed but if you wanted to save some typing.

galitskyd:
:+1:, @aspidites @almathew @JLMcLaugh how does it look to you guys?

myemails/forms.py
```diff
@@ -0,0 +1,28 @@
...
+from universal.helpers import get_company
+
+
+class EventFieldForm(ModelForm):
+    field = NoValidationChoiceField(choices=[('', '---------')])
+
+    def __init__(self, *args, **kwargs):
+        request = kwargs.pop('request')
+        super(EventFieldForm, self).__init__(*args, **kwargs)
+        setattr(self, 'request', request)
```
aspidites: Remind me agai, why do we need setattr instead of just `self.request = request`? This isn't the only place I've seen it, but I thought I'd ask since it was the most convenient example I have on hand.

tdphillips: Inertia. According to people, you should use setattr when you don't know what the attr is called.

static/emails.js
```diff
@@ -0,0 +1,50 @@
...
+                    // with that model.
+                    if (model_fields.hasOwnProperty(value)) {
+                        // We've already retrieved the fields for this model;
+                        // pull it from our cache.
+                        $select = model_fields[value];
+                    } else {
+                        // We have not retrieved this model's field list yet.
+                        $.getJSON('/emails/get-fields/', 'model=' + value,
+                            function (data) {
+                                fields = window.location.href.indexOf('valueevent') > -1 ? data['value'] : data['time'];
```
galitskyd: Not sure if we have to worry about IE8 or not. But you may have polyfill indexOf if we do care about IE8 and if this isn't going on secure.

tdphillips: I seem to recall some permutation of this conversation happening twice, each time after indexOf came back to bite us.

It is going on secure, but it will eventually be used by members to set up their own templates. That being the case, I'm guessing I should polyfill.

tdphillips: Actually, this should be String.prototype.indexOf. It's supposedly been in place since IE~6.

galitskyd: Ah.. yes, you are correct.

PR #1496, opened 2015-05-06
[PD-1301] Establish relationship between Contact and ContactRecord

aspidites:
## Description
This ticket's purpose is to establish a real foreign key relationship between the `Contact` and `ContactRecord` models. Doing so will allow for more efficient (and meaningful) queries for reports.

## Changes
- `ContactRecord` now has a `contact` field, which points to a contact
- `ContactRecord` no longer has a contact_name field. As there will never be a case where this would differ in the contact record, it makes sense to simply keep it as the `Contact` instance's name. Should it need to change, one can either associate the `ContactRecord` to a different contact or change the contact's name. 
- `contact_email` remains in tact, as it could very well be the case that a record was made for communication that was sen to an email address other than the one associated with the contact. Same logic for `contact_phone`. 
- myreports has been updated to take advantage of the new relationship (**not done**)
- `Contact`s cannot be deleted. Instead, they are simply archived. In doing so, we needn't worry about a contact being deleted causing information to be lost from a `ContactRecord`
- `Contact` now has an `archived_on` datetime field, which notes when the contact was archived. A Null signifies a contact which hasn't been archived.
- A contact is unarchived if a person attempts to create a contact with the same name and email address.

## Migrations
- Each of these migrations (including the data migration) have a backward migration associated with them, so things should be relatively safe. 
- The data migration is a set of raw queries, as doing so was way more efficient than the equivalent django query (feel free to prove me wrong).
- The removal of `contact_name` and the addition of `archived_on` are two separate migrations because of the fact that the data migration relies on the existence of contact_name to do it's thing. The migrations happen as follows:
    - `archived_on` field is added to `Contact`
    - contacts are created in archived status from `ContactRecord`s whose `contact_name` and `contact_email` can't be traced back to a `Contact`. All contact records will then be associated with a contact.
    - `contact_name` is deleted from `ContactRecord`


almathew:
Does the custom migration work on both sqlite and mysql?

aspidites:
@almathew dunno. I'll check that out later and write a sqlite specific one if need be.

aspidites:
@almathew addressed the sqlite thing. Just need to find out why these other tests are failing. Quite honestly, they are failing in apps I've never messed with, so that might take a while...

mypartners/models.py
```diff
@@ -187,10 +187,13 @@ class Contact(models.Model):
...
     name = models.CharField(max_length=255, verbose_name='Full Name')
     email = models.EmailField(max_length=255, verbose_name='Email', blank=True)
-    phone = models.CharField(max_length=30, verbose_name='Phone', blank=True)
+    phone = models.CharField(max_length=30, verbose_name='Phone', blank=True,
+            default='')
     locations = models.ManyToManyField('Location', related_name='contacts')
     tags = models.ManyToManyField('Tag', null=True)
-    notes = models.TextField(max_length=1000, verbose_name='Notes', blank=True)
+    notes = models.TextField(max_length=1000, verbose_name='Notes',
+                             blank=True, null=True)
```
tdphillips: null=True on a field containing text.

aspidites: Fixed.

PR #1498, opened 2015-05-06
At&t fix

mklauber:


almathew:
:+1:  @JLMcLaugh

import_jobs.py
```diff
@@ -613,6 +613,15 @@ def add_jobs(jobs, upload_chunk_size=1024):
     """
     conn = Solr(settings.HAYSTACK_CONNECTIONS['default']['URL'])
     num_jobs = len(jobs)
+    
+    # AT&T Showed that large numbers of MOCs can cause import issues due to the size of documents.
+    # Therefore, if the number of MOCs is above 10, arbitrarily use a lower chunk size.
+    max_onet_count = max([len(job.get('mapped_moc', [])) for job in jobs])
```
JLMcLaugh: Imports for large job collections are already slow from the large dictionary and xml objects being created. Let's avoid making another large list. Use a loop and break if we're over the max_one_count.

import_jobs.py
```diff
@@ -613,6 +613,16 @@ def add_jobs(jobs, upload_chunk_size=1024):
     """
     conn = Solr(settings.HAYSTACK_CONNECTIONS['default']['URL'])
     num_jobs = len(jobs)
+    
+    # AT&T Showed that large numbers of MOCs can cause import issues due to the size of documents.
+    # Therefore, if the number of MOCs is above 10, arbitrarily use a lower chunk size.
+    for job in jobs:
+        if len(job.get('mapped_moc', [])) > 100:
```
JLMcLaugh: Do we want 10 or 100 here?

almathew: I picked a random Company that had mapped mocs but not as many as ATT. They had ~260 per job.

tdphillips: I think this is referring to the comment, which implies len(mapped_moc) > 10.

PR #1500, opened 2015-05-08
Allow both event types to be queued [PD-763]

tdphillips:
Depends on #1495.

It's a little fuzzy what the ticket was asking for, so I opted to allow all current event types to be scheduled (ValueEvents will be scheduled immediately).

As this is based off of #1495, only the final commit is relevant to this PR.

This could use a good review to make sure I'm not being stupid.


aspidites:
Given [this](https://github.com/DirectEmployers/MyJobs/commit/7c66c272ce524248e7f6aee8f99116aa091bcacb), it seems fine. @almathew @mklauber @galitskyd @JLMcLaugh: objections?

PR #1502, opened 2015-05-08
[PD-1204] Partner, contact, communication reports

galitskyd:
Really only three major additions/changes and bunch of little minor architectural or code style changes.

* New Reports!
screen shot 2015-05-08 at 10 51 14 am
  * Partner
    * New View function (see renderViewPartner)
screen shot 2015-05-08 at 10 51 33 am
  * Contact
    * new View function (see renderViewContact)
      * screenshot had some phone numbers so I removed it.
* New way of how Reports are created.
* Form Object
  * Holds a list of Report Objects
  * Has functions that handle multiple Reports
    * Rendering a combined report is working, but may need improvements.
  * Currently not accessible through the UI as there are a lot of ?? of how to handle multiple reports. 
* Filtering improvements
* "Copy of" prepended to report names on clone.

This includes the commit where tabs were replaced with spaces so there are a fair amount of diffs for that.


myreports/views.py
```diff
@@ -29,6 +28,9 @@ def overview(request):
     report_count = reports.count()
     past_reports = reports[:10]
 
+    for report in past_reports:
```
aspidites: Not sure we should be doing this here. Seems like something only the front-end would be concerned with (eg. template or javascript). Granted, they aren't getting saved since you aren't doing `report.save`, but it still makes me nervous.

static/reporting.164-07.js
```diff
@@ -11,9 +10,28 @@ window.onpopstate = function(event) {
     renderOverview();
   } else if (state.page && state.page === 'new') {
     historyNew = function() {
-      report = new Report(state.report.types);
+      var types = state.types,
+          form,
+          report,
+          rs;
+
+      if (types.length > 1) {
```
aspidites: Why do we need to bother checking that the length is greater than 1? `[2].map()` will be just as effective as `[1, 2, 3].map()`.

galitskyd: You are right about map not being affected by the > 1 check.  It was more of a check to see if we need to create a Form object or not.

myreports/views.py
galitskyd: If I do it on the template level I'd have to make a template tag to do what we want, we've talked about that before.  Also, if I do it in javascript, once the file loads users will see the sidebar names "pop" as all the _'s turn to spaces.  That being said I wouldn't mind making this a property on the model though.

aspidites: Leave it for now. I intend to do what we talked about before and make the name distinct from the filename such that we won't have to do the name mangling anymore.

PR #1504, opened 2015-05-11
Remove candidates from nav [PD-1320]

tdphillips:
All tests pass.


templates/includes/topbar.html
```diff
@@ -451,7 +451,7 @@ <h3 id="topbar-logo">My.jobs</h3>
     {% else %}
```
tdphillips: This annoys me more than it should. I've removed lines 451-453 and tests failed. I couldn't find the matching `{% if %}` so I readded it.

PR #1512, opened 2015-05-13
Add site package support to ETL jobs [PD-1296]

tdphillips:
I think this does what we want. I'm probably wrong - let me know.

All tests pass locally.


import_jobs.py
```diff
@@ -45,6 +45,11 @@ def update_job_source(guid, buid, name, clear_cache=False):
     bu.save()
     add_company(bu)
 
+    # Determine what sites these jobs should be on
+    on_sites = set(bu.seosite_set.values_list('sitepackage', flat=True))
+    on_sites = filter(None, on_sites)
+    on_sites = on_sites or [0]
```
aspidites: Do we document somewhere that [0] means all?

tdphillips: It doesn't appear so. I grabbed it from postajob.

almathew: It's vague: https://github.com/DirectEmployers/MyJobs/blob/quality-control/postajob/models.py#L179

Although wherever the documentation would've gone it would be completely useless. There's not a single place I would think to look for that information, and I'm the one who wrote it.

aspidites: named "constant" would probably be better at this point. Something like:

```python
class Job:
    ALL_SITES = 0 # expand as necessary, not unlike an enum
...
on_sites or [Job.ALL_SITES]
```

aspidites:
Tests are failing.

tdphillips:
Tests failing are a known issue for this PR. It depends on the work @mklauber is doing for the AT&T fix.

aspidites:
I see. So this should be passing once #1514 is passing?

tdphillips:
Should be.

tdphillips:
I did not actually name [0] anywhere. To be honest, I forgot about it. If there are strong feelings on the matter, I'll add it somewhere (where, I don't know).

PR #1522, opened 2015-05-19
[PD-1370] Contact View shows duplicate States

galitskyd:
When viewing a Contact Report. If a contact has multiple locations, for example: Zionsville, IN; Carmel, IN; Fishers, IN; Austin, TX. In the table it would show up as IN, IN, IN, TX instead of IN, TX.

### DIFF
--- a/static/reporting.164-19.js
+++ b/static/reporting.164-19.js
@@ -1692,21 +1692,29 @@ function renderViewContact(id) {
     success: function(data) {
       var $span = $('<div class="span12"></div>'),
           $table = $('<table class="table table-striped report-table"><thead><tr>' +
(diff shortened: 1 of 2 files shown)


static/reporting.164-19.js
```diff
@@ -1692,21 +1692,29 @@ function renderViewContact(id) {
...
+            // for each location get state and trim whitespace
+            return (location.split(',')[1] || '').trim();
+            // Determine uniqueness
+          }).sort().filter(function(e, i, a) {
+            // Due to sort, duplicates will be together.
+            // Check to see if this element is not the same as the prior one.
+            return e !== a[i - 1];
+          }).join(', ');
+
+        return '<td>' + record.partner + '</td><td>' + record.name + '</td><td>' + record.phone + '</td>' +
```
aspidites: Comments missing for this line. I don't know how string concatenation works.

PR #1523, opened 2015-05-19
Tried a different way of replacing a location slug.

aspidites:
Rather than replacing the location slug text, I find out where the location slug starts and ends and replace that (by converting the location and location slug to lists).

I contemplated using string.find for this, but thought it would result in a similar issue to what we have now. Now we strip all occurrences of a substring. Find would strip only the first, but could produce a false positive. The alternative might be to match substrings wrapped with slashes, but I wasn't sure if that would be reliable, hence why the current method.

### DIFF
--- a/seo/breadbox.py
+++ b/seo/breadbox.py
@@ -215,12 +215,13 @@ class Breadbox(object):
             if ending_slug in slugs and len(location_filters) < 2:
                 slugs.remove(ending_slug)
 
-            location_slugs = location_slug.split('/')
-            left = slugs.index(location_slugs[0])
-            right = left + len(location_slugs)
+            if location_slug in new_path:
+                location_slugs = location_slug.split('/')
+                left = slugs.index(location_slugs[0])
+                right = left + len(location_slugs)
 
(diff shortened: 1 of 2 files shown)


aspidites:
As for tests, I still don't really get how the app works, so wouldn't be able to write a useful test (yes, I looked at existing tests).

seo/breadbox.py
```diff
@@ -211,11 +211,17 @@ def build_location_breadcrumbs_from_slugs(self):
...
+
+            if location_slug in new_path:
+                location_slugs = location_slug.split('/')
+                left = slugs.index(location_slugs[0])
+                right = left + len(location_slugs)
 
-            new_location_slug = "/".join(location_filters[1:])
-            new_path = new_path.replace(location_slug, new_location_slug)
+                new_path = '/'.join(slugs[:left] + location_filters[1:] + 
+                                    slugs[right:])
```
JLMcLaugh: Could you add some tests for this?

aspidites: See my previous comment about tests. I can write a test for the splitting of various combinations of strings which have slashes, if you want?

PR #1530, opened 2015-05-20
Data prod to qc

almathew:
I still have to do the added/removed columns part, but I wanted to get this pull request up because the code is a nightmare.

@aspidites I looked at Collector, but the relationship following was in the wrong direction and the information supplied wasn't quite good enough to confirm that relationships were created in the way that I needed them confirmed.


aspidites:
Bleh. sorry.

almathew:
As per @JLMcLaugh I'm not worrying about missing/additional columns.

copy_data

almathew:
Feel free to review/merge.

seo/admin.py
```diff
@@ -1244,3 +1244,40 @@ def check_inline_instance(obj, req):
...
+def copy_to_qc(modeladmin, request, queryset):
+    # This shouldnt't be used to copy a large number of objects at once.
+    if queryset.count() > 5:
+        error_message = 'You cannot copy more than 5 items at once.'
+        modeladmin.message_user(request, error_message)
+        return
+
+    queue = copy_following_relationships(queryset, copy_to='qc-redirect')
+
+    modeladmin.message_user(request, 'Below is a summary of the items copied. '
```
tdphillips: :-1:

"Stuff was done" was so concise.

almathew: I know. Unfortunately we'll probably be dubugging this in the future and I thought the debugging info might just be useful.

aspidites: Why 5? Sounds arbitrary. What if I have a legitimate need to copy 100 objects? Do I go in 20 times?

deploy/settings.dseo_qc_admin.py
```diff
@@ -17,6 +17,15 @@
         'HOST': 'db-redirectqc.c9shuxvtcmer.us-east-1.rds.amazonaws.com',
         'PORT': '3306',
     },
+    # Points to staging instead of QC for testing purposes.
```
aspidites: 1. Is this still supposed to point to staging?
2. By it pointing to staging, is our testing environment broken at all?

requirements.txt
```diff
@@ -24,14 +24,14 @@ django-storages==1.1.4
 django-taggit==0.12.1
 django-tastypie==0.9.15
 django-widget-tweaks==1.1.2
-factory_boy==2.4.1
+factory_boy==2.5.0
```
aspidites: Do we all need to update our dependencies before this will work locally, or is this just a version bump for the sake of it?

seo/admin.py
aspidites: As I think about it, can't we chunk the request into individual calls if this is the case?

seo/admin.py
```diff
@@ -1244,3 +1244,40 @@ def check_inline_instance(obj, req):
         return obj.get_inline_instances(req)
     else:
         return obj.inline_instances
+
+
+from queryset_copier import copy_following_relationships
+
+
+def copy_to_qc(modeladmin, request, queryset):
```
aspidites: Stylistic note: Isn't request usually the first argument?

seo/admin.py
almathew: Hal requested a small number. With some of the larger objects (by this I mean objects with a lot relationships) 5 really is the best it's going to be able to do in a reasonable amount of time.

seo/admin.py
almathew: Django determined the arguments here.

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import models, IntegrityError
+
+
+ERROR = -1
+NOT_STARTED = 0
+SAVED = 1
+COMPLETE = 2
+
+saved_states = [SAVED, COMPLETE, ERROR]
```
aspidites: PEP8 - Constants, so these should be capitalized.

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+                    # This point should never be reached. All associated
+                    # foreign key objects should be made before
+                    # we actually attempt to create the object
+                    # itself.
+                    raise ObjectDoesNotExist('The foreign key %s was not '
+                                             'already created when attempting '
+                                             'to create object %s.'
+                                             % (fk_object, obj))
+                kwargs[field.name] = new_fk_object
+        elif not isinstance(field, models.ForeignKey):
```
aspidites: else is sufficient here. There's no way to get past the previous block as a foreign key, so checking that it isn't a foreign key is redundant.

almathew: Yeah.. I kind of forgot to change that after I knocked the `if not include_null` part down to inside the first elif.

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+
+    :param queryset: A queryset of objects to be copied.
+    :param copy_to: The database the queryset is being copied to.
+
+    :return: The ordered queue dictionary containing the existing objects,
+             the newly created objects, and any errors that happened
+             during the process.
+
+    """
+    queue = populate_queue(queryset)
```
aspidites: You return queue from populate queue, so you can do `return create_in_order(copy_to, populate_queue(queryset))`

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+
+    :return: An updated queue with all the new objects created and/or
+             relevant errors recorded.
+
+    """
+    iterator = len(queue.items()) - 1
+
+    # Create all the objects with only the non-relationship fields and
+    # required foreign keys.
+    while [item for item in queue.items() if item[1]['status'] not in saved_states]:
```
aspidites: Is this guaranteed to terminate?

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+    :param copy_to: The database the objects are being copied to.
+    :param queue: An ordered dictionary of dictionaries, where the dictionries
+                  are mapped to the object. The expected format of the
+                  dictionary can be found in add_to_queue()
+
+    :return: An updated queue with all the new objects created and/or
+             relevant errors recorded.
+
+    """
+    iterator = len(queue.items()) - 1
```
aspidites: `iterator` in python has a very specific meaning. As this is an int, I think calling it an iterator is misleading. From usage below, it seems to just be an index.

requirements.txt
almathew: You'll need to update your dependencies. I needed the Meta `databases` field which was introduced in 2.5.0

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+    Saves a new object to the specified database.
+
+    :param copy_to: The database the object should be saved to.
+    :param model: The model for the object.
+    :param kwargs: The relevant kwargs for the object.
+
+    :return: The new object
+
+    """
+    try:
```
aspidites: Would using a set of objects to create instead of a list prevent us having to try/except for IntegrityError and obsolete the use for find_duplicate_object? 

That's fine if not. There are a few different calls going on here, so I'm sure I've missed something.

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+        return base_model.objects.using(copy_to).create(**kwargs)
+    except IntegrityError, e:
+        duplicate = find_duplicate_object(copy_to, base_model, e, **kwargs)
+        return update_existing_object(copy_to, duplicate, **kwargs)
+
+
+def find_duplicate_object(database, base_model, integrity_error, **kwargs):
```
aspidites: Feels odd that this is used upon receiving an error, yet it can fail in the same way as the error that prompted its being called. I assume the point there is that the error will 'bubble up'?

deploy/settings.dseo_qc_admin.py
almathew: Yes it's still supposed to point to staging.

It's a bit confusing, but in order for this to be testable the database can't point to itself. I planned to put an update in the ticket explaining that once this is merged.

seo/queryset_copier.py
almathew: No. This index error is a result of the object already existing with a different pk.

A really common instance of this is if your user account has different pks between prod and qc. The database will throw an integrity error for duplicate email addresses.

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+    :param null: True if you want to consider only foreign keys where
+                 null=True in the model definition. False if you only want
+                 foreign keys where null=False in the model definition.
+
+    :return: A list of fields objects that obj has a foreign key to.
+
+    """
+    objects = []
+
+    for fk_field in get_foreign_keys(obj, null=null):
```
aspidites: Your for loop is just appending to a list, so could be reduced to a comprehension as well. In fact, this whole function could be reduced to a list comprehension. If having to repeat get_foreign_key_object_for_field is too annoying, you can eliminate the final comprehension by conditionally appending instead of always doing so.

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+        objects.append(get_foreign_key_object_for_field(obj, fk_field))
+
+    return [obj for obj in objects if obj]
+
+
+def get_foreign_key_object_for_field(obj, field):
```
aspidites: I'm not even sure why a function is required here. getattr(obj, field.name) is far more concise on its own and doesn't require me to look up a definition.

seo/queryset_copier.py
almathew: Yes, the error bubbles up.

seo/queryset_copier.py
```diff
@@ -0,0 +1,513 @@
...
+    :return: The creation queue updated with the new object
+             and new status.
+
+    """
+    key = object_to_key(obj)
+
+    try:
+        queue[key]['new_object'] = get_or_create_object(copy_to, obj, queue)
+        queue[key]['status'] = SAVED
+    except Exception, e:
```
aspidites: PEP 8 - as I've mentioned before, this syntax is deprecated.

seo/tests/test_queryset_copier.py
```diff
@@ -0,0 +1,130 @@
+from seo import models
+from seo import queryset_copier as qc
+from seo.tests.setup import DirectSEOBase
+from seo.tests import factories
+
+
+class QuerysetCopier(DirectSEOBase):
```
aspidites: Isn't test required to be in the class name for discovery?

almathew: No

seo/queryset_copier.py
aspidites: Ah. Makes sense. Bummer.

seo/tests/test_queryset_copier.py
aspidites: Then, is there  a reason to be inconsistent, considering all of our test classes DO have test in their name?

seo/queryset_copier.py
almathew: This would be multiline as a list comprehension as well. I know a huge argument could happen over what's more readable, but I think explicit loops will always be more readable than multiline list comprehensions (especially to newer developers).

seo/queryset_copier.py
almathew: I wanted to make sure the way I was accessing objects was consistent every single time I did it. There's a few different ways to do it (another is `getattr(obj, field.attname)` and they all work just a little bit differently. With something like this the little differences are enough to really throw off some of the assumptions that I needed to make.

aspidites:
I don't see this getting any better until it's in stakeholders' hands in qc. Everything seems to function correctly locally. Anyone object to me merging this now?

PR #1532, opened 2015-05-21
[PD-1358] null=True -> default=''

aspidites:
Tests are failing locally, but only for me, so I wanted this on jenkins to see if I could reproduce the errors. That said, the seo branch fails for me unconditionally, regardless of which branch I'm on, so I assume something about my local database isn't identical to production/qc/staging databases.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -843,7 +843,7 @@ class ContactLogEntry(models.Model):
     contact_identifier = models.CharField(max_length=255)
     content_type = models.ForeignKey(ContentType, blank=True, null=True)
     delta = models.TextField(blank=True)
-    object_id = models.TextField('object id', blank=True, null=True)
+    object_id = models.TextField('object id', blank=True, default='')
     object_repr = models.CharField('object repr', max_length=200)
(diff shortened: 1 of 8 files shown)


mypartners/models.py
```diff
@@ -843,7 +843,7 @@ class ContactLogEntry(models.Model):
     contact_identifier = models.CharField(max_length=255)
     content_type = models.ForeignKey(ContentType, blank=True, null=True)
     delta = models.TextField(blank=True)
-    object_id = models.TextField('object id', blank=True, null=True)
+    object_id = models.TextField('object id', blank=True, default='')
```
tdphillips: I know you didn't add this, but I kind of have to wonder why an id is a textfield.

almathew: https://github.com/django/django/blob/master/django/contrib/admin/models.py#L32

almathew: I think the longer answer is that the id field doesn't have to be the typical autoincrement field, and textfield is a happier catch-all even in the worst cases solution.

tdphillips: Makes sense.

aspidites:
Abandoning this.

PR #1533, opened 2015-05-21
[ST-1258] more button work around for job listing

aspidites:
The issue is that there is no url for featured companies. The proper fix would involve touching pager.js, seo/urls/search_urls.py, seo/views/search_views.py, seo/helpers.py, and seo/filters.py at a bare minimum (I stopped digging once I realized there were this many things to change). 

As this template is only used by one company, I decided on a work around instead, which takes advantage of the fact that the second result set is already loaded. Thus, if a request comes for more companies, those companies are simply loaded. The implication of this work around is that there may be at a maximum, 20 featured companies, until a proper fix is implemented.

### DIFF
--- a/myblocks/models.py
+++ b/myblocks/models.py
@@ -302,7 +302,7 @@ class MoreButtonBlock(Block):
         }
 
     def required_js(self):
-        return ['%spager.163-24.js' % settings.STATIC_URL]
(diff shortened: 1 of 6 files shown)


tdphillips:
If this ticket isn't about a proper fix, what is?

aspidites:
@tdphillips  as the title and description says, it's a **workaround**.

PR #1535, opened 2015-05-26
Revert "Add site package support to ETL jobs [PD-1296]"

almathew:
Reverts DirectEmployers/MyJobs#1512


almathew:
The way this is done Pearson's jobs will now only show up on Person's website, not www.my.jobs

PR #1537, opened 2015-05-26
[PD-1372] Passed remaining query parameters to edit view.

aspidites:
Made sure all query parameters are being passed to the edit view.

### DIFF
--- a/mypartners/views.py
+++ b/mypartners/views.py
@@ -766,7 +766,7 @@ def prm_edit_records(request):
         if form.is_valid():
             form.save(user, partner)
             return HttpResponseRedirect(reverse('record_view') + '?' +
-                                        request.GET.urlencode())
+                                        request.META['QUERY_STRING'])
     else:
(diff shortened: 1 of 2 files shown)


mypartners/views.py
```diff
@@ -766,9 +765,8 @@ def prm_edit_records(request):
                                  partner=partner, instance=instance)
         if form.is_valid():
             form.save(user, partner)
-            return HttpResponseRedirect(reverse('record_view') +
-                                        '?partner=%d&page=%s' %
-                                        (partner.id, page))
+            return HttpResponseRedirect(reverse('record_view') + '?' +
+                                        request.GET.urlencode())
```
almathew: I don't mind either way - mostly I'm just wondering if there's some "standard" I don't know about, but is there any reason you're preferring rebuilding the GET params versus just using request.META['QUERY_STRING']?

aspidites: First thing I came up with. I'll probably switch it though as query_string is more obvious.

PR #1541, opened 2015-05-27
[PD-1381] Approval status

aspidites:
- default status is approved.
- can't use F objects or bulk create because of foreign keys
- couldn't find a reference to a user from partner, so i just set the approved_by for this model to the first company user of the owner of that partner


mypartners/models.py
```diff
@@ -194,6 +214,8 @@ class Contact(models.Model):
     notes = models.TextField(max_length=1000, verbose_name='Notes',
                              blank=True, default="")
     archived_on = models.DateTimeField(null=True)
+    approval_status = models.ForeignKey(
```
tdphillips: Should this be a OneToOneField?

mypartners/models.py
```diff
@@ -39,6 +39,26 @@
...
+    UNPROCESSED, APPROVED, DENIED = range(3)
+    STATUS_CODES = (
+        (UNPROCESSED, "Unprocessed"),
+        (APPROVED, "Approved"),
+        (DENIED, "Denied")
+    )
+    
+    code = models.PositiveSmallIntegerField(
+        default=APPROVED, choices=STATUS_CODES, verbose_name="Status Code")
+    last_modified = models.DateTimeField(verbose_name="Last Modified")
```
almathew: It might be worth tracking user now to save yourself the hassle of adding it later.

almathew: Also, you probably want `auto_now=True` 

https://docs.djangoproject.com/en/1.6/ref/models/fields/#django.db.models.DateField.auto_now

aspidites: @almathew I don't think we do want to track users. The original use case was creating records from emails, which wouldn't have users. I'll add an approved_by, though.

PR #1545, opened 2015-05-29
Single microsite job listing [PD-1296]

almathew:
Business Units can be tied to a site package, limiting all the jobs for that business unit to showing up ONLY on the sites specified in the site package.


almathew:
Should be done, but I'm getting a test failure that I think is probably related to my environment at the moment.

almathew:
Jenkins is satisfied. @JLMcLaugh or @mklauber might want to look at this?

mklauber:
+1

PR #1546, opened 2015-05-29
Fix for AT&T

mklauber:


almathew:
Since it was already merged without waiting once....

PR #1547, opened 2015-06-01
[PD-1402] Fixe topbar url for inbox

aspidites:
- used absolute url for topbar (since inbox doesn't exist on www.my.jobs). not the only absolute url that exists, so I don't feel bad about htis
- wrapped inbox in a row to fix styling issue


aspidites:
Experimenting with a few other options.

aspidites:
Done experimenting. Not sure how else to do this. This breaks testing the inbox locally, but it's not like testing topbar locally isn't already broken. Best alternative I can think of is to manually pass along the settings.PROJECT variable and conditionally change all links in the topbar which refer to absolute urls. 

That said, this will work on production as it points to secure.my.jobs/messages/inbox in both cases.

tdphillips:
Someone correct me if I'm wrong, but this seems like what the ABSOLUTE_URL setting was added for. Correctly setting it in the dseo prod web settings should fix all of the topbar links using it if they don't currently work.

tdphillips:
Ignore the final part of my last update. ABSOLUTE_URL should be all that's needed.

aspidites:
Top bar is ajaxed in (eg. not a template), so ABSOLUTE_URL isn't available. Even if we pass it along via <script>, it' still references the wrong url. 

From www.my.jobs:
- topbar should be ajaxed in from secure
- ABSOLUTE_URL refers to the current domain ("/"), thus would point to www, not secure
- ABSOLUTE_URL isnt' available from javascript, though it could be passed through with a <script> tag>

In other words, I'm fairly confident that ABSOLUTE_URL isnt' sufficient here, since we want to point to an external domain instead of an internal one (secure.my.jobs instead of www.my.jobs).

aspidites:
Trying again and failing miserably with ABSOLUTE_URL...

aspidites:
Starting from scratch...

PR #1550, opened 2015-06-01
Added migration to increase the length of the domain field on sites [ST-1271]

almathew:
This is numbered so it doesn't interfere with my other pull request (https://github.com/DirectEmployers/MyJobs/pull/1545)


almathew:
I really enjoyed putting the "CONTAINS MIGRATIONS" label on something that is literally just a migration.

PR #1551, opened 2015-06-01
[PD-1402] Tobar Fix for Inbox

aspidites:
# Note
You'll need to update ABSOLUTE_URL in your settings file to point to where ever myjobs is running, ensuring that it ends in a trailing slash. I also apparently forgot to add the message context processor to my local config, so be sure to do that if you haven't already.

- link to inbox now works
- links to individual messages now work
- formatting fixed for message inbox

Before:
2015-06-01-103544_1244x1056_scrot

After:
2015-06-01-145029_1916x1056_scrot


deploy/settings.dseo_staging_admin.py
```diff
@@ -69,7 +69,7 @@
     'current': 'http://ec2-23-20-67-65.compute-1.amazonaws.com:8983/solr/myjobs_test_current/',
     }
 
-ABSOLUTE_URL = '/'
+ABSOLUTE_URL = 'http://staging.my.jobs/'
```
JLMcLaugh: @aspidites Checking an old commit. Was this intentional, or should it also be pointing at staging.secure.my.jobs?

PR #1552, opened 2015-06-01
Mx record checks

almathew:
Just up for review. Don't merge until we can confirm `make_mx_record()` is working

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -38,6 +38,7 @@ mock==1.0.1
 newrelic==2.50.0.39
 ordereddict==1.1
 pep8==1.3.3
+pydns==2.3.6
 pyflakes==0.5.0
 pynliner==0.5.1
 pysolr==3.2.0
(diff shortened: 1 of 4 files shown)


aspidites:
> Don't merge until we can confirm make_mx_record() is working
And how do we do that?

almathew:
I should be testing it tomorrowish.

tdphillips:
I ran make_mx_record for indianapolis.jobs. `dig indianapolis.jobs MX` returns nothing (and probably won't for 24-48 hours) but specifying your name server makes the record show up (`dig @ns.1454.awsdns-53.org indianapolis.jobs MX`).

`dig @ns-1454.awsdns-53.org indianapolis.jobs MX`

tdphillips:
What do you know, it was more like ten minutes than 24 hours.

almathew:
Does it point to sendgrid correctly?

tdphillips:
Looks like it. It's the same as the record for my.jobs.

tdphillips:
Removed the DO NOT MERGE label but I would like #1605 to be looked at first.

PR #1553, opened 2015-06-01
[PD-1335] Reports - add regeneration as 4th column in report lists

galitskyd:
- Added new column for regenerate and updated JS to handle new rules
- Edit regenerate functionality to always regenerate (as opposed to checking to see if the file already exists and ignore the request) and delete the old report file if it already exists.

Sidebar:
regenerate

Archive:
screen shot 2015-06-01 at 4 23 34 pm


aspidites:
Works as expected locally. :+1:

PR #1554, opened 2015-06-02
[PD-1403] Parse query params as unicode instead of ascii.

aspidites:
Changed what's passed to urlencode to prevent errors.

### DIFF
--- a/universal/helpers.py
+++ b/universal/helpers.py
@@ -23,11 +23,14 @@ def update_url_param(url, param, new_val):
     outputs:
     The new url.
     """
+
     url_parts = list(urlparse(url))
     parts = copy(url_parts)
-    query = dict(parse_qsl(parts[4]))
+    query = {k.encode('utf-8'): v.encode('utf-8')
+             for k, v in parse_qsl(parts[4])}
     query[param] = new_val
     parts[4] = urllib.urlencode(query)
+
     return urlunparse(parts)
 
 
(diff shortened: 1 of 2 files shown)


universal/helpers.py
```diff
@@ -23,11 +23,14 @@ def update_url_param(url, param, new_val):
     outputs:
     The new url.
     """
+
     url_parts = list(urlparse(url))
     parts = copy(url_parts)
-    query = dict(parse_qsl(parts[4]))
+    query = {k.encode('utf-8'): v.encode('utf-8')
+             for k, v in parse_qsl(parts[4])}
```
JLMcLaugh: Could you write a regression test for this?

aspidites: Yeah.

aspidites:
@JLMcLaugh 
Hmmm... I have to remember how to reproduce this before I can write a regression test. I tried running unicode through the helper on qc and master vs this one and don't get errors, so it seems to be something in whatever uses this helper. Unfortunately, I'm having a heck of a time getting saved searches to send locally on staging. In other words, it could be a minute before I'm able to do as you ask. I'm not ignoring you.

aspidites:
OK. Finally able to reproduce this.

aspidites:
May already be fixed on QC. Closing this in the mean time.

aspidites:
Correctly. Apparently if I am the recipient of the saved search, QC works, if @almathew is, it breaks. This PR fixes both cases though.

fezick:
Did regression tests get written for this? I don't see it in the commit.

aspidites:
@fezick no. Walking over to explain why...

aspidites:
Figured it out (kind of). Writing a unit test now.

aspidites:
@fezick @JLMcLaugh all: There's now a test for this case.

PR #1555, opened 2015-06-02
Removed destructive statements from SQL.

aspidites:
Removed the overwriting of column values. Tested on mysql and sqlite. I'd rather this was tested on qc before hotfixing master, if possible (Which means we need to add these values to at least one record to see if migrating backward then forward again breaks anything as it does in production).


tdphillips:
We'll need to roll back to 12 before merging, won't we?

aspidites:
@tdphillips  I don't think it matters, since the backward migration is the same. We'd just want to update records before forward migrating.

PR #1574, opened 2015-06-04
[PD-1415] Fixed regenerate.

aspidites:
Didn't notice before because contactrecord report doesn't rely on json. The fix was to not pass values to the serialize function when regenerating. In other words, json represents all pristine records and should only be modified on export (which I thought I was doing but obviously wasnt).

### DIFF
--- a/myreports/models.py
+++ b/myreports/models.py
@@ -66,8 +66,7 @@ class Report(models.Model):
     def regenerate(self):
         """Regenerate the report file if it doesn't already exist on disk."""
         if not self.results:
-            values = json.loads(self.values)
-            contents = serialize('json', self.queryset, values=values)
+            contents = serialize('json', self.queryset)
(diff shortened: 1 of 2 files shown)


aspidites:
Wrote unit tests.

aspidites:
Spoke with hal, and he wants this out by tomorrow morning. I'll make the qc and staging versions after this goes out, as to not make the PR page too busy.

aspidites:
@JLMcLaugh  I don't think there are any more sensible tests that can be written for regenerate. The other case to worry about is when the format of the report has changed, in which case, there isn't any sensible way to regenerate, and thus I would expect an error. The only manageable way I can see to prevent such errors in the future is not through regression tests, but rather by ensuring that there is a data migration to accompany changes to the JSON structure for reports (or that a manually written query takes that into account).

PR #1575, opened 2015-06-04
[PD-1402] Properly display session state

aspidites:
Check that there is a valid session before showing options for authenticated users.


aspidites:
The topbar template relies on both user and cookies, which makes it non-trivial (perhaps impossible) to render it properly by just using `Template.render` or `render`. The best I can think of to do along the lines of a unit test is to write a dummy template and ensure that one a user is logged in and there is no session data that one thing gets rendered, and when either of those is false, another thing gets rendered.

TL;DR - There's a reason that no topbar tests exist for the seo app.

PR #1576, opened 2015-06-05
[PD-1420] Black-listed colmns so they dont show up in reports.

aspidites:
# Note
I based this off of #1579 since that PR also changes unit tests. You can either  merge that, then this, or merge this and close that.

The only changes relevant to this PR are [here](https://github.com/DirectEmployers/MyJobs/pull/1576/files#diff-eb9601953df6f5b354db064401d8015cR253) and [here](https://github.com/DirectEmployers/MyJobs/pull/1576/files#diff-ebd3cabe61d5dc9f762eb5c7c5138120R289)

Before:
2015-06-05-074346_628x665_scrot
2015-06-05-074400_636x425_scrot
2015-06-05-074418_625x321_scrot

After:
2015-06-05-074540_629x630_scrot
2015-06-05-074553_628x280_scrot
2015-06-05-074604_637x216_scrot

~~Still need to write a test to ensure that on the download page,  blacklisted columns don't show up, but I wanted this up to let it be known that this is being worked on.~~

Writing tests for this now will cause a merge conflict, so I'll write them once #1574 is merged.

### DIFF
--- a/myreports/tests/test_views.py
+++ b/myreports/tests/test_views.py
@@ -250,6 +250,21 @@ class TestDownloads(MyReportsTestCase):
         self.assertEqual(response.context['columns'], {
             'Partner': True, 'Contact Name': True, 'Contact Type': True})
 
+    def test_blacklisted_columns(self):
+        """Test that blacklisted columns aren't visible."""
+        blacklist = ['pk', 'approval_status']
+        response = self.client.post(
+            path=reverse('reports', kwargs={
+                'app': 'mypartners', 'model': 'contactrecord'}),
+            data={'values': ['partner', 'contact__name', 'contact_type']})
+
+        report_name = response.content
+        report = Report.objects.get(name=report_name)
+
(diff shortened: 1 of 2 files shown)


aspidites:
I basically moved and modified a unit test, since it makes more sense as a test for the download page rather than the regeneration. Also, by looking closely you'll see that the test is doing almost the same thing, except rather than test for the absence of 1 column, it checks for the absence of all blacklisted columns. 

This is more of a stop-gap measure until we've had enough time to figure out the best way to represent this. The new solution would be part of what Hal terms the "translation layer".

tdphillips:
>Still need to write a test to ensure that on the download page, blacklisted columns don't show up, but I wanted this up to let it be known that this is being worked on.

Is this still valid? I see that you messed with a test but I'm unsure if it's the test this is referring to.

aspidites:
@tdphillips, the test is written (back when I removed the do not merge tag). Sorry for not updating the description.

PR #1581, opened 2015-06-08
Event queue <PD-763>

tdphillips:
Email tasks do not get rescheduled if a run fails for some reason.

Other than that, I think this is done.


myemails/models.py
```diff
@@ -158,9 +207,129 @@ class EmailTask(models.Model):
...
+
+
+# I don't really like doing it this way (get from database pre save, set
+# attribute, get again post save), but we need to be able to 1) determine what
+# was changed and 2) only send an email if the save is successful. - TP
+def cron_post_save(sender, instance, **kwargs):
+    cron_event_kwargs = {'model': ContentType.objects.get_for_model(sender),
+                         'owner': instance.product.owner if hasattr(
+                             instance, 'product') else instance.owner}
+    events = list(CronEvent.objects.filter(**cron_event_kwargs))
```
aspidites: Any reason for the explicit conversion? Are you meaning for the query to be evaluated immediately?

myemails/models.py
```diff
@@ -158,9 +207,129 @@ class EmailTask(models.Model):
...
+        value_event_kwargs = {'model': ContentType.objects.get_for_model(sender),
+                              'owner': instance.product.owner if hasattr(
+                                  instance, 'product') else instance.owner}
+        events = ValueEvent.objects.filter(**value_event_kwargs)
+        original = sender.objects.get(pk=instance.pk)
+        for event in events:
+            old_val = getattr(original, event.field)
+            new_val = getattr(instance, event.field)
+            compare = getattr(operator, event.compare_using, 'eq')
+            if not compare(old_val, event.value) and \
```
aspidites: pep 8 with the `\`

myemails/models.py
tdphillips: It's meant to be evaluated immediately. The list is modified on line 241.

myemails/models.py
tdphillips: I think perhaps pep 8 needs to be updated. \ is valid, except when it's not.

aspidites: Thats fine. For the record, I never need `\`. That line wouldn't stop me from merging, but backslashes still make me sad.

aspidites:
@JLMcLaugh I spent a couple of hours looking at this yesterday and this morning. I haven't been able to confirm that the queuing works as I don't have rabbitmq and celery set up locally and apparently there are things I need to be cognisant of doing when attempting to do so (and thus have intentionally avoided trying). However, I've looked through the code a few times and played with the admin. Things seem to be functioning as expected, aside from usability concerns which I've raised elsewhere that don't directly affect this ticket.

aspidites:
It's been a few days and no one else has commented. I'm going to assume that if no one comments by lunch, everyone is ok with what's here.

@fezick @JLMcLaugh @galitskyd, @almathew @mklauber

myemails/models.py
```diff
@@ -2,10 +2,17 @@
...
 from django.contrib.auth.models import ContentType
 from django.contrib.contenttypes.generic import GenericForeignKey
-from django.db import models
+from django.core.mail import send_mail
+from django.db import models, OperationalError
+from django.db.models.signals import pre_save, post_save
 from django.template import Template, Context
 from django.utils.translation import ugettext_lazy as _
 
+from myemails.signals import cron_post_save, value_pre_save, value_post_save, \
```
almathew: It hurts.

But don't change it.

aspidites: Style: 
```python
from myemails.signals import (cron_post_save, value_pre_sve, value_post_save,
                              post_add_invoice, pre_add_invoice)
```

While I generally don't like `\` to begin with, I think here it's a legitimate complaint.

myemails/models.py
```diff
@@ -73,15 +81,88 @@ def __unicode__(self):
...
+        Returns:
+        :return: The created EamilTask.
+        """
+        return EmailTask.objects.create(
+            act_on=for_object,
+            related_event=self,
+            scheduled_for=self.schedule_for(for_object)
+        )
+
+    def schedule_for(self, for_object):
```
aspidites: The documentation says that unless overridden, this defaults to right now, but I don't see any mechanism to override the time.

myemails/models.py
```diff
@@ -2,10 +2,19 @@
...
 from django.contrib.auth.models import ContentType
 from django.contrib.contenttypes.generic import GenericForeignKey
-from django.db import models
+from django.core.mail import send_mail
+from django.db import models, OperationalError
+from django.db.models.signals import pre_save, post_save
 from django.template import Template, Context
 from django.utils.translation import ugettext_lazy as _
 
+from myemails.signals import (
```
tdphillips: I don't like lots of tabs. Is this acceptable, even though it sucks up a bit more vertical space?

myemails/models.py
tdphillips: Clarified the docstring a bit. Subclasses should override schedule_for to make use of for_object.

myemails/models.py
```diff
@@ -73,15 +81,88 @@ def __unicode__(self):
...
         Sends an email for a for_object.
 
-        :param for_object: The object an email is being sent for.
-
+        Inputs:
+        :for_object: The object an email is being sent for.
         """
-        email_content = self.email_template.render(for_object)
-        # TODO: Write actual send email logic
+        subject = ''
```
aspidites: A lot of duplication here, but until things are wired up and usable, I'm not sure how best to change this other than to factor out `recipient_company`, since it's only different in the last case. At the same time, given the comment above them, I wonder if such a  refactor would be confusing.

myemails/models.py
```diff
@@ -73,15 +81,88 @@ def __unicode__(self):
...
+            raise NotImplementedError('Add %s here somewhere!' %
+                                      self.model.model)
+
+        if not subject:
+            subject = 'An update on your %s' % self.model.name
+
+        recipients = CompanyUser.objects.filter(
+            company=recipient_company).values_list('user__email',
+                                                   flat=True)
+        if hasattr(sending_company, 'companyprofile'):
```
aspidites: is `companyprofile` a foreign key?

tdphillips: It is, but each company doesn't necessarily have one.

myemails/models.py
```diff
@@ -158,9 +227,67 @@ class EmailTask(models.Model):
...
+Binds the provided sender to pre_save and post_save signals.
+
+Inputs:
+:type_: Type of event being bound (cron, value, created)
+:sender: Model being bound
+:pre: Method to be bound to the pre_save signal; default: None
+:post: Method to be bound to the post_save signal; default: None
+"""
+
+try:
```
aspidites: I don't like the size of this try. While the caught exception is very specific, it seems like a lot of lines which could fail in a number of ways. I've personally been trying to avoid exceptional cases or minimize how many lines are exceptional when that's not possible (see [here](https://github.com/DirectEmployers/MyJobs/blob/pd-1431/seo/breadbox.py#L219).

myemails/signals.py
```diff
@@ -0,0 +1,123 @@
...
+    :instance: Specific instance saved
+    """
+    CronEvent = get_model('myemails', 'CronEvent')
+    EmailTask = get_model('myemails', 'EmailTask')
+    content_type = ContentType.objects.get_for_model(sender)
+    cron_event_kwargs = {'model': content_type,
+                         'owner': instance.product.owner if hasattr(
+                             instance, 'product') else instance.owner}
+    # Gets all CronEvents for this company and model...
+    events = list(CronEvent.objects.filter(**cron_event_kwargs))
```
aspidites: I feel like I've asked before, but don't see an answer anywhere (still only half-way through my coffee though, so apologies ahead of time): Is there a reason we're converting to a list here? Are we forcing the SQL evaluation on purpose?

myemails/models.py
aspidites: Right, that was what I was getting at. Now that I know that, I know that something like `if not sending_company.companyprofile` wouldn't have been sufficient here. Thanks.

myemails/models.py
tdphillips: You're right. How would you feel about just `Model=...` and `PurchasedProduct=...` being in the try and everything else being in else? That's all that's strictly necessary.

myemails/signals.py
tdphillips: We modify this list later on. QuerySet doesn't have .pop() or .index().

myemails/models.py
aspidites: Kinda like how I do my line continuations. Inconsistent with other imports, but I like it better than the `\`. I'm fine with it if no one else objects.

On a completely unrelated note, in personal code, I've been just importing modules and fully qualifying the namespace. In this case, it'd be `from myemails import signals` and `signals.pre_add_invoice`. I haven't looked, but I assume that was avoided to prevent hitting the 79 char limit too early.

myemails/models.py
aspidites: In response to the struck-through text -- That would have been perfect.

myemails/signals.py
aspidites: Ah...(see below)

myemails/signals.py
```diff
@@ -0,0 +1,123 @@
...
+    # ...and all tasks scheduled for this instance.
+    tasks = EmailTask.objects.filter(
+        object_id=instance.pk,
+        object_model=content_type)
+    triggered_events = {task_.related_event for task_ in tasks}
+    for event in triggered_events:
+        if event in events:
+            # If an event is already scheduled, remove it from the list of
+            # events to be scheduled.
+            events.pop(events.index(event))
```
aspidites: Why not `events.remove(event)`?

tdphillips: No reason. I do something that works with the intention of fixing everything later, then forget about it.

aspidites: TODO's are great for that ;-)

PR #1582, opened 2015-06-08
Allow larger domain submission

almathew:
Sorry

@tdphillips and I discussed this a bit, and determined this would be the best way to do it. Some undetermined validator is throwing the error, and it really shouldn't be.

### DIFF
--- a/seo/forms/admin_forms.py
+++ b/seo/forms/admin_forms.py
@@ -4,6 +4,7 @@ from taggit.forms import TagField, TagWidget
 
 from django.conf import settings
 from django import forms
+from django.core import validators
 from django.contrib import admin
 from django.contrib.auth.models import Group
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


seo/forms/admin_forms.py
```diff
@@ -242,6 +245,15 @@ def __init__(self, data=None, user=None, *args, **kwargs):
         super(SeoSiteForm, self).__init__(data, user, *args, **kwargs)
         this = kwargs.get('instance')
 
+        # Override some undesired default behavior
+        if hasattr(self, 'errors') and 'domain' in self.errors:
+            bad_error = 'Ensure this value has at most 100 characters'
+            try:
+                if self.errors['domain'][0].startswith(bad_error):
+                    del self.errors['domain']
```
aspidites: This will still fail if domain isn't part of the errors dict.

almathew: This part of the code should never be reached if it isn't. Hence this `'domain' in self.errors` part

PR #1585, opened 2015-06-09
[PD-1419] normalize responsive design widths

galitskyd:
After talking with @fezick we wanted to just fix the topbar as going to a two device setup would diverge from bootstrap too much, especially with redesigns looming.


galitskyd:
screen shot 2015-06-09 at 10 19 24 am
tests pass

aspidites:
Just wanted to note that `is_ajax` doesn't work because it's a jsonp request, which technically isn't AJAX, hence why @galitskyd is checking whether a callback exists, which serves the same purpose.

aspidites:
@galitskyd: My mistake. I had @mklauber keep code that I deleted before. The `test_all_columns_available` is redundant and superceded by `test_blacklist_columns`

JLMcLaugh:
I discussed splitting topbar css into 2 static files with @galitskyd in person.

PR #1591, opened 2015-06-11
[PD-1437] Deleting a user shouldn't delete related objects

aspidites:
Basically I set the affected foreign keys to none with a SET_NULL on user delete.


fezick:
Since this is a hotfix-to-staging, we will also need a PR into staging.

aspidites:
@fezick I was going to do that after you had a chance to play with it in QC. Being a bit more careful with hot fixes.

fezick:
:thumbsup:

PR #1593, opened 2015-06-11
[PD-1142] PRM and Reporting only available to member companies.

aspidites:
I created a members_only decorator which 404's when the current company isn't a member company. I also wrapped all PRM and report  views with it, per @fezick and I's discussion. Not sure if any other views are desirable.


aspidites:
I haven't written new tests yet because I want to see if any old tests break. After updating those, I'll write one for the new decorator.

mypartners/views.py
```diff
@@ -52,6 +53,7 @@
 
 
 @warn_when_inactive(feature='Partner Relationship Manager is')
+@not_found_when_not_member
```
JLMcLaugh: Wouldn't it be easier to make this part of the logic in company_has_access('prm_access')? Aren't we just changing the definition of who has prm access?

aspidites: I was under the impression that @fezick wanted a decorator specifically for checking for member status of a company. Besides that, reporting won't be restricted to prm users, so coupling this to prm users probably isn't a good idea.

JLMcLaugh: The definition of who has prm access has changed. The logic that decides if
someone has prm access  lives in the has_access decorator, and that's what
should change. It we need separate logic for reporting, then we we should
do it with has_access('reporting'). We should minimize business logic that
needs to get added to a dozen different views. If we need to add more
parameters to has_access to add different levels of access within app
later, that's fine, but this sounds like it's neatly divided between apps.

aspidites:
Didn't bother writing a test for this incarnation since it's the same decorator that's always been there.

aspidites:
Doing a bit of refactoring per @JLMcLaugh and I's discussion.

PR #1594, opened 2015-06-12
[PD-1442][PD-1440][PD-1443] Makes sure password change is required after invitation accepted

aspidites:
Basically, when a user accepts an invitation, they are given a temporary password, but never asked to change it. This corrects that. As for the hackish nature of how I accomplish that, please refer to the associated tickets. It's more of a stop-gap measure, given how much work *really* fixing account activation would be.

# new account activation template
2015-06-12-164539_1052x1056_scrot

### DIFF
--- a/registration/views.py
+++ b/registration/views.py
@@ -99,6 +99,9 @@ def activate(request, activation_key, invitation=False):
                 activated.in_reserve = False
                 password = User.objects.make_random_password()
                 activated.set_password(password)
+                # Without this, user isn't prompted for a new password when
+                # invited to use My.jobs. See PD-1444
+                activated._User__original_password = activated.password
                 activated.save()
                 ctx['password'] = password
     template = 'registration/%s/activate.html' % settings.PROJECT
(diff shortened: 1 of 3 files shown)


JLMcLaugh:
Do not merge until auth changes on https://github.com/DirectEmployers/MyJobs/pull/1593 are finished.

aspidites:
@JLMcLaugh how is #1593 related to this? This fixes something that's been broken where as that one changes what it means to have access to certain pages. Wait...looks like changes made it in that shouldn't have. That's what I get for working on too many tickets at once.

aspidites:
@JLMcLaugh  can you take another look at this? Removed the erroneous commits.

registration/views.py
```diff
@@ -99,6 +99,9 @@ def activate(request, activation_key, invitation=False):
                 activated.in_reserve = False
                 password = User.objects.make_random_password()
                 activated.set_password(password)
+                # Without this, user isn't prompted for a new password when
+                # invited to use My.jobs. See PD-1444
+                activated._User__original_password = activated.password
```
tdphillips: How does this line accomplish that? I'm seeing `__original_password` but no `_User__original_password`

aspidites: Python name mangling of `__*` members.

PR #1595, opened 2015-06-12
[PD-1437] Hotfix for staging

aspidites:
Just like #1591, but branched off of staging. Marking do not merge until fix can be confirmed in QC.


fezick:
Confirmed on staging. Deleting a user does not cascade to their saved search contact_record instances.

PR #1602, opened 2015-06-15
[PD-1449] More Unicode, More Problems

aspidites:
Tested this against the test cases in the ticket, @almathew's regression test, and the tests for pd-1403.

Will hot fix against staging/master (whichever is determined we need) once tested on the QC server b y someone other than me.


aspidites:
For the curious:

I reverted the code back to what it was before #1554 with  the two new unit tests and `test_unicode_in_saved_search` fails while this test passes. With the code as it is in master, the saved search test passes, though this test fails. With this PR, both pass.

PR #1603, opened 2015-06-15
Re-added fl params

almathew:
Re-adds an improvement I wrote and then had to quickly un-write because it didn't work properly. Added a test for syndication feed because that's where we saw all the problems.

### DIFF
--- a/myblocks/context_tools.py
+++ b/myblocks/context_tools.py
@@ -149,8 +149,7 @@ def get_job(request, job_id):
 def get_jobs_and_counts(request):
     # Text uses html_description instead of just description.
     fl = list(helpers.search_fields)
-    index = fl.index('description')
-    fl.pop(index)
+    fl.remove('description')
     fl.append('html_description')
 
     filters = get_filters(request)
--- a/myblocks/context_tools.py
+++ b/myblocks/context_tools.py
@@ -151,10 +151,7 @@ def get_jobs_and_counts(request):
     fl = list(helpers.search_fields)
     index = fl.index('description')
     fl.pop(index)
-    # Don't bother getting the description if we're not going to
-    # show the description to the user.
-    if get_query_string(request):
-        fl.append('html_description')
+    fl.append('html_description')
 
     filters = get_filters(request)
     site_config = get_site_config(request)
--- a/myblocks/context_tools.py
+++ b/myblocks/context_tools.py
@@ -147,13 +147,23 @@ def get_job(request, job_id):
 
 @Memoized
 def get_jobs_and_counts(request):
+    # Text uses html_description instead of just description.
+    fl = list(helpers.search_fields)
+    index = fl.index('description')
+    fl.pop(index)
+    # Don't bother getting the description if we're not going to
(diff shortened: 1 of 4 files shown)


almathew:
Going to mess around with description/html_description some.

almathew:
Ok, it's fixed. Turns out we only use `html_description` on the job detail page, which is only searching by guid (and therefore bypasses all the fl stuff) anyway.

@JLMcLaugh

seo/views/search_views.py
```diff
@@ -1706,8 +1706,17 @@ def search_by_results_and_slugs(request, *args, **kwargs):
...
                 facet_blurb_facet = active_facets[0]
 
+    # Text uses html_description instead of just description.
+    fl = list(helpers.search_fields)
+    index = fl.index('description')
+    fl.pop(index)
+    # Don't bother getting the description if we're not going to
+    # show the description to the user.
+    if query_path:
+        fl.append('html_description')
```
almathew: I got clever here, which given the history of this improvement may not have been a good idea, and stopped getting the description at all if we're not going to use it. Not sure if it's a good idea but *I* like it.

aspidites:
@almathew  is there a reference to the original improvement somewhere? Ticket number? Old PR?

almathew:
PD-1475 is the one I opened this morning. That one links to the original ticket. There are a bunch of hotfixes scattered around that undo the work from the original ticket.

myblocks/context_tools.py
```diff
@@ -147,13 +147,20 @@ def get_job(request, job_id):
 
 @Memoized
 def get_jobs_and_counts(request):
+    # Text uses html_description instead of just description.
+    fl = list(helpers.search_fields)
+    index = fl.index('description')
```
aspidites: `fl.remove('description')` is the same as `fl.pop(fl.index('description'))`

seo/views/search_views.py
```diff
@@ -1706,8 +1706,17 @@ def search_by_results_and_slugs(request, *args, **kwargs):
             if len(active_facets) == 1 and active_facets[0].blurb:
                 facet_blurb_facet = active_facets[0]
 
+    # Text uses html_description instead of just description.
+    fl = list(helpers.search_fields)
+    index = fl.index('description')
+    fl.pop(index)
+    # Don't bother getting the description if we're not going to
+    # show the description to the user.
+    if query_path:
```
JLMcLaugh: Can you tweak this comment to explain why we only show description highlights when we have query terms to match? It's not clear why having a query_path means we're not showing the description to the user unless you're familiar with the history of this search view.

almathew:
@JLMcLaugh 's concern has been addressed

aspidites:
My comment still hasn't been addressed and @almathew isn't here anymore. Anyone object to my changing it directly (which would then disqualify me from merging this myself)?

aspidites:
Made my change, so I can't merge this, but it looks like everything that was commented on has now been addressed. @mklauber @galitskyd, @JLMcLaugh, @tdphillips

PR #1604, opened 2015-06-16
[PD-1142] has_access decorator + data migration

aspidites:
- `has_access` decorator which takes a feature name and checks all permission requirements for that feature before displaying the view, otherwise 404ing. 
- data migration that sets all companies' `prm_access` field to whatever their `member` attribute is. 

Basically, the has_access factory maintains a dictionary of feature -> conditions which are checked before showing the  view. `company_has_access` is still around because postajob is using it and I didn't want to rock the boat that much.

### DIFF
--- a/seo/models.py
+++ b/seo/models.py
@@ -731,7 +731,7 @@ class Company(models.Model):
                                                     blank=True)
 
     # Permissions
-    prm_access = models.BooleanField(default=True)
+    prm_access = models.BooleanField(default=False)
     product_access = models.BooleanField(default=False)
     posting_access = models.BooleanField(default=False)
     user_created = models.BooleanField(default=False)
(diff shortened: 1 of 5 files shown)


seo/models.py
```diff
@@ -731,7 +731,7 @@ def company_user_count(self):
                                                     blank=True)
 
     # Permissions
-    prm_access = models.BooleanField(default=True)
+    prm_access = models.BooleanField(default=False)
```
almathew: This isn't available in admin, and your migration only handles existing members. How will this be handled for new members?

I'm in favor of adding it to the admin (right by the member checkbox) personally.

aspidites: Better question: @JLMcLaugh If we are saying that prm_access and member are essentially synonyms, why am I not just removing prm_access again?

aspidites: @almathew [this](https://github.com/DirectEmployers/MyJobs/pull/1604/files#diff-491aac1eccefcbab90ed7d8bf7dc2f66R772) should address your admin concern without having to modify admin.

PR #1605, opened 2015-06-16
A few improvements to mx records <PD-608>

tdphillips:
I'm currently running through all sites using a form of this and it appears to be working.

As such, it's probably quicker to wait for Jenkins to run tests than to wait for my process to finish so I can.

### DIFF
--- a/seo/models.py
+++ b/seo/models.py
@@ -533,7 +533,7 @@ class SeoSite(Site):
         Ensures that an MX record exists for a given domain, if possible.
         This allows the domain as an option for email_domain.
         """
-        if mail.outbox:
+        if not hasattr(mail, 'outbox'):
             # This is being run in a testing environment; don't try creating
             # an MX record.
             try:
--- a/seo/models.py
+++ b/seo/models.py
@@ -534,8 +534,7 @@ class SeoSite(Site):
         This allows the domain as an option for email_domain.
         """
         if not hasattr(mail, 'outbox'):
-            # This is being run in a testing environment; don't try creating
-            # an MX record.
+            # Don't try creating MX records when running tests.
             try:
                 can_send = can_send_email(self.domain)
                 if can_send is not None and not can_send:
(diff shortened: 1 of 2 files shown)


seo/models.py
```diff
@@ -525,6 +528,24 @@ def all_sites(self):
...
+        """
+        Ensures that an MX record exists for a given domain, if possible.
+        This allows the domain as an option for email_domain.
+        """
+        if mail.outbox:
+            # This is being run in a testing environment; don't try creating
+            # an MX record.
+            try:
+                can_send = can_send_email(self.domain)
+                if can_send is not None and not can_send:
```
aspidites: Instead of this, can we ensure that can_send_email always returns a bool then reduce this to `if not can_send`?

tdphillips: Then we have no way of telling if we own the domain or not. At the moment, can_send_email tells us one of three things - 1) We don't own the domain, 2) We own the domain and there is no MX record, 3) We own the domain and an MX record exists.

Put it another way, if can_send_email were to always return False any time there is no MX record, we may try to create a record for a domain we don't own and fail horribly.

aspidites: Ah. Seems like `can_send_email` is a bit too coarse then. Seems we would much rather something akin to `if we_own_domain and not mx_record_exists`. I'd rather two members that returned booleans being joined with a logical operator than a quasi tri-state-variable being used. 

Not sure if others agree, though. @mklauber, @galitskyd, @fezick @JLMcLaugh?

aspidites:
Other than my type complaints, this seems fine.

seo/route53.py
```diff
@@ -81,15 +81,11 @@ def can_send_email(domain):
 
     """
     if not domain_exists(domain):
-        return False
+        return None
```
JLMcLaugh: If we're using None to represent a different state than False, then that should be explained in the docstring.

PR #1607, opened 2015-06-18
Prevent saved searches with no recipient from being inserted in solr <PD-1465>

tdphillips:
Tests passed locally.


aspidites:
Looks right. Waiting on reply from IRC about regression test to ensure things aren't added to solr when user is None.

PR #1608, opened 2015-06-24
Rewrote the ETL-to-solr code to use generators

mklauber:
This will keep memory usage lower, as we should only need to have 
1024 jobs in memory at a time.


import_jobs.py
```diff
@@ -49,14 +49,31 @@ def update_job_source(guid, buid, name, clear_cache=False):
...
-        job['link'] = make_redirect(job, bu).make_link()
-    add_jobs(jobs)
-    remove_expired_jobs(buid, jobs)
+    jobs = (hr_xml_to_json(job, bu) for job in jobs)
+    jobs = (add_redirect(job, bu) for job in jobs)
+    
+    # AT&T Showed that large numbers of MOCs can cause import issues due to the size of documents.
+    # Therefore, when processing AT&T lower the document chunk size.
+    if int(buid) == 19389:
+        logger.warn("AT&T has large amounts of mapped_mocs, that cause problems.  Reducing chunk size.")
```
aspidites: Are we guaranteed to never receive a None or other non-digit? Otherwise int(buid) will fail. I looked as far as search_views.py where buid is retrieved from json, but don't know enough about those json files to know if buid will ever be nul.

import_jobs.py
```diff
@@ -64,6 +81,27 @@ def update_job_source(guid, buid, name, clear_cache=False):
         tasks.task_clear_bu_cache.delay(buid=bu.id, countdown=1500)
 
 
+def chunk2(iterable, chunk_size=1024):
```
aspidites: was the chain + islice version less performant here than the current chunk2 implementation? Also, can this just be chunk now?

import_jobs.py
```diff
@@ -64,6 +81,27 @@ def update_job_source(guid, buid, name, clear_cache=False):
...
+            for _ in range(chunk_size):
+                result.append(iterable.next())
+            yield result
+        except StopIteration:
+            yield result
+            raise StopIteration
+
+
+def add_redirect(job, bu):
+    job['link'] = make_redirect(job, bu).make_link()
```
aspidites: Does make_link ever fail?

import_jobs.py
```diff
@@ -601,33 +638,6 @@ def remove_expired_jobs(buid, active_jobs, upload_chunk_size=1024):
     return expired
 
 
-def add_jobs(jobs, upload_chunk_size=1024):
```
tdphillips: This gets imported from seo/views/search_vews.py, at the very least.

import_jobs.py
aspidites: Perhaps more detailed documentation, as "makes chunks from list" doesn't tell me much, especially since it's a generator (as opposed to returning a list of lists of a given chunk size). 

Random note: I suspect groupby would work as well, though I haven't played with it. Not a requirement though, since this works. Just noting.

import_jobs.py
mklauber: Technically, we should fail far before this, but another portion of the improve imports ticket was to put assertions on the guid/buid/name stuff, so add those to this PR.

import_jobs.py
mklauber: Dang.  I thought I'd checked on that.  Kk, I'll bring it back, and make sure it works with everything else.

aspidites:
Also, what's the significance of 1024, other than it's a nice number computer wise? (number of KB in a MB, etc)

mklauber:
@aspidites 1024?  Nothing especially other than the fact that it's the right order of magnitude.  bump it up too much and you start getting memory issues and solr starts dropping requests, going smaller adds more time spent opening new requests to solr.

aspidites:
@mklauber  re: 1024 - That's good enough for me.

tdphillips:
:+1:

@aspidites, any more feedback?

import_jobs.py
```diff
@@ -609,24 +631,20 @@ def add_jobs(jobs, upload_chunk_size=1024):
         :jobs: A list of solr-ready, json-formatted jobs.
 
     outputs:
-        The number of jobs loaded into solr.
+        The lids of jobs loaded into solr.
```
JLMcLaugh: typo

mklauber: fixed

import_jobs.py
```diff
@@ -609,24 +631,20 @@ def add_jobs(jobs, upload_chunk_size=1024):
         :jobs: A list of solr-ready, json-formatted jobs.
 
     outputs:
-        The number of jobs loaded into solr.
+        The ids of jobs loaded into solr.
     """
     conn = Solr(settings.HAYSTACK_CONNECTIONS['default']['URL'])
-    num_jobs = len(jobs)
     # AT&T Showed that large numbers of MOCs can cause import issues due to the size of documents.
     # Therefore, when processing AT&T lower the document chunk size.
```
JLMcLaugh: This comment should be removed also.

mklauber: Just have to keep finding things... :-P

import_jobs.py
```diff
@@ -37,6 +38,11 @@
 def update_job_source(guid, buid, name, clear_cache=False):
```
JLMcLaugh: Are there any tests on this function? We should get at least the same level of coverage we have on update_solr before making this many changes.

aspidites:
Sorry. Missed this entirely. Late now, but I would have been a :+1 anyway.

PR #1609, opened 2015-06-24
Merge redirect [PD-1466]

tdphillips:
Just opening for testing. Tests passed locally, however.


tdphillips:
Remaining: Potentially update all settings files to have an archive database.

README.md
```diff
@@ -0,0 +1,32 @@
+My.jobs URLs
```
tdphillips: This and the existing readme should be merged but they both contain references to DE Foundation and both say that the respective projects are alpha releases.

conf/etc/cron.hourly/logrotate
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
tdphillips: This and logrotate.hourly.conf should possibly be moved to deployment

default_settings.py
```diff
@@ -273,6 +273,7 @@
     'saved_search',
     'taggit',
     'fsm',
+    'compressor',
```
tdphillips: I could move everything django-compressor to redirect_settings.py but those settings wouldn't be loaded when running tests. This (and the others) should allow tests to be run using any settings file.

deploy/settings.dseo_prod_admin.py
```diff
@@ -24,6 +24,14 @@
         'HOST': 'db-redirectqc.c9shuxvtcmer.us-east-1.rds.amazonaws.com',
         'PORT': '3306',
     },
+    'archive': {
```
tdphillips: Archive for everyone! I don't know what all strictly needs access to the new database but SEO does use the Redirect model so it is needed somewhere..

deploy/settings.jenkins-single-test.py
```diff
@@ -33,7 +41,7 @@
 
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
+        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
```
tdphillips: Redirect actually caches stuff, which the dummy cache doesn't do.

redirect/migrations/0001_initial.py
```diff
@@ -0,0 +1,114 @@
+# -*- coding: utf-8 -*-
```
tdphillips: I didn't mark this as having migrations as all of these have already been applied.

redirect/models.py
```diff
@@ -0,0 +1,368 @@
+from django.conf import settings
```
tdphillips: Not all of these models are used and I would like to remove them but that's a task for a later time.

redirect/views.py
```diff
@@ -0,0 +1,366 @@
...
+from django.utils import text, timezone
+from django.views.decorators.csrf import csrf_exempt
+
+from redirect.models import (Redirect, CanonicalMicrosite,
+                             DestinationManipulation, CompanyEmail,
+                             EmailRedirectLog)
+from redirect import helpers
+
+
+def home(request, guid, vsid=None, debug=None):
```
tdphillips: We now have two views named home in different urls.

redirect/views.py
```diff
@@ -0,0 +1,366 @@
...
+    num_instances += dms.count()
+
+    data.pop('error')
+    data['updated'] = num_instances
+    data['new_bu'] = new
+    return HttpResponse(json.dumps(data),
+                        content_type='application/json')
+
+
+def redirect_404(request):
```
tdphillips: Redirect has its own 404 template that is distinct from secure and seo.

redirect/views.py
```diff
@@ -0,0 +1,366 @@
...
+    return HttpResponseNotFound(loader.render_to_string(
+        'redirect/404.html', context_instance=RequestContext(request)))
+
+
+def redirect_500(request):
```
tdphillips: Redirect has always pulled topbar with ajax. I've done the minimum possible to get this working, but it might be worth revisiting if we ever get rid of jcnlx.com.

Rather than overriding our overridden 500 template with the base 500 template, we could enable native topbar use.

seo/models.py
```diff
@@ -125,42 +125,6 @@ def sq_from_terms(terms):
...
-                                              'added')
-    expired_date = models.DateTimeField(blank=True, null=True,
-                                        help_text='Date that this job was '
-                                                  'marked as expired')
-    job_location = models.CharField(max_length=255, blank=True)
-    job_title = models.CharField(max_length=255, blank=True)
-    company_name = models.TextField(blank=True)
-
-    class Meta:
-        db_table = 'redirect_redirect'
```
tdphillips: This (and all of the other deletions from the seo app) referenced the Redirect model from the redirect app.

static/my.jobs.js
```diff
@@ -0,0 +1,19 @@
+$(document).ready(function() {
```
tdphillips: This could probably use a rename but it doesn't currently conflict with anything.

wsgi.py
```diff
@@ -0,0 +1,16 @@
+import os, sys
```
tdphillips: I'm not sure if we need two wsgi files, but they're both slightly different.

tdphillips:
There may be a few issues to be found but I feel confident enough that nothing would break if this got merged.

tdphillips:
Despite what this PR says, the last manually-started test run passed.
http://ec2-54-172-15-197.compute-1.amazonaws.com/jenkins/job/MyJobs-Single-Test/76/

README.md
aspidites: Any reason to not do so in this PR?

deploy/settings.jenkins-single-test.py
aspidites: Implications for local environments? Do we need to change anything?

redirect/models.py
aspidites: Did you make a ticket for the later removal of these models? Probably should lest we forget they can be/need to be removed.

redirect/views.py
aspidites: Fully qualified paths should take care of that. If we wanted to be more explicit, I guess we could use redirects_home instead.

static/my.jobs.js
aspidites: @galitskyd ?

aspidites:
@tdphillips did you ever merge the tests from the other PR into this one?

README.md
tdphillips: The descriptions aren't really compatible. I can dump one into the other but I bet someone would be unhappy by that decision.

deploy/settings.jenkins-single-test.py
tdphillips: I don't think there are any changes that need to be made. If there are, it would be limited to adding your own LocMemCache to settings.

redirect/models.py
tdphillips: I did not. When I wanted to previously, I was informed that it wasn't a priority.

Above my pay grade.

tdphillips:
@aspidites I did not.
https://github.com/DirectEmployers/MyJobs/pull/1614

README.md
tdphillips: Added the MyJobs URLs header to the secure README. Left everything else as-is.

PR #1614, opened 2015-06-29
Get job from db if it's not in solr <PD-1477>

tdphillips:
This could use a review. I don't think I missed anything but I just might be the most knowledgeable person on blocks now.

Depends on the merge.
PR without merge commits:
https://github.com/DirectEmployers/MyJobs/compare/merge-redirect...job-detail-fallback

All tests pass.


aspidites:
@tdphillips mind syncing this with qc so it doesn't look like such a monsterous PR?

tdphillips:
Synced.

myblocks/models.py
```diff
@@ -769,6 +770,14 @@ def handle_job_detail_redirect(self, request, *args, **kwargs):
         job = context_tools.get_job(request, job_id)
 
         if not job:
+            if job_id:
```
aspidites: I don't see an else clause, so I suspect this can be collapsed into `if not job and job_id`, which would also save you some indentation.

myblocks/models.py
```diff
@@ -769,6 +770,14 @@ def handle_job_detail_redirect(self, request, *args, **kwargs):
...
+                search_type = 'guid' if len(job_id) > 31 else 'uid'
```
aspidites: Really like the else here. More obvious which bits can fail. :+1:

seo/views/search_views.py
```diff
@@ -377,6 +378,12 @@ def job_detail_by_title_slug_job_id(request, job_id, title_slug=None,
         the_job = DESearchQuerySet().narrow("%s:(%s)" % (search_type,
                                                          job_id))[0]
     except IndexError:
+        try:
```
aspidites: Can we comment this? Something like "foo wasn't in list, so we try this other approach". Also, I wonder if instead of try/except that just passes, I think you can do `the_job = Redirect.objects.filter(**{search_type: job_id}).first() or the_job`.

```

Comment (tdphillips) [inline seo/views/search_views.py]: If we reach this point, the_job isn't defined. It would probably have to look a little more like

```
the_job = Redirect.objects.filter(...).first()
if the_job:
    return HttpResponseRedirect(...)
return dseo_404(...)
```

aspidites: Yep. Didn't see the line below with te dseo_404 return.

myblocks/models.py
tdphillips: There is an implicit else on `if job_id` however.

seo/views/search_views.py
tdphillips: I don't like duplicating this bit of code but I don't really know a good way around it.

PR #1615, opened 2015-06-30
[PD-1476] Stopped appending view source to job link if it's a mailto.

aspidites:
Bug is fixed, but I'm still trying to wrap my head around enough of how seo + solr + tests work in order to properly write a regression test for this. As far as I can tell, the Apply view isn't tested at all.


JLMcLaugh:
The apply link is build on the job detail page, so checking that it's the right format there should be all you need. The link itself is handled by redirects.

aspidites:
Have to properly call the view with prerequisites, since that's where the error happens (eg, a simple render of a template wont cut it).

aspidites:
Might be able to piggy back off of test_job_detail.

PR #1617, opened 2015-07-01
Selenium! <PD-1471><PD-1520><PD-1521><PD-1522>

tdphillips:
To be able to run Selenium tests, you need to first install phantomjs using whatever package manager you have. The Ubuntu package is called "phantomjs".

I'm probably going to add more checking to ensure this doesn't get run on production. In the meantime, please don't try.

To run on whatever you have in settings.py, execute `./manage.py test_posting` and admire the results.

To run on QC, execute `SETTINGS=QC ./manage.py test_posting`.

To run on Staging, execute `SETTINGS=STAGING ./manage.py test_posting`.

The tests in functional_tests rarely segfault and I'm unsure why.


functional_tests/test_mypartners.py
```diff
@@ -1,9 +1,51 @@
...
+
+@override_settings(DEBUG=True)
+class SeleniumTestCase(LiveServerTestCase):
+
+    """ Adds Selenium to LiveServerTestCase. """
+
+    @classmethod
+    def setUpClass(cls):
+        cls.browser = webdriver.PhantomJS()
+        cls.browser.set_window_size(1120, 550)
```
tdphillips: There's apparently a bug from 2013 whereby elements are inaccessible when using PhantomJS if you don't change the window size.

functional_tests/test_mypartners.py
```diff
@@ -33,25 +75,31 @@ def test_cant_log_in_without_account(self):
...
         # register
-        self.find('id_email').send_keys(user.email)
-        self.find('id_password1').send_keys(user.password)
-        self.find('id_password2').send_keys(user.password)
+        self.find('id_email').send_keys('foobar1@baz.com')
+        self.find('id_password1').send_keys('aaAA11..')
+        self.find('id_password2').send_keys('aaAA11..')
         self.find('register').click()
 
-        self.assertEqual(self.find('profile').get_attribute(
```
tdphillips: My assumption is that this test was written before the second page of profile creation was turned into an ajax result. The added block below waits for ajax to finish, or 10 seconds pass.

myemails/models.py
```diff
@@ -277,7 +277,7 @@ def send_email(self):
         content_type = model if model != 'invoice' else 'purchasedproduct'
         model_map[model] = ContentType.objects.get(
             model=content_type).model_class()
-except OperationalError:
+except (OperationalError, ProgrammingError):
```
tdphillips: Unrelated to the ticket at hand. I tested with sqlite. MySQL throws a ProgrammingError here.

functional_tests/test_mypartners.py
```diff
@@ -33,25 +75,31 @@ def test_cant_log_in_without_account(self):
...
+        # making it work. - TP
         self.assertEqual(username.get_attribute('placeholder'),
-                         'Please enter a correct email.')
+                         'Email')
 
     def test_user_registration(self):
         """
         As John, I should be able to register on My.jobs and log in.
         """
-        user = UserFactory.build(email='foobar1@baz.com')
```
tdphillips: You can't create a user if that user already exists.

functional_tests/test_mypartners.py
```diff
@@ -33,25 +75,31 @@ def test_cant_log_in_without_account(self):
...
         self.browser.get('/'.join([self.live_server_url, 'prm', 'view']))
 
         # register
-        self.find('id_email').send_keys(user.email)
-        self.find('id_password1').send_keys(user.password)
```
tdphillips: I'm not using an existing user anyway, but this would be passing a hashed password.

functional_tests/test_posting.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class JobPostingTests(TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.OVERRIDES = {}
+        cls.browser = webdriver.PhantomJS()
+        if 'qc' in sys.argv:
+            print 'Running test_posting with QC settings'
+            qc = imp.load_source('settings.myjobs_qc',
```
tdphillips: Good? Bad? Ugly?

functional_tests/test_mypartners.py
aspidites: I believe you, but link to the bug?

functional_tests/test_mypartners.py
aspidites: No, I just couldn't come up with a reliable way to ensure that the ajax was done at the time. Admittedly, I didn't try very hard as at the time, we weren't even serious about selenium anyway.

functional_tests/test_mypartners.py
tdphillips: https://github.com/ariya/phantomjs/issues/11637

I haven't actually tried without it, but it's only one line.

functional_tests/test_mypartners.py
```diff
@@ -33,25 +75,31 @@ def test_cant_log_in_without_account(self):
         self.find('id_password').send_keys(self.user.password)
         self.find('login').click()
 
+        # Useless check - placeholder before and after is Email. I'm just
+        # making it work. - TP
```
aspidites: Didn't used to be the case. The login page has changed a lot since I initially wrote this. I'd prefer that new checks were written that fulfill the docstring.

functional_tests/test_mypartners.py
aspidites: IIRC, the original intent was that there was a test that there was (or was going to be) a test that would ensure that you couldn't register with a username which already exists.

functional_tests/test_mypartners.py
aspidites: Oops. I guess I thought that this referred to factory, not the model instance. Obviously the factory proxies to a model instance.

functional_tests/test_mypartners.py
tdphillips: I should have noted that. I wasn't planning on leaving it like this.

functional_tests/test_posting.py
```diff
@@ -0,0 +1,42 @@
...
+
+from django.conf import settings
+
+from functional_tests.test_mypartners import SeleniumTestCase
+from myjobs.tests.factories import UserFactory
+from myjobs.models import User
+from seo.tests import patch_settings
+
+
+class JobPostingTests(TestCase):
```
tdphillips: My goal for the automated postajob tests is to enable us to run them on various relevant servers by just changing an environment variable.

Running `./manage.py test functional_tests.test_posting` will default to what's in your settings.py (hopefully local). Running `SETTINGS=QC ./manage...` will run against QC. Running `SETTINGS=STAGING ./manage...` will run against staging.

tdphillips: Though this brings up a point. If you were previously developing on production and then ran functional tests without specifying SETTINGS, bad things may happen. Should I perhaps default the base case to something else?

aspidites: I do something like this for my environment. If not specified, I default my settings to point to local stuff.

tdphillips: Everyone probably has different local database names. I'm thinking about just breaking if settings.ENVIRONMENT is set to "Production".

aspidites: Right. I guess I was just commenting on my approval of sane default sthat don't  break stuff.

functional_tests/test_mypartners.py
aspidites: OK. :+1:  then.

myemails/models.py
aspidites: If you merge with QC this will go away (I merged your other PR).

postajob/management/commands/test_posting.py
```diff
@@ -0,0 +1,191 @@
...
+    # It is less of a headache to set the password now and toggle
+    # password_change than it is to pass password into the user creation call,
+    # determine if the password was set or if we short-circuited, and then act
+    # on that determination.
+    user.set_password(password)
+    user.password_change = False
+    user.is_active = True
+    user.is_verified = True
+    user.save()
+    user.raw_password = password
```
aspidites: What is this? I've never seen it, nor can I get it to work ('User' object has no attribute 'raw_password').

aspidites:
Finally finished reviewing.

postajob/management/commands/test_posting.py
tdphillips: `raw_password` doesn't exist outside of this file.

We've complained before about using setattr to set new attributes rather than assigning them like this.

tdphillips:
After several runs, this mass appears to clean up after itself properly. Once my current in-progress test is completed, I would like to run this against QC.

Yea? Nay? Speak before the final test is completed or live with the consequences.

postajob/management/commands/test_posting.py
```diff
@@ -0,0 +1,459 @@
...
+from selenium.webdriver.support.wait import WebDriverWait
+
+from myjobs.models import User
+from postajob.models import SitePackage, Job, Product, ProductGrouping, \
+    ProductOrder
+from seo.models import Company, CompanyUser, SeoSite, Configuration
+from seo.tests import patch_settings
+
+
+class Command(NoArgsCommand):
```
aspidites: Think I mentioned this in one of @mklauber 's prs: NoArgsCommand is being deprecated.

@JLMcLaugh what's the likelihood that we will **ever** use django >= 1.8?

postajob/management/commands/test_posting.py
```diff
@@ -0,0 +1,459 @@
...
+@skipUnless('test_posting' in sys.argv, 'Selenium tests are incompatible '
+            'with Django tests')
+class JobPostingTests(TestCase):
+    OVERRIDES = {}
+    CREATION_ORDER = []
+    test_url = 'localhost'
+    test_port = ''
+
+    @classmethod
+    def setup_objects(cls):
```
aspidites: Reason this can't be done in an actual setup method?

postajob/management/commands/test_posting.py
tdphillips: That may be the case, but we're on 1.6 right now. According to the 1.8 docs, fixing it is as simple as replacing NoArgsCommand with BaseCommand. Trying that in 1.6, however, raises a NotImplementedError.

aspidites: Ah. So we'd have to fix it after the fact regardless. Lame. 

:+1:

postajob/management/commands/test_posting.py
tdphillips: I haven't needed to destroy/recreate anything yet. The first part is accurate, however.

postajob/management/commands/test_posting.py
```diff
@@ -0,0 +1,523 @@
...
+        self.get(view_request)
+
+        request = self.admin_company.request_set.get()
+
+        # The approve button submits a form whose action does not include
+        # a domain override. Add one. :(
+        request_approve = reverse('approve_admin_request',
+                                  kwargs={'pk': request.pk})
+        self.browser.execute_script(
+            '$("form[action=\'%s\']").attr("action", "%s");' % (
```
tdphillips: :sob:

tdphillips:
What has this taught us? Domain overriding is not fun.

tdphillips:
I think the code changes are finished. This could use a good review - it will touch qc and staging databases. If it doesn't clean up after itself (It has every time I ran it), things will happen.

If this is good to go, the next step is writing two Jenkins tasks.

postajob/management/commands/test_posting.py
```diff
@@ -0,0 +1,515 @@
...
+            cls.OVERRIDES = vars(qc)
+        elif environment == 'staging':
+            print 'Running test_posting with staging settings'
+            cls.test_url = 'staging.www.my.jobs'
+            staging = imp.load_source('settings.myjobs_staging',
+                                      'deploy/settings.myjobs_staging.py')
+            cls.OVERRIDES = vars(staging)
+        else:
+            assert (settings.DATABASES['default']['HOST'] !=
+                    'db-redirect.c9shuxvtcmer.us-east-1.rds.amazonaws.com'), \
```
JLMcLaugh: We should probably pull this name directly from a production settings file in the deploy folder.

postajob/management/commands/test_posting.py
```diff
@@ -0,0 +1,515 @@
...
+        """
+        data = data or {}
+        requested_url = 'http://{domain}{port}{path}'.format(
+            domain=cls.test_url, port=cls.test_port, path=path)
+        if domain:
+            requested_url += '?domain=%s' % domain
+        cls.browser.post(requested_url, data=data)
+
+    @classmethod
+    def get(cls, path='/', domain=None):
```
JLMcLaugh: We'll probably have to modify this to preserve query parameters in the future, but this is ok for now.

postajob/management/commands/test_posting.py
tdphillips: I pulled production settings to compare. Selenium tests passed.

PR #1618, opened 2015-07-01
[PD-1504] Tags should now work.

aspidites:
~~Extending a unit test, then done.~~ Test written. Basically, this removes the stringent requirements for tag auto-complete.


aspidites:
This is marked as a hotfix despite going to QC. Reason being that it needs to be looked at ASAP so that I can hotfix against master. Pointing to QC now so that there is somewhere to verify changes before pushing to master.

PR #1619, opened 2015-07-01
PD-1501 date_added field for solr, and mgmt command for checking tasks

mklauber:
start indexing the date_added field (already part of the production schema). 
Management command to check if tasks are backing up, but not making it to Solr.  (Dependent on the date_added field).

### DIFF
--- a/import_jobs.py
+++ b/import_jobs.py
@@ -118,7 +118,7 @@ def get_jobsfs_zipfile(guid):
     authheader = "Basic %s" % base64.encodestring('%s:%s' % ('microsites',
                                                              'di?dsDe4'))
     req.add_header("Authorization", authheader)
-    resp = urllib2.urlopen(req)
+    resp = urllib2.urlopen(req, timeout=30)
     return resp
 
 
--- a/seo/management/commands/check_tasks.py
+++ b/seo/management/commands/check_tasks.py
@@ -1,0 +1,50 @@
+from django.core.management.base import BaseCommand 
+import json
+from django.conf import settings
+import requests
+from pysolr import Solr
+from datetime import datetime, timedelta
+from django.core.mail import send_mail
+
+
+
+class Command(BaseCommand):
+    help = "Check if Solr has seen updates while tasks are in the queue."
+    
+    port = "15672"
+    endpoints = ["solr/", "/api/queues/dseo-vhost/priority/"]
+    msg_cutoff = 0
+    solr_cuttoff = 0
+    look_behind = timedelta(hours=1)
+    
+
+    def handle(self, port="15672", queues="solr,priority", *args, **options):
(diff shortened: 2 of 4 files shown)


import_jobs.py
```diff
@@ -118,7 +118,7 @@ def get_jobsfs_zipfile(guid):
     authheader = "Basic %s" % base64.encodestring('%s:%s' % ('microsites',
                                                              'di?dsDe4'))
     req.add_header("Authorization", authheader)
-    resp = urllib2.urlopen(req)
+    resp = urllib2.urlopen(req, timeout=30)
```
aspidites: Do we already/need to do something if a timeout occurs?

seo/management/commands/check_tasks.py
```diff
@@ -0,0 +1,50 @@
...
+    solr_cuttoff = 0
+    look_behind = timedelta(hours=1)
+    
+
+    def handle(self, port="15672", queues="solr,priority", *args, **options):
+        # Determine the number of tasks still in Rabbit 
+        msg_count = 0
+        for queue in queues.split(','):
+            # Get the queue data from the rabbit Management API
+            uri = "http://%(broker)s:%(port)s/api/queues/dseo-vhost/%(queue)s" % {'broker': settings.BROKER_HOST,
```
aspidites: nit: alignment

seo/management/commands/check_tasks.py
```diff
@@ -0,0 +1,50 @@
...
+                                                                                'port': port,
+                                                                                'queue': queue}
+            resp = requests.get(uri, auth=(settings.BROKER_USER, settings.BROKER_PASSWORD))
+            data = json.loads(resp.content)
+            msg_count += data["messages_ready"]
```
aspidites: Is this guaranteed to be an integer, or can it be something like "32"?

aspidites:
Looks mostly fine. The "nit" insn't really important.

import_jobs.py
mklauber: No.  We're currently using the default timeout, but we're getting strange hanging behavior.  If a timeout occurs, it should raise an exception, which should result in the task being cancelled and retried in 3 minutes.  Hopefully, by then whatever is causing connection issues should be resolved.

seo/management/commands/check_tasks.py
mklauber: Api returns an int in the json.  That should be stable.

mklauber:
Still want @JLMcLaugh to look it over to confirm it works for him.

import_jobs.py
aspidites: :+1:

aspidites:
I'm good with this. @galitskyd, @tdphillips , @JLMcLaugh ?

PR #1625, opened 2015-07-06
Random Reporting UI Improvements

galitskyd:
# What Randomly Changed?
#### Bootstrap
* Added Bootstrap's tooltip component (minified) to custom.js

#### Sidebar
Before:
beforetooltip

After:
aftertooltip

* Added tooltips to sidebar icons.
  * ~~@fezick, should tooltips be used for archive page as well? (I'm thinking yes)~~
* Better visual cues for the regeneration process.
  * With the addition of tooltips we can now dynamically update and show the new tooltip.
    * While regenerating the tooltip will show "Regenerating..." while hovered.
  * On completion the whole row, that is associated to the regenerated report, is briefly highlighted.
    * Affects both the sidebar and archive page.

#### Bug Fix
There was a bug in the popstate code where coming *BACK* from a download page to a new or cloned report would call the wrong function and would cause rendering errors.

#### Other stuff
* Added report type to the top of newly created and cloned reports.
screen shot 2015-07-06 at 3 36 33 pm

  * Ex: *User selects Partner and hits next*, the User gets reassured that the fields they now see are for "Partner Report"
  * Report object now has a .verboseType() that returns the user-facing name of the given report.
* "Create a new Report" is now "Create a New Report"


aspidites:
:+1:

aspidites:
Checked on Firefox. Will merge once we confirm IE 8 works.

PR #1628, opened 2015-07-08
[PD-1509] Contact Record Tags from Partner Saved Search

aspidites:
~~Not sure if I'm done yet. Can't run locally since some of our code base uses SOLR 3 and 4, and I've only set up SOLR 5.~~


aspidites:
Hmm.. the most recent error makes sense (from #a24ecd3), but it's not reproduced locally (sqlite). Made a change which should fix the index error, and now running locally with mysql to see if I can duplicate the erroneous behavior.

aspidites:
Local tests on mysql worked fine...

PR #1629, opened 2015-07-08
Reporting Javascript Improvements

galitskyd:
This PR should be merged **AFTER** #1625 as reporting.js has been touched in both PRs.

# Why?
While creating reporting.js I learned *a lot* about how Javascript uses objects and prototypes(!). So there were some choices made, like using six parameters to create a new FilteredList, that weren't the best which hurt maintainability/readability. With this PR I'm hoping to improve the maintainability and readability of this code.

# What?
* Prototype methods are created in a single object as oppose to being created separately and modifying the prototype object.
  * Prototype methods are more clearly grouped together.
  * Makes folding the code useful and shows what inherited methods are overwritten immediately apparent.
  * screen shot 2015-07-08 at 1 56 50 pm copy
* Field creation uses an object to set values as oppose to parameters.
  * Using an object is self-documenting.
  * People are more use to this style due to jQuery.
  * No longer have to pass undefined to retrieve default (just don't pass anything now!).
  * Is an important step towards creating fields from JSON which will happen once we get report forms in a db.  

```javascript
// Old Way
new TextField("Name", "name", true, "David");

// New Way
new TextField({
  label: "Name",
  id: "name",
  required: true,
  defaultVal: "David"
});

// Old Way
new TagField("Tags", "tags", false, undefined, "Use commas for multiple tags."),

// New Way
new TagField({
  label: "Tags",
  id: "tags",
  helpText: "Use commas for multiple tags."
});
```

### DIFF
--- a/static/reporting.166-06.js
+++ b/static/reporting.166-06.js
@@ -30,7 +30,7 @@ window.onpopstate = function(event) {
         form.renderReports(".rpt-container");
       } else {
         report.renderFields(".rpt-container", report.fields, true);
-        report.unbindEvent().bindEvents();
+        report.unbindEvents().bindEvents();
       }
       renderNavigation();
     };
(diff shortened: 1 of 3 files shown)


aspidites:
For the lazy (or scared): 
https://github.com/DirectEmployers/MyJobs/compare/reporting-random-ui-improvments...reporting-js-improvements

PR #1630, opened 2015-07-09
Update except clause

tdphillips:


aspidites:
# Context 
Running `./manage.py syncdb` will fail because myemails has code that relies on content type which, on a fresh install, doesn't exist. While sqlite would throw an OperationalError, mysql throws a ProgrammingError. This PR fixes that. Another side effect is that since new databases are created on each test run `./manage.py jenkins` would also fail.

PR #1631, opened 2015-07-10
Requeue tasks mgmt command.

mklauber:


seo/management/commands/requeue_failures.py
```diff
@@ -0,0 +1,20 @@
+from django.core.management.base import NoArgsCommand
+from djcelery.models import TaskState
+from datetime import datetime, timedelta
+from tasks import task_etl_to_solr
+
+class Command(NoArgsCommand):
+    args = 'None'
+    help = 'Checks for the existence of member logos on the content server'
```
tdphillips: fry

aspidites:
I don't like the 'None' and eval combo.

seo/management/commands/requeue_failures.py
```diff
@@ -0,0 +1,20 @@
...
+    def handle_noargs(self, **options):
+        eight_hours_ago = datetime.now() - timedelta(hours=8)
+
+        failed_tasks = TaskState.objects.filter(state__in=['FAILURE', 'STARTED', 'RETRY'], 
+                                                tstamp__gt=eight_hours_ago, 
+                                                name__in=['tasks.etl_to_solr',
+                                                          'tasks.priority_etl_to_solr'])
+
+        for task in failed_tasks:
+            task_etl_to_solr.delay(*eval(task.args))
```
aspidites: I don't think this is doing what you want:

```python
>>> from __future__ import print_function
>>> print(*eval("None"))
  File "<stdin>", line 1, in <module>
TypeError: print() argument after * must be a sequence, not NoneType
>>> 
```

JLMcLaugh: task.args is a string literal of the arguments passed into the original tasks. eval would split the string into 3 string arguments. But... this is really insecure. We should use celery's built in retry.

tdphillips: @aspidites: The args attribute on Command is not the same as task.args.

Not saying this is great, but eval("None") isn't happening.

aspidites: @tdphillips , oh, gotcha.

mklauber: For what it's worth, `TaskState` doesn't have a retry method (It's a record of the task execution, not the task object itself).

mklauber: But I may be able to try to parse and load the data via [`ast.literal_eval`](https://docs.python.org/2/library/ast.html#ast.literal%5Feval).

aspidites:
If it wasn't already obvious, I'm good with this now. Anyone else?

PR #1633, opened 2015-07-14
Use local states json file

tdphillips:
This file is both in the local repo and on s3, adding pointless network time which can fail.

### DIFF
--- a/states.py
+++ b/states.py
@@ -1,10 +1,9 @@
 import json
-import urllib2
 
 
 def get_state_map():
-    data_url = 'https://d2e48ltfsb5exy.cloudfront.net/myjobs/data/usa_regions.json'
-    data_list = json.loads(urllib2.urlopen(data_url).read())['regions']
+    states = open('jsondata/usa_regions.json')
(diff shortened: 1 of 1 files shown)


aspidites:
True. Any reason to not just use the bidict you made for postajob? (optionally relocating it to somewhere more appropriate). 

For the record, I meant to do so forever ago, but ended up having to do too many other things and abandoned my efforts (dynamic region/country selection in all PRM forms, IIRC).

tdphillips:
It should basically be a drop-in replacement after confirming that there aren't extra elements in one list vs the other.

I'm writing Selenium tests now, which is the only reason I didn't do that. The remote file seems to be inaccessible from home so I needed some form of workaround.

tdphillips:
You know what, we have a bit of technical debt. I'll work on reducing it by getting rid of one of these.

tdphillips:
Ignore that - there are a few differences that I'm not comfortable with.

Example:
location_data.py, ASM (American Samoa) is a country
usa_regions.json, AS (American Samoa) is a part of the US

PR #1640, opened 2015-07-17
Parse guid as a guid <PD-1535>

tdphillips:
Tests pass.

### DIFF
--- a/redirect/helpers.py
+++ b/redirect/helpers.py
@@ -4,6 +4,7 @@ from email.utils import getaddresses
 import urllib
 import urllib2
 import urlparse
+import uuid
 import markdown
 
 from jira.client import JIRA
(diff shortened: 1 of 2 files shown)


seo/tests/test_views.py
```diff
@@ -109,7 +110,7 @@ def test_job_detail_fallback(self):
         # When a job is not in solr but is in the database, we should
         # redirect to that job's apply url.
         self.conn.delete(q='*:*')
-        redirect = RedirectFactory(guid=self.job['guid'],
+        redirect = RedirectFactory(guid='{%s}' % uuid.UUID(self.job['guid']),
```
tdphillips: A UUIDField would be nice right about now.

redirect/helpers.py
```diff
@@ -759,6 +760,8 @@ def redirect_if_new(**kwargs):
     """
     Redirects to the job's url if new, otherwise returns None.
     """
+    if 'guid' in kwargs:
+        kwargs['guid'] = '{%s}' % uuid.UUID(kwargs['guid'])
```
tdphillips: Microsites use a 32-character guid. The key in the redirect database is 32 hex characters, four hyphens, and two curly braces.

PR #1642, opened 2015-07-21
Add subject and body to mailto links <PD-1494>

tdphillips:
Tests pass.

### DIFF
--- a/seo/tests/test_views.py
+++ b/seo/tests/test_views.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from datetime import datetime
 from copy import deepcopy
+import urlparse
 import default_settings
 import itertools
 import json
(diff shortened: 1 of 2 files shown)


tdphillips:
Tested this locally with settings pointed to production. I checked out one of DE's jobs (the only ones I knew of featuring mailto), clicked apply, and Gmail opened with the correct pre-filled fields.

aspidites:
Works fine, though I have a question: I thought view sources were numeric? When I run this locally, I get "View Source: Dot Jobs". Is that right?

tdphillips:
Interesting question. The thing I call a view source is called "view_source_id" by others. If you look in the seo ViewSource table, it's not even an id. What you're seeing as "Dot Jobs" is the view source with an id of 88 and a name of "Dot Jobs"

tdphillips:
"Dot Jobs" should be more meaningful to the recipient (and probably more understandable to the sender) than 88 would be.

tdphillips:
http://ec2-54-172-15-197.compute-1.amazonaws.com/jenkins/job/MyJobs-Single-Test/79/console

aspidites:
I agree. I just didn't know view sources had a name. :+1: all around, then.

PR #1643, opened 2015-07-22
Improve task failure logging in admin

mklauber:
This should show the full exception that causes failures in the admin, not just the retry.


aspidites:
I'm sold as soon as tests pass.

tdphillips:
Authorize.net transaction error.

PR #1644, opened 2015-07-22
Ensure we clean up the database if bad things happen <PD-1520>

tdphillips:
Found while debugging an actual issue with our db but it needs fixing anyway.

Tests pass.


tdphillips:
http://ec2-54-172-15-197.compute-1.amazonaws.com/jenkins/job/MyJobs-Single-Test/80/console

PR #1646, opened 2015-07-23
Ensure we only check verified secondary emails <PD-1534>

tdphillips:
Tests pass. I'm interested to see what people think. It's a slight perversion of get_email_owner.


tdphillips:
The only failure is a db issue that appears to have infected everything.

aspidites:
Coming over to ask directly, but wanted it documented:

1. Is there ever a case that we will ever want to get an email owner who *isn't* verified? 
2. If so, should True be the default rather than False, or is it more often the case that we call this without wanting to know about verification?

tdphillips:
I can't really answer that question. There are ~20 uses of get_email_owner and, without examining each one, I can't determine the validity of non-verified emails in those cases.

PR #1653, opened 2015-07-29
[PD-1537] Reporting API Refactor

aspidites:
# Why did I do all of this?
See the ticket and related ticket for the motivation behind this ticket. One departure from the original ticket is that this currently still has the flat object hierarchy, rather than the array of objects (which would represent AND/OR relationships). Given the amount of stuff this changes and the fact that the AND/OR widget isn't implemented yet, I decided to hold of on that. Luckily, I generally already know what needs to be changed in order to do that, and don't anticipate it being as big of a refactor as this one. 

# What did I do?

### Field keys
Before, we were assigning arbitrary IDs to fields and using those as object keys to pass as a request payload. In addition to making things confusing (eg, "was it contact.name or contact__name?"), it looked rather ugly. This PR allows for fields to optionally declare a key value, which if present, will be what's used for the payload. Those keys are specified as a string in familiar dot notation (eg. "contactrecord.contact.name").

### TagField
- I've decoupled the tag field from contact records, which made it possible to add tag fields to both the contact and partner report types. This was allowed by some refactoring I did to decouple fields from report type. It should now be possible to use all fields in all report types without restriction.

### FilteredList Refactor
- All hard-coded references to field type are removed. This allows us to not have to dynamically check the field type to remove AJAX payload data (such as when we previously dynamically changed `tags__name` to `contactrecord__tags__name`). 
- Dependencies are handled differently. Before, we had both a list of dependencies which noted which other lists needed to load before the current one should refilter, and also had an ignore list which noted which fields should not cause the current list to refilter. Now, there is a single dependencies object which maps field's to keys.
- Reduced the number of custom events down to one. Filtered lists emit a "filtered" event when filtering completes.
### Ajax changes
- JSON in, JSON out. Before, we would send query parameters and receive somewhat mangled JSON. Now we send query parameters as JSON and receive back pretty JSON.
- Approval status is no longer handled on the client, and is instead handled on the server.

### Model Simplification
- Models no longer have a `_parse_parameters` class method. Rather than having each model parse query parameters according to their needs, I've had the frontend ask for exactly what it wants. This has the gained benefit of localizing queries. That is, the frontend has more fine-grained control over what queries it can construct, and we no longer have to be weary about changing behavior for the sake of one use case affecting all others.
- So more field type identification - since the query parameters are exact, we no longer need to parse for the field type and assume what kind of query the user would like to perform. I suspect that queries will be made slightly faster by this, though I have yet to benchmark.
- I added a couple of (poorly named) helper functions which will convert a flat django query string to a nested dictionary suitable for serialization to JSON. For example, `{'foo__bar__baz': 7, 'foo__bar__buzz': 20}` would become `{'foo': {'bar': {'baz': 7, 'buzz': 20}}}`.
- state queries no longer use the synonym dictionary since state inputs are drop-downs across PRM.

### Refactored Helpers
-`universal.helpers` gained `query_to_json` and `json_to_query` which convert to and from a dictionary representing a Django query and a dictionary representing json
- `reports.helpers` lost parse_parameters, which was no longer needed.

## TODO
- [x] Update all unit tests to match the knew API format
- [x] Benchmark to ensure there haven't been any performance regressions
- [x] Update documentation
- [x] Write a data migration to deal with the renamed Report field.


mypartners/tests/factories.py
```diff
@@ -87,4 +98,4 @@ class Meta:
     address_line_one = factory.Sequence(lambda n: "%d Fake St" % n)
     city = fuzzy.FuzzyText()
     state = fuzzy.FuzzyText(length=2, chars=string.ascii_uppercase)
-    postal_code = fuzzy.FuzzyInteger(10000, 99999)
+    postal_approval_status = fuzzy.FuzzyInteger(10000, 99999)
```
tdphillips: Bad search/replace?

tdphillips:
Looks good to me but someone more familiar with reporting might be better.

aspidites:
The alleged failing test is passing locally:
2015-08-03-130526_1226x204_scrot

aspidites:
@galitskyd?

galitskyd:
Can't create any Contact Reports.

galitskyd:
Removing "Contact Email" on the download page then attempting to download breaks.

galitskyd:
Cloning a report throws errors.

galitskyd:
There is something weird going on with history API when going from a creating a report to doing anything with the sidebar then going back send you to the overview page as oppose to your report. The state is still created but is put in a strange place(?). For example, create a report, view archive, hit back (which should send you to overview page), hit back again, hit forward (now you should be on the report).

galitskyd:
Also FilteredLists are lying to me again... not sure if you can find anything to reproduce this.
screen shot 2015-08-06 at 8 52 37 am
screen shot 2015-08-06 at 8 52 52 am
screen shot 2015-08-06 at 8 53 38 am

aspidites:
@galitskyd:

> Can't create any Contact Reports.

Result of a typo. Fixed

> Edit: Correction, if you try to sort by a column then deselect it in the list for it not to show that column it breaks. This is the same on production. The selection is disabled if the column isn't checked but if they already select id doesn't switch to the default or something that will actually be used. 

Clarification: It's a bug, but one that already existed in production? Not

> Cloning a report throws errors.

I forgot to change it to a POST request with the new parameter format. Also
needed to send CSRF token since it's a post now. Fixed.

Still need to address your last two issues.

aspidites:
> There is something weird going on with history API when going from a creating a report to doing anything with the sidebar then going back send you to the overview page as oppose to your report. The state is still created but is put in a strange place(?). For example, create a report, view archive, hit back (which should send you to overview page), hit back again, hit forward (now you should be on the report).

Fixed. Callback wasn't being called as you suspected.

aspidites:
@galitskyd Addressed everything except for the "lying reports". Have to import your data first.

aspidites:
All of @galitskyd's issues addressed.

aspidites:
Updated migrations Friday, I'm not sure other issues remain that aren't better issued on a case-by-case basis. I suspect that unless there are other specific issues, the one-offs will be better addressed in qc, since there are a bunch of old reports there that I don't have access to.

galitskyd:
The Tag field you added to Partner Report doesn't autocomplete, works fin on Contact Report and Communication Report. Also, I don't know if you fix the migrations already but some of my reports get double json'd and 'icontains' for tags instead of 'in' (not sure if states and cities are suppose to be icontains as well).

aspidites:
@galitskyd : 
2015-08-10-095453_1244x1056_scrot
2015-08-10-095514_1244x1056_scrot

aspidites:
Worked through some migration issues that seemed to be caused by development environment. Corrected those and re-added a backward migration with documentation. I also integrated recent changes that converted "Contact" to "Communication" in certain places and verified that the produced csv is correct.

To be fair, @JLMcLaugh, the comments are larger than the actual PR.

PR #1655, opened 2015-07-30
Don't create a contact record until emails are sent <PD-1417>

tdphillips:
I can either do a quick fix (*weeps openly*) or I could (*sob*) rewrite email sending for the nth time. Preferences? If we go with "quick fix", I'll go back and edit comments. If we want a rewrite, I'll take down the PR.

Tests pass.


aspidites:
I don't know if a rewrite makes sense until after we've thoroughly (re-)discussed the expected behavior/use cases. Thus, I'd be fine with a (documented) quick fix.

tdphillips:
I'll keep that in mind but this is a bad PR. I should be looking at ContactRecord, not SavedSearchLog.

PR #1660, opened 2015-08-03
Fixed date issue for mypartners tests once and for all.

aspidites:
The original issue is that the N date format is a django only thing, so I was just Frankensteining the date together. After 3 or 4 failures to do so, I decided to just use a template renderer, which should (and does) render the date exactly as it would be in the template.


aspidites:
@tdphillips, this should make you happy.

aspidites:
This is also implemented in #1653, but I'm cherry picking the fix so that others can run tests without them failing.

PR #1662, opened 2015-08-04
Add Taiwan to country lists where it was missing. PD-1578

fezick:


fezick:
I can't test this locally, as the cloudfront URL appears to be hardcoded here: https://github.com/DirectEmployers/MyJobs/blob/693256e1b4c59a572256436263c568a153535dca/myprofile/templatetags/selection_boxes.py#L96

aspidites:
yet another reason to clean up the number of different files we use to check locations. @tdphillips didn't we make a list of such files somewhere?

tdphillips:
Possibly. postajob/location_data.py and seo/fixtures/countries.json already contain Taiwan while these two do not.

fezick:
@tdphillips, yes, that was the case.

fezick:
@haliscooper verified we are good to go with Taiwan. We are clear to merge.

PR #1663, opened 2015-08-04
[PD-1361] site redesign - navigation to existing items in megamenu, and replace items with SEOable pages (Part 1)

galitskyd:
For visibility of what I'm doing.

## MegaMenu Tabs to Static Pages
- [x] States
screen shot 2015-08-04 at 3 36 06 pm
- [x] Cities
screen shot 2015-08-05 at 2 29 02 pm
- [x] Companies
screen shot 2015-08-05 at 4 34 31 pm
- [x] Military Bases
screen shot 2015-08-07 at 11 05 40 am
- [x] Tags
screen shot 2015-08-07 at 4 52 53 pm

## Other Tasks after each page has been made
- [x] Better URLs
- [x] Remove MegaMenu
- [x] Refactor


seo/views/search_views.py
```diff
@@ -2022,3 +2023,70 @@ def __init__(self, **entries):
...
+
+
+def seo_states(request):
+    data_dict = {"states": sorted(states_with_sites.iteritems()),
+                 "other_locations": sorted(other_locations_with_sites.iteritems())}
+
+    return render_to_response('seo/states.html', data_dict,
+                              context_instance=RequestContext(request))
+
+states_with_sites = bidict({
```
tdphillips: We discussed this in person, but for everyone else...

The existing state/region lists are inadequate. The postajob one is almost right but we don't have sites for all of the US territories.

galitskyd: Thanks Troy. Also, I'm not entirely sure where to stick these two dicts. Wasn't sure if we wanted seo.search_views to reference something in postajob (if i were to put this with the other state/region lists) or if that is even that big of a deal. I was going to deal with this during the refactor step.

aspidites:
:-\ Even more location-based dictionaries?

templates/seo/cities.html
```diff
@@ -0,0 +1,96 @@
...
+{% block directseo_main_content %}
+    <div id="cities">
+        <div class="row">
+            <div class="span10"><h4>All Cities</h4></div>
+            <div class="span6"><h4>Popular Cities</h4></div>
+        </div>
+        <div class="span3 selectBox">
+            <div class="cityList" id="cityStatesDiv">
+                <ul role="menu" id="cityStatesList">
+                    <li role="menuitem">Alabama</li>
```
aspidites: Do we really want to hard-code this? In addition to a {% for %} loop reducing the size of the template, if it later turns out that we're missing a state or something, we'll have one less location to change if we just one of the ump-teen pre-existing state lists.

templates/seo/cities.html
```diff
@@ -0,0 +1,96 @@
...
+        <div class="span7 selectBox" id="cityCitiesSelect">
+            <div class="cityList" id="cityCitiesDiv">
+                <ul role="menu" id="cityCitiesList"></ul>
+                <div id="cityLoading"></div>
+            </div>
+        </div>
+
+        <div class="span6 selectBox">
+            <div class="cityList" id="popularCityList">
+                <ul role="menu" id="cityPopularCities" style="border: none;">
```
aspidites: What determines popular cities? Seems like it should be dynamic (no one is popular forever).

galitskyd: Be prepared to ask the same question about "Top Tags"

aspidites: Consider it asked.

tdphillips: They are a window into the past so that we may inspect the history of which cities were popular when this <ul> was created.

aspidites:
:+1: , though I only checked to see that the AJAX calls weren't 404ing and things were clickable. Also verified that the sites I went to for each one matched what I clicked on. Not exactly sure how tags are a supposed to work, but I clicked things and went places.

aspidites:
Worked through some migration issues that seemed to be caused by development environment. Corrected those and re-added a backward migration with documentation. I also integrated recent changes that converted "Contact" to "Communication" in certain places and verified that the produced csv is correct.

PR #1664, opened 2015-08-04
Stop marking things as being sent when they're not <PD-1363>

tdphillips:
- [x] Allow SavedSearchLogs to be created when send_email throws exceptions, including reasons for failure
- [x] Don't update last sent on saved searches when the above happens
- [x] Identify communication records for failed communication and display an indicator of such

I fear the third point may require a bit of the hackery from last week's aborted PR


tdphillips:
2015-08-07-083150_622x186_scrot
2015-08-07-083159_658x181_scrot

default_settings.py
```diff
@@ -77,11 +77,11 @@
 )
 
 TEMPLATE_LOADERS = (
-    ('django.template.loaders.cached.Loader', (
+    #('django.template.loaders.cached.Loader', (
```
aspidites: Is this intentional?

mypartners/migrations/0017_auto__add_field_contactrecord_successful.py
```diff
@@ -0,0 +1,449 @@
...
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Adding field 'ContactRecord.successful'
+        db.add_column(u'mypartners_contactrecord', 'successful',
+                      self.gf('django.db.models.fields.BooleanField')(default=True),
+                      keep_default=False)
+
+
+    def backwards(self, orm):
```
aspidites: Rather than adding a field to ContactRecord, could we just add an Action flag for ocntactlog entry? Something like "FAILURE"?

default_settings.py
tdphillips: Nope.

mypartners/models.py
```diff
@@ -728,6 +728,8 @@ class ContactRecord(models.Model):
     approval_status = models.OneToOneField(
         'mypartners.Status', null=True, verbose_name="Approval Status")
 
+    successful = models.BooleanField(default=True)
```
aspidites: See comment above about action flags.

mypartners/migrations/0017_auto__add_field_contactrecord_successful.py
tdphillips: It looks like that will mean we'll be replacing some "email" contact log entries with "failure" - is that okay?

It will also mean at least one more query per communication record.

aspidites: Yuck, nvm. I guess in the back of my head I was thinking these were bit fields that could be bitwise OR'd. Disregard that suggestion.

Does still seem like the field is on the wrong model. In my head, "failed email" sounds better than "failed record" (which would be different from my earlier suggestion, which would have moved the field to ContactLogEntry).

tdphillips:
If I'm going to update this, I'll basically be moving the successful flag from ContactRecord to ContactLogEntry and then doing a bunch of queries to display this in the UI. Acceptable?

tdphillips:
The old screenshots still stand.

mypartners/models.py
```diff
@@ -642,6 +642,12 @@ def get_record_url(self):
     def shorten_date_time(self):
         return self.date_time.strftime('%b %e, %Y')
 
+    @property
+    def contactlogentry(self):
```
tdphillips: I know this isn't pythonic but it follows Django name conventions for autogenerated keys.

mypartners/models.py
```diff
@@ -642,6 +642,12 @@ def get_record_url(self):
...
+        ct = ContentType.objects.get_for_model(self.__class__)
```
aspidites: Does grabbing by content type work better than using `django.db.models.loading.get_model`? Or is `get_model` insufficient here? After my foray into Django 1.7 and 1.8, I'm a bit leery about depending on ContentTypes, as they feel super brittle. Especially since they are only available post migration, though we have bits of code here and there (myemails for example) which depend on them before hand. 

If it's a must, I won't object, but I'd prefer an alternative. I *do* like the idea of a property though.

mypartners/models.py
```diff
@@ -733,6 +739,7 @@ class ContactLogEntry(models.Model):
     object_repr = models.CharField('object repr', max_length=200)
     partner = models.ForeignKey(Partner, null=True, on_delete=models.SET_NULL)
     user = models.ForeignKey(User, null=True, on_delete=models.SET_NULL)
+    successful = models.BooleanField(blank=True, default=True)
```
aspidites: Do we want blank=True here (as opposed to null)? Or is this to do with django representing booleans as a string instead of a small integer?

mypartners/models.py
tdphillips: I'm not using the content type to get the model. While ContactLogEntry doesn't have something explicitly called a generic foreign key, that's what the content_type and object_id fields are.

mypartners/models.py
tdphillips: I'm open to suggestions. I have no preference in this matter.

aspidites: Between null and blank, I'd probably prefer null. In fact, there's a [NullBooleanField](https://docs.djangoproject.com/en/1.6/ref/models/fields/#django.db.models.NullBooleanField) which accomplishes this. @JLMcLaugh , opinions?

JLMcLaugh: If they made a NullBooleanField then that seems like the recommended way to do it.

tdphillips: This is now a NullBooleanField.

aspidites:
Seems good to me. Didn't run locally as I'm not sure the best way to do so, and I'm not sure it would provide more than what the test which preempts a failure is doing. @JLMcLaugh, @tdphillips ?

tdphillips:
The screenshots were generated by creating a contact and assigning it a saved search, modifying the sendgrid password, and then trying to manually send the email.

aspidites:
I will do that locally then.

aspidites:
I don't think contact log entries are being created after send_email.  I call `send_email()` and check the saved search's `last_sent` attribute, which is correct, but after querying log entries (`ContactLogEntry.objects.last()`, I see that the action_time is far earlier than the saved search's last_sent date, which suggests that a new log entry wasn't made at all:

```pythoIn [1]: from mypartners.models import *

In [2]: from mysearches.models import *

In [3]: ss = SavedSearch.objects.last()

In [4]: l = ContactLogEntry.objects.last()

In [5]: ss.last_sent
Out[5]: datetime.datetime(2015, 8, 17, 12, 59, 20, tzinfo=<UTC>)

In [6]: l.action_time
Out[6]: datetime.datetime(2015, 8, 13, 12, 43, 26, tzinfo=<UTC>n

```

tdphillips:
@aspidites Were you sending a normal or partner saved search?

aspidites:
@tdphillips normal saved searches. I intentionally checked those, as I suspect that a saved search which wasn't sent properly should *never* be logged as sent. Surely we don't forgo logging regular saved searches?

tdphillips:
ContactLogEntry is PRM-specific logging and, as such, isn't on normal saved searches. Normal searches do have some logging via SavedSearchLog.

aspidites:
@tdphillips  totally right, considering the contact_type is `pssemail`. Not sure why that didn't occur to me, considering I've been spending a week+ with exactly that log type! Sorry about the extra noise, @tdphillips . I do wonder though, are those SavedSearchLogs tracking success status?

tdphillips:
@aspidites 
Nope. If an exception is raised when sending an email, we set was_sent=False and set the reason to whatever the exception message is. These should only be visible by us so having technical messages is okay.

aspidites:
That works for me. Just finishing up running or stuff locally with partner saved searches instead and will let you know how that goes.

aspidites:
@tdphillips  I don't think either of these should have the icon, given they aren't partner saved searches:
2015-08-17-161138_659x261_scrot

tdphillips:
Correct.

aspidites:
Finally got this to fail properly (pardon the oxymoron). I can confirm that the success flag is set as it should be and that the specific error that caused the failure is logged (in notes for now).

tdphillips:
Addressed @aspidites concerns. Success/failure only show up when the field is not null and also appears in the record's history.
2015-08-18-091725_377x264_scrot
2015-08-18-091738_244x160_scrot

aspidites:
Are we intending to mark partner saved searches which didn't technically fail as failed?

2015-08-19-143914_631x288_scrot
2015-08-19-143926_241x158_scrot

In the above case, I would either expect an error message in notes and history, or not see an icon at all if it couldn't be determined whether the email was sent successfully or not.

tdphillips:
Another dumb mistake on my part. I don't know how that one doesn't have a contact log entry but it appears that's the case. Because templates basically don't throw exceptions, accessing nonexistent attributes on None does not fail horribly.

tdphillips:
Tested on a communication record with no contact log entries. The "record history" box is still empty. It no longer shows as having failed.

PR #1668, opened 2015-08-05
Contact->Communication on report download <PD-1523>

tdphillips:
Fix provided by @aspidites with some modification. Objection?

Tests pass.

### DIFF
--- a/myreports/views.py
+++ b/myreports/views.py
@@ -299,17 +299,11 @@ def downloads(request):
                          if field not in blacklist[report.model]])
 
         values = json.loads(report.values) or fields
-        if 'contact_type' in values:
-            values[values.index('contact_type')] = 'communication_type'
+        for field_list in [values, fields]:
+            if 'contact_type' in field_list:
+                index = field_list.index('contact_type')
+                field_list[index] = 'communication_type'
         fields = values + [field for field in fields if field not in values]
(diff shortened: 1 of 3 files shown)


myreports/views.py
```diff
@@ -299,7 +299,17 @@ def downloads(request):
                          if field not in blacklist[report.model]])
 
         values = json.loads(report.values) or fields
+        if 'contact_type' in values:
+            values[values.index('contact_type')] = 'communication_type'
         fields = values + [field for field in fields if field not in values]
+        if 'contact_type' in fields:
```
aspidites: But you aren't appending. You're in-place editing. Just lines 306 and 312 should be sufficient, no?

tdphillips: What is now line 304 adds it back.

tdphillips: If I do this in reverse, however, (construct fields and then filter both), I think I could get by without this if.

aspidites: Ah, ok. Beware of the column order on the download page (selected values should come before other fields). Couple of other ways to get rid of the inner loop (using a set for instance).

tdphillips:
http://ec2-54-172-15-197.compute-1.amazonaws.com/jenkins/job/MyJobs-Single-Test/85/console

PR #1669, opened 2015-08-07
[PD-1595] Added disallow feeds to templates/robots.txt

aspidites:
Dejavu.


aspidites:
I think this is done then, since I've only changed the template one as agreed upon during our meeting.

PR #1672, opened 2015-08-10
More contact->communication kerfuffles <PD-1523>

tdphillips:
Tests pass.

### DIFF
--- a/static/record_overview.167-10.js
+++ b/static/record_overview.167-10.js
@@ -327,6 +327,9 @@ function build_data() {
     $(".sidebar > :input:has(option)").each(function() {
         if($(this).val() != 'all') {
             var data_key = $(this).prev('label').html().replace(":", "").replace(" ","_").toLowerCase();
+            if (data_key == 'communication_type') {
+                data_key = 'contact_type';
+            }
             data[data_key] = $(this).val();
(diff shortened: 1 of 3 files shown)


static/record_overview.167-10.js
```diff
@@ -327,6 +327,9 @@ function build_data() {
     $(".sidebar > :input:has(option)").each(function() {
         if($(this).val() != 'all') {
             var data_key = $(this).prev('label').html().replace(":", "").replace(" ","_").toLowerCase();
+            if (data_key == 'communication_type') {
```
tdphillips: I said I would do this on the Python side, but this posts to at least two urls at the moment. One location is better than >=2.

aspidites:
What's this look like tackling it from the other direction? That is, translating in python instead of javascript? More or less work? Looks legit otherwise.

tdphillips:
A bit more work. As the comment says, the js posts to at least two different urls. That would mean two places in Python, as opposed to one here.

aspidites:
I see. I think I started typing my comment before you submitted your first one. :+1:  in that case, I agree 1 location is better than one.

aspidites:
Confirmed working.

PR #1674, opened 2015-08-11
Rename countries.json PD-1578

fezick:
Rename countries.json to be versioned and change references in the python. PD-1578.

This gets around the issue with production using the same file as QC.


seo/tests/test_import_jobs.py
```diff
@@ -151,7 +151,7 @@ def test_remove_expired_jobs(self):
 
 
 class LoadETLTestCase(DirectSEOBase):
-    fixtures = ['countries.json']
+    fixtures = ['countries_167-24.json']
```
tdphillips: This is a separate file located in seo/fixtures.

fezick: Corrected.

tdphillips:
As the code is looking for this file on s3, tests can't pass. Ideally we would stop doing remote requests for local assets but uploading the new file to s3 would also allow tests to pass.

fezick:
I just uploaded the file.

PR #1676, opened 2015-08-14
Allow CC for ad-hoc compliance emails <PD-1603>

tdphillips:
Tests passed locally.

Test coverage is spotty, but I'm slowly increasing it.

### DIFF
--- a/redirect/helpers.py
+++ b/redirect/helpers.py
@@ -590,7 +590,7 @@ def get_job_from_solr(guid):
     return None
 
 
-def send_response_to_sender(new_to, old_to, email_type, cc=(), guid='',
+def send_response_to_sender(new_to, old_to, email_type, guid='',
                             job=None, solr_job=None):
     """
     Send response to guid@my.jobs emails
@@ -600,7 +600,6 @@ def send_response_to_sender(new_to, old_to, email_type, cc=(), guid='',
     :old_to: GUID address
     :email_type: no_job, no_contact, or contact; denotes what type of email
         is to be sent
-    :cc: Any addresses that were originally CC'd
     :guid: GUID portion of the incoming address
     :job: Job from database; includes basic job info (title, location, owner)
     :solr_job: Job from Solr; If this is passed, the job must not be expired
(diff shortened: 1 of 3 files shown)


redirect/helpers.py
```diff
@@ -607,6 +612,9 @@ def send_response_to_sender(new_to, old_to, email_type, guid='', job=None,
         old_to = old_to[0]
     email = EmailMessage(from_email=settings.DEFAULT_FROM_EMAIL,
                          to=new_to)
+    if cc:
```
aspidites: What happens if you set email's cc to None? If its the same effect as not specifying it at all, then this if clause will be redundant.

redirect/tests/test_views.py
```diff
@@ -1064,6 +1064,10 @@ def test_bad_email(self):
         self.assertEqual(len(mail.outbox), 0)
 
     def test_bad_guid_email(self):
+        """
```
aspidites: Yay, doc strings!

redirect/helpers.py
tdphillips: I'll check as soon as my tests finish.

redirect/tests/test_views.py
tdphillips: This is but a fraction of the undocumented tests, however. :sob:

redirect/helpers.py
tdphillips: EmailMessage.cc is defined as a list or tuple.

aspidites: Ah. then `cc = cc or []`. Then the if is redundant.

aspidites: Actually, tuples are immutable, so why not change the default argument to ()? Then you could remove the if statement.

tdphillips: Done.

aspidites:
Why is this marked to not merge?

tdphillips:
Waiting on something brought up on Jira.

redirect/helpers.py
tdphillips: This whole conversation proved moot. We're no longer CCing people on this particular email.

aspidites:
Code looks good. Easily testable locally, or are we better to try this out in QC, @tdphillips?

redirect/helpers.py
aspidites: Oh joy. I was wondering if I just glared over the cc code this morning or not. That makes sense now.

tdphillips:
Neither, unless we're going to set up MX records for QC.

PR #1678, opened 2015-08-17
404 redirect correction seo <PD-287>

koonsies:
Pardon the many modifications. My IDE had a linter installed and it went wild on some unneeded white space.

The changes made are in relation to the redirects on the my jobs search page. In short, if the member creates a request in an canonical order with erroneous search terms, the filters should be applied as usual and simply show a "0 results found" page. A 404 is returned when special characters are added to any search term and a 301 redirect will be issued when slugs are not in the canonical order. The 301 redirect reforms the URL in the proper order but otherwise returns the page as normal.

### DIFF
--- a/seo/breadbox.py
+++ b/seo/breadbox.py
@@ -156,6 +156,8 @@ class Breadbox(object):
     def build_company_breadcrumbs_from_slugs(self):
         company_slug_value = self.filters.get('company_slug')
         display_title = helpers.bread_box_company_heading(company_slug_value)
+        if not display_title:
+            display_title = company_slug_value
         breadcrumb = self.build_breadcrumb_for_slug_type('company_slug',
                                                          display_title)
         if breadcrumb:
@@ -219,7 +221,7 @@ class Breadbox(object):
             try:
                 left = slugs.index(location_slugs[0])
             except ValueError:
-                # location slug isn't part of the path 
+                # location slug isn't part of the path
                 pass
             else:
                 right = left + len(location_slugs)
--- a/seo/tests/test_views.py
+++ b/seo/tests/test_views.py
@@ -7,6 +7,7 @@ import default_settings
 import itertools
 import json
 from StringIO import StringIO
+from collections import OrderedDict
 
 from django.contrib.auth.models import AnonymousUser
 from django.conf import settings
@@ -169,8 +170,8 @@ class WidgetsTestCase(DirectSEOTestCase):
 
     """
     fixtures = ['seo_views_testdata.json']
-    
-    def test_get_widget(self):        
+
+    def test_get_widget(self):
         import widget_settings  # tuple of haystack slab sets
         resp = self.client.get("/")  # get the homepage
         site_config = resp.context['site_config'] # grab the current site config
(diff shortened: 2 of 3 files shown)


seo/tests/test_views.py
```diff
@@ -1588,9 +1589,19 @@ def test_view_all_jobs(self):
     def test_slug_tag_permanent_redirect(self):
         """Filter paths not in canonical order result in a 301 to the canonical
         path"""
+        slugs_and_paths = OrderedDict([
```
JLMcLaugh: This is dictionary seems very different from default_settings.SLUG_TAGS. I don't think this test is checking what you think it is. Are you sure we're not just ending up with a blank or garbeled path? The 301 below isn't a very strong assertion.

JLMcLaugh: Ok, using default_settings.SLUG_TAGS was an attempt to keep the test updated with any new fliter types, but it wasn't passing the correct 3-part path for moc_slug. What you have makes sense.

aspidites: @koonsies , just so I understand, was this in relation to what was discussed during the team meeting a few days ago whereby  the test was expecting a specific number of path segments, but didn't account for the fact that moc bread crumbs generate 3 path segments?

seo/breadbox.py
```diff
@@ -156,6 +156,8 @@ def build_breadcrumb_for_param(self, param):
     def build_company_breadcrumbs_from_slugs(self):
         company_slug_value = self.filters.get('company_slug')
         display_title = helpers.bread_box_company_heading(company_slug_value)
```
aspidites: Mostly random stream of consciousness, but I'm curious:
- `display_title = helpers.bread_box_company_heading(company_slug_value) or company_slug_value` would allow you to not have the extra check.
- Is there ever a reason that we'd want `helpers.bread_box_company_heading` to return None (or whatever falsey value it appears to be returning) rather than defaulting to `display_title`. That is, it seems like if we are always going to have to special case that function's return value, we should instead modify that function to always return a legit value.

seo/tests/test_views.py
koonsies: @aspidites, yes, that is the reason that I declared a dictionary for the test rather than using the one in settings. while not in the PR, I also added the location multiple value path.

seo/tests/test_views.py
```diff
@@ -1271,8 +1272,8 @@ def test_seo_sponsor_logo_network_site(self):
         request.user = AnonymousUser()
         settings.SITE_TITLE = "Acme"
         bb = factories.BillboardImageFactory.build()
-        bb.save()        
-        settings.SITE.billboard_images.add(bb)        
+        bb.save()
```
aspidites: If you replace `build` by class instantiation, the call to `save` becomes redundant:
```python
bb = factories.BillboardImageFactory()
settings.SITE.billboard_images.add(bb)
```

seo/breadbox.py
koonsies: * This is a good note. I'll make the modification
* I couldn't think of a reason. It seems that this is used for the text on the breadcrumb item, not for an existence check.

seo/tests/test_views.py
koonsies: this was a auto-lint change. don't worry, i removed the plugin that massacred my commit.

PR #1680, opened 2015-08-17
Delete job location instances from Solr [PD-939]

JLMcLaugh:
When a JobLocation is removed from a Job, it wasn't cleared from Solr. This is a post_save signal on Job that clears ALL orphaned locations from Solr. We're treating the relationship as a foreign key in Solr, but a m2m in the database/admin. This seemed like the simplest way to enforce that relationship, but I'd be up for a better solution.

### DIFF
--- a/postajob/models.py
+++ b/postajob/models.py
@@ -8,6 +8,7 @@ from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
 from django.core.validators import MinValueValidator
 from django.db import models, DEFAULT_DB_ALIAS
+from django.db.models import Count
 from django.db.models.query import QuerySet
 from django.db.models.signals import pre_delete
 from django.template.loader import render_to_string
@@ -252,7 +253,6 @@ class Job(BaseModel):
 
     def remove_from_solr(self):
         from import_jobs import delete_by_guid
-
         guids = [location.guid for location in self.locations.all()]
         delete_by_guid(guids)
 
--- a/postajob/signals.py
+++ b/postajob/signals.py
@@ -1,0 +1,19 @@
+from django.db.models import Count
+from django.db.models.signals import post_save
+
+from postajob.models import Job, JobLocation
+
+def remove_orphaned_job_locations(sender, **kwargs):
+    # If there are no Jobs related to a JobLocation, it shouldn't exist in Solr
+    #
+    # change_m2m signals detecting job locations being removed
+    # were complicated and unreliable. We're trying to enforce a foriegn key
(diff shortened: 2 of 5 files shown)


postajob/signals.py
```diff
@@ -0,0 +1,19 @@
+from django.db.models import Count
+from django.db.models.signals import post_save
+
+from postajob.models import Job, JobLocation
+
+def remove_orphaned_job_locations(sender, **kwargs):
+    # If there are no Jobs related to a JobLocation, it shouldn't exist in Solr
```
aspidites: This should probably be a docstring, especially if we ever decide to start using Sphinx in earnest.

postajob/signals.py
```diff
@@ -0,0 +1,19 @@
...
+    #
+    # change_m2m signals detecting job locations being removed
+    # were complicated and unreliable. We're trying to enforce a foriegn key
+    # with cascade delete from JobLocation to Job in the index, but not the
+    # database
+    from import_jobs import delete_by_guid
+    orphaned_job_locations = JobLocation.objects.annotate(
```
aspidites: I wonder if adding `.values_list('guid', flat=True) would be better here? I haven't checked, but I suspect the result is that you get a more specific query, rather than having to force evaluation of the query and traverse each job location.

postajob/tests/test_models.py
```diff
@@ -75,6 +75,28 @@ def test_job_creation_and_deletion(self):
...
         self.assertEqual(JobLocation.objects.all().count(), 0)
 
+    def test_add_remove_job_locations(self):
+        """
+        Regression test, job locations that were removed from a job weren't
+        being removed from Solr
+        Tests that adding or removing locations results in Solr updates after
+        a job save
+        """
+        job = JobFactory(owner=self.company, created_by=self.user)
```
aspidites: I think you can pas locations to the factory:

```python
locations = JobLocationFactory.create_batch(2)
job = JobFactory(owner=self.company, created_by=self.user, locations=locations)
```
Then we wouldn't have to do the `job.save()` bit.

aspidites:
@JLMcLaugh, can this be tested locally? I assume doing so would require using `import_jobs.py` some how, but am not sure. I'm trying to make a point of running PR's locally if they can be.

postajob/models.py
```diff
@@ -8,6 +8,7 @@
 from django.contrib.contenttypes.models import ContentType
 from django.core.validators import MinValueValidator
 from django.db import models, DEFAULT_DB_ALIAS
+from django.db.models import Count
```
koonsies: Is this being used?

PR #1681, opened 2015-08-17
Fix solr tests  [PD-1644]

darrint:
Some tests were attempting to override HAYSTACK_CONNECTIONS but haystack was still using django settings. In some cases the tests were still passing, leading to the ability to break staging.

HAYSTACK_CONNECTIONS is a dictionary and haystack takes a reference to the dictionary in django settings when it loads.

Test were replacing the dictionary in settings entirely. Haystack was referencing the old dictionary and therefore did not see the new dictionary.

I wrote a context manager which replaces the contents of a dict and used it in DirectSeoBase setup/teardown. This is needed in bunch of places so I made it available as a test decorator.

As an additional test I set my local dev_settings.py to point HAYSTACK to a completely bogus url. This url should have been ignored but it instead caused some tests to get 404 errors. I used these failures as an indication that some tests were referencing the HAYSTACK_CONNECTIONS setting. When all tests passed I knew that no more tests were using those settings.

All seo tests pass except two that I can't fix.


postajob/tests/test_models.py
```diff
@@ -64,6 +66,9 @@ def setUp(self):
             'name': 'Test Site Package',
         }
 
+    def tearDown(self):
```
JLMcLaugh: Was tearDown doing something unexpected? MyJobsBase.tearDown should get run if you don't override it.

postajob/tests/test_tasks.py
```diff
@@ -5,8 +5,10 @@
 from myjobs.tests.factories import UserFactory
 from postajob.models import Job, JobLocation
 from tasks import expire_jobs
+from seo.tests.setup import test_haystack_class_decorator
 
 
+@test_haystack_class_decorator
```
JLMcLaugh: Can we get this on MyJobsBase? We should probably keep the Solr overrides in as few places as possible. The decorator is convenient, but it will be harder to remember in the future.

seo/tests/setup.py
```diff
@@ -27,7 +27,71 @@ def search(self, *args, **kwargs):
...
 
+    def __exit__(self, type, value, traceback):
+        self.guard.__exit__(type, value, traceback)
+        self.reload()
+
+    def reload(self):
+        for key in self.new_haystack_connections:
+            haystack_connections.reload(key)
+
+def test_haystack_config():
```
aspidites: Why is this a method instead of a property?

postajob/tests/test_tasks.py
aspidites: I assume what @JLMcLaugh is getting at is that during test runs, we should never be using prod/stagign solr, and such rather than having to decorate tests accordingly, tests using solr should always use the test collection.

darrint: We can. However, there's more than one base class in play. DirectSeoBase inherits directly from django.test.TestCase. If we really want this airtight then I could look into making this part of some module that's always loaded.

postajob/tests/test_models.py
darrint: Added more here but didn't clean up before commiting.

postajob/tests/test_tasks.py
tdphillips: These base classes are artifacts from when we had three separate projects. Every test class inherits from seo base or myjobs base (or redirect base, but that doesn't add/remove documents).

I'm not sure if this would be considered MAGIC! down the line but decorating those two base classes should allow the settings modification to trickle down.

seo/tests/setup.py
aspidites: What I meant was that `test_haystack_config` returns a `NewHaystackOCnnection` instance. I'm wondering why this is a method (invoked as `test_haystakc_config()` rather than simply a property (referenced as `test_haystack_config`). 

Mostly a relevant comment if we are constantly calling test_haystack_config.

darrint: It's meant to be used as a method decorator.

seo/tests/setup.py
```diff
@@ -27,7 +27,71 @@ def search(self, *args, **kwargs):
...
+    return NewHaystackConnections(solr_settings.HAYSTACK_CONNECTIONS)
+
+def test_haystack_class_decorator(cls):
+    old_setUp = cls.setUp
+    old_tearDown = cls.tearDown
+
+    def setUp(self):
+        self.__settings_guard = test_haystack_config()
+        self.__settings_guard.__enter__()
```
aspidites: I think you can just decorate the test case with [this](https://docs.djangoproject.com/en/1.6/topics/testing/tools/#django.test.utils.override_settings) rather than manually calling __enter__ and __exit__.

aspidites:
Spoke in person, every comment addressed except for the bit about base classes vs decorators.

seo/tests/setup.py
JLMcLaugh: I'd be curious how this works with haystack. It looked like haystack may have been doing some initialization that accessed settings.HAYSTACK_CONNECTIONS before the test loaded. They have similar warnings for overriding Django internals that get loaded before any test runners and don't reimport from django.conf.settings. If I understand this correctly, it looks like @darrint 's solution was to replace that dictionary directly with the ReplaceDictContents.

darrint:
Will do this another way. No need to ever replace the haystack settings.

PR #1682, opened 2015-08-18
Removal of is_site_family boolean <PD-1606>

koonsies:
removed is_site_family boolean from seo->models and admin. also generated migration for the change.

note: whitespace was apparently being removed by default of Atom. messed with the configs and these polluted commits shouldn't be happening any longer... fingers crossed.

### DIFF
--- a/seo/admin.py
+++ b/seo/admin.py
@@ -51,15 +51,15 @@ class GroupListFilter(admin.SimpleListFilter):
     Limits group filter selection and results to only groups the user is a
     member of unless the user is a superadmin.
     """
-    
+
     title = _('Group')
     parameter_name = 'filtered_groups'
-    
+
     def lookups(self, request, model_admin):
         """
         Limits filter selection.
         """
-        
+
         if request.user.is_superuser:
             return [(group.name, group.name) for group in Group.objects.all()]
         return [(group.name, group.name) for group in request.user.groups.all()]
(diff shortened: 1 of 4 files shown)


darrint:
Did seo/admin.py change in any meaningful way? Seemed to be all whitespace changes.

seo/admin.py
```diff
@@ -1159,11 +1159,10 @@ class SpecialCommitmentAdmin(admin.ModelAdmin):
 
 
 class SiteTagAdmin(admin.ModelAdmin):
-    form = SiteTagForm 
+    form = SiteTagForm
     save_on_top = True
     fieldsets = [
-        (None, {'fields': ['site_tag', 'tag_navigation',
```
koonsies: @darrint here is the only meaningful change. removal of the "is_site_family" attribute.

seo/models.py
```diff
@@ -943,10 +943,6 @@ class SiteTag(models.Model):
                                          help_text='Tag can be used for '
                                                    'navigation by users. '
                                                    'Viewable by public.')
-    is_site_family = models.BooleanField('Tag represents site family',
-                                         default=False,
-                                         help_text='Site tag represents '
-                                                   'a family of sites.')
     parent = models.ForeignKey(SeoSite, blank=True, null=True,
```
JLMcLaugh: We should also be removing the parent field from SiteTag. It will be moved to SeoSite.

PR #1683, opened 2015-08-18
Remove the disabling/fixing of saved searches <PD-1680>

tdphillips:
Tests pass.

This effectively prevents site issues from causing saved searches from becoming disabled. As we no longer care if the page loads, we can close our eyes and hope for the best.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -84,9 +84,8 @@ def send_search_digest(self, search):
         if self.request.retries < 2:  # retry sending email twice
             raise send_search_digest.retry(arg=[search], exc=e)
         else:
-            # After the initial try and two retries, disable the offending
-            # saved search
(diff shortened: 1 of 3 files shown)


aspidites:
Should we not then have a test to ensure internal errors don't cause saved searches to become disabled?

tasks.py
```diff
@@ -84,9 +84,8 @@ def send_search_digest(self, search):
         if self.request.retries < 2:  # retry sending email twice
             raise send_search_digest.retry(arg=[search], exc=e)
         else:
```
aspidites: Cant you just delete the else block all together?

tdphillips: But where would I write the comment?

aspidites: Docstring explaining the failure case.

PR #1684, opened 2015-08-19
Add site parent fk <pd-1607>

koonsies:

### DIFF
--- a/seo/models.py
+++ b/seo/models.py
@@ -495,6 +495,10 @@ class SeoSite(Site):
     canonical_company = models.ForeignKey('Company', blank=True, null=True,
                                           on_delete=models.SET_NULL,
                                           related_name='canonical_company_for')
+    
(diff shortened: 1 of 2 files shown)


seo/models.py
```diff
@@ -495,6 +495,10 @@ def all_sites(self):
     canonical_company = models.ForeignKey('Company', blank=True, null=True,
                                           on_delete=models.SET_NULL,
                                           related_name='canonical_company_for')
+    
+    parent_site = models.ForeignKey('self', blank=True, null=True,
```
tdphillips: It might be good to include a related_name.
https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ForeignKey.related_name

As it is currently, grabbing a given site's parent is as simple as `site.parent_site` but getting all of a site's children would be `site.seosite_set.all()`

koonsies: Good note. I'll add that.

PR #1685, opened 2015-08-19
Fix solr test  [PD-1644]

darrint:
Some tests were attempting to override HAYSTACK_CONNECTIONS but haystack was still using django settings. In some cases the tests were still passing, leading to the ability to break staging.

HAYSTACK_CONNECTIONS is a dictionary and haystack takes a reference to the dictionary in django settings when it loads.

Test were replacing the dictionary in settings entirely. Haystack was referencing the old dictionary and therefore did not see the new dictionary.

**New:** I put code in `myjobs/__init__.py` to make sure that if we are testing, we permanently stomp the HAYSTACK_CONNECTIONS. We don't ever need a connection to a "real" solr instance in tests, anymore than we need a connection to a real database.

As an additional test I set my local dev_settings.py to point HAYSTACK to a completely bogus url. This url should have been ignored but it instead caused some tests to get 404 errors. I used these failures as an indication that some tests were referencing the HAYSTACK_CONNECTIONS setting. When all tests passed I knew that no more tests were using those settings.

All seo tests pass except two that I can't fix.

### DIFF
--- a/myjobs/__init__.py
+++ b/myjobs/__init__.py
@@ -2,6 +2,7 @@
 myjobs
 """
 
+import sys
 import solr.signals
 
 __version_info__ = {
@@ -25,3 +26,21 @@ def get_version():
     return ''.join(vers)
 
 __version__ = get_version()
+
+
+def hide_production_solr_from_tests():
+    """
+    This code here to protect production/staging systems from errant
+    unit tests. Including this code as a PROJECT_APP __init__.py
+    based on information found here:
+
+    http://stackoverflow.com/questions/6791911/execute-code-when-django-starts-once-only
+    """
+    import settings
+    import default_settings
+
+    settings.HAYSTACK_CONNECTIONS.clear()
+    settings.HAYSTACK_CONNECTIONS.update(default_settings.TEST_HAYSTACK_CONNECTIONS)
+
+if "test" in sys.argv or "jenkins" in sys.argv:
+    hide_production_solr_from_tests()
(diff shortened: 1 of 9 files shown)


myjobs/__init__.py
```diff
@@ -25,3 +26,21 @@ def get_version():
...
+
+    http://stackoverflow.com/questions/6791911/execute-code-when-django-starts-once-only
+    """
+    import settings
+    import default_settings
+
+    settings.HAYSTACK_CONNECTIONS.clear()
+    settings.HAYSTACK_CONNECTIONS.update(default_settings.TEST_HAYSTACK_CONNECTIONS)
+
+if "test" in sys.argv or "jenkins" in sys.argv:
```
JLMcLaugh: There's also a manage.py testapps command, and a recommended bash alias in the setup docs. How about we check for the string test in any substring of sys.argv? "test" in ''.join(sys.argv).lower()?

seo/tests/test_views.py
```diff
@@ -1916,18 +1913,17 @@ def test_syndicate_feed_paging(self):
...
-                resp = self.client.get(next_link)
-                tree = etree.parse(StringIO(resp.content))
-                self.assertEqual(resp.status_code, 200)
-                next_link = tree.find("link[@rel='next']")
-                if next_link is not None:
-                    next_link = next_link.get('href')
-            self.assertEqual(num_pages, 2)
+        next_link = '/feed/xml?num_items=1'
+        num_pages = 0
+        while next_link:
```
aspidites: Whiles always make me nervous. I see it was there before, so I suppose it's been well established that this will always terminate?

aspidites:
Seems good. Is the  best way to test this to change my solr settings and confirm that tests still run?

Also, do the test haystack connections settings care whether I'm running solr 3, 4, or 5?

darrint:
> Seems good. Is the best way to test this to change my solr settings and confirm that tests still run?

Yes, as far as I know.

>Also, do the test haystack connections settings care whether I'm running solr 3, 4, or 5?

No.

darrint:
Oops. closed by mistake.

PR #1686, opened 2015-08-19
[PD-1632] ui - improve layout of states network page

galitskyd:
## Before
screen shot 2015-08-24 at 8 21 21 am

## After
screen shot 2015-08-24 at 9 44 49 am


seo/views/search_views.py
```diff
@@ -2026,8 +2025,28 @@ def __init__(self, **entries):
 
 
 def seo_states(request):
-    data_dict = {"states": sorted(states_with_sites.iteritems()),
-                 "other_locations": sorted(other_locations_with_sites.iteritems())}
+    search = DESearchQuerySet()
```
aspidites: Does inlining this hinder readability?

seo/views/search_views.py
```diff
@@ -2026,8 +2025,28 @@ def __init__(self, **entries):
...
+    search = dict(search.facet('state').facet_counts()['fields']['state'])
+
+    # Mutates states by adding counts from search
+    def _add_job_counts(states):
```
aspidites: Seems your intent is that these will never be used outside of this function. Perfect use case for lambdas, especially since they are only used once.

universal/states.py
```diff
@@ -21,61 +20,62 @@ def get_state_map():
...
 
-other_locations_with_sites = bidict({
-    "District of Columbia": "districtofcolumbia.jobs",
-    "Guam": "guam.jobs",
-    "Puerto Rico": "puertorico.jobs",
-    "Virgin Islands": "usvirginislands.jobs"
-})
+other_locations_with_sites = [
+    {"state": "District Of Columbia",   "url": "districtofcolumbia.jobs"},
+    {"state": "Guam",                   "url": "guam.jobs"},
```
aspidites: Why all the extra white space?

galitskyd: I found it more readable.

seo/views/search_views.py
galitskyd: Assumption of intent is correct but I don't understand what you mean about lambdas and each function is used twice (which why i didn't kept my lambdas originally).

seo/views/search_views.py
galitskyd: I can change this. I was going to argue for consistency but search_views has both methods.

seo/views/search_views.py
tdphillips: This one in particular would be messy with a lambda. There is no one single expression that will allow you to update a dict while returning the result.

universal/states.py
aspidites: I'd like to use my PEP8 card:
2015-08-19-161559_741x301_scrot

galitskyd: my pep8 doesn't complain :fire:

seo/views/search_views.py
aspidites: Ah, eye-balling it, I only saw the single usage. Functions are fine, then.

aspidites:
:+1:

aspidites:
What's left? Just asking because this doesn't appear to have gained new commits since last Thursday.

aspidites:
What happened to "Other US Locations"?

galitskyd:
Combined into one list.

aspidites:
Couldn't run this locally until I synced with QC because of new migrations. Afterward, my screenshot looks nothing like what you posted.

aspidites:
The city links also don't go anywhere ("View Alabama Cities" for example).

aspidites:
Talked in person, the discrepancies seem to be caused by  individual site configurations. The missing cities links is a known issue (and I think part of why this is still marked DO NOT MERGE). :+1:  up until this point though.

Yay for not having to review the entire branch at once after @galitskyd is done.

PR #1687, opened 2015-08-19
Add help text and reorder form fields for address forms. PD-1488

fezick:
Adds help text to the address forms in profile, post a job, and PRM contacts. Also relabels "Address Label" as "Address Name" and moves it to the bottom of the form to avoid confusion that is occuring with users putting address line 1 in the address label field.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -177,14 +177,17 @@ class Contact(models.Model):
     # used if this partner was created by using the partner library
     library = models.ForeignKey('PartnerLibrary', null=True,
                                 on_delete=models.SET_NULL)
-    name = models.CharField(max_length=255, verbose_name='Full Name')
-    email = models.EmailField(max_length=255, verbose_name='Email', blank=True)
+    name = models.CharField(max_length=255, verbose_name='Full Name',
+                             help_text='Contact\'s full name')
+    email = models.EmailField(max_length=255, verbose_name='Email', blank=True,
+                             help_text='Contact\'s email address')
     phone = models.CharField(max_length=30, verbose_name='Phone', blank=True,
-            default='')
+            default='', help_text='ie 123-456-7890, (123)456-7890')
     locations = models.ManyToManyField('Location', related_name='contacts')
     tags = models.ManyToManyField('Tag', null=True)
     notes = models.TextField(max_length=1000, verbose_name='Notes',
-                             blank=True, default="")
+                             blank=True, default="", 
+                             help_text='Any additional information you want to record')
     archived_on = models.DateTimeField(null=True)
     approval_status = models.OneToOneField(
         'mypartners.Status', null=True, verbose_name="Approval Status")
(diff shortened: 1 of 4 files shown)


mypartners/models.py
```diff
@@ -134,20 +134,26 @@ def sort_by(self, *fields):
...
+                                           help_text='ie Chicago, Washington, Dayton')
+    state = models.CharField(max_length=200, verbose_name='State/Region',
+                                             help_text='ie NY, WA, DC')
     country_code = models.CharField(max_length=3, verbose_name='Country',
                                     default='USA')
     postal_code = models.CharField(max_length=12, verbose_name='Postal Code',
-                                   blank=True)
+                                   blank=True,
+                                   help_text='ie 90210, 12345-7890')
+    label = models.CharField(max_length=60, verbose_name='Address Name',
```
tdphillips: How many sites allow you to name your addresses? I usually see "home" and "billing" at the most.

JLMcLaugh: Business can have a lot more offices/branches to keep track of.

tdphillips: I should have had my comment on the corresponding line in myprofile, not mypartners. Does that still stand?

mypartners/models.py
```diff
@@ -134,20 +134,26 @@ def sort_by(self, *fields):
...
     address_line_two = models.CharField(max_length=255,
                                         verbose_name='Address Line Two',
-                                        blank=True)
-    city = models.CharField(max_length=255, verbose_name='City')
-    state = models.CharField(max_length=200, verbose_name='State/Region')
+                                        blank=True,
+                                        help_text='ie Suite 100')
+    city = models.CharField(max_length=255, verbose_name='City',
+                                           help_text='ie Chicago, Washington, Dayton')
+    state = models.CharField(max_length=200, verbose_name='State/Region',
```
aspidites: Help text on state is ineffective, as we use state drop-downs:
2015-08-19-145129_368x92_scrot

mypartners/models.py
```diff
@@ -134,20 +134,26 @@ def sort_by(self, *fields):
 
 
 class Location(models.Model):
-    label = models.CharField(max_length=60, verbose_name='Address Label',
```
aspidites: Does adding help text to the address fields without adding them to other form elements not make the whole form look incomplete/asymmetrical?:

2015-08-19-145342_422x580_scrot

aspidites:
Not sure how picky we want to be. In some places you give three examples, where in others you just give two. While you generally enumerate the examples, there is at least one case where "etc" is used.

mypartners/models.py
aspidites: Actually, it looks like profile units still use a free form text input. That should probably be a drop down as well...

I think we didn't add it originally because profile units aren't searchable and the complexity of the ticket which made those changes was already large.

mypartners/models.py
fezick: Users are allowed multiple addresses, and the open field was in line with the goal of giving full control to users.

mypartners/models.py
fezick: There is no help text on state for PRM, only on profile. It's open ended because profile allows for international input. PRM is restricted to US territories.

mypartners/models.py
fezick: Good point. I am adding it to the other forms.

fezick:
I tailored the help text to the field. Some work better with an example, others with instructions. I did keep them consistent from model to model and form to form.

aspidites:
@fezick  can you sync with quality-control?

fezick:
@aspidites ... done.

aspidites:
Are we ever going to allow filtering by phone number in reports? If so, encouraging inconsistent formatting may be a bad idea.

aspidites:
The help for tags is inaccurate:
2015-08-19-163308_343x105_scrot

Tags are separated by commas, not spaces:
2015-08-19-163417_338x143_scrot

fezick:
Good catch on tags, and I can't answer the phone number one way or the other, so I'll run under the assumption of maybe and only use one example.

aspidites:
I'm satisfied. Anyone else?

PR #1688, opened 2015-08-19
Requeue Failures using Celery

mklauber:


aspidites:
Good ol' literal_eval. Seems fine. 
@JLMcLaugh?

tasks.py
```diff
@@ -898,3 +900,18 @@ def send_event_email(email_task):
 
     email_task.completed_on = datetime.now()
     email_task.save()
+
+
+@task(name="tasks.requeue_failures", ignore_result=True)
+def requeue_failures(hours=8):
+    period = datetime.datetime.now() - datetime.timedelta(hours=hours)
+
+    failed_tasks = TaskState.objects.filter(state__in=['FAILURE', 'STARTED', 'RETRY'],
```
aspidites: Started and retry tasks are considered failures? Is this meant as something along the lines of "long running tasks that aren't finished by now are probably hung"?

JLMcLaugh: Yes, they can hang and get stuck in "STARTED" mode. We may get some unnecessary retries, but that won't hurt anything.

PR #1689, opened 2015-08-19
[PD-1676] Retroactively tag communication records

aspidites:
@JLMcLaugh, Hal requested that this not be merged/run until after a database snapshot is made for it. 

@mklauber I assume that if we clone prod into QC, we'll get a better idea of how this migration will perform on production.

@fezick, am I supposed to remove logging from the migration and run a query for orphan records, or just leave as is?


aspidites:
Ready to merge. Just need to run migrations (which, @mklauber , don't those run automatically on qc?).

aspidites:
Just want to double check that migration numbers don't clash.

aspidites:
OK. Good to go. Github was being dumb and still showing 0017 even though locally it was 0018.

tdphillips:
In the event that this happens to not work on QC, would this be reverted before the next staging deployment (a week from next Monday) or will a fix be merged on top of it?

aspidites:
@tdphillips sounds like a question for the Director of Architecture. @JLMcLaugh. 

That said, there have been enough runs (see the ticket), that I'm not sure what could go wrong.

aspidites:
It's also the reason we made a snapshot of the database.

PR #1690, opened 2015-08-19
CORS Real Middleware  [PD-1620]

darrint:
This replaces the code we have which does rudimentary CORS headers with an existing library. The library gives us full support for CORS with credentials, etc.

I have the tests passing and there are some rudimentary tests for CORS origin handling. I could use some help testing the site to make sure this doesn't break something.


myjobs/tests/test_cors.py
```diff
@@ -0,0 +1,42 @@
+import re
+
+from setup import MyJobsBase
+
+
+class CorsTests(MyJobsBase):
+    def setUp(self):
+        super(CorsTests, self).setUp()
+        self.resp = \
```
aspidites: Line breaks are allowed after open parentheses (`self.client.options('), so the `\` here is unnecessary.

default_settings.py
```diff
@@ -623,3 +624,6 @@
 COMPRESS_OFFLINE = True
 
 COMPRESS_ENABLED = False
+
+CORS_ORIGIN_ALLOW_ALL = True
```
aspidites: I don't think `CORS_ORIGIN_ALLOW_ALL` is a good idea here. While that was essentially what the old code was attempting to do, I don't think that was ideal either.

myjobs/tests/test_cors.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_credentials(self):
+        self.assertHeaderValue("true", "Access-Control-Allow-Credentials")
+
+    def test_methods(self):
+        self.assertAllIn(['POST', 'GET', 'OPTIONS', 'PUT', 'DELETE'],
+                         "Access-Control-Allow-Methods")
+
+    def test_allowed_headers(self):
+        self.assertAllIn(['content-type'], "Access-Control-Allow-Headers")
```
aspidites: `assertIn` is probably sufficient here.

myjobs/tests/test_cors.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def assertHeaderValue(self, expected, header_name):
+        header_dict = dict(self.resp.items())
```
aspidites: Does a response not implement enough of the dict protocol to allow you to check that the header_name is in it directly, rather than having to create a dictionary out of it's iterated items? The above seems analogous to `some_other_dict = dict(some_dict.items())`.

mysearches/helpers.py
```diff
@@ -86,14 +89,17 @@ def validate_dotjobs_url(search_url, user):
...
     except Exception, e:
-        print e
+        logger.error("Failed to parse dotjobs url: %s" % search_url)
+        logger.exception(e)
         return None, None
 
     try:
         page = urllib.urlopen(search_url).read()
         soup = BeautifulSoup(page, "html.parser")
     except Exception, e:
```
aspidites: Since you're fixing how these are logged, would you mind updating the syntax as well? (`except Exception, e` -> `except Exception as e`.

aspidites:
Have you tested that both the top bar and saved search widget still work? IIRC, those weren't properly accounted for in unit tests, so passing tests would tell you nothing about their proper functioning.

myjobs/tests/test_cors.py
darrint: Not that I could find. response.items() is a list of tuples.

mysearches/helpers.py
darrint: Added in another commit.

default_settings.py
darrint: I think I have to tell it *something*. The CORS_ALLOW_CREDENTIALS setting turns off the "*" behavior and replaces it with allow-whatever-origin-the-user-provided behavior we want. The test confirms this. In my own testing I observed that the allow credentials setting changed the behavior.

tdphillips:
This is merging into the fix-solr-test--pd-1644 branch. You can't point pull requests elsewhere after creating so this may have to be reopened.

PR #1691, opened 2015-08-20
CORS Real Middleware [PD-1620]

darrint:


darrint:
This replaces the code we have which does rudimentary CORS headers with an existing library. The library gives us full support for CORS with credentials, etc.

I have the tests passing and there are some rudimentary tests for CORS origin handling. I could use some help testing the site to make sure this doesn't break something.

darrint:
I'm still looking for input on ways to test this. I don't know how to tell if I just broke the application.

aspidites:
I'm going to pull this locally and give running the saved search widget a go. I suspect that will end up with a PR to not hard-code www.my.jobs, but we'll see. I'm satisfied with this otherwise.

aspidites:
I'm going to pull this locally and give running the saved search widget a go. I suspect that will end up with a PR to not hard-code www.my.jobs, but we'll see. I'm satisfied with this otherwise.

aspidites:
Ended up just resending the headers with a proper saved search url to test.

PR #1692, opened 2015-08-20
prevent grandchilden sites in seo family <pd-1608>

koonsies:
* added: helper function verify_parent_child_relationship to determine if a SeoSite has valid parent_site fields. 
* overwrite of the seosite clean function to report problems with parent_site directly on the admin form. 
* overwrite of the save function on seosite as a catch all if issues escape clean(). 
* added unit tests for all functionality.


seo/models.py
```diff
@@ -572,10 +573,45 @@ def email_domain_choices(self,):
...
+                error_flag = True
+                error_description = ('%s is a parent site and cannot' + \
+                                    ' have a parent') % self.name 
+            elif self.parent_site.parent_site:
+                error_flag = True
+                error_description = ('%s is a child of another site and' + \
+                                     ' cannot be a parent') % self.parent_site.name
+        return error_flag, error_description
+            
+    def clean(self, *args, **kwargs):
```
aspidites: Actually, wouldn't it be more appropriate to do a `clean_parent_site`? I know for model forms you can do per-field cleaning, though I'm having trouble finding similar for models. I do see `clean_fields` though, which would seem to be more appropriate in lieu of `clean_<field name>`.

aspidites:
@koonsies marked this as do not merge since you said you were experimenting with something else.

seo/models.py
```diff
@@ -57,6 +59,33 @@ def get_query_set(self):
...
+        super(NonChainedForeignKey, self).clean(value, model_instance)
+        
+        if value:
+            object_class = model_instance.__class__
+            potential_parent = object_class.objects.get(pk=value)
+            
+            if value == model_instance.pk:
+                raise ValidationError('%s cannot be at parent entity of itself'\
+                                        % model_instance.__unicode__())
+            elif object_class.objects.\
```
aspidites: If you break on the parenthesis or brackets, you don't need the ugly `\`:
```python
elif object_class.objects.filter(**{
        self.name: model_instance}).exists():
```

seo/models.py
```diff
@@ -57,6 +59,33 @@ def get_query_set(self):
...
+                            filter(**{self.name:model_instance}).exists():
+                raise ValidationError('%s is a parent entity and cannot be a child'\
+                                        % model_instance.__unicode__())
+            elif getattr(potential_parent,self.name):
```
aspidites: PEP8, space after the comma.

seo/models.py
```diff
@@ -57,6 +59,33 @@ def get_query_set(self):
...
+            object_class = model_instance.__class__
+            potential_parent = object_class.objects.get(pk=value)
+            
+            if value == model_instance.pk:
+                raise ValidationError('%s cannot be at parent entity of itself'\
+                                        % model_instance.__unicode__())
+            elif object_class.objects.\
+                            filter(**{self.name:model_instance}).exists():
+                raise ValidationError('%s is a parent entity and cannot be a child'\
+                                        % model_instance.__unicode__())
```
aspidites: Given string substitution,`.__unicode__()` is redundant here, I think.

aspidites:
Looks good, but I'd like to pull this locally and ensure it works as advertised. If someone else is more confident or has already done so, of course feel free to ignore this comment and merge.

aspidites:
Works as advertised.

PR #1693, opened 2015-08-20
[PD-1514] Fixed javascript for field parsing in reports.

aspidites:
-  the changes to reports/views.py prevent index errors when a report has 0 results
- the changes to static/reporting.js prevent adding of empty values to filter parameters and fix the key used for tags in contact reports
- the template file was just updated to reflect the new filename for reporting.js.

### DIFF
--- a/static/reporting.167-20.js
+++ b/static/reporting.167-20.js
@@ -319,8 +319,12 @@ Field.prototype = {
     return $("#" + this.id);
   },
   onSave: function(key) {
-    var data = {};
-    steelToe(data).set(key || this.key, this.currentVal());
+    var data = {},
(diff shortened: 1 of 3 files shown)


static/reporting.167-20.js
```diff
@@ -1020,14 +1025,6 @@ FilteredList.prototype = $.extend(Object.create(Field.prototype), {
     }
 
     return this;
-  },
```
aspidites: @galitskyd, unless I missed something, this was identical to the Field prototype's onSave, so I deleted it, as it's not accomplishing anything new.

galitskyd: you aren't missing anything.

PR #1696, opened 2015-08-21
Add task to check solr count 20 minutes after receiving SNS message.

mklauber:


tasks.py
```diff
@@ -726,6 +728,20 @@ def task_priority_etl_to_solr(guid, buid, name):
         raise task_priority_etl_to_solr.retry(exc=e)
 
 
+@task(name="tasks.check_solr_count", send_error_emails=True)
+def task_check_solr_count(buid, count):
+    conn = Solr(settings.HAYSTACK_CONNECTIONS['default']['URL'])
+    hits = conn.search(q="buid:%s" % buid, rows=1, mlt="false", facet="false").hits
+    if int(count) != int(hits):
```
aspidites: Two somewhat related questions:

1. Are we sure that count will never be passed in as None?
2. If so, is there ever a reason to believe it won't always be an int (and thus conversion is redundant)?

aspidites:
Ping @mklauber

tasks.py
mklauber: 1. Yes we are reasonably certain it won't be none.
2. No I'm not certain the data will be an int already.  

In addition, If either of these things are not true, I'd prefer it to fail here, and have the task fail, than to do type checking where we create the task, and have failures there instead.  

Sorry about the delay responding.

aspidites: Given your addition, it seems we should be wrapping the call to `int()` in a try/except with logging, no?

JLMcLaugh: Tasks can come in as strings or integers depending on their source, but it's ok to assume count will be in an integer-like format. There's no recovering if it's not, so the best we could do is change a ValueError to some other exception. Celery will log errors for us that point to the problem either way.

tasks.py
```diff
@@ -726,6 +728,20 @@ def task_priority_etl_to_solr(guid, buid, name):
         raise task_priority_etl_to_solr.retry(exc=e)
 
 
+@task(name="tasks.check_solr_count", send_error_emails=True)
+def task_check_solr_count(buid, count):
+    conn = Solr(settings.HAYSTACK_CONNECTIONS['default']['URL'])
+    hits = conn.search(q="buid:%s" % buid, rows=1, mlt="false", facet="false").hits
```
JLMcLaugh: I would like to validate anything going to Solr, however. Can we make sure buid looks like an integer?

mklauber: Sure

PR #1697, opened 2015-08-21
[PD-1090] Removed unused import.

aspidites:
Turns out I had already done most of this already. The only real thing that remains would be to pass the path and data into the constructor, but the tests are running, so I'm not inclined to change them.

### DIFF
--- a/myjobs/tests/test_forms.py
+++ b/myjobs/tests/test_forms.py
@@ -1,6 +1,5 @@
 from myjobs.forms import ChangePasswordForm
 from myjobs.tests.factories import UserFactory
-from myjobs.tests.test_views import TestClient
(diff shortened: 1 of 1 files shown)


tdphillips:
The tests are no longer running. Are you planning on changing something or would that come later?

aspidites:
@tdphillips I don't have any other changes going on top of this. I basically audited all test files and found this PEP8 change.

tdphillips:
Was unsure if "the only real thing that remains" meant you were going to do that or not.

aspidites:
Oh sorry. That was meant more as "It's not worth doing as part of this
ticket.". Should have been clearer.

PR #1698, opened 2015-08-21
Ensure redirects work on non-404 pages [ST-916]

tdphillips:
This works. If no one has vehement objections, I'll document everything, add tests, and call this done.


aspidites:
I still don't get the ticket, so no objections from me. Bonus points for using comprehensions though. :+1:

middleware.py
```diff
@@ -283,4 +284,17 @@ def custom_facets_ops_groups(site_facets):
...
+
+
+class RedirectOverrideMiddleware(object):
+    def process_request(self, request):
+        configuration = get_site_config(request)
+        if configuration.not_found_override.exists():
+            paths = [request.path,
+                     (request.path[:-1] if request.path.endswith('/')
+                      else request.path + '/')]
+            not_found = configuration.not_found_override.filter(
```
tdphillips: Breaks if the department that must not be named wants "/path" and "/path/" to go to different locations. If that happens... ¯\\_(ツ)_/¯

tdphillips:
Reiterating, my commit message is completely wrong yet the code does what I said it would in the ticket. :unamused:

seo/tests/setup.py
```diff
@@ -67,7 +67,8 @@ def setUp(self):
 
         self.base_middleware_classes = settings.MIDDLEWARE_CLASSES
         middleware_classes = self.base_middleware_classes + (
-            'wildcard.middleware.WildcardMiddleware', )
+            'wildcard.middleware.WildcardMiddleware',
+            'middleware.RedirectOverrideMiddleware')
```
tdphillips: Remove this middleware if you wish to confirm this for yourself.

aspidites:
@tdphillips the redirect override middleware is the one to remove for testing locally?

tdphillips:
@aspidites Correct.

tdphillips:
Tests passed locally but are failing on Jenkins. I'll look into it.

aspidites:
@tdphillips, the obligatory annoying question: You're running tests with MySQL locally?

tdphillips:
I did, yes.

seo/migrations/0021_auto.py
```diff
@@ -0,0 +1,424 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
```
aspidites: Would you mind renaming this migration to something more descriptive, please?

tdphillips: I'll do so. Yay, autonamed migrations.

aspidites:
Given bogus URL A, legit URL B, and redirect URL C, where A doesn't exist, B does, and C is where I want to go:

Without the middleware:
- I can go from A to C
- I can not go from B to C

Which, if I've understood correctly, is the expected (though undesired) behavior. However, with the middleware this behavior is unchanged. where as I would expect to be able to go from B to C.

tdphillips:
@aspidites Don't look too hard into it if you don't remember, but do you recall what the paths for A, B, and C were from your previous comment?

middleware.py
```diff
@@ -246,3 +247,24 @@ def custom_facets_ops_groups(site_facets):
...
+class RedirectOverrideMiddleware(object):
+    """
+    Does a redirect if one is set on the current configuration for the current
+    path. Default behavior was to not do redirects if the page did not result
+    in a 404.
+    """
+    def process_request(self, request):
+        # Configuration is cached. The effect on response time is
+        # hopefully minimal.
+        configuration = get_site_config(request)
```
JLMcLaugh: If we're going to start calling get_site_config more than once per request, we should memoize it.

tdphillips: What would memoizing get us that cache wouldn't?

https://code.djangoproject.com/ticket/21351#comment:6

JLMcLaugh: memoize is local, a call to memcached is fast, but not free. I assume it's microseconds vs milliseconds, but I could be wrong.

JLMcLaugh: I realized you were talking about Django's queryset caching, not memcached. You're correct.

PR #1699, opened 2015-08-21
Opt in to content negotiation with a decorator. [PD-1621]

darrint:
Set up for allowing browsers to use either JSON XHR or JSONP. This is a generic mechanism allowing our views to focus on providing data.

Stub in a method which will, in the future, do provide secure information for child sites. The hope is that this new method will replace most of the existing jsonp endpoints.

This is an opt in decorator so existing code should not be affected.

### DIFF
--- a/myjobs/autoserialize.py
+++ b/myjobs/autoserialize.py
@@ -3,7 +3,7 @@ import json
 from functools import wraps
 
 
-def autoserialize_view_decorator(fn):
+def autoserialize(fn):
     @wraps(fn)
     def handle_autoserialize(request):
         response = fn(request)
--- a/myjobs/autoserialize.py
+++ b/myjobs/autoserialize.py
@@ -6,19 +6,15 @@ from functools import wraps
 def autoserialize_view_decorator(fn):
     @wraps(fn)
     def handle_autoserialize(request):
-        if "callback" in request.GET:
-            callback = request.GET['callback']
-            content_type = "text/javascript"
-        else:
-            callback = None
-            content_type = "application/json"
-
         response = fn(request)
 
         payload = json.dumps(response)
-        if callback is not None:
+        if "callback" in request.GET:
+            callback = request.GET['callback']
             content = "%s(%s)" % (callback, payload)
+            content_type = "text/javascript"
         else:
             content = payload
+            content_type = "application/json"
         return HttpResponse(content=content, content_type=content_type)
     return handle_autoserialize
--- a/myjobs/autoserialize.py
+++ b/myjobs/autoserialize.py
@@ -1,0 +1,24 @@
+from django.http import HttpResponse
+import json
+from functools import wraps
+
+
+def autoserialize_view_decorator(fn):
+    @wraps(fn)
+    def handle_autoserialize(request):
+        if "callback" in request.GET:
+            callback = request.GET['callback']
+            content_type = "text/javascript"
+        else:
+            callback = None
+            content_type = "application/json"
+
+        response = fn(request)
+
+        payload = json.dumps(response)
+        if callback is not None:
+            content = "%s(%s)" % (callback, payload)
+        else:
+            content = payload
+        return HttpResponse(content=content, content_type=content_type)
+    return handle_autoserialize
--- a/myjobs/child_dashboard.py
+++ b/myjobs/child_dashboard.py
@@ -1,0 +1,8 @@
+from django.http import HttpResponse
+
+from myjobs.autoserialize import autoserialize_view_decorator
+
+
+@autoserialize_view_decorator
+def child_dashboard(request):
+    return {'start_here': 'fill in return data'}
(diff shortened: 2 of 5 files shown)


myjobs/urls.py
```diff
@@ -32,4 +33,5 @@
     url(r'^toolbar/$', 'toolbar', name='toolbar'),
     url(r'^cas/$', 'cas', name='cas'),
     url(r'^topbar/$', 'topbar', name='topbar'),
+    url(r'^child-dashboard/$', child_dashboard, name='child_dashboard'),
```
tdphillips: Was this intended to be used in testing?

darrint: That's the stub we will expand on in the future to return secure elements on child sites.

myjobs/autoserialize.py
```diff
@@ -0,0 +1,24 @@
+from django.http import HttpResponse
+import json
+from functools import wraps
+
+
+def autoserialize_view_decorator(fn):
+    @wraps(fn)
+    def handle_autoserialize(request):
+        if "callback" in request.GET:
```
aspidites: If you do a get on GET, you can avoid the `callback = None` bit:
```python
callback = request.GET.get("callback")
if callback:
    content_type = "text/javascript"
else:
    content_type = "application/json"
```
Could even forgo the else block if you set the content type and change it if there is/isn't a callback.

myjobs/autoserialize.py
```diff
@@ -0,0 +1,24 @@
...
+            callback = request.GET['callback']
+            content_type = "text/javascript"
+        else:
+            callback = None
+            content_type = "application/json"
+
+        response = fn(request)
+
+        payload = json.dumps(response)
+        if callback is not None:
```
aspidites: Maybe I'm wrong, but I think these conditionals could be combined such that you don't have to repeatedly check for the state of callback.

myjobs/child_dashboard.py
```diff
@@ -0,0 +1,8 @@
+from django.http import HttpResponse
+
+from myjobs.autoserialize import autoserialize_view_decorator
+
+
+@autoserialize_view_decorator
```
aspidites: Naming a decorator "decorator" seems a bit redundant. `@autoserialize_view`, or even `@autoserialize` should be sufficient.

myjobs/child_dashboard.py
```diff
@@ -0,0 +1,8 @@
...
+def child_dashboard(request):
```
aspidites: This is that stub method you mentioned, right?

myjobs/tests/test_autoserialize.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+urlpatterns = [
+    url(r'^ping/$', ping, name='ping'),
+]
+
+
+@override_settings(ROOT_URLCONF=__name__)
+class TestAutoserialize(TestCase):
+    def test_json(self):
```
aspidites: Rather than creating a test-only view, couldn't you use the `child_dashboard` here?

darrint: Child_dashboard's behavior will change.

myjobs/autoserialize.py
darrint: You are right. Earlier I was going to validate that callback and the accept header agreed. I can just do a single conditional at the end.

myjobs/child_dashboard.py
darrint: Yes.

myjobs/tests/test_autoserialize.py
aspidites: I don't see a problem with that. As it stands, the stubbed method is untested, which might lead to a false sense of security. Rather than testing a view that doesn't exist, I think we should be testing that stub method, and update said test as the method changes. By doing so we maintain full coverage, mitigate the risk of the behavior being tested differing from the behavior that's in production, and remove redundant code.

aspidites:
Spoke in person and I agree about not conflating testing the decorator with testing the decorated view. I would appreciate if you added a test for the stub method, though. Should make things easier later.

myjobs/autoserialize.py
```diff
@@ -0,0 +1,20 @@
+from django.http import HttpResponse
```
aspidites: This should probably live in `myjobs/decorators.py`.

myjobs/child_dashboard.py
```diff
@@ -0,0 +1,8 @@
+from django.http import HttpResponse
```
aspidites: I suspect this should be in `myjobs/views.py`.

myjobs/tests/test_autoserialize.py
```diff
@@ -0,0 +1,45 @@
+import re
```
aspidites: I don't mind this being in a separate module. If the other decorators had tests, I'd say it should be in `test_decorators.py`

myjobs/tests/test_child_dashboard.py
```diff
@@ -0,0 +1,12 @@
+import json
```
aspidites: I'm on the fence here. Other tests are in `myujobs/tests/test_views.py`, but I wonder if that's becoming a bear to maintain?

PR #1702, opened 2015-08-24
Ensure that the record's "successful" field is not none before showin…

tdphillips:
…g whatsits [PD-1363]

### DIFF
--- a/templates/mypartners/includes/record_card.html
+++ b/templates/mypartners/includes/record_card.html
@@ -2,7 +2,7 @@
 <div class="product-card">
     <div class="big-title">
         {% with record.contactlogentry as cle %}
-            {{ record.subject }} {% if record.contact.name %} {% if record.subject %}-{% endif %} {{record.contact.name}} {% endif %}{% if record.contact_type == "pssemail" and cle %} <i class="fa fa-{% if cle.successful %}check-circle green{% else %}times-circle red{% endif %}" title="{% if cle.successful %}Successfully sent{% else %}Email failed, view notes for more info{% endif %}"></i>{% endif %}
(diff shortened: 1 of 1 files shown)


templates/mypartners/includes/record_card.html
```diff
@@ -2,7 +2,7 @@
 <div class="product-card">
     <div class="big-title">
         {% with record.contactlogentry as cle %}
-            {{ record.subject }} {% if record.contact.name %} {% if record.subject %}-{% endif %} {{record.contact.name}} {% endif %}{% if record.contact_type == "pssemail" and cle %} <i class="fa fa-{% if cle.successful %}check-circle green{% else %}times-circle red{% endif %}" title="{% if cle.successful %}Successfully sent{% else %}Email failed, view notes for more info{% endif %}"></i>{% endif %}
+            {{ record.subject }} {% if record.contact.name %} {% if record.subject %}-{% endif %} {{record.contact.name}} {% endif %}{% if record.contact_type == "pssemail" and cle.successful != None %} <i class="fa fa-{% if cle.successful %}check-circle green{% else %}times-circle red{% endif %}" title="{% if cle.successful %}Successfully sent{% else %}Email failed, view notes for more info{% endif %}"></i>{% endif %}
```
tdphillips: In a Django template, `if cle.successful != None ` can basically double as `if cle and cle.successful != None` because exceptions are a theory that don't actually exist.

aspidites: Trying to remember, isn't `is not None` preferable to `!= None`?

tdphillips: This may be PyCharm being PyCharm or it could be Django's template system, but writing `cle.successful is not None` was displaying as an error.

aspidites: http://stackoverflow.com/a/3257957

Translation: PyCharm is dumb. That said, I don't *think* we do any custom comparisons on fields, so you should be fine.

tdphillips: I know it's valid Python but templating is a subset of that.

PR #1707, opened 2015-08-24
[PD-1702] Made sure label was passed when creating a location from the contact form

aspidites:
Local test run:
2015-08-24-135930_956x791_scrot

The first run, as you can see, is without my changes, the second, is with my changes. I also tested the form locally.


aspidites:
Moving this to a feature branch to coordinate deployment

PR #1710, opened 2015-08-24
[PD-1701] Moved the conditional logic around for diffing code.

aspidites:
Basically, when we are using attachments, the initial value should be set as a special case. As such, I moved the initial_val/new_val in an else that is only called when the two special cases are executed.

Given the complexity of testing this and the fact that it's affecting production immediately,  regression tests were moved to a separate task. 

That said, in order to test this properly, I'll need to use the mock package to mock a file, then create a PRMAttachment factory before getting to the meat of the test, which will need to create an attachment for a contact record then remove it.


JLMcLaugh:
@aspidites Thoughts on the failing test? It looks like it might be related to the communication record migration.

aspidites:
Not sure what about tagging records would cause that. I'm looking now.

aspidites:
@JLMcLaugh looks like it's directly related to my changes. Apparently "new_val" isn't properly being changed on each iteration,

aspidites:
Fixed another way.

PR #1712, opened 2015-08-25
[PD-1701] Removing attachments no longer causes validation errors.

aspidites:
Not a fan, but it was the fastest way to fix this that worked. Will re-evaluate this functionality and improve tests in PD-1704.

### DIFF
--- a/mypartners/helpers.py
+++ b/mypartners/helpers.py
@@ -156,7 +156,7 @@ def get_form_delta(form):
                 new_val = form.data.get(field, '')
                 new_val = form.fields[field].to_python(new_val)
             except ValidationError:
-                # MultipleFileFields break hear because the two place initial
+                # MultipleFileFields break here because the two places initial
                 # data can come from are formatted differently: one is a list,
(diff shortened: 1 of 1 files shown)


mypartners/helpers.py
```diff
@@ -145,15 +146,23 @@ def get_form_delta(form):
...
+                # There are two places that initial data can come from:
+                # form.initial or form.fields[field].initial. Django
+                # favors form.initial in their code, so this code prefers
+                # form.initial first as well.
+                initial_val = form.initial.get(field, form.fields[field].initial)
+                initial_val = form.fields[field].to_python(initial_val)
+                new_val = form.data.get(field, '')
+                new_val = form.fields[field].to_python(new_val)
+            except ValidationError:
+                # MultipleFileFields break hear because the two place initial
```
tdphillips: I have nothing valuable to say so I must resort to comments.

"break hear"

"place" should probably be plural

aspidites:
@tdphillips: fixed.

PR #1716, opened 2015-08-25
[PD-1710] Fixed clone report.

aspidites:
Despite recalling doing some checks agains the clone functionality during the reporting refactor, AJ discovered that it wasn't working properly. I got it working again in addition to commenting the functionality, as some of the code may not seem necessary on first glance.

@galitskyd  can you take a look at this?

### DIFF
--- a/static/reporting.167-25.js
+++ b/static/reporting.167-25.js
@@ -156,27 +156,45 @@ Report.prototype = {
     })[0];
   },
   createCloneReport: function(json) {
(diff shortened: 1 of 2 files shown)


aspidites:
Looks like a lot of changes, but essentially I just check for the existence of undefined`s first, rather than assuming that they aren't a problem and falling through, as I was doing before.

galitskyd:
:+1: anyone else?

PR #1717, opened 2015-08-26
[PD-1706] Create PRMAttachmentFactory

aspidites:
# Problem
- No factory exists for the `mypartners.PRMAttachment` model. 
- This model references a `partner` attribute, despite never declaring it
- The above two points mean that when we want an attachment during testing, we have to reference a file on disk and have to make sure to mutate the attachment object before associating it with a file. 

# Solution
- I've created a factory which uses a byte string as the attached file, which should eliminate unnecessary IO during unit tests
- I've made `partner` a property which then retrieves the attachment's contact record's partner
- No more mutation that you have to remember to do else risk exceptions.

# Caveats
- While there is technically a new unit test, it's more of a stub to ensure that I was able to instantiate the PRMAttachmentFactory. This test will be implemented in a separate PR
- Note that I'm merging this into a feature branch. The idea is that all subtasks get merged there, then I'll create a PR for the parent ticket (tagging subtasks along the way). I'm trying this out as a way to organize my work. We'll see how it goes.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -720,6 +720,12 @@ class PRMAttachment(models.Model):
     def partner(self):
         return getattr(self.contact_record, 'partner', None)
 
+    @partner.setter
+    def partner(self, partner):
+        # TODO: find all parts in code which attempt to set the partner
+        # attribute and remove them
+        pass
+
     def save(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


aspidites:
I'm not dealing with this...

PR #1718, opened 2015-08-26
Added django-extensions ajax select to parent_site form. [PD-1607]

koonsies:
Continuation of PD-1607.. made the field for parent_site use an ajax-assisted select rather than a long drop down.


aspidites:
@JLMcLaugh, the thing we were talking about in the meeting about items being unselected on form load wont be an issue here. I tested this by:

1. adding a parent site and saving
2. opening the same seo site and saving again
3. removing the parent site

After 1 and 2, the parent site was set correctly, and after 3, the parent site was None again. 

Seems to work as expected. I only tried with a couple domains, though I don't expect that will be an issue. 

@koonsies out of curiosity, have you timed (maybe using debug toolbar) how much this saves us in loading time?

PR #1719, opened 2015-08-26
[PD-1704, PD-1706, PD-1707, PD-1708] Refactor and write tests for form delta functionality

aspidites:
The original code (including the quickfix that I worked on yesterday) are erroneous in that for uploaded file removal, the change isn't actually logged. This corrects that by logging the ID of the deleted file on removal, and the filename on insertion. 

I'd have liked to have been consistent, but there are just enough special things that we are doing that enforcing some kind of consistency would have ballooned the complexity of this ticket. For instance, since we represent the attachments to remove as a multiple choice list and the files to add as a multiple file field, their representations are inherently going to be different.

### DIFF
--- a/mypartners/helpers.py
+++ b/mypartners/helpers.py
@@ -5,10 +5,12 @@ import os
 from urlparse import urlparse, parse_qsl, urlunparse
 from urllib import urlencode
 
-from django.db.models import Min, Max, Q
+from django.db.models import Min, Max, Q, Model
 from django.core.exceptions import ValidationError
+from django.core.serializers.json import DjangoJSONEncoder
 from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
+from django.forms import ChoiceField
 from django.http import Http404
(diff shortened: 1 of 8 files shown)


aspidites:
I know about the failures, but I'm having trouble running tests locally. Getting connection refused errors even on a pristine quality-control or master branch.

aspidites:
Tests are passing now. All should be right with the world.

mypartners/helpers.py
```diff
@@ -112,12 +114,17 @@ def log_change(obj, form, user, partner, contact_identifier,
         change_msg = get_change_message(form) if action_type == CHANGE else ''
     delta = get_form_delta(form) if action_type == CHANGE else {}
 
+    try:
+        delta = json.dumps(delta, cls=DjangoJSONEncoder)
+    except:
```
tdphillips: This shouldn't stay, should it?

aspidites: Nope. Good catch.

PR #1721, opened 2015-08-27
[PD-1635] ui - improve layout of cities network page

galitskyd:
Cities page:
screen shot 2015-08-27 at 9 44 21 am

New title for state page, it was just "State":
screen shot 2015-08-27 at 9 44 48 am


seo/views/search_views.py
```diff
@@ -2035,12 +2035,12 @@ def seo_states(request):
     all_link = all_link.format(intcomma(search.count()))
 
     # Turn search results into a dict formatted {state:count}
-    search = dict(search.facet_counts()['fields']['state'])
+    search = dict(search.facet_counts().get('fields').get('state'))
```
tdphillips: If "fields" isn't in the facet counts, this will still fail. You might want something like `.get('fields', {})`

PR #1722, opened 2015-08-27
Move country_short search to an exact string field. [PD-1581]

darrint:
Add a country_short_exact field.

Move the search on country_short to country_short_exact. This keeps country codes from being interpreted as search stop words.


tdphillips:
We discussed a possible procedure for this before @JLMcLaugh arrived this morning. The consensus was that, while it looks good, Jason should also take a look at it. Tests will never pass until Jenkins solr is updated and, barring word from higher up, that won't happen until after this gets merged.

JLMcLaugh:
Merge conflicts

PR #1725, opened 2015-08-31
Staging to qc

mklauber:


aspidites:
This will introduce a regression. The correct code is on QC

PR #1727, opened 2015-08-31
Added missing test [PD-1608]

koonsies:
Added a missing test that was causing issues previously. It is now corrected and should be added alongside the prevent grandchildren SEO Site code.

Sorry for the lateness!


koonsies:
Feel free to ignore this PR for this sprint. It is a "nice to have" rather than a required bit of code.

PR #1730, opened 2015-09-01
404 redirect correction seo [PD-287]

koonsies:
The following code is written to fix the hard redirect to "/" that was occurring when no results were found for a job search. 

The code should now present a 404 when given an invalid:
* company name
* moc slug
* job category ("new-jobs")

A 200 is return with "No results found" when:
* a proper company name is used, but no jobs exist for that company
* a proper MOC slug is used, but no jobs exist
* a proper category is used, but no jobs exist
* a location is provided that has no jobs (even if the location is made up.. i.e., /blamp/usa/jobs"


seo/tests/test_views.py
```diff
@@ -1587,9 +1588,19 @@ def test_view_all_jobs(self):
     def test_slug_tag_permanent_redirect(self):
         """Filter paths not in canonical order result in a 301 to the canonical
         path"""
+        slugs_and_paths = OrderedDict([
+            ('sample_title', '/jobs-in/'),
+            ('a/sample/loc', '/jobs/'),
+            ('test/moc/slug', '/vet-jobs/'),
+            ('sample_facet', '/new-jobs/'),
+            ('sample_company', '/careers/'),
```
JLMcLaugh: The reason for pulling default_settings.SLUG_TAGS directly was to make sure we're testing any new tags and accounting for any change in order. Could you assert that the lengths are the same so we're know to update this test in the future?

koonsies: Yup. I was thinking the same thing. I am currently comparing the length of the settings dict and the manufactured dict. There are not comparing order, though. 

The lines in question are 1600 and 1622

PR #1731, opened 2015-09-01
Use solr highlighting for search results. [PD-1438]

darrint:
This uses solr's native highlighting where possible to highlight search results.

Remove description from the solr query to improve Solr network use.

**Note 1:** some search behavior is changed.

**Note 2:** ~~Some tests are failing because they are expecting solr to avoid exact string matches entirely and only follow stems.~~ To get exact highlight matches I had to admit exact string matches.


aspidites:
Re: Note 1 - what behavior has changed?

darrint:
Ordering of some search results is different.

seo/helpers.py
```diff
@@ -532,7 +532,8 @@ def get_bread_box_headings(filters=None, jobs=None):
 def get_jobs(custom_facets=None, exclude_facets=None, jsids=None,
              default_sqs=None, filters={},  fields=None, facet_limit=250,
              facet_sort="count", facet_offset=None, mc=1,
-             sort_order='relevance', fl=search_fields):
+             sort_order='relevance', fl=search_fields,
+             additional_fields=[]):
```
aspidites: Mutable default arguments are a bad idea.

seo/helpers.py
```diff
@@ -567,8 +566,16 @@ def get_jobs(custom_facets=None, exclude_facets=None, jsids=None,
...
     sqs = sqs.bf('recip(ms(NOW/HOUR,salted_date),1.8e-9,1,1)')
 
+    # Limit the retrieved results to only fields that are actually needed.
     if fields:
-        sqs = sqs.fields(fields)
+        field_list = list(fields)
+    else:
+        field_list = list(fl)
+
+    if additional_fields:
```
aspidites: I'm sure I'm missing something. What does specifying field_list + additional_fields do that just specifying those additional fields with field_list wouldn't? Or is something happening to field_list in place that you don't want to happen to the additional fields?

seo/helpers.py
```diff
@@ -1001,8 +1010,11 @@ def prepare_sqs_from_search_params(params, sqs=None):
         if exact_title:
             sqs = sqs.filter(title_exact__exact=title)
         else:
-            sqs = sqs.filter(SQ(content=Raw(title)) | SQ(title=Raw(tb)))\
-                     .highlight()
+            # We have to query on description here so that highlighting
+            # matches the exact term and not a stem.
+            sqs = sqs.filter(SQ(content=Raw("((%s))^1" % title)) | SQ(title=Raw(tb)) |
+                             SQ(description=Raw(title))) \
```
aspidites: Ask @galitskyd , `\`s make me sad.

seo/search_backend.py
```diff
@@ -342,11 +342,16 @@ def search(self, query_string, sort_by=None, start_offset=0, end_offset=None,
...
+            kwargs['hl.alternateField'] = 'description'
+            kwargs['hl.maxAlternateFieldLength'] = \
+                settings.SEARCH_FRAGMENT_SIZE
+            kwargs['hl.requireFieldMatch'] = 'true'
+            kwargs['hl.fragsize'] = settings.SEARCH_FRAGMENT_SIZE
+            kwargs['hl.snippets'] = settings.SEARCH_SNIPPETS
             kwargs['hl.mergeContiguous'] = 'true'
-            kwargs['hl.simple.pre'] = '<b>'
-            kwargs['hl.simple.post'] = '</b>'
+            kwargs['hl.simple.pre'] = '###{{{###'
```
aspidites: Huh?

seo/helpers.py
darrint: Need to be able to add description to the field list without interfering with the default. Feeds need it. Probably others too. We don't want the description in the field list for search results.

seo/search_backend.py
darrint: see `highlight_solr`. For safety I strip the markdown/html tags then replace the highlight character sequences with bold tags.

seo/helpers.py
aspidites: :+1:

seo/search_backend.py
aspidites: Ah. I still think built-in string formatting would be safer/cleaner than the arbitrary escape sequence:

```python
kwargs['hl.simple.pre = "{hl_pre}"
kwargs['hl.simple.post'] = "{hl_post}"
...
highlited.format(hl_pre="<b>", hl_post="</b>")
```

If there are more substitutions, you could probably just save them all then pass that to format:
```python
substitutions = {
    "hl_pre": "<b>",
    "hl_post": "</b>"
}
...
highlighted.format(**substitutions)
```
```

darrint: Talked with @aspidites in person. We'll stick with the replace as we don't want to run untrusted data through format.

JLMcLaugh:
Merge conflicts

PR #1733, opened 2015-09-02
Removed code for PRM/Reporting maintenance

galitskyd:
Once we are ready to remove the maintenance splash pages from Reporting and PRM this can be merged.


templates/myreports/reports.html
```diff
@@ -12,43 +12,7 @@
 {% endblock %}
```
aspidites: Wouldn't it be better to keep this code, just in case we have to do something like this in the future?

galitskyd: I wouldn't mind keeping something like this in on our apps but this code and implementation leaves a lot to be desired

aspidites: Sounds like a ticket to me :P

galitskyd: I'll leave it up to @fezick on figuring out if we want a feature like this.

PR #1734, opened 2015-09-02
[PD-1618] CORS detection via OPTIONS call

aspidites:
Not exactly sure how I should bother testing this. I think that there is already a test for the context manager which is used to conditionally include the javascript. As such, a new test wouldn't do anything unless we were actually testing the javascript.

### DIFF
--- a/seo/cache.py
+++ b/seo/cache.py
@@ -150,7 +150,7 @@ def get_domain_parent(request):
     if request.user.is_staff and 'domain' in request.REQUEST:
         # filter() + first() will return None rather than raising an exception
         site = SeoSite.objects.filter(
-            domain=requset.REQUEST.get('domain')).first()
+            domain=request.REQUEST.get('domain')).first()
     else:
         # SITE might not be set on the settings object
         site = getattr(settings, 'SITE', None)
--- a/seo/cache.py
+++ b/seo/cache.py
@@ -148,12 +148,11 @@ def get_domain_parent(request):
         - SeoSite object or None
     """
     if request.user.is_staff and 'domain' in request.REQUEST:
-        try: 
-            return SeoSite.objects.get(domain=request.REQUEST.get('domain')).parent_site 
-        except:
-            return None
-    try:
-        return settings.SITE.parent_site
-    except:
-        return None
-     
+        # filter() + first() will return None rather than raising an exception
+        site = SeoSite.objects.filter(
+            domain=requset.REQUEST.get('domain')).first()
+    else:
+        # SITE might not be set on the settings object
+        site = getattr(settings, 'SITE', None)
+
+    return getattr(site, 'parent_site', None)
--- a/seo/context_processors.py
+++ b/seo/context_processors.py
@@ -3,4 +3,4 @@ from seo.cache import get_site_config, get_domain_parent
 
 def site_config_context(request):
     return {'site_config': get_site_config(request),
-            'domain_parent':get_domain_parent(request)}
+            'domain_parent': get_domain_parent(request)}
--- a/seo/models.py
+++ b/seo/models.py
@@ -84,7 +84,6 @@ class NonChainedForeignKey(ForeignKey):
                 raise ValidationError('%s is a child entity and cannot be a parent'
                                         % potential_parent)
         return value
-        
 
 
 def term_splitter(terms):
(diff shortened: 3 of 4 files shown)


seo/cache.py
```diff
@@ -148,12 +148,11 @@ def get_domain_parent(request):
         - SeoSite object or None
     """
     if request.user.is_staff and 'domain' in request.REQUEST:
-        try:
```
aspidites: Was originally going to fill out the except clause with expected failures before I realized that I could capture expected failures by not having any to begin with. Objections? I'm more than happy to revert if we don't like this style.

aspidites:
@darrint addressed after speaking with Hal.

aspidites:
Closing, as I discovered that a same-origin OPTIONS call doesn't accomplish what we want.

aspidites:
If you're wondering why the call doesn't do anything other than set a variable to false on failure, it's intentional. This is a stub which will be populated in PD-1624

seo/cache.py
koonsies: As we had discussed before, I'm a bit on the fence about this. I worry the first() method could cause some unexpected behavior in the future if two sites are somehow created with the same domain. I'm not sure throwing an exception is the right answer in this scenario either, but that's my hangup.

aspidites: Business rules should be such that no two sites should have the same domain. While I've raised similar arguments to yours in the past, the answer has always been of the form "That is an invalid use case as these sites are only created by a small handset of DirectEmployers employees, and if more than one domain is attached to a site, we have bigger problems."

Further, if there are two domains attached to the site, by what means would we reliably pick the "correct one"? Absent an answer to that question, I suspect that "The first one" is as good an answer as any.

koonsies: I suppose so. And so long as business rules are followed, this will work perfectly. I'm just skeptical of that always happening! But I have no major concerns with the way this is handled.

haliscooper:
Yep, just following a hopefully better approach with more granular tickets, all that's coming, tdphillips

haliscooper:
PD-1625, I think

tdphillips:
:+1: pending @aspidites and @koonsies discussion.

seo/cache.py
haliscooper: I like this answer, I don't think I've given it myself before, but it is very true :)
"That is an invalid use case as these sites are only created by a small handset of DirectEmployers employees, and if more than one domain is attached to a site, we have bigger problems."

But wouldn't be averse to a unique index on the domain column, or however it might be implemented in the django world, to avoid those bigger problems in the first place.

tdphillips: SeoSite has an implicit foreign key to Site, which is one of Django's models. Modifying those isn't pleasant but adding functionality like we did is easy.

haliscooper: But if we were to switch the django_site_domain index to being unique in the DB itself, that would throw an error if someone ever were to create a new row in django_site with an existing domain, right?

tdphillips: I'm not sure what Django would do for something unexpected like that but it would be an error of some type.

At a quick glance, we already have two entries in the base Site table for wrp.jobs. I'm not entirely sure how that happened but I suspect the solution would be to remove the one that isn't tied to a SeoSite.

aspidites:
@koonsies , sorry for the late reply - non-functional workstation most of the day. What if, as a compromise, we created a separate ticket to cover ensuring that the business rules are enforced? That is, we ensure that domain is unique as @haliscooper suggested?

JLMcLaugh:
I'd rather assume there may be multiple sites until its enforced in the code or database. There's at least 1 duplicate in the django site table, which tells me it's happened at last once before.

haliscooper:
Lets put an index on the domain column, delete that one entry, and enforce it

aspidites:
@JLMcLaugh, that still wouldn't address the question of what to do in the case of a duplicate. As it stands now, if there are duplicates, I assume the first is the correct one (how would I know otherwise).

aspidites:
@haliscooper,

> Lets put an index on the domain column, delete that one entry, and enforce it

As in right now, or a separate ticket?

aspidites:
Looking at the old code, the difference is that before we would set the parent site to None if there were multiples (that single case that currently exists in production), and my code returns the first match, which is analogous to getting the first one that was created. Also note that this is only the case for domain switching. That is, the erroneous case is only ever an issue when staff tries to use domain switching. For other cases, the site is grabbed by SeoSite instance, which means that even if there are two SeoSites with the same domain, it won't matter.

PR #1735, opened 2015-09-03
[PD-1668] blocks admin - integrate bootstrap 3 into the main codebase

galitskyd:
* Added bootstrap 3 to the repo.
* Bootstrap 3 requires 1.9+ jQuery so added the most recent version of 1.x, 1.11.3. jQuery 2.x does not support IE8.
* Removed the bootstrap 3 that only had the grid system being used by blocks. Updated reference to use the new bootstrap 3 file.

The front facing admin is using bootstrap 3 but rest of microsite is still using 1.8. Creating a new base with bootstrap 3 specifically for the front facing admin gives us more flexibility and doesn't require me to "fix up" our seo_base, or make functionality to switch between 1.8 and 3.

Pretty simple admin right now:
screen shot 2015-09-03 at 11 15 45 am

**Being logged in is required to view the dashboard**

Right now no links go anywhere. My next ticket I will be working on the Event Manager overview. Once that is done I will link to it. After that Blocks overview is next.

Later on I'd like to set up more site config specific UI elements. But for now just a simple base.


templates/myblocks/myblocks_base.html
```diff
@@ -8,7 +8,7 @@
         <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
         <meta name="msvalidate.01" content="B9E4E5CC00BB58B097061E7DBFC6FA6B" />
         <meta name="viewport" content="width=device-width" />
-        <link rel="stylesheet" type="text/css" href="{{ STATIC_URL }}bootstrap.min.css">
+        <link rel="stylesheet" type="text/css" href="{{ STATIC_URL }}bootstrap-3/css/bootstrap.min.css">
```
koonsies: Where can this template be viewed? I tried to look back through the code, but I don't understand in what instance the myblocks_base.html template is used.

fezick: It is used whenever a page that was built with blocks is viewed.

koonsies: Oh, alright. Is there an example of such a page that I could look at?

galitskyd: You'd have to create a myblocks page locally using microsite/admin. (Don't think any have been made at this point)

PR #1737, opened 2015-09-03
Migration fix for parent site field [PD-1608]

koonsies:
I discovered that South requires a line to be inserted that contains a placeholder for inspection rules whenever a custom field is declared. I included this line as well as the migration for this change.


tdphillips:
We can't make migrations on seo until this is resolved and the other PR has conflicts. I'm going to go ahead with this one.

PR #1738, opened 2015-09-04
[PD-1657] [PD-1658] events admin - templates

galitskyd:
Templates for Event Email Manager Admin. Delete doesn't actually delete anything since there is no way to make new templates in this system yet. The links don't go anywhere yet either because I haven't built out those templates.

This is the first real implementation of Bootstrap 3, which claims 'Mobile First'. This new way of going about building out templates is actually reverse in how we do things usually even though we've been saying "be mobile first". One clear indication of this is when looking at the styles. Mobile styles are the default and are overridden by the tablet, desktop, large desktop media queries.

There is still a decent amount of work left that affect these pages (the header needs a major overhaul). But I believe these pages are done besides changing out some of the placeholder templating work on the Manage Templates page. Currently the events only having one template was kind of unexpected so I kept in the logic to handle an iterable, which a foreignKey is not. I can change it if someone feels strongly that it should be changed.

## Event Overview
screen shot 2015-09-17 at 4 16 14 pm

screen shot 2015-09-17 at 4 16 20 pm

## Manage Header/Footer [PD-1657]
screen shot 2015-09-16 at 4 30 39 pm

screen shot 2015-09-17 at 4 40 35 pm

screen shot 2015-09-16 at 4 30 46 pm

screen shot 2015-09-17 at 4 40 15 pm

## Manage Templates [PD-1656]
screen shot 2015-09-16 at 4 35 54 pm

screen shot 2015-09-16 at 4 36 08 pm


aspidites:
I assume screenshots will go up after you're done? Just need them to verify that what I see is what you intended me to see. If so, I'll just come back to this one later when its done, rather than trying to stay ahead of the curve as usual.

aspidites:
Also, is this synced with QC? I'm seeing a bunch of @darrint's work here.

aspidites:
Looks like it's synced now, at least. Thanks

templates/myemails/manage_templates.html
```diff
@@ -0,0 +1,97 @@
...
+    <div class="col-md-12">
+        <h3 id="{{event.name.lower}}" style="margin-bottom: 0;">{{event.name}}</h3>
+        <p style="padding: 0;">Lorem ipsum dolor sit amet, consectetur adipisicing elit. Accusamus ad assumenda atque consequatur delectus dicta ipsum iste laudantium maxime nesciunt qui rerum sunt, tenetur voluptas voluptates. Blanditiis delectus iure quasi.</p>
+        <a class="awrap" href="">
+            <div class="add-row">
+                <i class="fa fa-plus-circle green"></i><span class="cta">Create a new template</span>
+            </div>
+        </a>
+        <table class="table table-striped">
+        {% for eh in "start" %}
```
tdphillips: Does this do something I'm not understanding? Looks like "eh" is getting each character from the string "start" one at a time. Later on, we're checking `eh.is_active`.

galitskyd: Placeholder right now. Eventually templates will be an iterable on events instead of a ForeignKey. If/Once that happens I'll just simply changes "start" and eh to something more meaningful.

PR #1742, opened 2015-09-09
[PD-1579] 4th wildcard facet for SEO

koonsies:
* Added model fields to support 4th wildcard
* Updated admin site to reflect change
* Updated code that pulls wildcard information to properly display 4th wildcard
* Updated tests to properly check for 4th wildcard functionality


tdphillips:
It looks like this contains everything from 1737 as well. Do you have a preferred merge order? If this one gets merged first, the other would have in effect already been applied.

koonsies:
@tdphillips Yup. I wasn't certain how development would go with this project, so I put #1737 out early (it was causing migration creations to fail). Either the other can be closed and this one merged or the other merged first, then this one.

koonsies:
Scratch that. I went ahead and closed #1737. It had failed tests anyway (I think it was a jenkins problem)

seo/cache.py
```diff
@@ -72,7 +72,7 @@ def get_facet_count_key(filters=None, query_string=None):
     )
 
 
-def get_custom_facets(request, filters=None, query_string=None):
+def get_custom_facets(request, filters=None, query_string=None):
```
aspidites: Was this intentional? Looks like space was added after the colon?

seo/helpers.py
```diff
@@ -1353,7 +1353,7 @@ def _facet_query_result_counts(tagged_facets, sqs):
 
 
 def get_solr_facet(jsids, filters=None, params=None):
-    custom_facets = settings.STANDARD_FACET
+    custom_facets = settings.STANDARD_FACET
```
aspidites: Same question here with the extra whitespace.

seo/models.py
```diff
@@ -23,6 +23,9 @@
 from haystack.inputs import Raw
 from haystack.query import SQ
 
+from south.modelsinspector import add_introspection_rules
+add_introspection_rules([], ["^seo\.models\.NonChainedForeignKey"])
```
aspidites: I'm sure this is necessary, but educate me: What does this actually do/fix?

seo/tests/test_filters.py
```diff
@@ -307,7 +307,12 @@ def get_num_of_widgets(*args, **kwargs):
             factories.SeoSiteFacetFactory(customfacet=facet, seosite=self.site,
                                           facet_group=3)
             facet_counts['facet_slab'].append((facet, 5))
-
+        for i in range(num_items):
```
aspidites: You aren't using the `i` here, so a call to `CustomFacetFactory.create_batch` is probably sufficient, in lieu of the for-loop.

seo/models.py
tdphillips: [South doesn't like custom fields.](http://south.readthedocs.org/en/latest/customfields.html) As NonChainedForeignKey is basically a ForeignKey with a funny clean, we can tell South about it and South will happily continue creating migrations with that field.

aspidites: Ah. Cool. Thanks, Troy.

tdphillips:
Conflicts.

koonsies:
Weird. I merged QC into the branch from command line with no conflicts.

tdphillips:
I went ahead and merged #1737 in the meantime. It's a little concerning that this still lists the migration as new.

koonsies:
Good looking out. I merged with QC again and it appears that the migration is no longer flagged as such.

PR #1743, opened 2015-09-10
Increase required email failures to 3 [PD-217]

tdphillips:
Instead of inspecting the most recent unprocessed email log, we need to look at the last three of them, processed or not, to determine if we should stop sending emails.

Tests pass.


tasks.py
```diff
@@ -214,23 +214,29 @@ def process_user_events(email):
     Processes all email events for a given user.
     """
     user = User.objects.get_email_owner(email=email)
-    logs = EmailLog.objects.filter(email=email,
-                                   processed=False).order_by('-received')
+
+    logs = EmailLog.objects.filter(email=email).order_by('-received')
```
tdphillips: As this runs daily, filtering on processed=False may never give us more than a couple of events - especially if someone has saved searches that are sent monthly. We'll later filter on processed=False when we're updating the queryset.

tasks.py
```diff
@@ -214,23 +214,29 @@ def process_user_events(email):
...
     newest_log = logs[0]
 
-    filter_by_event = lambda x: [log for log in logs if log.event in x]
+    filter_by_event = lambda x, num=None: [log for log in logs
+                                           if log.event in x][:num]
```
tdphillips: This logic is off. I think the list splitting should be on the `logs` list and not the generated one.

We need to grab the last `num` events and determine if they're all failures, not get the maximum number of failures we can.

tdphillips:
Tests failed but that appears to be a Jenkins issue.

myjobs/tests/test_tasks.py
```diff
@@ -7,12 +7,35 @@
 
 
 class TaskTests(MyJobsBase):
+    @staticmethod
+    def make_email_logs(email, event, received, processed, number):
+        for _ in range(number):
```
aspidites: Do we care about `bulk_create` in this case?

tasks.py
aspidites: :+1:

myjobs/tests/test_tasks.py
tdphillips: I guess I probably should.

tdphillips: Done.

tdphillips:
As this is checking three events and not three days worth of events, it's still possible for an account to get disabled on one day.

This task runs at midnight, however, and each send has multiple retries. Both of these when taken together gives us multiple chances per day for temporary issues to resolve themselves.

JLMcLaugh:
Do we know what a typical failure looks like? If the receiver's mail server is down, is SenGrid doing a dozen retries? All I can find is that they keep retrying for 24 hours and don't keep more than 100,000 retries in the queue.

tdphillips:
SendGrid doesn't appear to retry sending emails - the only retries I see are the ones where they retry submitting event logs to us.

If the receiver's mail server is down, we get a bounce posted back at some point. That event will then be processed by us at the following midnight.

If I need to check the last three days of activity rather than the last three pieces of activity, that's going to increase complexity somewhat.

aspidites:
@JLMcLaugh is this good to go? I see @tdphillips  answered your question, but don't know if that's enough to merge or not.

PR #1745, opened 2015-09-14
Reorganize unclear dependencies  [PD-1739]

darrint:
This doesn't remove circular dependencies, but clarifies some mystery situations. It should be easier to fix up circular deps in the future.


seo/admin.py
```diff
@@ -44,7 +44,7 @@
                         GoogleAnalyticsCampaign, SeoSite, SeoSiteFacet,
                         SeoSiteRedirect, SiteTag, SpecialCommitment, ViewSource)
 from seo.queryset_copier import copy_following_relationships
-from seo.signals import check_message_queue
+from seo.models import check_message_queue
```
aspidites: Doesn't this make things more confusing, considering check_message_queue is defined in seo.signals, not seo.models?

aspidites:
I feel like this PR is doing more than advertised. In particular, the changes to seo.models seems to be about more than simply reorganizing dependencies.

darrint:
To move the signals import I have to do what I did, moving it to the models code as Troy's links point out. And yes, for 1.8 there are new and better options available.

seo/admin.py
aspidites: Disregard. Now that it isn't 5pm, this makes sense.

JLMcLaugh:
Merge conflicts

PR #1748, opened 2015-09-14
[PD-1554] First go at reproducing Hal's ERD as Django models.

aspidites:
# Description
This ticket essentially attempts to convert @haliscooper's [ERD diagram](https://directemployersfoundation.atlassian.net/browse/PD-1380) into Django models. At this point, I'm more concerned with getting the relationships correct. As such, while I'm more than willing to take suggestions on naming, I'd prefer to kee pthe bike shedding to a minimum. You will also notice that I am merging this into a feature branch. The intent there is that once we are sure things are faithfully represented and addressed any and all concerns (including any shedding of bikes which needs to occur), we can merge this into QC.

# Changes
I created 11 new models
- Column
- ColumnFormat
- Configuration
- ConfigurationColumn
- DataType
- InterfaceElementType
- PresentationType
- ReportPresentation
- ReportType
- ReportingType
- UserType

This in turn generated 15 MySQL Tables:
- myreports_column
- myreports_columnformat
- myreports_configuration
- myreports_configurationcolumn
- myreports_configurationcolumn_column_formats
- myreports_datatype   
- myreports_interfaceelementtype 
- myreports_presentationtype
- myreports_reportingtype
- myreports_reportingtype_report_types
- myreports_reportingtype_user_types 
- myreports_reportrepresentation 
- myreports_reporttype
- myreports_reporttype_data_types
- myreports_usertype

# Visualizations

### Django Models
myreports

### ERD
pd-1554

### Django Models
models2

### ERD
erd2

### Django Models
models3

### ERD
erd3

### DIFF
--- a/myreports/models.py
+++ b/myreports/models.py
@@ -81,15 +81,19 @@ class Report(models.Model):
 class UserType(models.Model):
     user_type = models.CharField(max_length=10)
     is_active = models.BooleanField(default=False)
+    reporting_types = models.ManyToManyField(
+        'ReportingType', through='UserReportingTypes')
 
 
 class ReportingType(models.Model):
     reporting_type = models.CharField(max_length=50)
     description = models.CharField(max_length=500)
+    report_types = models.ManyToManyField(
+        'ReportType', through='ReportingTypeReportTypes')
     is_active = models.BooleanField(default=False)
 
 
-class UserReportingType(models.Model):
+class UserReportingTypes(models.Model):
     user_type = models.ForeignKey('UserType')
     reporting_type = models.ForeignKey('ReportingType')
     is_active = models.BooleanField(default=False)
@@ -98,10 +102,12 @@ class UserReportingType(models.Model):
 class ReportType(models.Model):
     report_type = models.CharField(max_length=50)
     description = models.CharField(max_length=500)
+    data_types = models.ManyToManyField(
+        'DataType', through='ReportTypeDataTypes')
     is_active = models.BooleanField(default=False)
 
 
-class ReportingTypeReportType(models.Model):
+class ReportingTypeReportTypes(models.Model):
     reporting_type = models.ForeignKey('ReportingType')
     report_type = models.ForeignKey('ReportType')
     is_active = models.BooleanField(default=False)
@@ -113,7 +119,7 @@ class DataType(models.Model):
     is_active = models.BooleanField(default=False)
 
 
-class ReportTypeDataType(models.Model):
+class ReportTypeDataTypes(models.Model):
     report_type = models.ForeignKey('ReportType')
     data_type = models.ForeignKey('DataType')
     is_active = models.BooleanField(default=False)
@@ -125,7 +131,7 @@ class Configuration(models.Model):
 
 
 class ReportPresentation(models.Model):
-    report_data = models.ForeignKey('ReportTypeDataType')
+    report_data = models.ForeignKey('ReportTypeDataTypes')
     presentation_type = models.ForeignKey('PresentationType')
     configuration = models.ForeignKey('Configuration')
     display_name = models.CharField(max_length=50)
@@ -169,4 +175,6 @@ class ConfigurationColumn(models.Model):
     interface_element_type = models.ForeignKey(InterfaceElementType)
     alias = models.CharField(max_length=100)
     multi_value_expansion = models.PositiveSmallIntegerField()
+    column_formats = models.ManyToManyField(
+        'ColumnFormat', through='ConfigurationColumnFormats')
     is_active = models.BooleanField(default=False)
--- a/myreports/models.py
+++ b/myreports/models.py
@@ -83,24 +83,52 @@ class UserType(models.Model):
     is_active = models.BooleanField(default=False)
 
 
-class DataType(models.Model):
-    data_type = models.CharField(max_length=100)
+class ReportingType(models.Model):
+    reporting_type = models.CharField(max_length=50)
     description = models.CharField(max_length=500)
+    is_active = models.BooleanField(default=False)
+
+
+class UserReportingType(models.Model):
+    user_type = models.ForeignKey('UserType')
+    reporting_type = models.ForeignKey('ReportingType')
     is_active = models.BooleanField(default=False)
 
 
 class ReportType(models.Model):
     report_type = models.CharField(max_length=50)
     description = models.CharField(max_length=500)
-    data_types = models.ManyToManyField(DataType)
     is_active = models.BooleanField(default=False)
 
 
-class ReportingType(models.Model):
-    reporting_type = models.CharField(max_length=50)
+class ReportingTypeReportType(models.Model):
+    reporting_type = models.ForeignKey('ReportingType')
+    report_type = models.ForeignKey('ReportType')
+    is_active = models.BooleanField(default=False)
+
+
+class DataType(models.Model):
+    data_type = models.CharField(max_length=100)
     description = models.CharField(max_length=500)
-    user_types = models.ManyToManyField(UserType)
-    report_types = models.ManyToManyField(ReportType)
+    is_active = models.BooleanField(default=False)
+
+
+class ReportTypeDataType(models.Model):
+    report_type = models.ForeignKey('ReportType')
+    data_type = models.ForeignKey('DataType')
+    is_active = models.BooleanField(default=False)
+
+
+class Configuration(models.Model):
+    name = models.CharField(max_length=50)
+    is_active = models.BooleanField(default=False)
+
+
+class ReportPresentation(models.Model):
+    report_data = models.ForeignKey('ReportTypeDataType')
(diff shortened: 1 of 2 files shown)


myreports/models.py
```diff
@@ -76,3 +76,75 @@ def regenerate(self):
...
+class ReportType(models.Model):
+    report_type = models.CharField(max_length=50)
+    description = models.TextField(max_length=500)
+    data_types = models.ManyToManyField(DataType)
+    is_active = models.BooleanField(default=False)
+
+
+class ReportingType(models.Model):
+    reporting_type = models.CharField(max_length=50)
+    description = models.TextField(max_length=500)
```
koonsies: Correct me if I'm wrong, but I think putting max_length on a TextField doesn't do anything. Or is it different in MySQL?

myreports/models.py
```diff
@@ -76,3 +76,75 @@ def regenerate(self):
 
         self.results.save('%s-%s.json' % (self.name, self.pk), results)
         self._results = contents
+
+
+class UserType(models.Model):
+    user_type = models.CharField(max_length=10)
+    is_active = models.BooleanField(default=False)
```
koonsies: In the ERD, the various intermediary tables (User_Reporting_Type, ReportType_ReportingType) have an "Active" flag as well. Without declaring those intermediary tables in the models, we won't have those flags.

tdphillips: What does it mean for an intermediate table to be deactivated that shouldn't be accomplished by removing the relation?

koonsies: The only thing I can think of is preservation of data. I wanted to point it out just in case it was designed this way for a reason. In previous projects we kept inactivated relationships in order to see where relationships had formed in the past. In this context, though, that may not be a concern.

myreports/models.py
```diff
@@ -76,3 +76,75 @@ def regenerate(self):
...
+    configuration = models.ForeignKey(Configuration)
+    column = models.ForeignKey(Column)
+    interface_element_type = models.ForeignKey(InterfaceElementType)
+    alias = models.CharField(max_length=100)
+    multi_value_expansion = models.PositiveSmallIntegerField()
+    column_formats = models.ManyToManyField(ColumnFormat)
+    is_active = models.BooleanField(default=False)
+
+
+class ReportRepresentation(models.Model):
```
koonsies: I'm confused on this model. The closest thing I can find in the ERD is Report_Presentation, which is a intermediary table for PresentationType, Configuration, and ReportTypeDataType (The intermediary for report type and data type). Is this a design change that went over my head?

myreports/models.py
aspidites: @koonsies I'll ask hal, but my immediate response is that I didn't want to have to manually create the through tables. Also, I'm curious as to what is accomplished by having it on the through table, like Troy asked.

koonsies: That's totally fair. I'd say it is worth following up on. In all fairness, every table has an active flag.. So he may have just painted with a very wide brush.

myreports/models.py
aspidites: That it doesn't is a [bug](https://code.djangoproject.com/ticket/22206). I'll change it though.

myreports/models.py
aspidites: @koonsies : 
2015-09-16-165512_189x216_scrot

aspidites: Typo, it is indeed ReportPresentation. `data_type` I got from `Report_Data_ID` in the ERD. From the name it seemed to be a foreign key to `Data_Type` in the ERD, but the fact that the arrow is going to the through table may suggest otherwise. I'll ask hal about this.

I'll be tagging each of the iterations so that they are easier to spot.

aspidites:
# Update
Made a few changes based on @koonsies and @haliscooper's feedback. 

# Changes
- Removed all many to many fields from django models
- Added models for through tables

# Missing Things
- I need to re-add the many-to-manys on the Django side, but wanted to make sure the base ERD was create first, and also wasn't sure which models said many-to-many's should be on, so wanted to experiment with that on a separate tag just in case I need to revert
- Hal and I had talked about removing the active bits from models which are part of a many-to-many. I wanted that change recorded separately from trying to figure out the base ERD as well.

aspidites:
@tdphillips  reference?

tdphillips:
We're using MySQL ~5.6 so I guess it's higher than that.

aspidites:
Reverted back to char fields with max_length > 255 because of [this](https://dev.mysql.com/doc/refman/5.6/en/char.html). That is:
> Values in VARCHAR columns are variable-length strings. The length can be specified as a value from 0 to 65,535.

aspidites:
@JLMcLaugh, as the through tables were manually creatd, I wasn't sure where to put the `ManyToManyField`s on the django side. For instance, does it make more sense to say that a `UserType` has `reporting_types` or that a `ReportingType` has `user_types`? Naturally, either one willb e queriable, but I'm curious as to which should be the primary mode of operation. @haliscooper asked whether or not it was possible to use both. While I haven't tried yet, the question then would become would it make sense to do both? 

Other than the validation of the schema, those are the only remaining questions for this ticket. I'll report back after I experiment with putting them on both ends to see if that route is even feasible.

JLMcLaugh:
@aspidites I don't think the location of the m2m relationships matters too much. It might be a little cleaner on whichever model you think will be getting queried by the views. e.g., we'll have a user tied to a request, so we're probably querying UserType more often than ReportingType. That way, when someone checks the model related to the view, they'll see that UserType is connected to ReportingType.

aspidites:
# Update

The same as the last time, except I implemented `ManyToManyField`s as follows:
- `UserType` has a many to many called `reportint_types` through `UserReportingTypes`
- `ReportingType` has a many to many called `report_types` through `ReportingTypeReportTypes`
- `ReportType` has a many to many called `data_types` through `ReportTypeDataTypes`
- `ConfigurationColumn` has a many to many called `column_formats` through `ConfigurationColumnFormats`

I also went ahead and pluralized the name of all the through tables to be consistent with django. Results of these changes have been updated in the OP under "Third Iteration".

aspidites:
# Update
This should be the last update, pending feedback. This final iteration(*) removes the extra active bits @haliscooper  and I talked about. As such, the final result should be some permutation of revisions 3 or 4.

(*) Tentatively

myreports/models.py
```diff
@@ -76,3 +76,105 @@ def regenerate(self):
...
+
+class ReportType(models.Model):
+    report_type = models.CharField(max_length=50)
+    description = models.CharField(max_length=500)
+    data_types = models.ManyToManyField(
+        'DataType', through='ReportTypeDataTypes')
+    is_active = models.BooleanField(default=False)
+
+
+class ReportingTypeReportTypes(models.Model):
```
koonsies: Should probably be "Type" rather than "Types"

myreports/models.py
```diff
@@ -76,3 +76,105 @@ def regenerate(self):
...
+    is_active = models.BooleanField(default=False)
+
+
+class DataType(models.Model):
+    data_type = models.CharField(max_length=100)
+    description = models.CharField(max_length=500)
+    is_active = models.BooleanField(default=False)
+
+
+class ReportTypeDataTypes(models.Model):
```
koonsies: "Types"

myreports/models.py
```diff
@@ -76,3 +76,105 @@ def regenerate(self):
...
+    is_active = models.BooleanField(default=False)
+
+
+class ColumnFormat(models.Model):
+    name = models.CharField(max_length=50)
+    format_code = models.CharField(max_length=2000)
+    is_active = models.BooleanField(default=False)
+
+
+class ConfigurationColumnFormats(models.Model):
```
koonsies: "Formats"

myreports/models.py
tdphillips: Pluralization was mentioned [here.](https://github.com/DirectEmployers/MyJobs/pull/1748#issuecomment-141477180)

koonsies: Oh, I missed that. I don't think I've ever seen that as a standard before.

tdphillips:
This is @haliscooper's baby but he doesn't seem keen on looking after it.

haliscooper:
Not a matter of keenness, a matter of not knowing django, and needing the time to sit and figure this #&$%#&$% out ;-)

aspidites:
@haliscooper,

I made the ERD diagrams precisely because you don't know django.

haliscooper:
Yep, just want to understand how we get from model to the DB structure with the models and the ORM, I'm approving the model code as well, correct?

aspidites:
Django's innards generate the sql table etc from the model. I then reverse
engineered the resulting tables to come up with the ERD and manually moved
the generated boxes to closely match your ERD. In other words, verifying
the ERD is indeed verifying the models, as the ERD was programmatically
generated from those models. I addressed Jason M about model-specific
concerns, which he approved earlier last week.

PR #1749, opened 2015-09-15
Create a QueryRedirect model and the middleware to use it [PD-1715]

tdphillips:
Adds a QueryRedirect model to allow redirects when the original destination does not 404, query strings optional.

2015-09-16-094118_1238x515_scrot

Tests pass.

### DIFF
--- a/middleware.py
+++ b/middleware.py
@@ -276,13 +276,12 @@ class RedirectOverrideMiddleware(object):
             params = filter((lambda x: x[1] != ''), params)
             if len(params) > 0:
                 # This QueryRedirect has query params that we need to check
-                for param in params:
+                for param, value in params:
                     # Determine if the current value is multi-valued and
                     # normalize it to a list regardless
-                    param = (param[0], param[1].split('|'))
+                    values = value.split('|')
                     # Turn the previous tuple into a list of param=value strings
-                    test_params = (lambda x: ['='.join([x[0], y])
-                                              for y in x[1]])(param)
+                    test_params = ['='.join([param, x]) for x in values]
                     if any(test_param.lower()
                            in request.META['QUERY_STRING'].lower()
                            for test_param in test_params):
--- a/middleware.py
+++ b/middleware.py
@@ -259,12 +259,4 @@ class RedirectOverrideMiddleware(object):
         # Configuration is cached. The effect on response time is
         # hopefully minimal.
         configuration = get_site_config(request)
-        if configuration.not_found_override.exists():
-            # Check both the current path with and without a trailing slash
-            paths = [request.path,
-                     (request.path[:-1] if request.path.endswith('/')
-                      else request.path + '/')]
-            not_found = configuration.not_found_override.filter(
-                old_path__in=paths).first()
-            if not_found:
-                return redirect(not_found.new_path, permanent=True)
+        # TODO: Do stuff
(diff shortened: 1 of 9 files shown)


middleware.py
```diff
@@ -251,20 +252,71 @@ def filter_custom_facets_by_production_status(custom_facets):
...
-        if configuration.not_found_override.exists():
-            # Check both the current path with and without a trailing slash
-            paths = [request.path,
-                     (request.path[:-1] if request.path.endswith('/')
-                      else request.path + '/')]
-            not_found = configuration.not_found_override.filter(
-                old_path__in=paths).first()
-            if not_found:
-                return redirect(not_found.new_path, permanent=True)
+        paths = [request.path,
```
aspidites: is `rstrip` not sufficient here?:
```python
>>> "http://secure.my.jobs/reporting/admin".rstrip("/")
'http://secure.my.jobs/reporting/admin'
>>> "http://secure.my.jobs/reporting/admin/".rstrip("/")
'http://secure.my.jobs/reporting/admin'
>>> url = "http://secure.my.jobs/reporting/admin/"
>>> url2 = url[:-1]
>>> url2
'http://secure.my.jobs/reporting/admin'
>>> url.rstrip("/") == url2.rstrip("/")
True
>>> 
```

tdphillips: I want the path with and without an end slash. Using rstrip would not remove the need for the if - if the path doesn't have an end slash, we'll still need to check and add one.

seo/models.py
```diff
@@ -1229,8 +1231,54 @@ def show_sites(self):
     what_helptext = models.TextField(blank=True)
     where_helptext = models.TextField(blank=True)
 
-    not_found_override = models.ManyToManyField('redirects.Redirect', null=True,
-                                                blank=True)
+
+@python_2_unicode_compatible
+class QueryParameter(models.Model):
```
tdphillips: Looking back now that this is basically done, I see some improvements.

This originally had a foreign key to QueryRedirect, meaning any number of parameters were allowed. Then we moved to three explicit parameters and no extras, which is a OneToOne to QueryRedirect. It really could have just been a charfield on QueryRedirect. Meh. It looks nice this way.

middleware.py
```diff
@@ -251,20 +252,71 @@ def filter_custom_facets_by_production_status(custom_facets):
...
+            # defined query parameters for the current QueryRedirect object,
+            # in the form of a list of tuples of (parameter, value).
+            params = [getattr(r, param).value
+                      if hasattr(r, param) else ''
+                      for param in param_names]
+            params = zip(param_names, params)
+            params = filter((lambda x: x[1] != ''), params)
+            if len(params) > 0:
+                # This QueryRedirect has query params that we need to check
+                for param in params:
```
aspidites: You can iterate over the first and second items independently (and avoid the index lookups) in a list of tuples.

middleware.py
aspidites: Ooh, I see what you're doing now, and order doesn't matter. Saw brackets and immediately thought it was just a list comprehension.

middleware.py
tdphillips: I just got what you're saying. I'll do that now.

tdphillips:
Quote from Seth:
> this looks great.

aspidites:
I wanna run this locally before I merge. Otherwise :+1:

aspidites:
Good to go once @tdphillips  remmoves the DO NOT MERGE flag again. Confirmed it worked on his machine and mine. Tested with pipes as well.

tdphillips:
Now that I know what the blank at the top of the foreign key widget is for, I've now filled it with appropriate text.

2015-09-17-111702_1229x349_scrot

PR #1750, opened 2015-09-15
[PD-1758] added immutable defaults for search_transformer.py

koonsies:
remove mutable defaults from the search_transformer.py


tdphillips:
I merged but did not delete the branch. I'm assuming this also needs to go to QC?

PR #1751, opened 2015-09-16
[PD-1758] added immutable defaults for search_transformer.py

koonsies:
Good call @tdphillips 

PR for keeping staging in sync with QC


aspidites:
Totally makes sense.

PR #1752, opened 2015-09-16
[PD-1747] Added models for non-user email outreach.

aspidites:
This adds two new models:
- CommonEmailDomain: Essentially a list of common email domains which we don't want to allow for outreach.
- OutreachEmailDomain: a mapping from company to email domain (unique together) which will be used to determine from which email domains outreach efforts will be processed from.

I also included a couple of signals:
- save_outreach_email_domain: checks that a domain can't be added for use as outreach if it is too common
- ~~save_common_email_domain: removes email domains that were added in the past if a new common email is added.~~

~~The second signal is probably contentious (and I am waiting to hear back about whether it should exist).~~

Just spoke with Hal. Retroactively removing domains is not desired. Nor is allowing a user to explicitly say that despite being common, we want to add the domain.

I did not make any forms or the like, and expect to address any validation concerns then. I do still need to write tests for this, but wanted it up for visibility/commentary.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -841,10 +841,3 @@ def save_outreach_email_domain(sender, instance, **kwargs):
         raise ValidationError(
             "The domain %s has been blacklisted, as it is too common." %
             instance.domain)
-
-@receiver(post_save, sender=CommonEmailDomain,
-          dispatch_uid='post_save_common_email_domain_signal')
-def save_common_email_domain(sender, instance, **kwargs):
-    # Remove outreach email domains which should now be blacklisted do to the
-    # new common domain.
-    OutreachEmailDomain.objects.filter(domain=instance.domain).delete()
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -561,7 +561,7 @@ class ContactRecord(models.Model):
     company_ref = 'partner__owner'
     objects = ContactRecordManager()
 
-    created_on = models.DateTimeField(auto_now=True)
+    created_on = models.DateTimeField(auto_now_add=True)
     created_by = models.ForeignKey(User, null=True, on_delete=models.SET_NULL)
     partner = models.ForeignKey(Partner, null=True, on_delete=models.SET_NULL)
     contact = models.ForeignKey(Contact, null=True, on_delete=models.SET_NULL)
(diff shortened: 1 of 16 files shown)


mypartners/models.py
```diff
@@ -811,3 +812,39 @@ def natural_key(self):
...
+          dispatch_uid='pre_save_outreach_email_domain_signal')
+def save_outreach_email_domain(sender, instance, **kwargs):
+    if CommonEmailDomain.objects.filter(domain=instance.domain).exists():
+        raise ValidationError(
+            "The domain %s has been blacklisted, as it is too common." %
+            instance.domain)
+
+@receiver(post_save, sender=CommonEmailDomain,
+          dispatch_uid='post_save_common_email_domain_signal')
+def save_common_email_domain(sender, instance, **kwargs):
```
tdphillips: If this one is kept, I suspect someone will need to be informed (perhaps with a message).

aspidites:
"DO NOT MERGE" is because I need to add a data migration to pre-populate the common emails table.

aspidites:
Added admin support. Decided to go with a custom save as pre_save hook was a bit hostile in the user interface. I didn't bother with custom filtering options in the admin, as nothing else that's part of mypartners even works in the django admin, and I think there are tickets to give this a user-friendly admin later down the road.

aspidites:
I doubt collation will change anytime soon, but I went ahead and changed the check to an iexact just in case.

mypartners/models.py
```diff
@@ -811,3 +812,46 @@ def natural_key(self):
...
+class CommonEmailDomain(models.Model):
+    """
+    Common email domains which should not be allowed for outreach email
+    domains.
+    """
+
+    class Meta:
+        ordering = ["domain"]
+
+    domain = models.CharField(max_length=255, unique=True)
```
koonsies: Since this is using a simple charfield, it doesn't have specific validation for ensuring you enter 'google.com' rather than  ' google.com  ' or so on. Since we're doing string comparison to ensure a domain entered is not a common domain, perhaps we should be doing a strip on those values? Might avoid confusion in the future when someone copy+pastes a email and it contains an extra space.

aspidites: I'd rather do that at the form level. Much better to have good data enter the database in the first place, rather than mold the database to check for garbage values. Also, iterating through each domain to strip them would be slower than the generated query from an `iexact`.

koonsies: That makes sense. I am mentioning it at a model level because if these entries are added via the admin, updating one of the "clean" methods on the model would also allow you to strip inputs from the admin form.

PR #1753, opened 2015-09-17
Child site checking  [PD-1610]

darrint:
Add a decorator that will allow child sites to call secure parent sites to show blocks that should be secure.

Integrate with stubbed child dashboard call.

### DIFF
--- a/myjobs/cross_site_verify.py
+++ b/myjobs/cross_site_verify.py
@@ -1,0 +1,151 @@
+import re
+import logging
+from django.http import HttpResponse
+
+from functools import wraps
+from universal.helpers import get_domain
+from seo.models import SeoSite
+from urlparse import urlparse, parse_qs
+
+
+XRW = 'XMLHttpRequest'
+logger = logging.getLogger(__name__)
+
+
+class DomainRelationshipException(Exception):
+    pass
+
+
+def extract_hostname(url):
+    if url is None:
+        return None
+    else:
+        return urlparse(url).hostname
+
+
+def extract_hostname_from_header(header):
+    if header is None:
+        return header
+    match = re.match(r'(.*):[0-9]+$', header)
+    if match:
+        return match.group(1)
+    else:
+        return header
+
+
+def parse_request_meta(meta):
+    method = meta.get('REQUEST_METHOD', '')
+
+    host = extract_hostname_from_header(meta.get('HTTP_HOST'))
+
+    origin = extract_hostname(meta.get('HTTP_ORIGIN'))
+
+    referer = extract_hostname(meta.get('HTTP_REFERER'))
+
+    xrw = meta.get('HTTP_X_REQUESTED_WITH')
+
+    qs = meta.get('QUERY_STRING', '')
+
+    qsreferer = parse_qs(qs).get('referer', [None])[0]
+
+    return (method, host, origin, referer, qsreferer, xrw)
+
+
+def guess_child_domain(host, origin, referer, qsreferer):
+    if origin is not None:
+        return origin
+
+    if referer is not None:
+        return referer
+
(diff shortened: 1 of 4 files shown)


myjobs/management/commands/testfast.py
```diff
@@ -0,0 +1,52 @@
+import sys
```
aspidites: is this module supposed to be included in this PR?

myjobs/tests/test_child_dashboard.py
```diff
@@ -7,6 +7,14 @@
 
 class TestChildDashboard(MyJobsBase):
     def test_child_dastboard(self):
```
aspidites: dastboard?

myjobs/management/commands/testfast.py
tdphillips: If by included you mean "a file named sys.py in project root", sys is a builtin.

myjobs/cross_site_verify.py
```diff
@@ -0,0 +1,151 @@
...
+            raise DomainRelationshipException('qsreferer-not-host')
+
+    raise DomainRelationshipException('no-child-info')
+
+
+def fail():
+    return HttpResponse(status=404, content="not found")
+
+
+def is_permitted_method(method):
```
aspidites: Why aren't we using [`CORS_ALLOW_METHODS`](https://github.com/ottoyiu/django-cors-headers) here, since we already rely on that library?

myjobs/management/commands/testfast.py
aspidites: @tdphillips  I meant `myjobs/management/commands/testfast.py`. Given the PR is about child site checking, I was confused as to why a new command about running tests fast was included in it.

darrint: I can do a separate PR.

myjobs/cross_site_verify.py
darrint: As far as I know CORS_ALLOW can only instruct browsers. The intent here is to disallow from the server.

myjobs/cross_site_verify.py
```diff
@@ -0,0 +1,151 @@
...
+def is_self_or_child(host_site, origin_site):
+    return (is_parent_self_call(host_site, origin_site) or
+            is_child_of_parent(host_site, origin_site))
+
+
+def xrw_ok(xrw):
+    return xrw == XRW
+
+
+def get_site(domain):
```
koonsies: @aspidites did we decide that we were simply taking the "first" match of a SeoSite in the event of multiple of the same domain? If so, this should not return a None on duplicates.

myjobs/cross_site_verify.py
```diff
@@ -0,0 +1,151 @@
+import re
+import logging
+from django.http import HttpResponse
+
+from functools import wraps
+from universal.helpers import get_domain
```
koonsies: Unused import  (Thanks, PyCharm!)

myjobs/cross_site_verify.py
aspidites: `CORS_ALLOW_METHODS` is used in the middleware, which is used by the server, no?

myjobs/cross_site_verify.py
aspidites: Pretty sure, given it's done in a few other places. Ping @JLMcLaugh. Begs the question though, what is get_site supposed to do that isn't done by grabbing `settings.SITE`? Only real use case for getting a site by domain ambiguously would be when domain switching as a staff user.

myjobs/cross_site_verify.py
darrint: As far as I know it just sends a header back.

PR #1756, opened 2015-09-17
Add a command for quickly running all non-Django tests. [PD-1771]

darrint:
Moved this from another PR.


myjobs/management/commands/testfast.py
```diff
@@ -0,0 +1,52 @@
...
+            self.loader.loadTestsFromNames(names, module))
+
+    def suite_without_django(self, suite):
+        tests = (t for t in self.iterate_over_suite(suite)
+                 if not self.is_django(t))
+        return unittest.suite.TestSuite(tests)
+
+    def iterate_over_suite(self, suite):
+        for test_like in suite:
+            if not isinstance(test_like, unittest.suite.TestSuite):
```
aspidites: I'm just trying to understand the recursive case. When test_like isn't an instance of unittest.suite.TestSuite, what could it be? I assume either an instance of django's test case or another iterable.

aspidites:
One concern I have over this command is that it tells me nothing about which tests are being run. With the normal test runner, I know which test is being run along with which module it is in. With this, I just get a bunch of logging lines which seem to be showing translations from input text to lucene queries. How hard would it be to improve the output of this thing?

Also, it seems that this is only useful for your solr tests, is that right? Do o intend to move tests which don't technically require the Django framework outside of django-specific tests so that the yield from including this is greater?

darrint:
It only tells you about failed tests just like Django. I left some logging on from when I did the search stuff. If it turns out to be annoying I'll happily turn it off.

I made this because I have similar tests for cross site authentication. They were together with this in another PR until I separated them at your request.

I didn't want to duplicate the hack at the bottom of the search tests.

PR #1758, opened 2015-09-18
Address cleanup [PD-1489]

koonsies:
One-time script for updating incorrect label fields in the My Partners -> Location table.

A disclaimer.. This script is far more involved than it needs to be. I spent a bit of personal time researching the method with which Darrin did his search transformer and decided to use that method to create this simple parser. 

* If a label is found to be an address, attempt to shift values down (Label -> AddLine1 -> AddLine2)
* "Score" if location.label's probability of being an address line
* If score is 0-10, prompt user to verify address (~150)
* If score is 20+, automatically verify (~600)


seo/address_clean_script.py
```diff
@@ -0,0 +1,935 @@
...
+                        'office','ofc',
+                        'room','rm',
+                        'suite','ste',
+                        'trailer','trlr',
+                        'lobby','lbby',
+                        'hangar','hngr',
+                        'floor','fl',
+                        'basement','bsmt']
+
+po_box_terms = ['po','p.o.', 'p o','po','post office box']
```
tdphillips: No biggie. Is there a reason that "po" appears twice?

koonsies: I really like that acronym. Other than that... No.

aspidites:
When is this getting ran?

koonsies:
No time set. I just need to get prod write access temporarily in order to run it.

tdphillips:
@JLMcLaugh This is something we've never done before. What's our procedure? Does @haliscooper need to review as well since this is modifying user data?

haliscooper:
As long as JM or JS or I look at it and are good, then it's good.  Possibly something that JM should run, though, just so a limited set of people are messing with prod data.

seo/address_clean_script.py
```diff
@@ -0,0 +1,935 @@
...
+        if location.address_line_one:
+            if not location.address_line_two:
+                location.address_line_two = location.address_line_one
+        location.address_line_one = location.label
+        return location
+
+def determine_label_is_address():
+    count_converted = 0
+    count_skipped = 0
+    for location in Location.objects.raw("select * FROM mypartners_location where label != '' and label != address_line_one "
```
JLMcLaugh: What's label != address_line_one  looking for? Are we assuming that if someone put the same address twice, they're deliberately using the street address as a label?

koonsies: That is one way to consider it. Another reason I added that line is in case (somehow) the script gets run twice, it doesn't keep pushing addresses down (label -> line 1.. then label -> line 1 -> line 2 on second run)

JLMcLaugh: You could probably add a filter(contacts_library=None) to this.

PR #1759, opened 2015-09-18
[PD-1756] Added model for in-coming outreach email addresses.

aspidites:
This model contains the email that a non-user would send an outreach email request to in order to be processed. The lack of admin access is intentional, as per @haliscooper , we still haven't decided exacxtly how we want this table to populated. As such, I didn't want to blindly give admin access if this turned out to be something that would be determined programatically, rather than by human intervention.

As I'm using default django functionality and this model isn't referenced anywhere (part of another ticket), I didn't see any benefit to writing a code. If such a test were to fail, we have bigger problems than this model.


aspidites:
Depends on feedback for #1761. If we decide to go with that, then I'll need to update this migration. If not, this will be ready to go.

PR #1761, opened 2015-09-18
Updated domain field to be a URL field and use autocomplete for django admin.

aspidites:
@koonsies had mentioend the use of CharField for domain. While there isn't a `DomainField`, I wonder if this is still an improvement, since it comes with a validator. After all, the domain for an email address is a valid URL, right?

I ran schemamigration with --update to avoid having to make another migration.

Update - This hadn't been merged yet, so I went ahead and updated the admin form as well, rather than openning up yet another pull request with dependencies.


aspidites:
@mklauber  if we decide to include this, you'll want to migrate qc down to mypartners/0019 first.

aspidites:
@koonsies  changing this to a URLField also takes care of your concerns with whitespace. I just saved "      bacon.org" and saw "bacon.org" when I re-opened the form.

koonsies:
Excellent. I had wondered if those field types would make a difference.

tdphillips:
Any preference, @mklauber?

PR #1762, opened 2015-09-18
[PD-1761] Outreach email workflow states

aspidites:
These states will represent the status of email processing. It was determined to not use the existing Status table, as these statuses are of a different kind. See @haliscooper or @fezick  for a run down of how they are different. This is not exposed to the admin intentionally, and the same argument for a lack of tests apply. Tests for this model are bound to be made once the different pieces actually need to talk to each other.


aspidites:
Dependes on #1759 and potentially #1762 (which would require me to update the migration).

PR #1765, opened 2015-09-21
Add searching and filtering of query redirects [PD-1773]

tdphillips:
This duplicates Django's Redirect admin, which we don't test.

Tests pass.

### DIFF
--- a/seo/admin.py
+++ b/seo/admin.py
@@ -1244,10 +1244,13 @@ class MocParameterAdmin(admin.TabularInline):
 
 
 class QueryRedirectAdmin(ForeignKeyAutocompleteAdmin):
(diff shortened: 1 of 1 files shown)


aspidites:
Would using an autocomplete box for SeoSite domains allow the page to load faster? I only ask because I was just asked to do something similar for the non-user outreach domain admin (with respect to companies). 

"Not what I was asked to do" is a valid response. Just curious, considering the trend of what's being asked for in django admins (@koonsies was asked to do something similar for the parent site functionality).

@fezick?

tdphillips:
Autocomplete like this?

2015-09-16-094118_1238x515_scrot

aspidites:
re: autocomplete - yes. 
re: list on the right side - I Guess I don't really know what that list is for, other than scrolling to find a domain.

PR #1766, opened 2015-09-21
Update myemail models [PD-1655]

darrint:
This is a literal implementation of the request in [PD-1655]

Tests pass except for the admin tests in myemails.

I need some direction as to where to take this next.


myemails/models.py
```diff
@@ -19,7 +19,6 @@
 class EmailSection(models.Model):
     SECTION_TYPES = (
         (1, 'Header'),
-        (2, 'Body'),
```
tdphillips: If anyone's concerned, the current event structure is completely unused. We have no risk of invalidating existing rows.

aspidites:
Looks like according to @tdphillips' most recent comments, the ValueEvent and CronEvent models should disappear.

tdphillips:
I'll leave that up to whoever works on it. My understanding was that none of the current models fit with the intended direction.

darrint:
@tdphillips @aspidites I held off on removing those types until I understood better. Two questions now:

1. If there's no member user expectations to preserve, should we remove the existing admin entirely? i.e. `myemails/admin.py`

2. Is there any data migration to do at all? If not, I see what @tdphillips meant by drop existing tables.

aspidites:
There isn't any data to migrate, unless you are expecting to prepopulate some tables. Either way, dropping tables should be fine.

myemails/models.py
```diff
@@ -155,65 +150,22 @@ def send_email(self, for_object):
...
-
-class CreatedEvent(Event):
-    model = models.ForeignKey(ContentType,
-                              limit_choices_to={'model__in': [
-                                  'invoice'
-                              ]})
+        if len(templates) == 0:
+            return None
+        else:
+            return templates[0]
```
galitskyd: templates.first()?

tdphillips: .first() would handily duplicate the entire if/else as it returns None if there's nothing there.

galitskyd: :+1:

PR #1767, opened 2015-09-21
Re-add delete button on communication records [PD-1735]

tdphillips:
The context variable being referenced as `object_id` is actually `item_id`. Pointing at the correct var makes the delete button show up as it should.

Running tests now.


tdphillips:
Deleting until archiving stuff happens.

fezick:
Research into why the button went away in the first place has revealed that the delete button that is not being displayed is a true delete, not a soft delete that archives the record on the backend. As such, we cannot merge this until archiving is in place for CRs.

PR #1772, opened 2015-09-22
[PD-1780] Update company search filter logic

koonsies:
### Issues Fixed:
* Company filtering used incorrect method for determining if company exists
   * Company filtering now checks the model itself for existence and not a separate function
* Company filtering was returning 404 for companies that had jobs (but did not exist in DB)
   * Company filtering now only returns a 404 if company does not exist in DB and has no jobs
* Function call had ambiguous name
   * Added comment to function call to prevent future misunderstandings

_Tests updated to verify new logic._

### DIFF
--- a/seo/helpers.py
+++ b/seo/helpers.py
@@ -1443,6 +1443,12 @@ def jobs_and_counts(request, filters, num_jobs, fl=search_fields):
 
 
 def get_company_data(filters):
+    '''
+    Return the thumbnail for a company if it exists. Returns None if company does
+    not have a thumbnail.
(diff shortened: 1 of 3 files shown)


aspidites:
If this is a hotfix, it's pointing toward qc, unless you intended on hotfixing master after confirming that qc was properly working?

koonsies:
That was more or less the idea. Get it to QC to ensure it's going to function properly.

aspidites:
Is solr stuff properly in QC so this can be checked against it?

PR #1775, opened 2015-09-23
[PD-1791] Resolve some circular imports

aspidites:
# Rationale
Attempting to use any third-party tools (sphinx, pytest, or even [this code snippet](https://gist.github.com/aspidites/c9ee5e4e4c0ad5dacaa8) results in a circular import error. This PR attempts to do the minimum amount of work possible to resolve the circular import without disrupting the rest of the ecosystem or resorting to hacks.

# Changes
- I moved `PartnerEmailChoices` from `mypartners.forms` to `mysearches.forms` as `partner_email_choices`
-- it is only used in mysearches.forms
-- I needed to use the User model, which is already imported in `mysearches.forms`, but not `mypartners.forms`
- I moved the logic in `mypartners.models` which attaches a user to a contact on save
-- Unless I misunderstand why we were doing so, we only care if a user with an email exists when modifying contact, as the email becomes read only in that case. Further, we only automatically create users when partner saved searches are created
-- As the code was only executed on save, a user created after a contact would not show a read-only email address, as the user didn't exist on time of save of that contact; subsequent saves of that contact, however, would then mark the email as read-only since the user would exist and thus be created on that subsequent save
- All references to `User` in `mypartners.model` were replaced with `"myjobs.User"` and the import of `User` was removed.

I'm relying on jenkins to run tests, as I'm not currently able to run the entire suite. 

~~# Feedback~~
~~I'm especially curious as to if my assumptions about the `Contact.save` functionality are misguided, as that single assumption is what makes this PR possible.~~

Spoke with Jason S. The only time a user should be attached to a contact is if a partner saved search is created. Originally, when a partner saved search is created a new account is created as well. The intended behavior here is that if a user already exists, that one should be associated with the contact. On the other hand, if a contact is created using an email that happens to belong to a user, that user should NOT be associated with the contact UNLESS a partner saved search is created for that contact.

# Updated Changes
- ~~I rationalized (please correct me if I'm wrong) that if a contact was attached to a user, that user's email would be the same as the contact's email, so there was no need to iterate and attach the user's email instead of the contact's email as they wouldn't differ. Once a user is associated with a contact, that contact can't change their email anyway~~
- ~~I attached the user to the contact only when the form is saved to avoid the case where a user is attached to a contact merely because a partner starts creating a saved search for that contact. I suspect that if the saved search isn't created, the association shouldn't happen.~~
- ~~PartnerEmailChoices has been reduced to a comprehension that has been inlined.~~

Apparently some other voodoo happens with how we make the contact choices and such, because when I made the changes above, tests about the initial email started failing. Rather than waste time investigating that, I decided to compromise with at least knowing that the user isn't associated with the contact in all cases.

### DIFF
--- a/mypartners/forms.py
+++ b/mypartners/forms.py
@@ -331,19 +331,6 @@ class PartnerForm(NormalizedModelForm):
         return instance
 
 
-def PartnerEmailChoices(partner):
-    choices = [(None, '----------')]
-    contacts = Contact.objects.filter(
-        partner=partner, archived_on__isnull=True)
-    for contact in contacts:
-        if contact.user:
-            choices.append((contact.user.email, contact))
-        else:
-            if contact.email:
-                choices.append((contact.email, contact))
-    return choices
-
-
 class ContactRecordForm(NormalizedModelForm):
     date_time = SplitDateTimeDropDownField(label='Date & Time')
     length = TimeDropDownField()
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -14,7 +14,6 @@ from django.db import models
 from django.db.models.signals import pre_delete, pre_save, post_save
 from django.dispatch import receiver
 
-from myjobs.models import User
 from postajob.location_data import states
 from universal.helpers import json_to_query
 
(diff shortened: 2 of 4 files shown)


tdphillips:
Is this worthy of story points given how many people were in the conversation at various times?

aspidites:
@tdphillips: I was originally going to say no (I asked @fezick), but given that I'd like to have the functionality backed by unit tests and add some helpful documentation, yes (thus, the DO NOT MERGE).

aspidites:
I attach the user when the contact list is being created, since that's essentially what the old code was doing. I wonder, however, if I should instead be attaching the user in save? It would be available when iterating through contact choices in that case, but the idea is that the emails should be the same anyway, right?

PR #1776, opened 2015-09-23
Add soft delete to partners and communication records [PD-1775][PD-1776]

tdphillips:
I've overridden the manager on these models so the default is to grab non-archived rows. This should take care of every pre-existing use of these particular models, with the exception of saved searches tied to an archived contact/partner. I have an open query on that in the ticket but that shouldn't block review.

I don't think there's a good way of determining if a given partner or communication record is a duplicate of an existing one so I have not implemented anything resembling @aspidites' Contact unarchiving.

Tests pass.


mypartners/tests/test_views.py
```diff
@@ -1144,8 +1144,8 @@ def test_email_forward_parsing(self):
             self.assertEqual(self.data['text'], record.notes)
             self.assertEqual(Contact.objects.all().count(), 2)
 
-            Contact.objects.get(email=record.contact_email).delete()
-            record.delete()
+            Contact.objects.filter(email=record.contact_email).delete()
+            ContactRecord.objects.filter(id=record.pk).delete()
```
tdphillips: As .delete() has been overridden to archive, we need to delete using a queryset.

darrint: The delete-which-actually-deletes is just needed to make this one test pass?

tdphillips: It's in a loop. I can either keep track of the count and modify it for future loops or I can clean up at the end of the loop.

But yes, it's currently allowing the test to pass.

mypartners/models.py
```diff
@@ -134,6 +151,20 @@ def sort_by(self, *fields):
         return self.get_query_set().sort_by(*fields)
 
 
+class ArchivedModel(models.Model):
+    archived_on = models.DateTimeField(null=True)
+
+    objects = SearchParameterManager()
+    all_objects = SearchParameterManager(archived=None)
```
aspidites: Should that be `archived=False`?

koonsies:
+1 over my head is anything is wrong with it

PR #1778, opened 2015-09-24
[PD-1779] Fix Search Transformer Asst. Bugs

koonsies:
## Sort Transformer Bug Resolution
* Improper handling of single quotes
 * Added single quotes to list of characters to be considered part of their connected term
* Improper handling of field declarations with quotes or parentheses
 * Added handling to identify field:"value value" format and field:("value value") format.

Test cases added for these examples. 
### To run test cases, either... 
* use ./manage.py testfast
* from shell: 
 * from seo.search_transformer import transform_search
 * transform_search("your string here")
* runserver and try some searches!
 * add a print line
 * at seo -> helpers.py line 1012-1013
 * print title
 * or use ipdb...
 * or cross your fingers and click Merge!

### DIFF
--- a/seo/search_transformer.py
+++ b/seo/search_transformer.py
@@ -76,9 +76,7 @@ class Peekable(object):
 
     def next(self):
         try:
-            next_val = iter(self).next()
-            print next_val.token_type, next_val.token, next_val.flags
-            return next_val
+            return iter(self).next()
         except StopIteration:
             return self.last
 
@@ -179,7 +177,7 @@ term_punctuation_pattern = "|".join([re.escape(c)
                                      for c in term_punctuation])
 
 # Need to identify ':' NOT followed by space or controlled character as delimiter for specific field search term
-field_re = re.compile(r'(\w+:)(?!\s|%s)(.*)' % term_punctuation_pattern)
+field_re = re.compile(r'(\w+:)(?=\"|\()(.*)')
 
 # Need to include dashes as part of search terms.
 # Also want to include : and ^ just in case those need
@@ -574,11 +572,9 @@ def optimize_tree(tree, root):
     ]
 
     for optimization in optimizations:
-        print optimization
         new_tree = optimization(tree, root)
         if new_tree is not None:
             tree = new_tree
-        print tree
 
     for i, child in enumerate(tree.children):
         if isinstance(child, AstTree):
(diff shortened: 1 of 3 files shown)


seo/search_transformer.py
```diff
@@ -480,7 +506,6 @@ def prepend_term_prefix(tree, root):
         new_children.extend(tree.children[2:])
         return AstTree('and', children=new_children)
 
-
```
darrint: Minor nit: This blank line is for pep8.

seo/search_transformer.py
```diff
@@ -298,6 +309,8 @@ def string(self):
                 return self.head_string()
         elif self.node_type == 'paren':
             return "(%s)" % self.head_string()
+        elif self.node_type == 'field':
+            return "%s" % "".join(self.child_strings())
```
darrint: I'm a little suspicious of how this will serialize in weird situations. All I can think to do is just keep an eye on it though.

aspidites:
Unless stated otherwise, my comments are more me trying to learn this thing, not things that need to change.

koonsies:
@aspidites Makes sense. Thanks for being vigilant. There's a lot going on, so careful review will help keep us from having to yank this out of production again.

aspidites:
Are we worried at all about unbalanced symbols like parentheses? There are cases where I'd get a partial translation (eg. a trailing "AND ") and other times where I would just get escaped parentheses. 

In other words, are we shooting for "good enough" or bullet proof?

seo/search_transformer.py
```diff
@@ -286,7 +297,7 @@ def __init__(self, node_type, head=None, tail=None, flags=None, children=None):
 
     def string(self):
         if self.node_type == 'or':
-            return "%s" % " OR ".join(self.child_strings())
+            return " OR ".join(self.child_strings())
         elif self.node_type == 'and':
```
aspidites: I think this was the bit that was causing the trailing "AND " I mentioned in another comment. Does that have negative consequences, or does solr ignore the AND in that case?

aspidites:
Could we document AstTree? Specifically, I don't have any way of knowing what the allowed node types are, and thus whether or not something is being parsed correctly. Also, I can't see where head is actually used.

seo/search_transformer.py
haliscooper: As bullet proof as possible

PR #1781, opened 2015-09-25
Call child dashboard from microsites [PD-1625]

darrint:
* Put child dashboard endpoint on microsites in addition to secure.
* Plumb in the rest of the client code for calling the correct microsite.
* Lots of hand testing in a browser to make sure `@cross_site_verify` is doing the right thing in context.

This appears to correctly accept/reject all the situations I can throw at it. I don't have access to ie8 for testing and I haven't covered adding a referer querystring for ie8.

__Note:__ If we haven't discussed it enough, Django's CSRF middleware is incompatible with what we are doing here. Our `@cross_site_verify`, assuming we have correctly implemented it, is sufficient to protect against CSRF according to the OWASP cheat sheet.

### DIFF
--- a/myjobs/child_dashboard.py
+++ b/myjobs/child_dashboard.py
@@ -1,9 +1,12 @@
-from django.http import HttpResponse
+from django.views.decorators.csrf import (
+    csrf_exempt as django_csrf_exempt)
 
 from myjobs.autoserialize import autoserialize
 from myjobs.cross_site_verify import cross_site_verify
 
 
+# The django csrf exemption should stay first in this list.
+@django_csrf_exempt
 @cross_site_verify
 @autoserialize
 def child_dashboard(request):
--- a/seo/cache.py
+++ b/seo/cache.py
@@ -139,13 +139,14 @@ def get_site_config(request):
                 site_config = Configuration.objects.get(id=2)
         cache.set(config_cache_key, site_config, timeout)
     return site_config
-    
-def get_domain_parent(request):
+
+
+def get_secure_blocks_site(request):
     """"
-        Returns the parent site for the currently selected domain (if exists)
-        
+        Returns the secure blocks site for the current site
+
         Returns:
-        - SeoSite object or None
+        - SeoSite object or None if we can't figure out the current site.
     """
     if request.user.is_staff and 'domain' in request.REQUEST:
         # filter() + first() will return None rather than raising an exception
@@ -155,9 +156,11 @@ def get_domain_parent(request):
         # SITE might not be set on the settings object
         site = getattr(settings, 'SITE', None)
 
+    # We can't figure out the current site. Give up.
     if site is None:
         return None
 
+    # Return parent if there is one. This site, otherwise.
     parent_site = getattr(site, 'parent_site', None)
     if parent_site is None:
         return site
--- a/seo/cache.py
+++ b/seo/cache.py
@@ -160,6 +160,6 @@ def get_domain_parent(request):
 
     parent_site = getattr(site, 'parent_site', None)
     if parent_site is None:
-        return site.domain
+        return site
     else:
         return parent_site
--- a/seo/cache.py
+++ b/seo/cache.py
@@ -155,4 +155,11 @@ def get_domain_parent(request):
         # SITE might not be set on the settings object
         site = getattr(settings, 'SITE', None)
 
-    return getattr(site, 'parent_site', None)
+    if site is None:
+        return None
+
+    parent_site = getattr(site, 'parent_site', None)
+    if parent_site is None:
+        return site.domain
+    else:
+        return parent_site
--- a/static/dashboard.100-10.js
+++ b/static/dashboard.100-10.js
@@ -1,0 +1,50 @@
+
+function read_cookie(cookie) {
+    var nameEQ = cookie + "=",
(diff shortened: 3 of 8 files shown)


seo/cache.py
```diff
@@ -155,4 +155,11 @@ def get_domain_parent(request):
         # SITE might not be set on the settings object
         site = getattr(settings, 'SITE', None)
 
-    return getattr(site, 'parent_site', None)
+    if site is None:
+        return None
+
+    parent_site = getattr(site, 'parent_site', None)
+    if parent_site is None:
+        return site.domain
```
tdphillips: The docstring for this method is no longer true - it returns an SeoSite object, a string, or None.

darrint: Good catch. Would you agree that it's just String or None?

tdphillips: Line 165 returns an SeoSite object.

darrint:
I'm marking this do not merge until we have the IE8 stuff figured out.

static/dashboard.100-10.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+function dashboard(dashboard_url) {
+  // need to explicitly marshal the result data into this deferred
+  // because we are on old jQuery.
+  var result = $.Deferred();
+  $.ajax({
+    type: "OPTIONS",
+    crossDomain: true,
+  }).then(function(data, textStatus, xhr) {
```
aspidites: If I'm not mistaken, `then` is an alias for `pipe`, which is essentially an asyncronous `$.map`. It doesn't look like you're iterating over the result set, so I think `done` is more appropriate here. If we are using jquery < 1.8, `then` indeed aliases `done`, otherwise what I just said should hold. Even then, it seems wiser to be future proof, lest we upgrade jquery and wonder why things are randomly behaving differently down the road.

aspidites:
The version number for the new static file is wrong. It should be (today) 168-30, or (when you created this PR) 168-25.

I obtained the  version number by runing `./version -b` in the root of the MyJobs project.

static/dashboard.100-10.js
darrint: > Deprecation Notice:As of jQuery 1.8, the deferred.pipe() method is deprecated. The deferred.then() method, which replaces it, should be used instead.

I want both success and failure here. Done only receives success.

aspidites: I'm objecting to the use of `then` all together unless you're needing some kind of filtering behavior. As for catching both success and failure, there is`always`.

myjobs/child_dashboard.py
```diff
@@ -1,9 +1,12 @@
-from django.http import HttpResponse
+from django.views.decorators.csrf import (
+    csrf_exempt as django_csrf_exempt)
 
 from myjobs.autoserialize import autoserialize
 from myjobs.cross_site_verify import cross_site_verify
 
 
+# The django csrf exemption should stay first in this list.
+@django_csrf_exempt
```
aspidites: is there a case where we have cross_site_verify and not django_csrf_exempt or autoserialize? I just remember objections being raised on code I wrote a while ago for stacking too many decorators. I don't personally care so long as each doesn't have duplicate functionality and they are necessarily decoupled.

aspidites:
As for testing on IE8, [modern IE](https://www.google.com/search?q=modern+ie&ie=utf-8&oe=utf-8). Wouldn't feel safe merging this until IE 8 and perhaps 10 were tested.

myjobs/child_dashboard.py
darrint: They aren't coupled or duplicate, at least not yet. If we add a second endpoint and the same decorators are needed then we should definitely combine them.

static/dashboard.100-10.js
darrint: Talked offline. Line 22 will be a done/fail. 23/25 will remain as then -> deferred.

myjobs/child_dashboard.py
aspidites: :+1:

darrint:
IE10 works too.

seo/cache.py
```diff
@@ -155,4 +155,11 @@ def get_domain_parent(request):
         # SITE might not be set on the settings object
         site = getattr(settings, 'SITE', None)
 
-    return getattr(site, 'parent_site', None)
```
aspidites: Given the following:
- {{ domain_parent }}, per [this ticket](https://directemployersfoundation.atlassian.net/browse/PD-1609) was meant to return an SeoSite's parent site's domain
- By definition, per [this ticket](https://directemployersfoundation.atlassian.net/browse/PD-1608) and [this code](https://github.com/DirectEmployers/MyJobs/pull/1692), an SeoSite can not be it's own parent

The code, as you've modified it, is misleading, as if an SeoSite doesn't have a parent site, it returns the site itself, which, per the above, is inconsistent.

If the context processor has been re-purposed, I think it would be better to rename it to remove confusion. 

That said, I think your goal in these edits could be met by using the `|default` filter on top of the old context processor. In doing so, `domain_parent` remains true to its name and you get your desired output of having either the domain parent or the site itself.

darrint: I've renamed this function and it's downstream variables with new names to clarify its purpose.

PR #1786, opened 2015-09-29
[PD-1808][PD-1812] Added posting_access to admin and updated how we redirect to login

aspidites:
In order to set posting_access on a company without first needing to get write access to the database, I need to enable it in admin. This will also prevent member services from needing developer intervention in order to set up postajob access in the future. 

Update; Added a regression test for login redirects and made all postajob tests use seo settings.

### DIFF
--- a/postajob/tests/test_views.py
+++ b/postajob/tests/test_views.py
@@ -7,7 +7,7 @@ from django.conf import settings
 from django.core import mail
 from django.core.urlresolvers import reverse
 
-from myjobs.tests.setup import MyJobsBase
+from seo.tests.setup import DirectSEOBase
 from mydashboard.tests.factories import (BusinessUnitFactory, CompanyFactory,
                                          CompanyUserFactory, SeoSiteFactory)
 from myjobs.tests.factories import UserFactory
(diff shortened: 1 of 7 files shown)


universal/decorators.py
```diff
@@ -26,7 +26,7 @@ def wrap(request, *args, **kwargs):
             # with this url as the next url.
             if request.user.is_anonymous():
                 params = {'next': request.get_full_path()}
-                next_url = build_url(reverse('home'), params)
+                next_url = build_url(reverse('login'), params)
```
JLMcLaugh: Can we get a regression test for PD-1812?

aspidites:
Regression test written. I haven't updated all the other tests, as I wasn't able to run these tests locally, so I'm waiting on Jenkins.

universal/decorators.py
aspidites: Yeah. For the record, you just opened a can of worms. :-P

aspidites:
Readded the url to urls.py because it was causing 301's where code was still referring to home when it probably should have been using login.

postajob/tests/test_views.py
```diff
@@ -23,41 +24,48 @@
...
 
 
-class PostajobTestBase(MyJobsBase):
+class PostajobTestBase(DirectSEOBase):
     def setUp(self):
         super(PostajobTestBase, self).setUp()
         self.user = UserFactory(password='5UuYquA@')
         self.company = CompanyFactory(product_access=True, posting_access=True)
 
+        #configuration = ConfigurationFactory(
```
tdphillips: Is this coming back at some point or is it gone for good?

aspidites: Would have come back if I found a way to make the configuration matter, which I didn't. I'll go ahead and delete it.

PR #1788, opened 2015-09-29
[PD-1782] Fix None Returns for Company Titles

koonsies:
## The Issue
Companies listed in jobs may or may not exist in our database. When the company header function is called with a company slug that does not exist in our DB, try to pull the title from the job. If that fails, return the company slug as a last resort

### Updates
1. Added handling to attempt to match company slug to a given job and to default to provided company slug if job is unavailable.
2. Created new tests for function to retrieve company title
3. Created new test base class (pulled from test_views.FallbackTestCase) that contains a configured site and a job.

### DIFF
--- a/seo/helpers.py
+++ b/seo/helpers.py
@@ -388,13 +388,10 @@ def bread_box_company_heading(company_slug_value, jobs=None):
     if not company_slug_value:
         return None
     kwargs = {'title_slug': company_slug_value}
-    business_unit = BusinessUnit.objects.filter(**kwargs)
+    business_unit = BusinessUnit.objects.filter(**kwargs).first()
 
-    try:
-        return business_unit[0].title
-    except (IndexError, TypeError):
-        # No business unit found
-        pass
+    if business_unit:
+        return business_unit.title
 
     try:
         return jobs[0].company
@@ -404,7 +401,7 @@ def bread_box_company_heading(company_slug_value, jobs=None):
 
     # this is unlikely to happen as companies that do not exist in DB
     # and do not have associated jobs are often 404'd.
-    return company_slug_value
+    return company_slug_value.replace('-', ' ').title()
 
 
 def location_from_job(job, num_locations):
--- a/seo/helpers.py
+++ b/seo/helpers.py
@@ -541,7 +541,7 @@ def get_bread_box_headings(filters=None, jobs=None):
             bread_box_headings['moc_slug'] = moc
 
         company_slug_value = filters.get("company_slug")
-        company = bread_box_company_heading(company_slug_value)
+        company = bread_box_company_heading(company_slug_value, jobs)
         if company:
             bread_box_headings['company_slug'] = company
 
--- a/seo/helpers.py
+++ b/seo/helpers.py
@@ -383,7 +383,7 @@ def bread_box_company_heading(company_slug_value, jobs=None):
     otherwise, return company string from job or company slug itself
     :param company_slug_value: company filter provided
     :param jobs: jobs matching the provided company
-    :return:
+    :return: business unit title or company slug value parameter
     """
     if not company_slug_value:
         return None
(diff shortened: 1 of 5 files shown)


seo/tests/setup.py
```diff
@@ -122,6 +124,38 @@ def tearDown(self):
...
+        Attributes:
+            job - job from solr_settings.SOLR_FIXTURE
+            buid -  business unit id from provided job
+            site - site attached to business unit id from job
+            config - configuration of selected site
+    """
+    def setUp(self):
+        super(DirectSeoTCWithJobAndSite, self).setUp()
+
+        self.job = self.solr_docs[1]
```
JLMcLaugh: I like having this class. There's too much copy-and-paste setting up sites in our tests.

But, DirectSeoTestCase already adds jobs to solr and creates a business unit. The SeoSiteFactory and ConfigurationFactory should be all you need to get a site tied to that business unit with a config that makes the jobs browsable

koonsies: I agree. I modified this code to better fit that idea. Let me know what you think.

seo/helpers.py
```diff
@@ -377,18 +377,34 @@ def _page_title(crumbs):
...
+    otherwise, return company string from job or company slug itself
+    :param company_slug_value: company filter provided
+    :param jobs: jobs matching the provided company
+    :return: business unit title or company slug value parameter
+    """
     if not company_slug_value:
         return None
-
     kwargs = {'title_slug': company_slug_value}
     business_unit = BusinessUnit.objects.filter(**kwargs)
```
tdphillips: If `.first()` were to be tacked onto the end of this, one of the try/excepts below could be a simple `if business_unit is None` check.

seo/tests/test_views.py
```diff
@@ -2802,22 +2770,23 @@ def test_invalid_company_behavior(self):
...
-        company_from_job = Company.objects.filter(name__iexact=self.job['company'])
+        job = self.solr_docs[1]
+        company_from_job = Company.objects.filter(name__iexact=job['company'])
         if company_from_job:
             slug_value = company_from_job[0].company_slug
-            company_from_job.delete() #make sure company doesn't exist in DB
+            company_from_job.delete()  # make sure company doesn't exist in DB
         else:
-            slug_value = self.job['company'].lower()
+            slug_value = job['company'].lower()
```
JLMcLaugh: Slugs would also replace spaces with dashes, but I can't think of a reason this if/else should be here in a test anyways.

seo/helpers.py
```diff
@@ -377,18 +377,34 @@ def _page_title(crumbs):
...
+
+    try:
+        return jobs[0].company
+    except (IndexError, TypeError):
+        # No job provided
+        pass
+
+    # this is unlikely to happen as companies that do not exist in DB
+    # and do not have associated jobs are often 404'd.
+    return company_slug_value
```
JLMcLaugh: We should deslugify this. Removing hypens and title casing should look better in most cases.

koonsies: Is there a deslugify method? So far, I only see "slugify" in the slugify app.

JLMcLaugh: No, and it won't be perfect. Title case and replace hyphens with spaces. Like you said, there really shouldn't be a reason for us to get here anyways.

aspidites: Should it? I mean, if we decide to deslugify in the future, it might be nice to have the function just to be certain that we always deslugify the same way. Plus, if we change how that should happen for some reason, changing one function will be easier than changing X call sites.

aspidites:
Should deslugify exist? I mean, if we decide to deslugify in the future, it might be nice to have the function just to be certain that we always deslugify the same way. Plus, if we change how that should happen for some reason, changing one function will be easier than changing X call sites.

PR #1791, opened 2015-09-30
[PD-1624]  call secure blocks

darrint:
This branch is purposefully not based on quality-control at the moment.

This is my current understanding of how we will load blocks.

It currently lacks the ability to pass parameters to blocks and probably other stuff too.

This diff reads best in reverse. Start at the template to see the intended interface. Then the machinery works backwards through the javascript, into child_dashboard, then finally into the myblocks system.

In my testing I have element_id of a search block set to the id in the template.


static/dashboard.100-10.js
```diff
@@ -20,31 +20,49 @@ function dashboard(dashboard_url) {
...
     url: dashboard_url,
     type: "GET",
     dataType: "jsonp",
+    data: JSON.stringify({ blocks: json }),
   });
 }
 
+function load_secure_blocks(dashboard_url) {
+  var request = {};
+  $("*[secure-block]").each(function(i, block) {
```
fezick: This seems overly complex for someone to implement on the front end, as they would have to include two properties on each widget placeholder. Instead, we could use $("div[id^='dewidget']").each(...) or something similar. This would make it so only one property was needed, and leave all additional properties as optional configuration settings.

fezick:
With the exception of my one comment, this looks correct from a high level perspective.

static/dashboard.100-10.js
darrint: Why is this important? No one outside this department will be typing these in. In myblocks proper this secure placeholder vs. rendered content will handled transparently to the user composing pages from blocks.

I would prefer to tweak this based on selector performance, etc. (Which I have not clue about...)

PR #1792, opened 2015-09-30
[PD-1769] Added last_modified where appropriate.

aspidites:
See conversation on Jira for details, but the TL;DR is that I added last_modified to the following:
- SavedSearch
- Partner
- Contact
- ContactRecord


tdphillips:
Depending on if we want the new field to be tracked in Solr (I doubt it as no one uses dashboard), we'll need to either update the schema or add last_modified to solr/signals.py somewhere. I'll play with it a bit once my own tests finish to see if I can get a more definitive location.

koonsies:
Are these tables going to be the ones that will be impacted by PD-1770? And are you intending to grab that ticket as well, Edwin?

mysearches/models.py
```diff
@@ -50,6 +50,7 @@ class SavedSearch(models.Model):
                              on_delete=models.DO_NOTHING)
 
     created_on = models.DateTimeField(auto_now_add=True)
+    last_modified = models.DateTimeField(auto_now=True, default=datetime.now)
```
koonsies: Is adding default=datetime.now redundant with auto_now=True? AFAIK, auto_now makes the field uneditable and defaults to datetime.now (essentially) on any save of any data.

aspidites:
@koonsies, I assume that these tables will impact PD-1770, but no, I have zero intention of grabbing that ticket myself.

aspidites:
@tdphillips thanks, keep me posted. I've not done either of those, so would need some guidance.

mysearches/models.py
aspidites: South complains without it. Upgrading to 1.8 would allow (and in fact warn) that default is redundant.

haliscooper:
Ummm, making these autonow recreates the same issue we had in the past.  LastModified should be set by the code when a record is changed, only when done through the system by the user.

tdphillips:
Maybe it should have been communicated earlier that not all modifications are modifications.

haliscooper:
Perhaps so, Troy, although it seemed inherent to the situation that a modification we make to correct something isn't a modification a user would expect to see presented in the interface, as mentioned in the planning meeting.
The discussion in the planning meeting on adding a more "user storyish" element to tickets will help, but at the same time, hoping people will think of the purpose behind the ticket and ask if it's not obvious to them.  So, in this case, we had a public facing created_on field that was being updated each time a row was updated, which fed misinformation (in the eyes of the user and in reality) to the user.  We create tickets to reset those back to the original date, and add a last modified column that would later become visible to the user, which again, the user would only expect to reflect when THEY last updated it.  The only reason to flatten out modified dates per table instead of joining to the log table is the processing time and bottleneck of accessing the the log table looking for the most recent item for a number of different types, when it's already taking an insert for every change/entry to the system. Since we're not updating dates in the log table when we go into the backend to modify data, I'd assume we'd combine that concept with the fact that autoupdating the date on previous backend data modifications has caused false info displayed to the user (in their eyes) and avoid it.  As well, adding tags, locations, etc, which are modifications to other tables rather than the core table (contacts, in this case) won't get marked as a modification by autonow when in the eyes of the user, that is a modification they should see represented.

aspidites:
@haliscooper I wasn't at that planning meeting, so assuming that I agree  with your assessment of what a modification is, I wouldn't have reached the same conclusion from a ticket description of " Discuss with various people regarding where " on a ticket titled "my.jobs - database - add last_modified column where appropriate".

haliscooper:
It's OK to reach a different conclusion.  However, when I point out the issues that conclusion raises, I don't expect the reaction I got.  
That said, I do expect you to put yourself in a few different roles every time you do anything, the user, admins, the system itself (front end and back).  If you went into a system and the last modified was set to a date more recent than when you last logged in (and there was no scheduled activity that would have changed it), you'd be surprised.  We just went through that with created_on.  Not only that, from a system end of things, the last modified wouldn't match what's recorded in contact_log_entry, and again that would confuse a user if presented in reporting.  True, some data is system oriented, and then user concerns don't factor in, but if you're uncertain as to the purpose of the last_modified, ask, the Jasons and I discussed it in the backlog meeting. Jason M chose the flattened version rather than joining to contact_log_entry, because it made sense technically, not as a potentially different value than what is in the log.  Even then, one of us might give you a different answer, that's the price of having more than one person involved in anything. And if something changes after a change is made, that's life, in every occupation, we try to mimimize it, but it's unavoidable.  If we knew everything up front, we'd be doing waterfall, but then real world changes would make it far more painful than removing an autonow.

PR #1793, opened 2015-09-30
Add docs and improvements to how things are archived [PD-1793]

tdphillips:
This moves the logic in the previously overridden delete methods to a new archive method and modifies the delete view to instead call this new method.

We now filter out contacts/communication records as appropriate when any parent/grandparent contacts/partners have been archived.

All tests pass.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -11,7 +11,7 @@ from django.core.files.storage import default_storage
 from django.core.urlresolvers import reverse
 from django.core.exceptions import ValidationError
 from django.db import models
-from django.db.models.signals import pre_delete, pre_save, post_save
+from django.db.models.signals import pre_delete, pre_save
 from django.dispatch import receiver
 
 from postajob.location_data import states
@@ -124,7 +124,10 @@ class SearchParameterManager(models.Manager):
     def __init__(self, archived=False):
         """
         Adds a new argument to the manager constructor to allow us to specify
-        if a given query should include archived models
+        if a given query should include archived models.
+
+        If this isn't doing what you want, examine the model using it and pay
+        special attention to what value of "archived" is being passed in.
 
         Inputs:
         :archived: True: query archived only, False: exclude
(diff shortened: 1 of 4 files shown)


mypartners/models.py
```diff
@@ -156,13 +159,21 @@ def sort_by(self, *fields):
 class ArchivedModel(models.Model):
     archived_on = models.DateTimeField(null=True)
 
+    # These two managers do different things and may not exactly do what you
```
tdphillips: If you run into weirdness, would this description (and the related test) be sufficient as explanation?

koonsies: It explains the issue well enough to me. Being that we're replacing the base object manager, I'm sure there will be some confusion somewhere along the line.. but this should do a lot to clear it up.

mypartners/models.py
```diff
@@ -242,7 +253,10 @@ def __unicode__(self):
     def delete(self, *args, **kwargs):
         pre_delete.send(sender=Contact, instance=self, using='default')
         super(Contact, self).delete(*args, **kwargs)
-        self.primary_contact.clear()
+
+    def archive(self, *args, **kwargs):
+        #pre_delete.send(sender=Contact, instance=self, using='default')
+        super(Contact, self).archive(*args, **kwargs)
```
koonsies: Are we intending to uncomment the pre_delete signal at some point? Otherwise, I don't see the point of this block.

mypartners/tests/test_models.py
```diff
@@ -163,11 +163,20 @@ def test_contact_archived(self):
...
         self.assertFalse(self.contact.archived_on)
-        self.contact.delete()
+        self.contact.archive()
         self.assertEqual(Contact.objects.count(), 0)
         self.assertEqual(Contact.all_objects.count(), 1)
         self.assertTrue(self.contact.archived_on)
 
+    def test_archived_manager_weirdness(self):
+        """
+        Demonstrates a gotcha with our custom model managers.
```
koonsies: Perhaps make this a bit more specific. I liked the comment on the main logic block, but this might be better served as "Demonstrates that archived models are not returned by a standard ArchivedModel.objects call".. Or something like that.

mypartners/models.py
tdphillips: That was me trying to debug why Contacts were being deleted when Partners were. This is still a work in progress but I think this line should work just fine if uncommented.

mypartners/tests/test_models.py
tdphillips: Good call. I'm also improving that test in general.

fezick:
Is this pull or PD-1797 or PD-1793? I think PD-1793, but I want to verify before I pass 1797.

tdphillips:
1793. I'll go back and re-tag the mistagged commits.

PR #1794, opened 2015-10-01
[PD-1815] Fixed the HTML view for Contact Record Reports

aspidites:
Before:
2015-10-01-140511_1052x1056_scrot

After:
2015-10-01-140535_1052x1056_scrot

Note that communication records always refers to records which aren't of type "job".

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -546,7 +546,7 @@ class ContactRecordQuerySet(SearchParameterQuerySet):
         to the resulting objects.
         """
 
-        contacts = self.values(
+        all_contacts = self.values(
             'partner__name', 'partner', 'contact__name',
             'contact_email').distinct()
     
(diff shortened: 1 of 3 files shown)


mypartners/models.py
```diff
@@ -538,12 +538,21 @@ def referrals(self):
...
-                        '-records')
+        """ 
+        Returns a queryset annotated by the number of referrals (contact_type =
+        'job') and number of records (contact_type != 'job'). Django doesn't
+        allow annotated values to be filtered without also filtering the base
+        result set, which is why we annotate separately then attach the values
+        to the resulting objects.
+        """
+
+        contacts = self.values(
```
tdphillips: Nothing complains about this sharing the name of the method?

aspidites: contacts is a local variable, where as `def contacts` is an instance variable. Shadowing would only happen if I used "self.contacts". I'll rename to reduce confusion though.

PR #1795, opened 2015-10-02
[PD-1624] Populate pages with blocks securely on demand

darrint:
With this PR the block replacement is functional.

```diff
diff --git b/templates/job_listing.html a/templates/job_listing.html
index d86bb6f..179f812 100644
--- b/templates/job_listing.html
+++ a/templates/job_listing.html
@@ -59,16 +59,12 @@
 
 {% if default_jobs or featured_jobs %}
     {% if "network" in site_tags or site_config.show_saved_search_widget %}
-        <script src="//d2e48ltfsb5exy.cloudfront.net/myjobs/tools/def.myjobs.widget.153-05.js"></script>
         <div id="direct-saved-search-box" class="direct_rightColBox direct-action-box">
-            <div id="de-myjobs-widget"></div>
+            <div data-secure-block-id="saved-search-widget-24"></div>
         </div>
     {% endif %}
     <a name="filters"></a>  
-    <div id="direct_shareDiv" class="direct_rightColBox direct_shareSolo" role="menu">
-        <h3>{% trans "Share" %}</h3>
-        {% include 'includes/add_this.html' %} <!--replace with my.jobs sharing when ready. JPS 10-4-12-->
-    </div>
+    <div data-secure-block-id="share-block-1"></div>
     <a href="#top" class="direct_mobileJumpLink">{% blocktrans %}Return to top{% endblocktrans %}</a>
 
 {% endif %}
```

I'm still __unsure about adding the element_id field__. Is it really necessary or should we just put the block records's primary key id in data-secure-block-id?

To test this use the admin add/edit blocks and set element id's. I've tested only with the share block (key 1 in my database) and the saved search widget which I had to add to the database.

The flow of this mechanism works like this:

* Template contains an element with data-secure-block-id="$element_id" (see patch above)
* Document onready fires and calls secure block with parent_domain as a parameter. (seo_base.html)
* Secure block js looks for secure blocks, makes a request. (secure-blocks.nnn-nn.js)
* Secure block view looks up the ids given and calls blocks to render content with given parameters. (myblocks/views.py)
* Blocks figure out their context from the parameters. (myblocks/models.py)
* Secure block js gets response from view and replaces DOM content. (secure-blocks.nnn-nn.py


myblocks/views.py
```diff
@@ -65,4 +75,34 @@ def set_page(self, request):
...
+@django_csrf_exempt
+@cross_site_verify
+@autoserialize
+def secure_blocks(request):
+    try:
+        if request.method == 'POST':
+            blocks = json.loads(request.body)[u'blocks']
+        else:
+            blocks = json.loads(request.GET['blocks'])
+    except:
```
tdphillips: Should we be catching anything here?

aspidites:
> I'm still unsure about adding the element_id field. Is it really necessary or should we just put the block records's primary key id in data-secure-block-id?
I agree. The same strategy works just fine in reporting (inspect the view, clone, regenerate, or download buttons).

aspidites:
> Document onready fires and calls secure block with parent_domain as a parameter. (seo_base.html)

This should be `secure_blocks_domain`. I assume the description hasn't  changed since the discussions on the other PR.

myblocks/models.py
```diff
@@ -119,6 +124,14 @@ def get_template(self):
                                                       self.bootstrap_classes(),
                                                       self.template)
 
+    def render_for_ajax(raw_self, request, params):
+        self = raw_self.cast()
```
aspidites: Huh?

myblocks/models.py
```diff
@@ -362,10 +375,36 @@ def submit_btn_name(self):
...
 
+    def context(self, request, **kwargs):
+        saved_search_url = kwargs.get('url')
+        success_email = kwargs.get('success_email')
+        search = None
+        user = request.user if request.user.is_authenticated() else None
+
+        if user:
+            try:
+                search = SavedSearch.objects.filter(user=user,
```
aspidites: `.first()` would accomplish the same without the try/except/pass.

myblocks/tests/test_secure_blocks.py
```diff
@@ -0,0 +1,23 @@
...
+
+from django.core.urlresolvers import reverse
+
+from seo.tests.setup import DirectSEOBase
+
+from seo.tests.factories import SeoSiteFactory
+
+
+class TestChildDashboard(DirectSEOBase):
+    def test_secure_blocks(self):
```
aspidites: Do we have a test for when the content isn't just "{}"?

myblocks/views.py
```diff
@@ -1,9 +1,19 @@
+import json
+import logging
+
 from django.conf import settings
 from django.http import Http404, HttpResponse
 
 from django.views.generic import View
+from django.views.decorators.csrf import (
+    csrf_exempt as django_csrf_exempt)
```
aspidites: Why the alias? I don't think it clashes with anything else we use.

darrint: I originally left it as is but it confused me. The method is not exempt from csrf, it is exempt from Django's CSRF handler.

myblocks/models.py
darrint: See the original cast method. The model subclassing doesn't work like real python subclassing.

PR #1796, opened 2015-10-02
[PD-1769] Create date_modified columns on various tables

koonsies:
Slight modification of @aspidites original work on #1792 

Removed auto_now from each column and regenerated migrations.

Modified a few forms to exclude the field (otherwise, tests were failing)


darrint:
+1

PR #1799, opened 2015-10-05
[PD-1763] Added models to track outreach results.

aspidites:
Realized that the one big difference between Partner.library and these models is that multiple partners, contacts, and communication records can be made from a single non-user outreach email.


mypartners/models.py
```diff
@@ -900,7 +900,23 @@ class OutreachWorkflowState(models.Model):
...
-    from_email = models.EmailField(max_length=255, verbose_name="Email", 
-                                   help_text="Email outreach effort sent from.")
+    from_email = models.EmailField(
+        max_length=255, verbose_name="Email",
+        help_text="Email outreach effort sent from.")
     email_body = models.TextField()
     current_workflow_state = models.ForeignKey("OutreachWorkflowState")
+
+
+class OutreachPartner(models.Model):
```
koonsies: I'm a bit confused. What makes this better than just having a standard many to many relationship between Outreach and Partner?

haliscooper: Came out of a discussion that I remember Jason M saying something and me going, "Oh yeah, those should be separate table" but can't remember why.  Could very well be a column on partner, contacts and contact record that is null 99% of the time, inefficient to index it.  Jason is searching his memory...

aspidites: If I understand correctly, this was addressed and I'm keeping this as-is? @JLMcLaugh?

JLMcLaugh: There are some comments on the ticket now. @koonsies is right though. We don't need separate models if we're just trying to get a many-to-many relationship.

aspidites:
Now that the dust has settled, I updated the migration to simply add 3 many to many's on the outreach model (which subsequently make the through tables that the ticket was asking for).

darrint:
+1

mypartners/models.py
```diff
@@ -896,11 +896,22 @@ def __unicode__(self):
...
+    def __unicode__(self):
+        return self.state
+
 
 class NonUserOutreach(models.Model):
     date_added = models.DateTimeField(auto_now_add=True)
     outreach_email = models.ForeignKey("OutreachEmailAddress")
-    from_email = models.EmailField(max_length=255, verbose_name="Email", 
-                                   help_text="Email outreach effort sent from.")
+    from_email = models.EmailField(
```
tdphillips: Do we also need to track whatever's in "to" or "cc"?

JLMcLaugh: +1, I don't know how we'd be able to tie it to a contact or suggest a partner otherwise.

aspidites: outreach_email IS the to email address (not that it's a foreign key to an outreach email address). CC is intentionally being ignored, per conversation by @haliscooper

tdphillips:
There's now a duplicate mypartners migration.

aspidites:
@tdphillips, fixed.

darrint:
:+1:

PR #1801, opened 2015-10-06
Ensure QueryRedirect doesn't trigger if there are extra query parameters (QC) [PD-1823]

tdphillips:


middleware.py
```diff
@@ -256,16 +257,20 @@ class RedirectOverrideMiddleware(object):
     strings, and any entries in the QueryRedirect table for the current domain.
     """
     def process_request(self, request):
+        if not settings.SITE.queryredirect_set.exists():
```
JLMcLaugh: Does anyone know if this triggers another database query on prefetched objects? .all() would be guaranteed not to.

middleware.py
```diff
@@ -256,16 +257,20 @@ class RedirectOverrideMiddleware(object):
     strings, and any entries in the QueryRedirect table for the current domain.
     """
     def process_request(self, request):
+        if not settings.SITE.queryredirect_set.all():
```
JLMcLaugh: For those curious, it looks like exists would not have triggered another query. Either option is fine.

PR #1804, opened 2015-10-06
Re-applied Search Transformer Asst. Bugs patch.

aspidites:
Taking a quick look at the PRs since last week, this seems to be the only thing different.


aspidites:
Should be able to remerge staging into master after this.

seo/helpers.py
```diff
@@ -1443,10 +1442,6 @@ def jobs_and_counts(request, filters, num_jobs, fl=search_fields):
 
 
 def get_company_data(filters):
-    """
-        Return the thumbnail for a company if it exists. Returns None if company does
-        not have a thumbnail.
-    """
```
koonsies: This comment does not need to be removed. It was added in the company hotfix to give a better description to get_company_data.. Though I should create a PR to rename that function anyway.

aspidites: I didn't remove it. Some commit that I used to get this from did.

koonsies:
Ha.. Ha.. I think I know why one part of this is behaving erratically. I did not hotfix to QC for the company hotfix that went into master and staging. Many of these modifications are from that hotfix (most of them, actually). I will look into putting up that hotfix for QC and see what comes from it. For now, I'm adding a Do Not Merge flag.

PR #1805, opened 2015-10-06
[PD-1555] First cut at dynamic reports api and throwaway ui.

darrint:
This provides a basic UI showing some API interaction from a real-ish page.

~~I don't yet have a strategy for mapping users to usertypes.~~ There are two usertypes and there is a single function and tests for making the determination.

To test this, load the provided fixture:

1. python manage.py loaddata myreports class_and_pres.json
2. Then navigate to https://qc.secure.my.jobs/reports/view/dynamicoverview
3. There should be buttons. Click them to see the flow.
4. Also watch network activity in the browser's debugger to see the nature of the api as it stands.

Edit: Now with working user types.


myreports/views.py
```diff
@@ -344,3 +345,56 @@ def download_report(request):
     response.write(serialize('csv', records, values=values, order_by=order_by))
 
     return response
+
+
+@has_access('prm')
```
koonsies: This currently only allows this module to be accessed by users belonging to a member company, not superusers or users who are otherwise staff. Is this intentional?

darrint: I think it is supposed to be that way.

haliscooper: It used to be, but now other entities will be in there (state personnel, other types of employers, etc).  So, prm access won't be the determining factor as to whether one can access reporting, anyone will be able to access it, what's available to them will be determine by their user_type, app access, and permissions.

haliscooper: That's why it's it's own app, instead of being part of PRM, reporting from all other apps  (and reporting currently on us.jobs) will come through this one interface.

aspidites:
1. Merge conflicts
2. If this doesn't account for all UserTypes, then it wouldn't seem that this really addresses PD-1555. If it isn't possible to account for those, then it would seem that whatever ticket does account for those user types should block this one. If we don't want tickets to block this one, it would seem the scope of this ticket should be narrowed and possibly redefined. While we @fezick, @haliscooper, and I have had conversations which narrowed the scope from what it is now (exposing the entire ERD structure to the API), that doesn't seem to have been an update that is reflected in the current ticket.

In regards to 2), the reason I'm being a stickler is that in tagging it at PD-1555, if it doesn't actually account for user types, that ticket will be marked as merged and qc will happen on it, even though we can't technically account for different user types.

If it's the case that user types are accounted for, but only one set of users (member company users) exists or has reports available for them, please note that somewhere.

darrint:
There are only two known usertypes, even in the ERD doc on the epic: PD-1380. Both are accounted for. I'm ok with holding up the ticket if we're unsure, but we should still merge this.

PR #1807, opened 2015-10-07
Create default login pages on request [PD-1666]

tdphillips:
I see three options.

- Do this
- Do the bits I commented out (grab a known generic login page)
- Instead of automagically converting and loading the fixture, parse said fixture and manually construct the component objects

I've done no testing yet - I'm curious if most of `create_login_page` could be one transaction.


registration/helpers.py
```diff
@@ -0,0 +1,94 @@
...
+                'myblocks.columnblock']['latest']
+        if 'page' in thing['fields']:
+            thing['fields']['page'] += model_map['myblocks.page']['latest']
+
+    # Re-dump this updated fixture as json.
+    contents = json.dumps(contents)
+    with NamedTemporaryFile(suffix='.json') as fixture:
+        # Write this new text as a temp file and flush its buffer. In testing,
+        # we would sometimes get truncated files without the flush().
+        fixture.write(contents)
```
darrint: Shouldn't we be calling git add somewhere around here?

tdphillips: I don't think so. NamedTemporaryFile defaults to creating the file in /tmp and the file is deleted when the context manager exits.

darrint: We're still going to have to figure out how to get these through a qa/staging process. Maybe we can itegrate this with JIRA.

tdphillips: We don't run any other object creation through the qa/staging process.

darrint: It can be automated.

PR #1808, opened 2015-10-07
[PD 1624] Part 2: Populate pages with blocks securely on demand 

darrint:
https://github.com/DirectEmployers/MyJobs/pull/1795

Some comments weren't addressed yet when the PR above was merged. This PR addresses those comments.


koonsies:
:+1:

myblocks/models.py
```diff
@@ -382,18 +382,16 @@ def context(self, request, **kwargs):
         user = request.user if request.user.is_authenticated() else None
 
         if user:
-            try:
-                search = SavedSearch.objects.filter(user=user,
-                                                    url=saved_search_url)[0]
-            except IndexError:
-                pass
+            search = (SavedSearch.objects
```
aspidites: Surrounding parentheses are redundant here:

```python
search = SavedSearch.objects.filter(
    user=user,
    url=saved_search_url).first()
```

myblocks/tests/test_secure_blocks.py
```diff
@@ -7,17 +7,34 @@
 from seo.tests.factories import SeoSiteFactory
 
 
-class TestChildDashboard(DirectSEOBase):
-    def test_secure_blocks(self):
-        url = reverse('secure_blocks')
+class TestSecureBlocks(DirectSEOBase):
```
aspidites: Missing documentation for tests.

myblocks/views.py
```diff
@@ -88,12 +89,11 @@ def secure_blocks(request):
...
-        logger.warn('secure block parse error: %r %r',
-                    request.body,
-                    request.GET,
-                    exc_info=True)
-        return {"error": "failed to parse request"}
+    except ValueError:
+        message = ('secure block parse error: %r %r' %
+                   (request.body,
+                    request.GET))
+        raise SuspiciousOperation(message)
```
aspidites: Do we want the user to know that their request was denied, or should it be more of a jedi trick (This is not the page you were looking for). I ask because there are examples throughout the code base were we could technically raise 403's, but we opt instead for 404s, citing information leakage as the reason. (Not that you were ever part of any of those conversations, @darrint).

@JLMcLaugh, thoughts?

darrint: It's a fair point. I'm using SuspiciousOperation since that's the easiest way to get Django to give a 400. I was returning 200 before. @JLMcLaugh @fezick @haliscooper

myblocks/models.py
darrint: It's for whitespace. If I leave them out it's hard to keep the lines short enough for flake8; or I have to use backslashes.

aspidites: My snippet only goes to column 48 (class > method > if > my snippet), so it's still short enough. That's why I didn't make it a one liner.

myblocks/views.py
JLMcLaugh: 400 seems like the right code. I can't think of a case where it's a security issue unless we were trying to use secret fields as some kind of half-baked security measure.

PR #1810, opened 2015-10-08
Rename function for retrieving company thumbnail

koonsies:
Simply rename get_company_data to get_company_thumbnail as the former name is misleading.


darrint:
:+1:

aspidites:
You can push the button, @darrint , it's not going to bite you.

PR #1811, opened 2015-10-08
Removed prm_access as it was only really being used as an alias to member status.

aspidites:
A while ago, @JLMcLaugh  decided that `prm_access` was redundant, and that access to PRM was really denoted by whether or not a company was a member company. Fear about how much work would go into having to change views, models, etc in addition to not having ironed out exactly what to do during a data migration resulted in us not deleting the field. At the time, what we did instead was create a post save signal which set prm_access to match member status. 

Unfortunately, I didn't document the change, which caused some confusion today. In particular, @koonsies  was confused as to why prm_access, on save, would always return False. The answer was that he was trying to access it directly and for a company who wasn't a member. Rather than go back and document the seemingly random behavior, I decided to see how much work it would be to simply remove prm_access. Turns out it wasn't that hard. 

As for data migrations, it would not be a good idea to promote all users with prm_access to members as at one point it was possible to have one set to false and the other set to true. At the same time, the fact that the alias has been in production seems to indicate that all companies who need PRM access are already enabled as a member. As such, I decided no migration was needed at all. If a company who had PRM access no longer has it (some how), they simply need to be promoted to member. If they aren't supposed to be a member, then they shouldn't have ever had access to PRM anyway.

As for the removed decorator, it was a generalized version of the company_has_access decorator, which I had intended on removing, but never did. I suspect that the new permissions system will deprecate the old behavior, however. That said, I figured I'd do some house cleaning in the meantime.

@JLMcLaugh, sound reasonable? @fezick: Do I even need a ticket for this? Took maybe 5 minutes to do with the help of ack and vim.

### DIFF
--- a/mypartners/views.py
+++ b/mypartners/views.py
@@ -29,7 +29,7 @@ from urllib2 import HTTPError
 from email_parser import build_email_dicts, get_datetime_from_str
 from universal.helpers import (get_company_or_404, get_int_or_none,
                                add_pagination, get_object_or_none)
-from universal.decorators import has_access, warn_when_inactive
+from universal.decorators import company_has_access, warn_when_inactive
 from myjobs.models import User
 from mysearches.models import PartnerSavedSearch
 from mysearches.helpers import get_interval_from_frequency
@@ -52,7 +52,7 @@ from mypartners.helpers import (prm_worthy, add_extra_params,
 
 
 @warn_when_inactive(feature='Partner Relationship Manager is')
-@has_access('prm')
+@company_has_access('member')
 def prm(request):
     """
     Partner Relationship Manager
@@ -89,7 +89,7 @@ def prm(request):
 
 
 @warn_when_inactive(feature='Partner Library is')
-@has_access('prm')
+@company_has_access('member')
 def partner_library(request):
     company = get_company_or_404(request)
 
@@ -119,7 +119,7 @@ def partner_library(request):
                               RequestContext(request))
 
 
-@has_access('prm')
+@company_has_access('member')
 def create_partner_from_library(request):
     """ Creates a partner and contact from a library_id. """
     partner = new_partner_from_library(request)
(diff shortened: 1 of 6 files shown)


koonsies:
This also impacts @darrint as his PR #1808 uses the has_access('prm') decorator

aspidites:
@darrint  can use "has_company_access('member') and be compatible with whether or not this gets merged.

JLMcLaugh:
I think I said the prm_access bit/attribute on the model was unnecessary. The decorator still seems useful. This pull request moves business logic (all members and only members have access to prm) out of a single function and into every decorator and template we check prm access status on. There's no reason the has_access decorator has to return the matching attribute name on the model as its only option. The reason you split it from company_has_access was to make it more flexible. In this case it can return get_company(req).get('member', False) or similar.

aspidites:
@JLMcLaugh I don't follow. I vaguely remember that conversation, but given that upcoming authentication system and the fact that the only thing using the has_access decorator is prm (via prm_access), I don't thing we gain anything by keeping the decorator around. 

Succinctly, I planned to:
- get rid of has_access decorator
- create the activities decorator (which would be more granular)
- get rid of company_has_access decorator

Unless I'm missing something more obvious, these changes don't operationally change anything for the user. The only thing that I'm removing is flexibility which is soon to be deprecated in the near future anyway.

JLMcLaugh:
@aspidites, the company-level app access is either going to be implemented with a class based view or a decorator that's almost exactly like has_access. This is actually a good example of why we should keep the business logic in the has_access method. has_access can be updated to check company-level app permissions for prm or reporting in one spot. There's no reason that has_access or company_has_access can only check bits on the company model.

aspidites:
Closing this. I think what I've done with the new activities system will in fact deprecate the need for has_access (and company_has_access), but won't be obvious until you see tests/code. I haven't disclosed everything in this conversation, but as the code is already written, figure showing it will be more illuminating than trying to argue for their removal here.

PR #1812, opened 2015-10-08
Add doctype and language code to myblocks.Page [PD-1736]

tdphillips:
The PR that added these fields to Configuration doesn't include tests and I'm not sure what value writing tests for this one would have.

Not specifying doctype or language code on a Page results in the one from the underlying Configuration to be used.

All tests pass.

### DIFF
--- a/myblocks/models.py
+++ b/myblocks/models.py
@@ -16,6 +16,7 @@ from django.template.loaders.filesystem import Loader
 from django.utils.safestring import mark_safe
 
 from myblocks import context_tools
+from myblocks.context_tools import get_site_config
 from myblocks.helpers import success_url
 from myjobs.helpers import expire_login
 from myjobs.models import User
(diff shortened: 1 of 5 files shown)


myblocks/models.py
```diff
@@ -619,6 +621,15 @@ class Page(models.Model):
 
     updated = models.DateTimeField(auto_now=True)
 
+    add_blank = lambda choices: (('', 'Inherit from Configuration'),) + choices
+
+    doc_type = models.CharField(max_length=255, blank=False,
```
tdphillips: These are technically blank=True but we're providing our own blank values so we can customize the display text (the lambda a few lines up from here).

myblocks/models.py
```diff
@@ -679,7 +695,8 @@ def get_head(self):
     def get_template(self, request):
         filters = context_tools.get_filters(request)
 
-        context = {
+        context = self.context(request)
+        context.update({
```
tdphillips: This should probably be moved to the `context` method but I wanted to do the minimum to get this working.

universal/accessibility.py
```diff
@@ -0,0 +1,21 @@
+LANGUAGE_CODES_CHOICES = (
```
tdphillips: Both of these tuples technically have accessibility implications, so here they are.

PR #1813, opened 2015-10-08
Attempt to fix authorize.net tests.

aspidites:
Randomized more values to reduce the probability of duplicate transactions.

### DIFF
--- a/postajob/tests/factories.py
+++ b/postajob/tests/factories.py
@@ -1,4 +1,5 @@
 import datetime
+import random
 
 import factory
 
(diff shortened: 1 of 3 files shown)


postajob/tests/factories.py
```diff
@@ -66,7 +67,8 @@ class Meta:
     package = factory.SubFactory(SitePackageFactory)
     owner = factory.SubFactory(CompanyFactory)
     name = 'Test Product'
-    cost = '5'
+    # prevent duplicate transactions by not using the same cost
+    cost = str(random.randint(1, 5000))
```
koonsies: What happens if there is a duplicate cost? There is technically still a chance of a duplicate here.

tdphillips: The comment is a little off but all of these modifications are intended to reduce the chance for duplicate transactions.

In most cases, it probably won't matter as the cvn and card number will be different. In the event that all three are the same, we'll get a test failure due to a duplicate transaction.

aspidites: @tdphillips is right. The difference between the current code in master and this code is that the probability for duplicates is much lower. Before we had a choice of 899 cvns, 1 value, and however many developers were running tests. That number is now 999 cvns, 5000 values, and however many developers. 

The correct fix would be to mock the API and have separate tests for ensuring that the endpoints were still live, but that would have required a much larger set of changes (not the least of which is refactoring a large set of postajob tests).

PR #1817, opened 2015-10-13
Small reporting improvements

galitskyd:
* Added name to the top of the page when viewing a report.
* Clicking on 'Company Name' will send the user back to the reporting overview. Feels similar to users that work with PRM.

screen shot 2015-10-13 at 4 42 44 pm


aspidites:
Still no indication of what kind of report you're clicking on in the sidebar?

aspidites:
Seems fine otherwise. Maybe @lemniscate316 wants to peek at this? Probably won't be able to say one way or another if its good enough to merge, but may be a good example of what some of our JS looks like.

PR #1818, opened 2015-10-14
[PD-1383]  dynamic reporting interface

darrint:
This is a first cut at enough code to make a trivial report generate CSV. It uses some of the models present, not all yet.

Changes so far:

* Switch from table_name to app_name/model_name.
* Sample data for a contact records report.
  * So far only the model_name/app_name of the columns is used.
* A model to represent a report that exists.
* Some views to run a report and download csv, etc.

To test:

1. run manage.py loaddata myreports class_and_pres
2. Go to secure.my.jobs/reports/view/dynamicoverview
3. Enter 3 in configuration id, Run.
4, Click the report link to download csv.

What I expect to do before pushing:

* Use the configuration columns for query values.
* Build a trivial filter interface around configuration columns.

I marked XXX in a few spots in the code where I was unsure about design choices.

### DIFF
--- a/myreports/models.py
+++ b/myreports/models.py
@@ -174,7 +174,8 @@ class PresentationType(models.Model):
 
 
 class Column(models.Model):
-    table_name = models.CharField(max_length=50)
+    app_name = models.CharField(max_length=50, null=True)
+    model_name = models.CharField(max_length=50, null=True)
     column_name = models.CharField(max_length=50)
     is_active = models.BooleanField(default=False)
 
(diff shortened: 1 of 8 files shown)


myreports/models.py
```diff
@@ -209,3 +210,78 @@ class ConfigurationColumn(models.Model):
...
+
+    objects = SearchParameterManager()
+
+    def __init__(self, *args, **kwargs):
+        super(DynamicReport, self).__init__(*args, **kwargs)
+        self._results = '{}'
+        # XXX: Where will this come from?
+        self.filters = {}
+        # XXX: name report better
+        self.name = "Dynamic Report Name"
```
tdphillips: Could this be pulled from the Configuration name or will those names be different?

darrint: Ended up adding the name to the Dynamic Report model itself.

PR #1820, opened 2015-10-15
[PD-1848] New Permissions System

aspidites:
This is mostly up here for visibility so that people can start reviewing  before I get any further. This isn't marking any particular Jira ticket because the tickets are split as such that either I haven't (and can't complete) everything requested in the ticket, or confirming that what is requested in the ticket requires another ticket to be completed first.

A large part of what is done is documented in the [epic](https://directemployersfoundation.atlassian.net/browse/PD-1743), but the TL;DR is essentially this:
- added Activity, AppAccess, and Role models
- added requires decorator
- wrote tests for the above items
- documented above items thoroughly
- added the ability to assign app access within django admin (this is a permanent change)

I'm explicitly not naming their role as that should become apparent by reading the docstrings for each of the new items. If not, I'm considering that a bug which should be addressed before this is merged.  Questions a bout intent should have been answered in one of our various meetings or the individual tickets themselves.


aspidites:
@koonsies, the changes aren't as scary as the diff makes them seem. The most important bits are the new models in myjobs, the updated models in myjobs and seo (user and company respectively), and the new decorator. The only thing that remains after that is migrations to support the models, factories to support the tests, tests to assert behavior of the models and decorator, and admin changes to access app-level access.

myjobs/decorators.py
```diff
@@ -96,3 +97,85 @@ def wrap(request, *args, **kwargs):
...
+            # the required_access we have
+            company_access = company.app_access.values_list( 'name', flat=True)
+            user_activities = request.user.roles.values_list(
+                'activities__name', flat=True)
+
+            # the required_access we need
+            required_access = AppAccess.objects.filter(
+                activity__name__in=activities).values_list(
+                    'name', flat=True)
+
```
darrint: I'm having trouble following this. Activities is a mixed set of app access names and user activities?

If so, I'd prefer to use separate decorator arguments for app access and user activities.

aspidites: Review the models. Each activity is designated an app_access. As such, aa separate decorator argument would be redundant and erroneous. Redundant because each activity knows to which app_access it belongs. Erroneous because you would then be able to say that a view requires app  access for activities that are in no way related- requiring that a company have user management access in order to view a contact record, for instance.

Could you explain which bit was hard to follow. I took extra time to ensure that not only were the models and decorator thoroughly documented, but that the intent of each block of the conditional was as well. If even after all of that the intent is still not clear, I would like to address that immediately.

darrint: Here's an example. I'm adding new views to myreports, In my case existing apps and activities in your new models are sufficient to express what I need. For security I need to decorate those my new views with this decorator. So does that look like this?

```python
@requires(['myreports', 'delete_report_data'], MissingAppAccess, MissingActivityAccess)
def nuke_reports_view(request, ...):
    # Wow!
```

aspidites: Doh! How could I forget a usage example?! I'll add that to the documentation for the decorator. 

As for your example, it would look like this:
```
@requires(['delete_report_data'])
def nuke_reports_view(requiest, ...):
    # Wow
```
Specifying the callbacks is optional and defaults to MissingAppAccess and MissingActivity respectively, so no need to specify. I assumed that myreports is the app_access tied to `delete_report_data` (I suspect activities will be verb phrases, not nows), so it isn't necessary.

aspidites: On a side note, I planned on making the activities list positional arguments, but following positional arguments by variable keyword arguments (**kwargs) is disallowed in Python 2.7.

aspidites:
@darrint , do the [examples](https://github.com/DirectEmployers/MyJobs/blob/activities/myjobs/decorators.py#L148-L181) I've added make the use of this decorator clearer?

universal/helpers.py
```diff
@@ -98,7 +98,7 @@ def get_company(request):
 
     # If settings.SITE is set we're on a microsite, so get the company
     # based on the microsite we're on instead.
-    if settings.SITE.canonical_company:
+    if hasattr(settings, "SITE") and settings.SITE.canonical_company:
```
aspidites: In theory, settings.SITE should always set when MyJobs is properly set up, but in the case of unit tests, using things like RequestFactory cause middleware to not be run (which is what sets settings.SITE). I tried numerous ways of adding SITE to settings within the unit tests (look back a few commits), but none of them worked, as there was a chicken-egg problem. I decided to nip this in the bud.

That is, settings.SITE can still not be assigned, but at least now, we fail gracefully.

darrint:
The examples are perfect. Thank you.

tdphillips:
Lots of comments. :+1:

koonsies:
:+1:  Good work. Great documentation.

PR #1821, opened 2015-10-15
Allow requeuing of all unsent saved searches [PD-1811]

tdphillips:
Awaiting responses on the ticket. In the meantime, the new task can be examined.

All tests pass.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -168,17 +168,23 @@ def requeue_missed_searches():
                            [choice[0] for choice in DOM_CHOICES])
 
     q = Q()
+    # There are 36 distinct last_sent dates we need to check - the six previous
+    # days of the week for weekly searches and the last 30 days for monthly
+    # searches. We can combine all of these into a single Q.
     for day in days_of_week:
+        # Math: today is Monday (1) and the search should have sent on
+        # Saturday (6). 1 - 6 = -5. -5 + 7 = 2 days ago.
+        # Keep in mind that isoweekday is 1-indexed and starts on Monday.
         days_ago = today.isoweekday() - int(day)
-        if days_ago < 0:
-            days_ago = 7 - days_ago
+        if days_ago <= 0:
+            days_ago += 7
         q = q | (Q(frequency='W', day_of_week=day) &
                  (Q(last_sent__lt=today - timedelta(days=days_ago)) |
                   Q(last_sent__isnull=True)))
     for day in days_of_month:
         days_ago = day_of_month - day
-        if days_ago < 0:
-            days_ago = 31 - days_ago
+        if days_ago <= 0:
+            days_ago += 31
         # As a happy little side effect of requeuing saved searches, this will
         # resend monthly searches for, for example, the 31st of the month
         # on the first day of the following month if the previous month doesn't
--- a/tasks.py
+++ b/tasks.py
@@ -154,7 +154,7 @@ def requeue_missed_searches():
     Determine which saved searches should have been sent in the past but have
     not. Send those individually or in digests as appropriate.
     """
-    today = datetime.today()
+    today = date.today()
     day_of_week = str(today.isoweekday())
     day_of_month = today.day
 
--- a/tasks.py
+++ b/tasks.py
@@ -199,9 +199,9 @@ def requeue_missed_searches():
                                        flat=True).distinct()
     digests = SavedSearchDigest.objects.filter(pk__in=digests)
 
-    for digest in digests.filter(is_active=True):
-        send_search_digest.s(digest).apply_async()
-
+    # Accounting for digests greatly increases the complexity of preventing
+    # duplicate emails and there are plans to remove them eventually. As such,
+    # we are ignoring digests here.
     for digest in digests.filter(is_active=False):
         searches = all_searches.filter(user=digest.user)
         for search in searches:
--- a/tasks.py
+++ b/tasks.py
@@ -159,13 +159,12 @@ def requeue_missed_searches():
     day_of_month = today.day
 
     # We have lists of every day in the week and every day of the month.
-    # To exclude searches that will be sent later today, we need to filter
-    # today out of those lists.
-    not_today = lambda day: (lambda choice: choice != day)
-    days_of_week = filter(not_today(day_of_week),
-                          [choice[0] for choice in DOW_CHOICES])
-    days_of_month = filter(not_today(day_of_month),
-                           [choice[0] for choice in DOM_CHOICES])
+    # To exclude searches that will be sent later today, we need to remove
+    # today from those lists.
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -147,6 +148,66 @@ def update_partner_library():
             added, skipped, source)
 
 
+@task(name='tasks.requeue_missed_searches', ignore_result=True)
+def requeue_missed_searches():
```
tdphillips: This is a little messy and potentially more time consuming than normal sends. Not much I can do about that.

tasks.py
```diff
@@ -147,6 +148,66 @@ def update_partner_library():
...
+        # on the first day of the following month if the previous month doesn't
+        # have that many days in it.
+        q = q | (Q(frequency='M', day_of_month=day) &
+                 (Q(last_sent__lt=today - timedelta(days=days_ago)) |
+                  Q(last_sent__isnull=True)))
+
+    all_searches = SavedSearch.objects.filter(user__opt_in_myjobs=True,
+                                              user__is_disabled=False,
+                                              is_active=True)
+    all_searches = all_searches.filter(q)
```
tdphillips: If you would like a count, this is making use of (3*6) + (3*30) + 1 (109 total) Q objects AND/ORed together.

tasks.py
tdphillips: This logic could possibly have been rolled into the normal send_search_digests task with an increase in complexity but one of the requests was being able to run this separately. Win-win.

aspidites:
Q's always make me dizzy. It's a bit dense for me to say whether it's good or not. The new tests pass though, so I take that as a good indication.

tasks.py
```diff
@@ -147,6 +148,66 @@ def update_partner_library():
...
+        days_ago = today.isoweekday() - int(day)
+        if days_ago < 0:
+            days_ago = 7 - days_ago
+        q = q | (Q(frequency='W', day_of_week=day) &
+                 (Q(last_sent__lt=today - timedelta(days=days_ago)) |
+                  Q(last_sent__isnull=True)))
+    for day in days_of_month:
+        days_ago = day_of_month - day
+        if days_ago < 0:
+            days_ago = 31 - days_ago
```
koonsies: I think this should be day_of_month + day. Unless I am mistaken, this is subtracting a negative. For instance, a monthly extract with a DOM of 30th ran on today would have a "days_ago" value of 42.. So below, it would only look for emails that haven't been sent since September 7th. 

The other way would have the result of "days_ago" = 20, with the filter being last_sent__lt=September 29th.

Same goes for the week days.

tdphillips: I don't think I would have caught that. The same erroneous math is happening for days_of_week as well.

koonsies: Yeah, I've done that a hundred times. Takes having someone else reading the code to notice something like that.

tdphillips: I think I've fixed the math.

tasks.py
```diff
@@ -147,6 +148,72 @@ def update_partner_library():
...
+    not_today = lambda day: (lambda choice: choice != day)
+    days_of_week = filter(not_today(day_of_week),
+                          [choice[0] for choice in DOW_CHOICES])
+    days_of_month = filter(not_today(day_of_month),
+                           [choice[0] for choice in DOM_CHOICES])
+
+    q = Q()
+    # There are 36 distinct last_sent dates we need to check - the six previous
+    # days of the week for weekly searches and the last 30 days for monthly
+    # searches. We can combine all of these into a single Q.
```
tdphillips: I have a copy of this query from yesterday. It seems correct after a quick skim but I can pass it along if anyone is interested.

tasks.py
```diff
@@ -147,6 +148,72 @@ def update_partner_library():
...
+    not. Send those individually or in digests as appropriate.
+    """
+    today = datetime.today()
+    day_of_week = str(today.isoweekday())
+    day_of_month = today.day
+
+    # We have lists of every day in the week and every day of the month.
+    # To exclude searches that will be sent later today, we need to filter
+    # today out of those lists.
+    not_today = lambda day: (lambda choice: choice != day)
```
aspidites: I think these would be clearer without the lambdas and filter:
```python
days_of_week = [choice[0] for choice in DOW_CHOICES if choice[0] != day_of_week]
```
In fact, it may be clearer to just use `list.remove`:
```python
days_of_week = [choice[0] for choice in choices]
days_of_week.remove(day_of_week)
```
It's one more expression, but still the same number of lines if you consider wrapping.

tasks.py
```diff
@@ -147,6 +148,72 @@ def update_partner_library():
...
+                          [choice[0] for choice in DOW_CHOICES])
+    days_of_month = filter(not_today(day_of_month),
+                           [choice[0] for choice in DOM_CHOICES])
+
+    q = Q()
+    # There are 36 distinct last_sent dates we need to check - the six previous
+    # days of the week for weekly searches and the last 30 days for monthly
+    # searches. We can combine all of these into a single Q.
+    for day in days_of_week:
+        # Math: today is Monday (1) and the search should have sent on
```
aspidites: :+1:  for explaining the calculation.

tasks.py
tdphillips: But my nested lambda

tdphillips: I'm working on it. I don't know why I did it this way.

tasks.py
```diff
@@ -147,6 +148,72 @@ def update_partner_library():
...
+        # Saturday (6). 1 - 6 = -5. -5 + 7 = 2 days ago.
+        # Keep in mind that isoweekday is 1-indexed and starts on Monday.
+        days_ago = today.isoweekday() - int(day)
+        if days_ago <= 0:
```
aspidites: If you modulo 7 you don't have to use an if statement. `(1 - 6) % 7 == 2`.

tasks.py
tdphillips: Removed.

tasks.py
tdphillips: Replaced both instances in this file with modulo. The tests still have the ifs in them, however.

tasks.py
```diff
@@ -173,6 +239,8 @@ def filter_by_time(qs):
         monthly = qs.filter(frequency='M', day_of_month=today.day)
         return chain(daily, weekly, monthly)
 
+    requeue_missed_searches.apply_async()
```
JLMcLaugh: Is last_modified updated when the task actually executes? With multiple workers is this going to create duplicates by requeuing tasks before the initial emails have a chance to send?

tdphillips: In the requeue_missed_searches task, we're looking for and excluding saved searches that are to be sent today. There shouldn't be any duplicates.

PR #1823, opened 2015-10-15
[PD-1862, PD-1847] Activity migration

aspidites:
Created activities for the following AppAccesses:

PRM

    Contact
    Partner
    Communication Record
    Partner Saved Search

User Management

    User
    Role

Each of the above has CRUD activities (create, read, update, delete).

**Depends on #1820**. Only real changes are [these](https://github.com/DirectEmployers/MyJobs/compare/activities...activity-migration)

## Update
In addition, I wrote another data migration that creates a "Role Admin" role for each company and adds all activities to that role. ~~I have a few open questions, yet, which is why the data migration is still incomplete and this is marked as DO NOT MERGE.~~ A standalone script will be used to assign users to the roles created by this migration.


aspidites:
Want to play with some of the default role stuff before these go in.

koonsies:
:+1:

PR #1824, opened 2015-10-16
[PD-1383]  dynamic reporting interface

darrint:
With this change many of the myreporting new models are being used as intended to create a dynamic reporting system. Only one report works: Contacts, although more report types appear in the UI.

* The UI is still throwawy.
* Using the UI you can walk all the way through to running a report.
* There is no filtering, etc.
* The column list in the model is respected.

OT for this PR but needs discussion: At this point the Javascript is a disaster. If we are serious about building single page apps we need to discuss the future of how we will implement them. Piling up loads of jQuery as I have done here won't scale.

To test this I recommend running `manage.py reset_dynamic_data` in qc. **Do not** run it in production though, at least not without understanding exactly what it does. It   deletes all records from some tables.


myreports/report_query.py
```diff
@@ -0,0 +1,39 @@
+from mypartners.models import Status, Contact
```
aspidites: No documentation.

myreports/report_query.py
```diff
@@ -0,0 +1,39 @@
...
+def get_report_query(report_query_name):
+    return report_queries[report_query_name]()
+
+
+def list_of_str(qs):
+    return [str(f) for f in qs.all()]
+
+
+@report_query("Contacts")
+class ContactReportType(object):
```
aspidites: I thought the point of the new models was to not hard code report types, but rather to have them represented entirely by database table relations. Did you and @haliscooper come to some other understanding that might alleviate my confusion?

myreports/tests/factories.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+class InterfaceElementTypeFactory(DjangoModelFactory):
+    class Meta:
+        model = 'myreports.InterfaceElementType'
+
+    is_active = True
+
+
+def create_full_fixture():
```
aspidites: Can't place my finger on it, but this whole function feels counter intuitive. Part of that feeling stems from the fact that you are calling directly into the factory's `_meta` attribute just to get at the underlying model.

darrint: It was _meta or have another 11 imports.

myreports/report_query.py
darrint: Yes. As a result of discussion, queries originate with reporting types and are somehow refined or modified by data type.

myreports/tests/factories.py
aspidites: What exactly are these lines doing? Between the lack of comments and the short variable names, I can't exactly tell.

aspidites: Kinda feels like the kind of thing that would be in a setUp method.

myreports/tests/setup.py
```diff
@@ -5,6 +5,19 @@
 from mypartners.tests.factories import PartnerFactory
 from seo.tests.factories import CompanyFactory, CompanyUserFactory
 
+from myreports.tests.factories import (
```
aspidites: Much better. I don't expect you to change this again, but usually if I know I'm going to need a lot of symbols, I'll just import the module and access what I need to qualified. This works though, and takes less effort to ascertain what's going on in those fixtures.

myreports/tests/test_views.py
```diff
@@ -419,9 +417,47 @@ def test_report_types_api(self):
...
-        self.assertEquals("Communication Records", data['2']['name'])
+        self.assertEquals("Contacts", data['2']['name'])
+        self.assertEquals("Contacts Report", data['2']['description'])
+        self.assertEquals("Communication Records", data['3']['name'])
         self.assertEquals("Communication Records Report",
-                          data['2']['description'])
+                          data['3']['description'])
+
+    def test_data_types_api(self):
+        resp = self.client.post(reverse('data_types_api'),
```
aspidites: Docstrings.

myreports/report_query.py
aspidites: Is this conversation logged somewhere? Interested to see the logical steps that led to the exceptions to "everything should be data driven". In particular, the current reporting system is able to query on arbitrary models, where as this seems to be model specific.

koonsies: +1 to curiosity about this

aspidites:
Bad news:
2015-10-21-155540_956x420_scrot

I get this when clicking next a few times (Final button is "Contact CSV").

aspidites:
And no navigation. Will @lemniscate613 be taking care of that once backend stuff works?

darrint:
I see what you mean about the unicode now. Yes, if you could identify some data that causes this I would greatly appreciate it.

tdphillips:
People love to paste from MS Word et al into web forms. \u2019 is a fancy right single quotation mark but we've also had issues with \u2013 (en-dash) and whatever the code point for bullet is.

darrint:
@tdphillips Thanks. That allowed me to reproduce the exception in a view test.

PR #1825, opened 2015-10-16
[PD-1850] Automatically add activities to Admin roles.

aspidites:
*This depends on #1823 , which in turn depends on #1820.*
This makes it so that creating a new activity automatically adds that activity to the "Role Admin" role. In my tests, it takes ~5 seconds to add a new activity to ~30,000 "Role Admin" roles (in 4 queries).

~~The actual differences are [here](https://github.com/DirectEmployers/MyJobs/compare/activity-migration...pd-1850?expand=1)~~

# Hotfix
*update:* I changed Role Admin -> Admin. I didn't create another migration since this hasn't hit production, and I didn't want to run two nearly identical data migrations with the second undoing the work of the first.

The other migration wouldn't work on QC because the max packet size there is 4M vs 16M locally. With @mklauber 's guidance, I decided to batch the bulk create at 450 rows.

### DIFF
--- a/myjobs/migrations/0007_create_admin_roles.py
+++ b/myjobs/migrations/0007_create_admin_roles.py
@@ -7,11 +7,11 @@ from django.db import models
 class Migration(DataMigration):
 
     def forwards(self, orm):
-        """Create a 'Role Admin' role for existing companies."""
+        """Create a 'Admin' role for existing companies."""
 
         # create the required roles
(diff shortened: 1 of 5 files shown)


koonsies:
:+1:

aspidites:
I've marked this as a hotfix since it changes a migration, per @JLMcLaugh 's request, and thus needs to be merged before other stuff gets to qc.

PR #1827, opened 2015-10-19
[PD-1817] Add handling for last_action_time to applicable forms on PRM

koonsies:
## Changes

* Rename last_modified to last_action_time
* Add last_action_time handling to Partner, Contact, ContactRecord, PartnerSavedSearch forms
* Add tests for last_action_time changes on above forms
* Correct misc. comments/import errors
* Hand testing of PRM modules (see below for steps)

## Testing By Hand

* Clone branch pd-1817_last_modified_handling
* If you haven't done so, link a member company to your user account to access PRM (CompanyUser)
* Run django server, access site through any url that you have overwritten in hosts file (my.jobs:8000/prm/ for me)
* Add/Edit the forms relating to Partner, Contact, ContactRecord and PartnerSavedSearch
* Watch server log for print statements declaring that the last_action_time has been updated
 * ex. PARTNER-EDIT:  2015-10-19 09:01:25.123915

### Copied from Ticket..

The update for this column is based on saving the following forms: 
* NewPartnerForm
 * Handled via default value for last_action_time
* PartnerForm
 * Handled on form.save().
* ContactForm
 * Handled on form.save().
 * Connected partner is NOT updated when Contacts added/modified/deleted
* LocationForm
 * Handled on form.save().
 * Connected Contact IS updated when location added/modified/deleted
* ContactRecordForm
 * Handled on form.save().
* PartnerSubSavedSearchForm
 * Handled on form.save().
* PartnerSavedSearchForm
 * Handled on form.save().

There is an important caveat to the above functions that are handled on a form.save(): The save functions for logging as well as updating this value do NOT check to ensure there are actually changes made on the form. Opening the Partner form and clicking "Save" will create a change log and update last_action_time just the same as if you had changed the name, tags, or so on.

Another note: modifying tags through tag manager does not update any of the four models. For instance, if you tag a few objects as "Foo", go to tag manager and change "Foo" to "Fo" (thus changing it for all of the connected models), none of those models' last_action_date will change. If, instead, you go to a connected objects form (Contact, for instance) and delete the tag, it will reflect as a change on that Contact.


aspidites:
@fezick, @haliscooper, I thought at one point we said that fields such as last_modified, archived_on, etc shouldn't be displayed in reporting exports? If so, I think this + recent changes that added archived_on to more models introduced a regression:

2015-10-20-102933_1916x1056_scrot
@koonsies, this would be a blocker assuming my memory isn't failing me. I can point you in the right direction as far as correcting this goes. Should be a one line change (a hack that's been there a while, but hasn't been removed since reporting was getting overhauled anyway).

aspidites:
Another regression, though I don't know if it originates here or elsewhere: If a partner saved search is made for a user and that user opts out of myjobs emails, that saved search is marked as in active when viewing "My Saved Searches", but not so when viewing the partner saved search directly as a partner.

The reason I'm commenting here is I'm curious as to whether a partner saved search becoming inactive due to a user unsubscribing or a user opting out, should this last_action_time be changed or not? At the moment it is not.

aspidites:
There are a few inconsistent cases that I've talked with @koonsies about that I disagree with, but seem to be conscious design choices. Barring those and what I've most recently commented on, this seems to be working as expected.

haliscooper:
Last modified is intended to be visible in reporting, archived on would not (since it's a fancy name for "you deleted this, you don't even know it exists anymore").  Last Modified, as implemented, is a user facing piece of info, relevant to the user.

aspidites:
I'll open a separate ticket then.

aspidites:
Still waiting on an answer about the partner saved search issue I identified above. @haliscooper , @fezick , @JLMcLaugh, @koonsies ?

aspidites:
Looks like you can recover from an active/unsubscribed state assuming you can reach out to the contact and let them know that they aren't receiving saved searches because they unsubscribed from myjobs communication. In that case, it's not a regression, though I think it remains questionable if that is the desired behavior. 

Question about if that counts as a change still remains though.

Also, I don't feel right merging this myself until a member of mangement, @haliscooper  in particular confirm that the caveats are those which are expected. I'd like to avoid another case of "This is exactly not what we asked for even though we had conversations that aren't documented about how this is exactly what we asked for".

mypartners/migrations/0027_auto__del_field_partner_last_modified__add_field_partner_last_action_t.py
```diff
@@ -0,0 +1,532 @@
...
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Deleting field 'Partner.last_modified'
+        db.delete_column(u'mypartners_partner', 'last_modified')
```
JLMcLaugh: I know most of these don't have data, but I'd rather have this be a rename than a destructive delete/add.

koonsies: Done.

mysearches/migrations/0013_change_last_modified_to_last_action_time.py
```diff
@@ -0,0 +1,502 @@
...
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+        # Deleting field 'SavedSearch.last_modified'
+        db.delete_column(u'mysearches_savedsearch', 'last_modified')
```
JLMcLaugh: Rename here too.

koonsies: Goodness. I think I forgot to hit "save" before committing haha. I had that changed at one point!

PR #1828, opened 2015-10-19
Improve mypartners factories [PD-1863]

tdphillips:
Fixing all factories is a bigger beast than I thought, so I'll leave it at an app-by-app basis as issues crop up.

This can still result in odd relationships as demonstrated here but this is an issue that can be fixed through correct use of the factory.
```
    assert Partner.objects.count() == 1
    contact = Contact.objects.get(pk=...)
    record = ContactRecordFactory(contact=contact)
    assert Partner.objects.count() == 2
    assert contact.partner != record.partner
```

All tests pass.

### DIFF
--- a/mypartners/tests/factories.py
+++ b/mypartners/tests/factories.py
@@ -58,7 +58,8 @@ class ContactRecordFactory(factory.django.DjangoModelFactory):
     notes = 'Some notes go here.'
     date_time = datetime.now()
 
-    contact = factory.SubFactory(ContactFactory, name='example-contact')
+    contact = factory.SubFactory(ContactFactory, name='example-contact',
+                                 partner=factory.SelfAttribute('..partner'))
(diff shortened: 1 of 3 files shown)


koonsies:
:+1:

aspidites:
What exactly is demonstrably odd? I feel like the assertions are assuming something that I'm not understanding, so I can neither agree nor disagree as to whether the output is odd.

tdphillips:
I haven't checked, but I'm fairly certain that contactrecord.contact.partner and contactrecord.partner are the same in our database.

There's no enforcement of this in the database or code, however, so the factory is working as designed. If you don't want additional partners to be created and you're creating a record from a contact, you'll need to specify the partner as well.

I'm not saying it's a problem, just noting it as something to keep in mind when working with factories and foreign keys.

PR #1829, opened 2015-10-20
[PD-1869] Added archived_on to the global blacklist for reporting.

aspidites:
Results:
2015-10-20-132753_1916x1056_scrot

If you go to production now and try to create a CSV for a communication record report, you'll see that the "archived on" field is present, when it shouldn't be. This corrects that. I basically promoted that column to the common blacklist.

### DIFF
--- a/myreports/views.py
+++ b/myreports/views.py
@@ -261,10 +261,10 @@ def downloads(request):
         report = get_object_or_404(
             get_model('myreports', 'report'), pk=report_id)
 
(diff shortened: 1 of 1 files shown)


aspidites:
This is NOT a hotfix. No tests, as this was a sort of adhoc feature whereby whats included was decided upon arbitrarily. Furthermore, the new models should account for this kind of thing.

PR #1830, opened 2015-10-21
[PD-1866] Wrapped PRM views with requires on QC and Staging

aspidites:
# Description
I've modified the `requires`, and `warn_when` family of decoraties (including `has_access`) decorators as follows:
- `requires` returns the original view if `settings.DEBUG` is `False`, otherwise it does its thug thizzle.
- `warn_when` is unconditionally enabled unless the `enabled` argument is set to `False`. In which case, it is disabled when `settings.DEBUG` is `True`.

The effect of the above is that if a view is decorated with both `has_access` and `requires`, requires will only work when debugging is enabled, where as `has_access` will only work when debugging is disabled.

# Manually Testing
In order to test this yourself, modify the `DEBUG` variable in your `settings.py` and observe the following expected behavior:
- When `DEBUG` is `True`:
-- visiting any PRM page for a company that doesn't have "PRM" app access should result in a 404 page saying that app-level permissions are missing
-- visiting any PRM page for a company that *does* have "PRM" app access, but as a user whose roles don't have all of the PRM-related activities should result in a 404 saying that activities are missing.
- When `DEBUG` is `False`:
-- app access, roles, and activities shouldn't matter. Instead, you should see a warning when attempting to access PRM for a company who is not a member. 

# Caveats
- I am aware that it is possible to assign a user to a role that belongs to a company for which they do not have access. As roles will eventually accomplish what having a row on the company user table does, becoming a so-called company user would be a matter of being assigned a role in that company with the appropriate activities. In the mean time, I think we should do any role filtering at the UX level.

# Relevant Changes
I based this off of my pd-1850 branch. If you're curious about what is different between that one and this one, go [here](https://github.com/DirectEmployers/MyJobs/compare/pd-1850...pd-1866). Should be less intimidating than what's currently showing in the "Files changed" tab.


universal/decorators.py
```diff
@@ -166,4 +171,4 @@ def has_access(feature):
     return not_found_when(
         condition=lambda req: all(c(req) for c in conditions.get(feature, [])),
         message="Must have %s access to proceed." % feature,
-        feature=feature)
+        feature=feature, enabled=False)
```
koonsies: So, if the purpose of this is to ensure that has_access('prm') is on in production, but off everywhere else, and the exact inverse is true for "@requires", why does one use an enabled flag while the other does not? Would there be an instance where you would change enabled=True, but not remove the overwrite for "@required"? Not saying it's wrong, just curious.

aspidites: The `enabled=True` is a bit of a hack needed because `warn_when_inactive` isn't translatable to activities, so I need a way to be able to say "don't do special things with DEBUG". Since has_access is a decorator factory, the alternative would have been to check debug in there directly then construct a decorator that did nothing for the case where debug is false. Decorators are tricky enough that I am more confident that a new function parameter is a more benign change than drastically changing the body of a decorator. 

tl;dr - the enabled=False thing is an escape hatch and perhaps poorly named, but I wasn't too concerned about it as it'll be removed sooner than later. That's also why it's not a documented parameter.

tdphillips:
Looks like merging the previous PR in the chain rendered this one unable to be merged.

PR #1833, opened 2015-10-21
[PD-1836 PD-1837] company role/user overview and listing interface

lemniscate613:
[My previous PR was based off of Edwin's PD-1866, but it should have been based off of quality-control.]

Here's a draft of the interface for managing roles and users. Any feedback welcome (especially UI). Still need to make fully interactive with Activities.

Relevant changes: https://github.com/DirectEmployers/MyJobs/compare/pd-1866...pd-1837

overview

roles

edit role

users

edit user


aspidites:
- Do we want a separate app for this? Given that User is in myjobs, I added the roles and activities models to myjobs, so would expect this to be an extension of that. 
- If we do want a separate app for this, I don't think it should be called manage. I'm not looking forward to typing `./manange.py test manage` and similar.

haliscooper:
re Edwin's comment above, it should be where the other user stuff is, ie, my.jobs
Jason S noted a single page app in the definition of done in PD-1836, any reason why we strayed from that?
The role column on manage users is likely to be pretty crowded, as Role names aren't likely to be as short as "Role 1".
Jason's "The UI presents Rebecca with a form to customize the invitation message, and then sends Joe an invite when she clicks save" isn't accounted for here.
Lastly, a user needs to be able to assign or unassign roles to multiple users at once, it's a very common use case, you create a new role, and need to add it to many or all users at once, to work your way through 200 users each time you add a role or split it up or whatever manipulation you do (whenever we release a new feature that isn't just an extension of an existing feature, or, more pressing, day one of the rollout of this system when the main user is trying to create roles and assign them to multiple users).  Whether that's a separate interface (preferred), as trying to change one user's roles becomes a chore is roles are always assigned in a bulk attach interface) or adapts the manage users list page, either works, but it's a common use case.  Can be another ticket, but accounted for in this one.

PR #1836, opened 2015-10-23
[PD-1748] Changed domain back to a charfield.

aspidites:
@haliscooper  discovered that `URLField`s require a scheme to be passed in (eg. "http://"), which is unacceptable. Changed this back to a char field with 200 char max.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -872,7 +872,7 @@ class OutreachEmailDomain(models.Model):
         ordering = ["company", "domain"]
 
     company = models.ForeignKey("seo.Company")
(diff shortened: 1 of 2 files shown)


mypartners/models.py
```diff
@@ -872,7 +872,7 @@ class Meta:
         ordering = ["company", "domain"]
 
     company = models.ForeignKey("seo.Company")
-    domain = models.URLField()
+    domain = models.CharField(max_length=200, unique=True)
```
tdphillips: This one shouldn't be unique.

aspidites: I'm getting sloppy. Thanks for catching this.

PR #1837, opened 2015-10-23
[PD 1837 PD 1837] company role/user overview and listing interface

lemniscate613:
Work in progress. This is here for visibility.

Seeking feedback especially on:
* Anything UI, user flow
* How to explain what activities are (need an Activities page?)
* How to make the Overview page useful

Big updates with this PR:
* Single page app built on ReactJS. It's a basic setup. It will work locally without you installing anything (e.g. npm/browserify/gulp stuff). I wanted to build this before our discussions on build/deploy processes firmed up (which it looks like they are). Next to do is integrating these build tools.
* Bootstrap 3. David mentioned using 3 for his Blocks Admin, so I used it for this new project as well. Using 3 will also make it possible to use prebuilt React components built according to 3: https://react-bootstrap.github.io/ There are caveats with IE8 support, but from what I've researched it looks like we can account for them. 
* This PR has no new app, unlike my previous one ( https://github.com/DirectEmployers/MyJobs/pull/1833 ). It's simply a new view in myjobs/views.py and a new line in myjobs/urls.py.
* When adding / editing a role, you will be able to update associated users. This is based on the insight from @haliscooper here: https://directemployersfoundation.atlassian.net/browse/PD-1836#comment-172226
* Edit invitation email when creating a new user.
* Updated users and roles layout. Should scale visually better (e.g. if a user is assigned to many roles).
* Added an Activities page. I'm sure there's a better way than this to explain to users what each Activity is.

Main things to do:
* Integrate build tools (most likely will use a Yeoman generator)
* Install React-Bootstrap and use it's Components
* Interact with real data

overview

roles

add role

edit role

activities

users

add user

edit user

### DIFF
--- a/myjobs/urls.py
+++ b/myjobs/urls.py
@@ -34,4 +34,5 @@ urlpatterns = patterns(
     url(r'^toolbar/$', 'toolbar', name='toolbar'),
     url(r'^cas/$', 'cas', name='cas'),
     url(r'^topbar/$', 'topbar', name='topbar'),
+    url(r'^manage-users/$', 'manage_users', name='manage_users'),
 )
--- a/myjobs/views.py
+++ b/myjobs/views.py
@@ -31,6 +31,7 @@ from myprofile.forms import (InitialNameForm, InitialEducationForm,
 from myprofile.models import ProfileUnits, Name
 from registration.forms import RegistrationForm, CustomAuthForm
 from tasks import process_sendgrid_event
+from universal.helpers import get_company_or_404
 
 logger = logging.getLogger('__name__')
 
@@ -592,3 +593,16 @@ def topbar(request):
     response.content = "%s(%s)" % (callback, json.dumps(html.content))
 
     return response
+
+
+def manage_users(request):
+    """
+    View for manage users
+    """
+    company = get_company_or_404(request)
+    ctx = {
+        "company": company
+        }
+
+    return render_to_response('manageusers/index.html', ctx,
+                                RequestContext(request))
--- a/myjobs_urls.py
+++ b/myjobs_urls.py
@@ -38,6 +38,12 @@ urlpatterns = patterns(
 
 
 urlpatterns += patterns(
+    'myjobs.views',
+    url(r'^manage-users/$', 'manage_users', name='manage_users'),
+)
+
+
+urlpatterns += patterns(
     '',
     url(r'^accounts/', include('registration.urls')),
 )
--- a/myjobs_urls.py
+++ b/myjobs_urls.py
@@ -36,13 +36,6 @@ urlpatterns = patterns(
     url(r'^emails/', include('myemails.urls')),
 )
 
-
-urlpatterns += patterns(
-    'myjobs.views',
-    url(r'^manage-users/$', 'manage_users', name='manage_users'),
-)
-
-
 urlpatterns += patterns(
     '',
     url(r'^accounts/', include('registration.urls')),
--- a/static/custom.css
+++ b/static/custom.css
@@ -1,0 +1,3998 @@
+/*Document level classes*/
+
+.topbar h3 a {
+    padding: 0 20px;
+}
+
+.btn {
+    padding-top: 6px;
+    padding-bottom: 7px;
+}
+
+.clear {
+    clear: both;
+}
+
+.btn.pull-left {
+    margin-left: 10px;
+}
+
+#focusBox {
+    margin-bottom: 10px;
+}
+
+html,
+body {
+    height: 100%;
+    background: transparent;
+}
+
+body {
+    font-family: Arial, Helvetica, sans-serif;
+    height: auto;
+    background-color: #DDD;
+}
+
+header,
+footer {
+    background-color: transparent;
+
+    margin: 0;
+}
+
+footer {
+    padding-top: 0;
+    border-width: 0;
+}
+
+.wrapper{
+    margin: 0 auto;
+    width: 960px;
+    display: block;
+}
+
+/*Header Level Elements*/
+h1 {
+    color: #900;
+}
+
+h2 {
+    line-height: 1;
+    font-size: 1.7em;
+    color: #333;
+}
+
+.h2block {
+    background-color: #fff;
+    padding: 5px 20px;
+    border-width: 1px;
+    border-color: #d4d0c5;
+    border-style: solid;
+    -moz-border-radius:6px;
+    -webkit-border-radius:6px;
+    border-radius: 6px;
+    margin-bottom: 20px;
+}
+
+.h2block h2 {
+    margin-bottom: 10px;
+}
+
+p + h2 {
+    margin-top: 20px;
+}
+
+/*Forms*/
+input,
+textarea{
+    width: auto;
+}
+
+/*Tables*/
+.rowTitle {
+    font-weight: bold;
+}
+
+/*Lists*/
+li {
+    color: #000;
+}
+
+dd {
+    margin-bottom: 10px;
+}
+
+/*Blockquote*/
+.section blockquote {
+    margin: 10px;
+    padding: 5px;
+    border-width: 2px;
+    border-style: solid;
+    border-color: #999;
+    border-color: rgba(127,127,127,.5);
+    background-color: rgba(255,255,255,.5);
+}
+
+/*Contant Containers*/
+.hero-unit h1 {
+    background: #333;
+    color: #fff;
+    margin: -60px -60px 0 -60px;
+    -webkit-border-top-left-radius: 6px;
+    -webkit-border-top-right-radius: 6px;
+    -moz-border-topleft-radius: 6px;
+    -moz-border-topright-radius: 6px;
+}
+
+.roundedCorners6 {
+    -webkit-border-radius: 6px;
+    -moz-border-radius: 6px;
+    border-radius: 6px;
+}
+
+.roundedCorners6 > h2 {
+    -webkit-border-top-left-radius: 6px;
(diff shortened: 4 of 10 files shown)


aspidites:
I'll take a closer look at this between meetings today, but my initial feedback is rather terse:
- Seems to capture all the functionality provided by the models I introduced, so :+1: 
- You do have an activity description to lean on, so that should take care of non-descriptive activities
- It seems like for assigning activities to a role you have two lists (not selected/selected), but for assigning that role to users, there is only one list, which to me feels a bit inconsistent. 
- On desktops with this kind of select mechanism, there are usually buttons for "add one", "add all", and the "remove" analogs. I think this is the case in django admin, as well. Would that make sense here as well?
- The overview page looks a bit vacuous at the moment. A lot of empty whitespace.
- Would it be more consistent if the roles listing was in the card style that we use elsewhere (see David's most recent UI update).

lemniscate613:
Excellent, excellent points.

"The overview page looks a bit vacuous at the moment. A lot of empty whitespace."

Totally agree. Wondering what information would make it useful... If it's not needed, we can just make the Roles view the default view.

aspidites:
# Detailed Feedback
Finally had a chance to take this for a test spin. I might repeat some points above, but I'll include more details, which should be a bit more actionable for you than "this is wrong".

## Caveats
- I blatantly ignored most tablet/mobile stuff. In particular, if it isn't a regression, I didn't mention it.
- I ignored anything that would be affected by the back-end. For instance, top-level navigation (which would be controlled by activities).

# Areas of Opportunity
## Overview
- The top bar is broken in that the hamburger button is shorter than the arrow button:
2015-10-26-111716_133x64_scrot
- I think the overview would be more functional if something were clickable. For instance, being able to click the 1 in pending to see who still hasn't accepted an invitation, or clicking the green 3 in active to see users.
- Being able to use the history api to go back to the overview would be nice (back and forward buttons for those not in the know). Alternatively (especially for IE 8), an explicit link to the overview would be great

## Roles
- As we are moving away from tables, I wonder if it would be best to use cards here, or if, from a designers perspective, it make sense to stay with tables. I think I heard an argument for tables once with respect to the profile page from @fezick, but positive feedback seemed to be unanimous with respect to the recent My.jobs face lift @galitskyd implemented
- The cursor for the Edit link is still a bar, rather than a pointer
- It might be useful to use the popups we have in the old reporting interface as tips for activities. For instance, hovering over "Access Analytics" would show  that activity's description.

### Edit Role
- I'm unsure if the "Activities Assigned to this Role" bit is unfinished or not. If unfinished, I suspect my upcoming comments are redundant. 
- As I mentioned before, it would be nice if there were some buttons between the two boxes to add/remove activities from a role.
- Alternatively, is there any reason two select boxes is better than a list of check boxes? Something like:
2015-10-26-113646_298x232_scrot
- No modal when you delete a role, where as elsewhere we  tend to prompt the user before doing destructive actions
### Activities
- What purpose is there in the activities page, given that a user cannot modify them in any way? As we provide activities, the only interaction users will have with them is to assign them to roles.
- Assuming the answer to the above is "documentation for the user", I think we'd be better served by showing descriptions as they are trying to decide which activities to add to a role. Otherwise, they end up having to keep two tabs open to see what an activity does while they are assigning them to roles.

## Users
- It might be nice to be able to sort by role, status, and user email. 

### Edit User
- Does the disabled button with the @ symbol do anything? I'm not sure it adds anything to the page, especially with the existence of the label.
- Roles aren't mandatory. That is, you can have a user who isn't assigned to any role (and thus has no permissions nor access to any company). That said, they have to at least be assigned a role with no activities to have access to the company*, so I wonder if that is still worth enforcing, or if that's a matter of educating the user? 

# Successes
- It feels generally snappy and while resources are large, I suspect we only have to load them once.
- Even though we are using a different grid system, I can't, as a user, tell.
- The widget choice by and large feels consistent with what we currently use elsewhere
- Compared to how long new interfaces tend to take, this seemed to come out rather quickly, which makes me confident in our (read: your) ability to iterate on design  and remain agile
- While the tablet view and topbar need some work, the mobile views seem to have mostly taken care of themselves (perhaps thats a point for bootstrap 3?)

* We currently determine company access via a through table that joins companies to users, but my intention was that the same could be accomplished by simply assigning a user a role which has no activities. That is:
before: User <-> CompanyUser <-> Company
after: User <-> Role <-> Company

haliscooper:
I actually prefer the checkboxes with the nested approach, I've used that before to great benefit.  THe purpose of an activities page is (my guess, not sure I noted that req) is for users to see what permissions they have in a nice summary, ie, it's the read only version of what's available to them?  Just a guess, it's necessary to be able to have a user see what they can do, though.  
Can't run any of your code to see any of this, just commenting on your notes...

haliscooper:
And yes, a user must be assigned to a role, as you noted, you want to lose the company user table in the future, at which point, going through roles is your join.

haliscooper:
Ahh, saw your note, yep, exactly.

lemniscate613:
Nice! Thank you for taking the time to give such *substantial* feedback. I'll process it after our Sprint Planning meeting.

Also, thank you mentioning (offline) the idea of caveats. In future PRs I'll be more careful to give context about what's done / isn't to make reviewing more efficient.

More later...

haliscooper:
Just removed some of my comments, ignore what isn't there anymore.

lemniscate613:
> Roles aren't mandatory. That is, you can have a user who isn't assigned to any role (and thus has no permissions nor access to any company). That said, they have to at least be assigned a role with no activities to have access to the company*, so I wonder if that is still worth enforcing, or if that's a matter of educating the user? We currently determine company access via a through table that joins companies to users, but my intention was that the same could be accomplished by simply assigning a user a role which has no activities.

Ideally, we do as much behind the scenes as possible. If we have an empty role (with no activities) that every user gets added to automatically (so we don't get orphaned users) we just don't list that role in the list of roles. Did I read that correctly?

aspidites:
## TL;DR
You can safely ignore my comment about making roles mandatory. For this use case, they should be (though it will still not be mandatory at the database level).

## Full disclosure

> Ideally, we do as much behind the scenes as possible. If we have an empty role (with no activities) that every user gets added to automatically (so we don't get orphaned users) we just don't list that role in the list of roles. Did I read that correctly?

@haliscooper , correct me if I'm wrong.
@lemniscate613 , Ideally, we offload as much of the work to the user as possible without it feeling disruptive. That is, the only role which is automatically created is the admin role*, because we need to have a way for users to create other roles without having first seen the system. From then on, it is the responsibility of the admin to create and assign new roles. 

> Roles aren't mandatory. That is, you can have a user who isn't assigned to any role (and thus has no permissions nor access to any company). That said, they have to at least be assigned a role with no activities to have access to the company*, so I wonder if that is still worth enforcing, or if that's a matter of educating the user? 

While this technically true that roles are not mandatory, @haliscooper wanted to enforce the assigning of a role much as you already had to prevent the orphan user case you just mentioned**

### Notes
* Technically, there will be another role which has all but role/activity privileges which will be assigned to existing companies, but not new ones. The point of this particular role is to facilitate migration in not forcing them to recreate and assign roles for functionality they have already been using for the last year and a half.

** Technically, it is already possible to create orphan users. I think it's personally a misnomer to consider them orphan, as a person need not belong to a company to use saved searches and build a profile. In essence, they are simply job seekers. Rather, I think the incongruence (which @haliscooper ) mentioned, comes from inviting a user to a company only for that user to not have access to that company.

haliscooper:
aspidites is correct on all counts

lemniscate613:
Just for the record I asked @aspidites in chat,

>To summarize our conversation... Should I make it so that -- in order to save a user -- at least one role must be selected?

And he said, "yep."

haliscooper:
A resounding "Yep"

lemniscate613:
@fezick mentioned making the Roles and Users functionality and visually distinct. Two questions for you. 1) Could you confirm the current design meets your intentions? They are under the same umbrella, "Manage Users:"

screen shot 2015-10-27 at 9 00 17 am

But separated by the shared side menu:

screen shot 2015-10-27 at 9 01 15 am

I ask because I could imagine them being even _more_ distinct than they are now. For example, presented to the user as two separate applications:

screen shot 2015-10-27 at 9 02 59 am

The only catch is, @haliscooper said it's important that users be able to -- while adding/editing a role -- modify many users at once. So the functionality is understandably a bit mixed.

2) Currently, everything PD-1836 and PD-1837 happens under this URL which does not change: http://127.0.0.1:8000/manage-users/

Even though it's a single page app, we can use ReactJS URL routing (instead of Django routing) to get URLs like:

http://127.0.0.1:8000/manage/
http://127.0.0.1:8000/manage/roles/
http://127.0.0.1:8000/manage/roles/add
http://127.0.0.1:8000/manage/roles/edit
http://127.0.0.1:8000/manage/users/
http://127.0.0.1:8000/manage/users/add
http://127.0.0.1:8000/manage/users/edit

This is how I designed the URLs originally, but @aspidites didn't like the idea (understandably) of the generic "manage" bit.

So maybe we can instead have:

http://127.0.0.1:8000/manage-roles/
http://127.0.0.1:8000/manage-roles/add
http://127.0.0.1:8000/manage-roles/edit
http://127.0.0.1:8000/manage-users/
http://127.0.0.1:8000/manage-users/add
http://127.0.0.1:8000/manage-users/edit

Do you have strong opinions about how we structure the URLs on this project?

aspidites:
@lemniscate613  to be clear, I had a problem with an app called manage. I don't care what the urls are. In other words `./manage.py migrate manage` is an eye sore, navigating to `/manage/{users,roles}/' is perfectly fine.

PR #1835, opened 2015-10-23
Use PRM styles with old apps (homepage, myprofiles, mysearches)

galitskyd:
This is an attempt to unify secure.my.jobs styles and to give the older apps some :heart:

Also fixed the homepage that had a few issues with style and alignment issues.

## Homepage -- Logged Out
screen shot 2015-10-23 at 9 11 49 am

screen shot 2015-10-23 at 9 11 55 am

## Homepage -- Logged In
screen shot 2015-10-23 at 9 12 32 am

screen shot 2015-10-23 at 9 12 52 am

## MyProfiles
screen shot 2015-10-23 at 9 23 53 am

screen shot 2015-10-23 at 9 24 07 am

## MySearches
screen shot 2015-10-23 at 9 24 42 am

screen shot 2015-10-23 at 9 24 50 am

## MySearches -- View Feed (affects PRM as well)
screen shot 2015-10-23 at 9 25 09 am

screen shot 2015-10-23 at 9 25 18 am

PR #1839, opened 2015-10-26
Registration redesign tweak

aspidites:
In reviewing #1835, I made a few comments which I didn't think needed to hold back that PR. This PR addresses those points. Namely:

- On the registration form, the placeholder's seemed redundant. According to @lemniscate613 

## Before
2015-10-26-093049_1148x1056_scrot

## After
2015-10-26-093117_1148x1056_scrot

- On saved searches, despite each job listing having its own link, there was no hover effects to indicate that, which to me felt like clicking anywhere in the table would go to the same link

## Before
2015-10-26-094914_1052x1056_scrot

## After
2015-10-26-095046_1052x1056_scrot

There isn't a ticket for this as it very well be that no one else finds these points troublesome.

### DIFF
--- a/registration/forms.py
+++ b/registration/forms.py
@@ -27,7 +27,6 @@ class CustomSetPasswordForm(SetPasswordForm):
                                               'Password is required.'},
                                   label=('Password'), required=True,
                                   widget=forms.PasswordInput(attrs={
-                                      'placeholder': _('Password'),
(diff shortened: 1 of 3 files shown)


lemniscate613:
:+1: 

Point #1 jives with the principle of not forcing the user to process unnecessary information.

PR #1843, opened 2015-10-27
[PD-1895]  React build system complete with unit tests

darrint:
This is reworked from last week based on feedback.

# What's new

* Unit tests based on the jasmine framework. There are, as usual, 4000 node.js test frameworks to choose from. Once @aspidites figured out how to get ES6 modules working reliably I was able to make this work. Jasmine seems to be well supported and a complete solution out of the box, so I went with it.
* UI Business logic implemented in idiomatic ES-latest JS. It should read very easily now.

# Testing

```
cd gulp/
npm install
node_modules/.bin/gulp build
```

go to https://secure.my.jobs/reports/view/dynamicoverview

To tweak around with the js files, run:
nodejs node_modules/gulp/bin/gulp.js watch

* reporting.js will quickly and automatically rebuild.
* unit tests will run

# Some code standards I'm following:

* As much as possible write idiomatic ES-latest Javascript.
  * I even adopted ES7 async/await and used it where deferred processing was needed. (util/spec has an exception)
  * ES6 classes with constructors constrain some of the JS madness.
* Don't this.state in react components. Be stateless. Make ordinary code do as much heavy lifting as possible.
* Dependency Injection. See the unit test for "sends an error action..."
  * ActionCreators makes api calls and does a lot with the results.
  * The logic for what to do with api results, including errors, is what we want to test, not the api itself.
  * Therefore we should separate out the api calls and "inject" them into ActionCreators. There are several ways to inject.
    * The best way is to make api a class, and pass an instance of that class into ActionCreator's constructor.
    * The second best way is to pass a function reference into ActionCreator's constructor. We've done that for console.error and store.dispatch.
    * Don't pass ActionCreator's constructor the ability to get what it needs. Pass it what it needs.
    * Don't have ActionCreator get what it needs on its own.
  * In unit tests mock/fake/stub the dependencies of the thing we are testing.
    * ActionCreator is fed functions and a fake api. Once we do that it's easy for us to test any part of the ActionCreator code that we want to reach.
    * It is unfortunate that I had to mock as much as I did. A better design might result in even less test setup. However, there is no magic module wrangling or anything like that needed to make the tests work. Just JS classes and objects.
* Redux is essentially acting as a database.
  * React is displaying the data.
  * React event handlers are generally one liners that call action creators.
  * Action creators dispatch actions into redux.
  * Redux handles actions one at a time, replacing old state with new state for each action.
* All our imports have to be relative. No matter how stupid this seems, I can't find a way around it. Thanks Node.js. :-1:


.gitignore
```diff
@@ -16,6 +16,10 @@ dev_*
 CACHE
 static/redirect.css
 
+# Gulp stuff
+gulp/node_modules/
+static/bundle/
```
aspidites: Will this mess with how we deploy static files?

gulp/gulpfile.js
```diff
@@ -0,0 +1,84 @@
+require('babel/register');
```
aspidites: :+1:  * :100:

gulp/gulpfile.js
```diff
@@ -0,0 +1,84 @@
...
+// Splitting vendor libs into a separate bundle improves rebuild time from 8
+// seconds to <500ms.
+gulp.task('vendor', function() {
+    return browserify([], { debug: true, list: true, })
+    .require(vendor_libs)
+    .bundle()
+    .on('error', function(error, meta) {
+        util.log("Browserify error:", error.toString());
+    })
+    .pipe(source('vendor.js'))
```
aspidites: This series of pipes seem to be duplicated in package as well. Any way to encapsulate them and remove duplication?

gulp/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "de-build",
+  "version": "0.0.0",
```
aspidites: `version` and `description`, please. I'm a stickler for documentation.

gulp/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
```
aspidites: Author and license. DirectEmployers Association and whatever seems prominent in the rest of the code base for license.

gulp/src/util/freeze.js
```diff
@@ -0,0 +1,23 @@
+// Copied from public domain:
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze
+
+// To make obj fully immutable, freeze each object in obj.
+// To do so, we use this function.
+export function deepFreeze(obj) {
```
aspidites: <trolling>
You mean you didn't es-latest-ify the port of deepFreeze?
</trolling>

Immutable data is always good. :+1:

gulp/src/reporting/actions.js
```diff
@@ -0,0 +1,136 @@
...
+            const rts = await this.api.getReportingTypes();
+            this.dispatch(Actions.nextPage("reportingTypes", rts, "reportingTypes"));
+        } catch(e) {
+            this.errorLog(e);
+            this.dispatch(Actions.unexpectedError("Error while getting reporting types."));
+        }
+    }
+
+    async nextAfterReportingType(reportingTypeId) {
+        try {
```
aspidites: Wasn't there conversation a while back about not dispatching multiple actions in a single creator, or am I simply misremembering?

gulp/src/reporting/actions.js
```diff
@@ -0,0 +1,136 @@
...
+                ...state,
+                pageIndex: newPage,
+                loading: false,
+            };
+            newState[dataKey] = newData;
+            return Object.freeze(newState);
+        }
+        case 'SELECT': {
+            const {key, id} = action;
+            const newState = {...state};
```
aspidites: Seems like a false promise to declare something as `const` only to mutate it right after. I understand `const` only guarantees the object reference remains static, not its properties, but I think `let` here would communicate your intent more clearly.

gulp/src/reporting/api.js
```diff
@@ -0,0 +1,99 @@
+// based on:
+// http://stackoverflow.com/questions/24710503/how-do-i-post-urlencoded-form-data-with-http-in-angularjs
+function ancientSerialize(obj) {
```
aspidites: This seems to fall over on nested objects:

```javascript
function ancientSerialize(obj) {
  var str = [];
  for(var p in obj)
    str.push(encodeURIComponent(p) + "=" + encodeURIComponent(obj[p]));
  return str.join("&");
}

console.log(ancientSerialize({
  foo: "bar",a
  baz: {
    buz: "Feed"
  }
}));
```
Output: `foo=bar&baz=%5Bobject%20Object%5D`.

gulp/src/reporting/actions.js
darrint: The current guideline is "Don't dispatch more than one action creator from a react event handler."

gulp/src/reporting/api.js
darrint: I can disallow nested objects.

.gitignore
darrint: Yes. @mklauber and @JLMcLaugh are working on it on another ticket.

gulp/src/util/freeze.js
darrint: Notice the `export` keyword. Case closed.

gulp/gulpfile.js
darrint: I'd like to hold off on that a little longer. Once @lemniscate613 gets his admin work merged there will be a good opportunity to factor this better.

gulp/package.json
darrint: I can fix the description. With version I'm trying to express "not versioned" without breaking any tools.

darrint: Will put 0.1.0

gulp/src/util/freeze.js
aspidites: But arrow functions...court back in session.

aspidites:
@darrint, talked to @fezick  and @JLMcLaugh. We'll need to restrict access to the new system to staff only before we can merge this. There's already a restrict_to_staff decorator in myreports/decorators if you want to slap that around your views.

tdphillips:
Will there be an addition to the setup doc once this is merged?

darrint:
That can be done.

tdphillips:
The testing instructions seem to conflict with the instructions in said docs already in that you can't set this up for testing if they have been followed.

gulp/src/reporting/bootstrap.js
```diff
@@ -0,0 +1,20 @@
+import {polyfill as es6PromisePolyfill} from 'es6-promise';
+import 'fetch-polyfill';
+
+// Put as many polyfills here as we can.
+//
+// ES5-shim/sham and babel/polyfill probably have to be part of the bundling
+// environment.
+
+export function bootstrap() {
```
koonsies: Is this something we named? If so, I'm a little leery of naming the same name as the HTML/CSS framework we use. Not that it should be confusing under close scrutiny, but it can cause some double takes. Can it be named polyfill.js, perhaps?

darrint:
@tdphillips What problem did you see in testing?

gulp/src/reporting/bootstrap.js
darrint: I can rename it if someone has a good suggestion.

tdphillips:
Namely "npm install" didn't install gulp. When I installed it manually, it was in a location our project didn't care for.

gulp/src/reporting/bootstrap.js
JLMcLaugh: I agree bootstrap might be confusing, even though it's accurate. polyfill probably has more chance of collision though. import_polyfill, run_polyfill, bootstrap_polyfill?

PR #1844, opened 2015-10-27
[PD-1885] Created management command for converting company users to PRM Users.

aspidites:
# Description
~~This scans the current company user table and assigns all entries in that table the "PRM User" role. As a management command, it isn't  run automatically, which means we can redefine what such a user is before running it in production. A SQL variant of this will be developed by @haliscooper in the future which does almost the same thing, but I wanted to get this into QC so that people could continue to...QC things.~~
This has been promoted to ticket status. That is, it's not just for QC, it *is* what makes the "temporary" roles referred to in the linked ticket. That said, "temporary" is a misnomer, as there is nothing temporary about the role. What this accomplishes is creating a `PRM User` role to be assigned to all current `CompanyUser`s who are not also assigned the `Admin` role. The end effect is that when the new system is enabled (as it is in QC and Staging), anyone who could access PRM before can do so again, assuming the involved companies, which this ticket also takes care of.

# Output
[redacted, as the script is being modified]


lemniscate613:
:+1:

aspidites:
Update: I can haz moar fast?:
2015-10-27-165926_849x69_scrot

lemniscate613:
Fast?

:+1:

aspidites:
Concerns were raised during the meeting about this being a management command as opposed to a migration.  A few points:
- I was expressly advised not to make this a migration
- It's only a management command because that was the easiest way to create a script that didn't need to either be imported within  console and have it's functions invoked manually or require django initialization which I have yet to do consistently and without error in the past
- Even if management commands can be repeated, there is at least one instance of a management command that we only ever run once - `syncdb` -- database changes tend to be invoked via the `migrate` command
- I'm not convinced that we won't want to do this again in the future
- It's easy enough to open a PR which removes this command later on down the road

All things considered, I'd prefer this was a data migration as The content isn't too dissimilar to the data migration we used to create admin users. We'd also then be able to automate dependency checking, rather than requiring intervention from the user if they forget a step.

aspidites:
@lemniscate613, your previous :+1:  is probably invalid now, considering the conversations and changes that went into this after you cast that vote. Mind checking this again once the DO NOT MERGE is removed?

aspidites:
@lemniscate613  scratch that. I'm opening a new PR as I've done a lot of refactoring and don't want any current commentary to mislead anyone.

PR #1851, opened 2015-10-29
Properly wrapped my decorator with @method_decorator.

aspidites:
Without this, reports is broken on QC and Staging.


aspidites:
@koonsies

PR #1853, opened 2015-10-29
[PD-1905] Fix number of contacts displayed in PRM

aspidites:
The issue was that doing somethign akin to `instances = Model.objects.values().distinct()`, then later evaluating it with something like `for instance in instances`, you get different results. Here, `.count()` after first initializing the query set would display a fewer number of results than runing `len()` on that same result. 

~~My solution (which I'm not at all happy with) is to convert the query set into a set of tuples, then back into a list of dictionaries.~~

Functional programming for the win. I went with a fold  over the query set (`reduce`), which is both more elegant and efficient than what I was doing before.

Before:
2015-10-29-113404_1920x1080_scrot

After:
2015-10-29-141923_1916x1056_scrot


koonsies:
I like what you did with reduce. Pretty clever.. and something I'd not have thought to do.

 In review of the issue, the root cause of the distinct() call not working is that distinct in MySql works on a row level. On that particular model, there is an implicit order_by on the contactrecord.date_time field, causing it to be included in the selected row (thus that row not being found as distinct).

Instead of this, the following change will fix distinct()

`
        all_contacts = self.values(
            'partner__name', 'partner', 'contact__name',
            'contact_email').distinct().order_by('partner__name')
`

aspidites:
Very nice. Going with your solution. May even be able to remove another hack that I was using before....

PR #1854, opened 2015-10-29
[PD-1906] Admin role for new companies

aspidites:
A previous PR added the admin role to all existing companies (but didn't assign them). This one makes sure that newly created companies have the admin role.

@JLMcLaugh , should this be a hotfix to prevent the week between where admin roles are automatically created and new companies are made without this being in place?
Alternatively, we can fake that migration backward and re-run it to catch up any stragglers (faking all the way forward after that).


aspidites:
I think we determined that we would just rerun the management command to avoid having to make this a hotfix.

PR #1855, opened 2015-10-30
[PD-1910] Address issues with JSON parser in PRM

koonsies:
https://directemployersfoundation.atlassian.net/browse/PD-1910

Modifying contacts in a contact record was causing a serializer error. This is due to changes to foreign key relationships (under the current model) list the entire related model in the delta object. Code would change this relationship to a more readable format (<classname>: pk: <pk> desc: <ob. unicode>).

Feel free to critique the string format or any other issues with this change.

### DIFF
--- a/mypartners/helpers.py
+++ b/mypartners/helpers.py
@@ -153,7 +153,8 @@ def get_form_delta(form):
 
         if initial or new:
             # do not return complete models for foreign keys, only PK
-            def convert_object(x):  x.pk if isinstance(x, Model) else x
+            def convert_object(x):
+                return x.pk if isinstance(x, Model) else x
             delta[field] = {'initial': convert_object(initial), 'new': convert_object(new)}
 
     return delta
--- a/mypartners/helpers.py
+++ b/mypartners/helpers.py
@@ -28,13 +28,6 @@ from mypartners.models import (Contact, ContactLogEntry, CONTACT_TYPE_CHOICES,
 from mypartners.widgets import MultipleFileField
 
 
-class ModelSafeJSONEncoder(DjangoJSONEncoder):
-    def default(self, o):
-        if isinstance(o, Model):
-            return "%s: pk:%s desc: %s" % (o.__class__.__name__, o.pk, str(o))
-        return super(ModelSafeJSONEncoder, self).default(o)
-
-
 def prm_worthy(request):
     """
     Makes sure the User is worthy enough to use PRM.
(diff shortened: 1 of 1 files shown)


aspidites:
If this is the solution, I'd rather not have commented out code lingering around. We are using git after all, so rolling back is easy enough.

koonsies:
Oh, good call. I didn't even notice I had left that in.

aspidites:
Would a better fix be to ensure that the get_form_delta function returns ids instead of instances? If I inspect the delta on error, I have:
```javascript
{
    'contact': {
        'new': <Contact: Jen Bernhardt>,
        'initial': 2191L
    }
}
```

aspidites:
Also:
```python
from djagno.core.serializers import serialize
model = Contact.objects.get()
print serialize('json', [model])
```
Only point being it's possible. Only does a shallow serialization though. I forget what my reason was for not using that directly.

koonsies:
Updated code to move the model handling to the get_form_delta function. Entire model objects should no longer be passed back within the returned dict.

mypartners/helpers.py
```diff
@@ -152,7 +152,9 @@ def get_form_delta(form):
             new = [fd.name for fd in new or [] if fd]
 
         if initial or new:
-            delta[field] = {'initial': initial, 'new': new}
+            # do not return complete models for foreign keys, only PK
+            def convert_object(x):  x.pk if isinstance(x, Model) else x
```
aspidites: @koonsies , my suspicions were correct:
```python
>>> def convert_dict(x): x.keys() if isinstance(x, dict) else x
... 
>>> convert_dict(4)
>>> q = convert_dict(4)
>>> q
>>> q = convert_dict({'foo': 'bar'})
>>> q
>>> def convert_dict(x): x.keys() if isinstance(x, dict) else x
... 
KeyboardInterrupt
>>> def convert_dict(x): return x.keys() if isinstance(x, dict) else x
... 
>>> q = convert_dict({'foo': 'bar'})
>>> q
['foo']
>>> 
```
Basically, as you're defining a function and not returning anything, any assignment from an invocation of it will result in a None. The solution is either to convert this back into a lambda (in which case the return keyword is erroneous), or add the return keyword. That said, if you do add return, I'd prefer that this was defined on two lines (PEP-8).

aspidites:
Comments  aside, I do appreciate the new approach more. :+1:  for that.

PR #1859, opened 2015-11-02
Master to QC

mklauber:


seo/tests/test_views.py
```diff
@@ -2865,6 +2865,72 @@ def test_invalid_moc_404(self):
                                HTTP_HOST=self.site.domain)
         self.assertEqual(resp.status_code, 404)
 
+class FilterTestCase404(DirectSEOTestCase):
```
tdphillips: This is the only serious conflict I see. There are two classes with this name here and in QC. Both of these subclass DirectSEOTestCase while one of the ones in QC subclasses a different test case.

registration/views.py
```diff
@@ -150,7 +150,6 @@ def merge_accounts(request, activation_key):
         status.approved_by = existing_user
         status.save()
 
-
```
mklauber: This one was me.  I made a pep8 change on the staging branch to trigger a new build.

aspidites:
I'm a little concerned with the fact that changes which aren't changes are being picked up, in addition to the bidirectional merging we seem to be doing (QC -> Staging -> Master -> QC).

mklauber:
Typically, we aren't doing master->QC.  I dropped this in right now, just to see what changes we had, and was rather surprised at them.

mypartners/views.py
```diff
@@ -91,7 +91,7 @@ def prm(request):
         'company': company,
         'user': request.user,
         'partner_ct': ContentType.objects.get_for_model(Partner).id,
-        'view_name': 'PRM',
+        'view_name': 'PRM'
```
mklauber: Looks like this was a change that David made, that was never merged back to QC.  https://github.com/DirectEmployers/MyJobs/commit/0507be7aa70345f8fd83fa4af032edb4479d0fbf

myreports/views.py
```diff
@@ -135,7 +135,7 @@ class ReportView(View):
     app = 'mypartners'
     model = 'contactrecord'
 
-    @requires(PRM, missing_activity, missing_access)
+    @method_decorator(requires(PRM, missing_activity, missing_access))
```
mklauber: I think the cause of this (despite the fact that the code is in fact present on QC) is that the code change was merged to staging from a different branch (https://github.com/DirectEmployers/MyJobs/pull/1852) than the code merged to quality control (https://github.com/DirectEmployers/MyJobs/pull/1851).  Not sure why this confused git other than the commit tree cannot be resolved simply?

seo/tests/test_views.py
mklauber: The original cause of this looks like it was my fault.  I think it was probably a merge conflict that was improperly resolved.

wsgi.py
```diff
@@ -4,7 +4,7 @@
 newrelic.agent.initialize('/home/web/MyJobs/MyJobs-urls/newrelic.ini')
 
 
-os.environ['DJANGO_SETTINGS_MODULE'] = 'MyJobs-urls.settings'
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings'
```
mklauber: I think the cause of this (despite the fact that the code is in fact present on QC) is that the code change was merged to staging from a different branch (#1847) than the code merged to quality control (#1845). Not sure why this confused git other than the commit tree cannot be resolved simply?

myreports/views.py
aspidites: Hmm...I don't think this has ever caused issues before. I'd be a bit sad if git(hub) wasn't able to see that two files, regardless of which commit they were changed in, were similar.

wsgi.py
aspidites: Same comment as above.

aspidites:
I'm fine with this being merged even though it makes me feel dirty. I wonder if some experimentation on a test repo might uncover exactly what pattern of behavior causes these results? Should be easier to inspect than one with a ton of history like this. 

I don't think doing so should be a priority, however, as you've got way more important things on your plate, but it could be an interesting experiment none-the-less.

mklauber:
test this please

PR #1861, opened 2015-11-02
Stop using seo/secure middleware for redirect [PD-1560]

tdphillips:
Test on QC:
2015-11-02-095942_722x1215_scrot

Test on redirect-middleware (not sure why this is a 301 but it resolves in a browser):
2015-11-02-101501_744x1208_scrot

Once any issues are dealt with and we decide on the final change set, I'll open PRs for staging/master.

Tests  pass with the exception of those involving imports.


aspidites:
:+1: .

PR #1868, opened 2015-11-03
[PD-1913, PD-1914] Deprecate the use of CompanyUser

aspidites:
# Note
I am not intending here to completely remove the CompanyUser model, but rather make its inclusion redundant.

# Description
With the introduction of roles, there is no need for a company user table, as a user can be tied to a company merely by seeing if that user is assigned a role which belongs to the company in question. That said, as the activities system isn't deployed, it would be unwise to simply remove any use of the company user model. Instead, what this PR does is ensure that when debugging is enabled, roles are checked and when debugging is disabled, the original behavior which relies upon company users is preserved.

# Branch
This is based off of #1854. The core changes are [here](https://github.com/DirectEmployers/MyJobs/compare/pd-1906...pd-1913?expand=1).

# Caveats
- There are a few things done with CompanyUser that I am intentionally ignoring for the moment, such as post save hooks, as my primary concern with these edits is ensuring that functionality can be properly reached without first having to create a company user.
- Per @JLMcLaugh, I'm ignoring anything to do with selenium/functional tests.
-  I intentionally ignored all of postajobs as it is tightly coupled to company user.
- MyDashboard admin is only to expose company users to django admin, which doesn't make sense to write alternate behavior for

# Progress
This is probably going to take a while, as there are 13 files which reference company user in some way and I need to write regression tests to ensure both behaviors. As such, I'll be updating this PR with my progress as I tackle these issues.

## Files to update
Each will be marked complete once occurrences are resolved and a regression test is written.
- [x] seo/models.py (7 occurrences)
- [x] universal/helpers.py (1 occurrence)
- [x] myjobs/templatetags/common_tags.py (1 occurrence)
- [x] myemails/models.py (2 occurrences)
- [x] mydashboard/views.py (2 occurrences)

### DIFF
--- a/mydashboard/views.py
+++ b/mydashboard/views.py
@@ -75,14 +75,17 @@ def dashboard(request, template="mydashboard/mydashboard.html",
 
     authorized_microsites, buids = get_company_microsites(company)
 
-    admins = CompanyUser.objects.filter(company=company.id)
+    # roles are only enabled during development
+    if settings.DEBUG:
+        admins = User.objects.filter(roles__company=company)
+    else:
+        admins = User.objects.filter(company=company)
 
-    # Removes main user from admin list to display other admins
-    admins = admins.exclude(user=request.user)
+    admins = admins.exclude(pk=request.user.pk)
     requested_microsite = request.REQUEST.get('microsite', '')
-    requested_date_button = request.REQUEST.get('date_button', False)    
-    candidates_page = request.REQUEST.get('page', 1)    
-          
+    requested_date_button = request.REQUEST.get('date_button', False)
+    candidates_page = request.REQUEST.get('page', 1)
+
     # the url value for 'All' in the select box is company name
     # which then gets replaced with all microsite urls for that company
     site_name = ''
@@ -447,7 +450,7 @@ def export_csv(request, candidates, models_excluded=[], fields_excluded=[]):
                 num += 1
             else:
                 break
-        
+
         instance = getattr(unit, unit.content_type.name.replace(" ", ""))
         fields = retrieve_fields(instance)
 
--- a/myjobs/templatetags/common_tags.py
+++ b/myjobs/templatetags/common_tags.py
@@ -90,7 +90,10 @@ def get_company_name(user):
 
     # Only return companies for which the user is a company user
     try:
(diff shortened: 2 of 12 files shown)


myjobs/templatetags/common_tags.py
```diff
@@ -90,7 +90,10 @@ def get_company_name(user):
 
     # Only return companies for which the user is a company user
     try:
-        return user.company_set.filter(companyuser__user=user)
+        if settings.DEBUG:
+            return Company.objects.filter(role__user=user)
+        else:
+            return user.company_set.filter(companyuser__user=user)
```
koonsies: Mm.. This code. Filter the list of companies with which this user belongs to only return those companies with which this user belongs.

koonsies:
:+1:  sleek, stylish, and sexy

mydashboard/views.py
```diff
@@ -75,14 +75,17 @@ def dashboard(request, template="mydashboard/mydashboard.html",
 
     authorized_microsites, buids = get_company_microsites(company)
 
-    admins = CompanyUser.objects.filter(company=company.id)
+    # roles are only enabled during development
+    if settings.DEBUG:
+        admins = User.objects.filter(roles__company=company)
+    else:
+        admins = User.objects.filter(company=company)
```
tdphillips: myemails.models has some code similar to this but retains the use of CompanyUser. Should we prefer one over the other?

aspidites: I just wrapped what was already there. As for switching from what filter was there before, it was redundant. The original, as @koonsies noted, was the equivalent of saying "Give me a list of companies associated with this user restricted by only those associated with this user".

myjobs/templatetags/common_tags.py
```diff
@@ -90,7 +90,10 @@ def get_company_name(user):
 
     # Only return companies for which the user is a company user
     try:
-        return user.company_set.filter(companyuser__user=user)
+        if settings.DEBUG:
+            return Company.objects.filter(role__user=user)
+        else:
+            return user.comapny_set.all()
```
tdphillips: comapny

koonsies: Looks like this is corrected in his followup PR, #1874

PR #1869, opened 2015-11-03
[PD-1757] outreach inbox management

koonsies:
## Goal

* Create back end handling for management of non-user outreach inboxes.
* Develop basic front end for modification
 * Can be revised/updated later, or used as-is

## To Test

* Check out branch
* Log in to a user with PRM access
* Head to /prm/view/nonuseroutreach/inboxes

## Notice

The format of the front end code is a bit off. There is a tag with "@my.jobs" at the end of the input textbox. As is, the model expects a full email to be entered into that textbox. This is being addressed in a following ticket. Until then, testing should be done with the current model design in mind.


darrint:
Reopened this to trigger a Jenkins retry.

mypartners/tests/test_views.py
```diff
@@ -1672,3 +1672,30 @@ def test_location_delete(self):
         # reload contact from DB, as it is holds "expired" data
         reload_contact = Contact.objects.get(pk=self.contact.pk)
         self.assertNotEqual(original_action_time.date(), reload_contact.last_action_time.date())
+
+
+class OutreachViewTests(MyPartnersTestCase):
```
aspidites: Do we only care that the end point is reachable, or is there something more significant to assert? Or is it that whatever we could assert we are hoping would better be tested with selenium or the like?

mypartners/views.py
```diff
@@ -1322,6 +1323,44 @@ def process_email(request):
 
 @requires(PRM, missing_activity, missing_access)
 @has_access('prm')
+def manage_outreach_inboxes(request):
```
aspidites: May be a good reason to reach for a class-based view here, considering the explicit checks for GET and POST. The only place we've used them consistently is postajob. There's one class based view in myreports as well (others not being so either because I didn't write them or the overhead of introducing the class wasn't worth it).

I don't feel strongly one way or another, but if it cleans things up, I'd be all for it.

mypartners/tests/test_views.py
koonsies: I was torn on that. I had considered putting in an assertContains to check the response for certain characteristics.

templates/mypartners/nonuseroutreach/outreach_inbox_management.html
```diff
@@ -0,0 +1,85 @@
...
+    @media all and (max-width: 500px) {
+        #inbox-form input[type='checkbox'] {
+            float:right;
+            height:auto;
+        }
+    }
+</style>
+{% endblock %}
+
+{% block content %}
```
aspidites: Caveat: I *hate* form_error_highlight.

Should we be using form_error_highlight to keep error handling consistent?

koonsies: I am looking into reproducing the same style of error message. I can't directly use form_error_highlight because it is specifically used for the standard bootstrap form. Since my layout is different and consists of a number of forms in a formset, I'll have to write my own version.

mypartners/views.py
koonsies: That's a good point. For this code, I don't see much benefit from changing it at this point, but good note on the get/post handling.

templates/mypartners/nonuseroutreach/outreach_inbox_management.html
koonsies: Updated to match what is seen on other forms. Not exact, but close.

mypartners/tests/test_views.py
koonsies: Added a check on response to verify the page being returned is the page we're looking for.

templates/mypartners/nonuseroutreach/outreach_inbox_management.html
aspidites: Following the "spirit" of form-error-highlight without actually using it is a great compromise.

aspidites:
Finally got around to running this locally.

Comments:
- The New Outreach Email Inbox widget probably needs the "Save" button on it. It's redundant, but how we generally handle these things (see Partner Details for an example).
- `Save` Should probably be `Add` to be consistent.
- It feels odd that the existing inbox doesn't have a header (The div that holds the "New Outreach Email Inbox" header). 
- The delete checkbox  (which I understand is supposed to be used in conjunction with the Save button) feels inconsistent. We've typically used a button (see PRM) with a modal in this case
- Editing an existing email inbox prefix inline is convenient, but inconsistent with the rest off PRM, where we have an explicit Edit button. If we decide to keep the inline editing, I think it should show as read only then change to an input once clicked (see the "per page" button in PRM).
- Typically, we've marked Navigation, whereas now it's just the empty ribbon above the Save button.
- @fezick , did we decide to move navigation to the top of the side bar, or is that something waiting on the site redesign
- This probably needs to be wrapped in @restrict_to_staff

General notes (that you don't really have control over)
- I hate how everything in this project is named. "Non-User Outreach Inbox Management" is a mouthful, and I'm not sure if, as a user, I would know what that meant without looking it up every time.

koonsies:
* The New Outreach Email Inbox widget probably needs the "Save" button on it. It's redundant, but how we generally handle these things (see Partner Details for an example).
 * I can understand that. I will add it.
* Save Should probably be Add to be consistent.
 * If you're talking about the "Save" button added above, I agree. Otherwise, I don't think it reflects what is actually occurring, being as it could be saving a change to an existing email.
* It feels odd that the existing inbox doesn't have a header (The div that holds the "New Outreach Email Inbox" header).
 * I can try it out with an empty header. Other than that, any idea on what to put in there?
* The delete checkbox (which I understand is supposed to be used in conjunction with the Save button) feels inconsistent. We've typically used a button (see PRM) with a modal in this case
 * I don't believe we do formsets in PRM anywhere else. Using a modal with the current design scheme doesn't feel right, as it would force a submission of the entire formset to do an individual delete. 
* Editing an existing email inbox prefix inline is convenient, but inconsistent with the rest off PRM, where we have an explicit Edit button. If we decide to keep the inline editing, I think it should show as read only then change to an input once clicked (see the "per page" button in PRM).
 * I will look into this. Where is the "per page" button?
* This probably needs to be wrapped in @restrict_to_staff
 * Will this be a staff-related use case? Or perhaps activity related? If I'm understanding correctly, a user granted access to PRM may not necessarily have the staff flag.

aspidites:
> I can try it out with an empty header. Other than that, any idea on what to put in there?
"Navigation". It's what we use everywhere else, though that might be because a new page is opened.

> I will look into this. Where is the "per page" button?

Clicking the 10 changes the input:
2015-11-06-163714_637x159_scrot
2015-11-06-163707_124x36_scrot

> Will this be a staff-related use case? Or perhaps activity related? If I'm understanding correctly, a user granted access to PRM may not necessarily have the staff flag.

No. This feature isn't ready, as such it should only be viewable by staff members for the sake of QCing it. 

```python
from myreports.decorators import restrict_to_staff

@restrict_to_staff()
def view(request):
    pass
```

This way, a user just putting in the URL won't happen to see a feature they can't use. On release, we'd simply remove the decorator.
```

aspidites:
> I don't believe we do formsets in PRM anywhere else. Using a modal with the current design scheme doesn't feel right, as it would force a submission of the entire formset to do an individual delete.

We don't in PRM, but we do in postajob.

koonsies:
In review of this ticket, I made the following changes:

* Add header to all inbox forms, labeling current inboxes as "Existing Outreach Inbox" and the new slot as "New Outreach Inbox"
* Add "Navigation" text to side bar
* Add requires_staff decorator to view

The remainder of the changes, while valid in their own rite, may be out of scope of this ticket. This isn't necessarily meant to be final design, but rather a good cut at the functionality/layout. It can later be cleaned up if it is deemed necessary.

Do  note that the changes are not going to be in this ticket, but the PR for 1912

tdphillips:
If the delete checkboxes are staying, I think it would be nice if they were labels to allow clicking on the word.

PR #1870, opened 2015-11-04
Fix saved search failures [PD-1915]

tdphillips:
This could use a good review. I'm worried we're treating symptoms rather than a problem but I haven't detected one.

Most of the modifications to jobs.html revolve around turning it into one big multi-line format string (adding query string support to feeds, doubling up on all single `%` characters).

All tests pass.


myjobs/tests/helpers.py
```diff
@@ -38,7 +38,9 @@ def _inner(url, *args, **kwargs):
         elif 'mcdonalds/careers/' in url or \
                 url.endswith('?location=chicago&q=nurse'):
             file_ = 'careers.html'
-        elif 'www.my.jobs/jobs' in url or 'www.my.jobs/search' in url:
+        elif any(domain+path in url
```
tdphillips: mypartners uses jobs.jobs for some reason. It was add another clause here or modify most mypartners tests.

mysearches/helpers.py
```diff
@@ -93,7 +92,7 @@ def validate_dotjobs_url(search_url, user):
         return None, None
 
     try:
-        page = urllib.urlopen(search_url).read()
+        page = urllib2.urlopen(search_url).read()
```
tdphillips: The patcher has always patched urllib2. This was changed from urllib2 to urllib in 65afb01a2956858329a5bbc127675edefe0cb225 but the patcher wasn't updated.

aspidites:
:+1: . Not sure where I'd begin to look for a problem rather than a symptom, though.

PR #1872, opened 2015-11-06
[PD-1912] modify outreach email handling

koonsies:
## The Issue

When originally designed, the Outreach Email Address model was created to store an email address (feelgoodinc@gmail.com). Instead, we need to only store a prefix (feelgoodinc) and append '@my.jobs' or whichever domain we decide on to the end of that prefix. 

## Solution

This PR covers the migration to fix the model as well as the custom handling to ensure values entered into the model field do not contain special characters.

## Steps

* Change email field to CharField on Outreach Email Address model
* Add custom regex validator to ensure we are receiving usable input
* Update forms to display proper guidance for inputs

## Differences

This PR follows PD-1757. To view the changes specific to this PR, compare here: http://git.io/v8sBf

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -10,6 +10,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.files.storage import default_storage
 from django.core.urlresolvers import reverse
 from django.core.exceptions import ValidationError
+from django.core.validators import RegexValidator
 from django.db import models
 from django.db.models.signals import pre_delete, pre_save
 from django.dispatch import receiver
@@ -39,6 +40,8 @@ ACTIVITY_TYPES = {
     4: 'sent',
 }
 
+email_validator_local = RegexValidator('^([0-9a-zA-Z]+[\.\_]?[0-9a-zA-Z]+)*$',
+                                       'Invalid email identifier. Allowed characters [A-Z] [0-9] [ , . ]')
 
 class Status(models.Model):
     """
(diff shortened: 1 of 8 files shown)


mypartners/models.py
```diff
@@ -39,6 +40,8 @@
     4: 'sent',
 }
 
+email_validator_local = RegexValidator('^([0-9a-zA-Z]+[\.\_]?[0-9a-zA-Z]+)*$',
```
aspidites: This rejects many valid email addresses*. Given the amount of stuff that we don't validate (addresses, phone numbers, email addresses for communication records, the invite system, etc), do we really need to do so here?

* http://stackoverflow.com/questions/2049502/what-characters-are-allowed-in-email-address

tdphillips: If the intention is to make it so they can only enter the local part of an address, disallowing "@" should be good enough.

koonsies: @JLMcLaugh @haliscooper Do either of you have input on this? Essentially, the form will be asking for the members to input the "username" portion of the email (the portion before the @). I currently restrict it to [0-9] [A-Z] [ _ . ] .. With a caveat being that you cannot have two of the underscores or periods in a row. I may add hyphen in there as well. As @aspidites points out, technically the username field is pretty wide open in terms of the characters it can have in it.. The question is, what characters do we WANT in it.

JLMcLaugh: You could add a temporary domain and use an out of the box email validator. django.core.validators has an email validator.

koonsies: Good call. Investigating the potential solutions, I found that EmailValidator used a regex specifically for username which I could import and use without issue.

koonsies:
Reopen to merge/complete working form. UX updates will follow on a future PR

PR #1875, opened 2015-11-09
Remove invalid options from ActivationProfile.sent [PD-1548]

tdphillips:
While looking into upgrading Django, I ran into a wall. SOP is to either delete the migrations directory or rename it to south_migrations and then let Django recreate its own initial migrations. It will see that nothing has been changed when running migrations and mark them as already applied.

That's all well and good but ActivationProfile.sent has both auto_now_add and a default value. This causes Django's migration generation to fail because that's invalid.

If we're ever going to upgrade, it looks like we'll need to fix it in South first.

All tests pass.

### DIFF
--- a/registration/models.py
+++ b/registration/models.py
@@ -67,7 +67,7 @@ class ActivationProfile(models.Model):
     user = models.ForeignKey('myjobs.User', verbose_name="user")
     activation_key = models.CharField(_('activation_key'), max_length=40)
     email = models.EmailField(max_length=255)
-    sent = models.DateTimeField(auto_now_add=True, default=datetime_now)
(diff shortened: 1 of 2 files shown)


registration/models.py
```diff
@@ -67,7 +67,7 @@ class ActivationProfile(models.Model):
     user = models.ForeignKey('myjobs.User', verbose_name="user")
     activation_key = models.CharField(_('activation_key'), max_length=40)
     email = models.EmailField(max_length=255)
-    sent = models.DateTimeField(auto_now_add=True, default=datetime_now)
+    sent = models.DateTimeField(default=datetime_now, editable=False)
```
JLMcLaugh: Isn't this a method call?

PR #1877, opened 2015-11-10
[PD-1837, PD-1923] Role interface and role API

lemniscate613:
I addressed too much in previous PRs. This one is solely about:

* Interface for managing roles (PD-1837)
* Associated APIs (PD-1923)

# Most Relevant Files

Thank you @aspidites. I'll try to point this out in the future.

[gulp/gulpfile.js](https://github.com/DirectEmployers/MyJobs/pull/1877/files#diff-ab729026c3d4a4364ad28d329d544093)
[gulp/src/manageusers/manageusers.js](https://github.com/DirectEmployers/MyJobs/pull/1877/files#diff-5b20481ecac9619deecc75649b273934)
[myjobs/views.py](https://github.com/DirectEmployers/MyJobs/pull/1877/files#diff-5419a48adf710fe19aaa02f5ffaf9eee)

# What's New

* Incorporated David's card styles
* CRUD for roles
* Different multiselects (they are React components, not the previous FSM widget)
* Various confirmations to user (e.g. a role must contain an activity, confirm deletion of role, etc.)
* Nice API error messages
* Modularizing more (e.g. cookie.js util by @darrint)

# How to Test

Make sure gulp is installed and:
```
cd /gulp/
npm install
gulp
```
Finally, simply CRUD roles in the browser (JS and more developed Python tests for future PRs): localhost:PORT/manage-users/

# Future Work

* Hide the navigation link to /manage-users/ based on... what criteria?
* Most of the interface for user editing (PD-1836) is done, but that will be in a separate PR.
* All of this is built in Boostrap 3! Kinks will be worked out (tablet/mobile) in future PRs as well as continuing to clean up templates like base-new.html.
* Possibly refactor with Flux (i.e. Actions and the Dispatchers). My code is simple and works, but it can be more elegant and testable.

# You Are Awesome

It's true.


.gitignore
```diff
@@ -16,7 +16,11 @@ dev_*
 CACHE
 static/redirect.css
 
+<<<<<<< HEAD
+Gulp stuff
+=======
```
koonsies: Merge error here

lemniscate613: Resolved. Thank you!

gulp/gulpfile.js
```diff
@@ -22,6 +24,8 @@ var vendor_libs = [
 
 var dest = '../static/bundle';
 
+var strip_debug = true;
+
```
darrint: per @aspidites, I think we agreed not to do this.

gulp/src/manageusers/manageusers.js
```diff
@@ -0,0 +1,778 @@
...
+
+        <div className="row">
+          <Content page={this.props.page} action={this.props.action} role_to_edit={this.props.role_to_edit} role_id={this.props.role_id} disappear_text={this.props.disappear_text}/>
+          <Menu />
+        </div>
+        <div className="clearfix"></div>
+      </div>
+    );
+  }
+});
```
koonsies: There is a lot of code here to create the header (which is essentially the same as mypartners/includes/prm_header.html)

Is there a reason we don't use the template include to bring in this header rather than creating it via an "include" statement on the template? As it stands, if we want to modify the header for all pages, we have to go to each .JS file and modify it.

koonsies: Or, in the spirit of React, make the title a reusable component. Then we have to make changes in, at most, two places when modifying the header.

gulp/gulpfile.js
lemniscate613: You're right -- we agreed not to leave logging in committed code. This node package strips any _just in case_.

If you want to leave them in while working locally, you can:
```
gulp watch-no-strip
```

gulp/src/manageusers/manageusers.js
```diff
@@ -0,0 +1,778 @@
...
+var OverviewPage = React.createClass({
+  render: function() {
+    return (
+      <div className="row">
+        <div className="col-xs-12 ">
+          <div className="wrapper-header">
+            <h2>Overview</h2>
+          </div>
+          <div className="product-card no-highlight">
+            <p>What should go here?</p>
```
koonsies: As this is an overview of the module itself, I'd think a brief blurb about the purpose and use of the module should belong here. Furthermore, there should be a way to get back to this overview after going into one of the modules (Roles, Activities). As it stands, you can't get back to Overview without refreshing your page.

gulp/src/manageusers/manageusers.js
```diff
@@ -0,0 +1,778 @@
...
+      <div>
+        <div className="row">
+          <div className="col-sm-12">
+            <h1><a href="/manage-users/" title="Back to Manage Users">DirectEmployers</a></h1>
+          </div>
+        </div>
+
+        <div className="row">
+          <div className="col-sm-12">
+            <div className="breadcrumbs">
```
koonsies: I'm assuming that this is because nothing currently links to /manage-users/, but breadcrumbs will need to eventually do something. Perhaps it can be determined by what page you are on? If you're on "Roles", it could link back to "Overview", if you're on "Overview", it links back to whatever page will link to this one.

gulp/gulpfile.js
aspidites: As I've said before, I get leery any time I see "just in case" mentioned in reference to keeping anything in code: [YAGNI](http://yagni.net/).

Further, consciously deciding not to do something but keeping this just-in-case raises two other concerning scenarios:

1. I know that its there, so as a lazy developer, I don't bother removing my debugs because I know they will be stripped anyway
2. I don't know (or remember) that it's there, so I add debug statements which disappear when building, which completely thwarts my debugging attempts. 

I would suggest that if this improves your workflow, you do as I (and others) do for tools that don't get used by everyone else on the python side -- install it without adding it to a build file. For instance, I use IPDB, but that's not included in requirements.txt. Similar case for a few profiling plugins and middleware that I have installed.

aspidites:
re: flux - is it at least safe to say that @darrint  has ruled out redux? As for flux, I suppose that's still on the table. If I have any objection to vanilla react, it's how state is handled, but given the environment we are in, it's not much of a hard complaint.

lemniscate613:
@aspidites Good observation. I put this as future work (in a different PR):

> "Possibly refactor with Flux (i.e. Actions and the Dispatchers). My code is simple and works, but it can be more elegant and testable."

aspidites:
Just a list of usability questions/concerns I had. I'll do my code critique inline.

- When are we going to implement history API/navigation? Clicking on "Roles", then navigation, I would expect to land back on the overview, but I don't. 
- It's a bit hard to compare roles at the moment. I have a "User" role in addition to a "PRM User" role (artifact of me testing this tagging stuff), but have to scroll vertically to review the list of activities in each role. 
- Interface not usable on mobile:
2015-11-12-111119_1920x1080_scrot
- Error message placement is a bit odd. I think we've been using a yellow/red banner at the top of the page elsewhere for system messages, and red text above the element for input-specific errors.
- having to click, scroll, then shift+click to select all activities (to add or remove) is cumbersome, and non-obvious (no help text) for individuals who don't interact with a computer very often (or perhaps only do so to use Facebook and the like).
- Is the multiselect meant to feel like FSM widget? How concerned are we with having a unified field?
- If I have multiple roles, I have to scroll to get to the "Add Role" button
- These views need to be wrapped. At the moment, without belonging to any roles, I can still access the overview page, rather than only being able to view it if I am assigned a role for the company being viewed which has the "read role" activity. While I have a ticket to wrap views, I'd greatly appreciate not having to go back and wrap brand new  views
- Was it decided that the URLs should not change? I thought you and @fezick  talked about ,despite this being a single-page app, the URL changing anyway. I don't care either way, but want to be sure my expectations line up with this ticket's.
- There seems to be some flickering when you click on the cancel button, as though things get rendered twice. Same thing for the delete button.

aspidites:
@lemniscate613, I saw your future work section, which is what prompted me to comment on redux :P

gulp/src/manageusers/manageusers.js
```diff
@@ -0,0 +1,778 @@
+import React from "react";
```
aspidites: No documentation. Would be nice to not simply have to infer which bits of the view this corresponds to.

lemniscate613:
@aspidites I intended to work on tablet/mobile in a future PRs ("Kinks will be worked out (tablet/mobile) in future PRs"), but I went ahead and played with it a second ago... Check this out. I have a solution I really like for displaying wide tables on mobile:

This:

screen shot 2015-11-13 at 3 22 29 pm

Becomes this:

screen shot 2015-11-13 at 3 17 52 pm

aspidites:
@lemniscate613 much better. Working on mobile later is one thing, a completely unusable experience is another. Your recent modifications restore that usability, so :+1: 

Also, for the instructions, you may want to add that you need to `npm install` to get dependencies, especially if gulp isn't already installed. In other words, the instructions become:

```
cd ./gup
npm install
gulp
```
(npm should already be installed, per the MyJobs installation guide.

myjobs/views.py
```diff
@@ -592,3 +595,336 @@ def topbar(request):
...
+def api_get_activities(request):
+    """
+    Retrieves all activities
+    """
+
+    activities = Activity.objects.all()
+    return HttpResponse(serializers.serialize("json", activities, fields=('name', 'description')))
+
+
+def api_get_roles(request):
```
aspidites: I receive an error from the JSON that is returned by this view function. That said, I think we are doing way too much work here. I can comment line by line, but I wonder if submitting  a diff and offering general comments would be more productive?

aspidites: Hmm... maybe I'll do both.

myjobs/views.py
```diff
@@ -592,3 +595,336 @@ def topbar(request):
...
+        response_data[role_id] = {}
+
+        response_data[role_id]['role'] = {}
+        response_data[role_id]['role']['id'] = role.id
+        response_data[role_id]['role']['name'] = role.name
+
+        response_data[role_id]['activities'] = {}
+        # This company has access to various apps by means of multiple app_access_id's
+        # Retrieve all activities with these app_access_id's
+        available_activities = Activity.objects.filter(app_access__in=company.app_access.all())
```
aspidites: These lines are way too long. In JS we fudge it as is convenient, but in Python, we stick to the PEP8 79 char limit rather faithfully, with rare exceptions.

gulp/src/manageusers/manageusers.js
aspidites: :+1: @koonsies

gulp/src/manageusers/manageusers.js
```diff
@@ -0,0 +1,778 @@
...
+      <ul>
+        {associated_activities_list}
+      </ul>
+    );
+  }
+});
+
+var RolesList = React.createClass({
+  handleEditClick: function(role_id) {
+    ReactDOM.render(
```
darrint: This subtly breaks React, as @koonsies discovered this morning. Rendering ~~should~~ **must** only be done in render. In these event handlers you can only fire off background work and/or setState({...}). If you need to affect some other part of the DOM with your event then you can call back to a component closer to the root of the hierarchy. There is no other way.

templates/base-new.html
```diff
@@ -0,0 +1,177 @@
...
+          <div class="clearfix"></div>
+      </footer>
+      <script>
+          var static_url = "{{ STATIC_URL }}";
+      </script>
+      <!--[if IE]>
+      <script type="text/javascript" src="{{ STATIC_URL }}es5-shim.min.js"></script>
+      <![endif]-->
+      <script type="text/javascript" src="{{ STATIC_URL }}custom.166-06.js{{ gz }}"></script>
+      <script type="text/javascript" src="{{ STATIC_URL }}my.jobs.160-30.js{{ gz }}"></script>
```
koonsies: This jQuery import may not be required. If you're not using the code here, I'd remove it. Notably, it enabled/disables buttons automatically when doing AJAX. This made my module break as it was clearing the "disabled" state of some of my react components.

templates/base-new.html
```diff
@@ -0,0 +1,177 @@
+{% load i18n %}
```
koonsies: Are we intending on using this in other places than on the react single-page apps? If not, perhaps we could rename it to base-react.html?

gulp/src/manageusers/manageusers.js
darrint: To be fair, this turned out not to be causing the problem. It was a bad interaction with some existing jquery code and use of jquery.get. Although, I still would strongly argue that we not use this technique.

gulp/gulpfile.js
lemniscate613: @aspidites You convinced me. Gone!

gulp/src/manageusers/manageusers.js
lemniscate613: @koonsies Excellent idea. I added "break headers out into an included file" to a list I'm maintaining for future work.

gulp/src/manageusers/manageusers.js
lemniscate613: Here's what I drafted:

"Use this tool to create users and permit them to perform various activities.

First, create a role. A role is a group of activities (e.g. view existing communication records, add new communication records, etc.). Then, create a user and assign them to that role. Once assigned to a role, that user can execute activities assigned to that role."

templates/base-new.html
lemniscate613: @koonsies This template introduces many significant changes: Bootstrap 3, the beginnings of consolidated styles, and React (instead of jQuery). I renamed the template base-react.html, but we should remember that it does more than that.

templates/base-new.html
lemniscate613: @koonsies Good point. I'm removing them, but keeping them as comments for historical reasons. They contain a lot of custom jQuery (e.g. captcha forms, tooltips, alerts to confirm disabling account, etc.) that doesn't appear relevant to our new apps. We will (probably) need to rewrite them if we want their functionality.

tdphillips: If the code is deleted, it's still in Git.

templates/base-new.html
koonsies: If the intent is to use it in places other than with react apps, I wouldn't rename it as I mentioned.

templates/base-new.html
lemniscate613: Ah, yes. My latest work is in a different branch, yet to be submitted.

Since there has been so much conversation around this application, and because I heard some wanted to use a draft frontend for basic testing, I've been trying to break it into different PRs.

For example, I wanted to address tablet/mobile, "as well as continuing to clean up templates like base-new.html" in a different PR.

I understand not wanting to merge a PR with significant work yet to do into quality-control though...  What's the best workflow for sharing chunks of features/UI? Just keep opening/closing PRs (and not mark something to be merged until it's absolutely, completely done)?

PR #1885, opened 2015-11-17
Re-add try/catch [PD-1939]

tdphillips:
I was able to syncdb with this.

All tests pass.

### DIFF
--- a/myemails/models.py
+++ b/myemails/models.py
@@ -4,7 +4,7 @@ from django.contrib.auth.models import ContentType
 from django.contrib.contenttypes.generic import GenericForeignKey
 from django.core.mail import send_mail
 from django.conf import settings
-from django.db import models
+from django.db import models, OperationalError, ProgrammingError
 from django.db.models.signals import pre_save, post_save
 from django.dispatch import receiver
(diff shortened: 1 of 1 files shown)


myemails/models.py
```diff
@@ -211,39 +211,40 @@ def event_receiver_decorator(fn):
 # I don't really like doing it this way (get from database pre save, set
 # attribute, get again post save), but we need to be able to 1) determine what
 # was changed and 2) only send an email if the save is successful. - TP
+try:
+    @event_receiver(pre_save, 'invoice', 'created')
```
tdphillips: I could have wrapped something in event_receiver but I'd rather not have these signal methods floating around if anything fails.

aspidites:
Only thing I'd say is to maybe replace "pass" with a comment explaining why the try/except is required in the first place. Otherwise :+1:, being as to how you explained why you did something you wouldn't prefer to have to do.

aspidites:
Fun Fact: With a comment, you don't need `pass`. I'm merging this anyway, as it doesn't harm anything.

Confirmed syncdb (--all) and migrate (--fake) work.

PR #1886, opened 2015-11-17
[PD-1922] Wrap Tags in Activities, pt 2

aspidites:
# Description
The previous PR failed to properly wrap the partner tagging page's autocomplete (/prm/view/tagging).  In addition to correcting this, I've also fixed the top bar, which was previously expecting a mixture of company user and activities stuff.


aspidites:
@fezick after this is merged, you should be able to QC PD-1922 without waiting on PD-1892

universal/helpers.py
```diff
@@ -117,18 +114,18 @@ def get_company(request):
...
 
     if not company:
-        try:
-            # If the company cookie isn't set, then the user should have
-            # only one company, so use that one.
-            company = request.user.get_companies()[0]
-        except IndexError:
-            company = None
+        # If the company cookie isn't set, then the user should have
+        # only one company, so use that one.
```
koonsies: This is more for my info, but is this an established business rule now? That a user cannot have roles assigned from multiple companies? I had heard it mentioned before, but never got to a resolution.

aspidites: Neither. This code isn't enforcing one role per multiple  companies, but rather getting the current company to check activities against. You can change which company this references by selecting another one from the company drop down in the topbar under "Employers".

koonsies: Alright. My confusion mostly came from the comment above. I read it to mean that if a user currently does not have a company set in the context, he should only possibly have a role in one company, so use that company.

aspidites: I didn't write the comment.

"should" is more like "probably". In other words, we only set the company cookie when a user has multiple companies, so if the cookie isn't set, it's probably because they are associated with more than one company.

koonsies: Cool, cool. Thanks for the explanation!

PR #1887, opened 2015-11-18
[PD-1880]  Reporting Interface with Filtering

darrint:
# Test Guide

* `manage.py reset_dynamic_data`
* https://secure.my.jobs/reports/view/dynamicoverview
* Only the Contacts report type is active.

# Expectations for this PR vs other tickets

There are 4 layers present in this PR

## UI as seen by the user, view.js/wizard.js

These are placeholders to demonstrate backend capability and "flow" for the user. They have some major visual flaws. This is ok for this PR. See tickets below.

## JS application code

The only thing I reserve the right to change is the overall arrangement of the code in files and directories. I expect the current directory structure and import structure to change as we learn.

Everything else, the classes, functions, gulpfile, build, tests, docs, etc. is expected to be production standard and I should fix problems now.

## API (views)

This should be production quality.

## Datasources (Contacts)

What I have for contacts and it's associated tests should be ready for production as it stands. I expect to identify some ways to better factor common functionality when I add more datasources. 

In particular, the filtering and help functions should be complete.

# Review Guide

I would recommend starting with the API in `myreports/views.py`.

The user walks through:
* reporting_types_api
* report_types_api
* data_types_api
* presentation_types_api
* filters_api
* help_api <- lots of times
* run_dynamic_report
* list_dynamic_reports <- automatically refreshed by client
* download_dynamic_report

From there you can proceed into `myreport/datasources.py` and it's associated modules.

The client code starts in `gulp/src/reporting/main.js`. From there you can follow the code to reportEngine.js, which contains the business logic in a unit testable form.

The view code is at view.js and wizard.js.

# Todo
- [x] fix bug in tags after adding then deleting all tags (no results)
- [x] add locations widget
- [x] add date range widget
- [x] limit contacts to "approved," whatever that means. See earlier code.
- [ ] ~~name reports and show names~~
- [x] rewrap views per: #1889
- [x] docstrings party!

# Defer to future tickets
- [pd-1950] Admin for new report entities and playing with columns
- [pd-1383] Add datasources for Partners and Communication Records
- [pd-1951] Audit API carefully looking for security holes. i.e. Get dynamic report for other company, etc.
- [pd-1952] Name report runs.
- [pd-1970] Date range component
- [pd-1958] Tags component
- [pd-1956] Partners component
- [pd-1957] Contacts component

Those will address properly getting UI from placeholder to ready, and getting enough data available to make this system viable.


aspidites:
Do you mind either rebasing off of #1884 or rewrapping the views manually? As they are now, you'd be undoing the work I did in aforementioned PR, since the PRM shortcut I introduced forever ago was more restrictive (though in keeping with how production operates) than we need/want.

aspidites:
>  limit contacts to "approved," whatever that means. See earlier code.

Did you ask anyone what it means? [Here](https://directemployersfoundation.atlassian.net/browse/PD-1381) you go.

aspidites:
Biggest complaint here is our use of ES6, which doesn't seem to have been agreed upon (see @lemniscate613's PRs). While I personally think some things are clearer in ES6 than ES5, I don't like the idea of having to context switch between the two depending on what part of the code I'm in. Especially when you consider things like the fact that `this` means something different depending on which standard you are  following.

darrint:
As far as ES6/7/2015/??? is concerned, we're using using Babel's default of TC39 stage 2 features. https://tc39.github.io/process-document/ I believe that the ES7 async/await feature is still at stage 2. I'm going to have a very difficult time giving that up.

aspidites:
I'll have a hard time merging this considering how much of a headache
maintaining these new features is. I recently had to update two different
versions of topbar because we decided to deviate from what was already
established. I'd really not like to also context switch between two
different code standards which have very different implications with
respect to browser support and the amount of tooling we need to maintain.

aspidites:
Oh, I really appreciate you annotating future work with related tickets. :+1: x :100:. Really helpful to assure other developers that a) it actually will be worked on, and b) what the context of that future work is. Well done.

aspidites:
Reserving my comments until I get clarification on Jira.

darrint:
I am willing to compromise on babel settings. (ES6/7/2015/+++) There are a few bits, such as promises + async/await, that will be pried from my cold dead fingers. If you don't like, say, arrow functions, I guess we could lose that but I'd hate to see them go.

@JLMcLaugh If we need to meet on this let's do so.

aspidites:
1. Promises exist in ES5
2. If a developers choice of language choice were reason enough to increase disparity in a  code base such that it negatively impacts further development by increasing the maintenance overhead of code in general and the comprehension of that code in general, then I'd have started writing PureScript long ago. That you like a feature is *not* in my opinion a  good enough for introducing such burdens. 
3. As I said before, if you can convince the others to convert to ES6, I might be amenable to such an endeavor, though even the recent introduction of bootstrap 3 over 2 has caused enough headaches that it wouldn't be an easy sell.

I apologize if this all sounds harsh, but I'm not a fan of the "But I like it, so it's staying" tone which doesn't take into consideration the cost of introducing a language implementation which is both incongruent with the rest of the code base and hardly standardized across browsers. Indeed, in supporting ES6, we are introducing shims which increase the size of the JS payload needed on each page, which directly conflicts with what we've been meaning to do in reducing response time across all of our platforms.

darrint:
Those are all valid concerns and you are right to want them addressed in this PR. @JLMcLaugh can you set up a meeting for me, @aspidites, and "The Right People (tm)" discuss and deside this?

gulp/gulpfile.js
```diff
@@ -13,12 +13,32 @@ var stripDebug = require('gulp-strip-debug');
...
+// For development run the default target, then leave the watch target running.
+
+// Future: Add jshint to the application bundle builds in the hope that we
+//     can use it in a manner similar to how use use flake8 on python code.
+
+// These go in vendor.js and are left out of app specific bundles.
 var vendor_libs = [
     'react',
     'react-dom',
     'redux',
```
aspidites: Didn't you say we weren't using this anymore?

gulp/src/reporting/api.js
```diff
@@ -65,6 +84,14 @@ class Api {
         return this.parseJSON(this.checkStatus(response));
     }
 
+    async listReports() {
+        var response = await this.getFromReportingApi(
+            "/reports/api/list_reports", {
+                method: 'GET',
```
koonsies: Why is method: 'GET' passed in here? It is hardcoded above.

PR #1889, opened 2015-11-19
Invalid activities

aspidites:
Makes invalid activities throw an error. Just putting this up to run tests.

Not as bad as it looks, it's only [these](https://github.com/DirectEmployers/MyJobs/compare/pd-1892...invalid-activities) changes


aspidites:
Just keeps getting worse...I'm over it.

PR #1890, opened 2015-11-19
Converted to built-in decorator.

aspidites:
@lemniscate613 discovered a decorator built into django that does almost what we do with `myreports.decorators.restrict_to_staff`. As we are only concerned with hiding views from a non-staff user, I don't think it's a big deal that we are redirecting instead of 404ing.

I also wrapped a few API endpoints that weren't previously wrapped despite belonging to an unreleased feature.

Just curing some NiH*.

* Well, not really Not invented Here syndrome, as I didn't realize staff_member_required existed.


lemniscate613:
:+1:

koonsies:
Looks good, but there's a merge conflict currently.

aspidites:
@koonsies , done. This isn't quite as trivial as it was before though. See updated description. Still a simple PR.

PR #1895, opened 2015-11-19
[PD-1948] Can no longer protect a view against an activity that doesn't exist

aspidites:
# Description
Before now, passing an activity which didn't exist would cause a page to be unreachable. This causes a ValueError when the view is called instead.

[These](https://github.com/DirectEmployers/MyJobs/compare/pd-1892...pd-1948) are the actual changes.

# Caveat
I could have moved the check up a level such that as soon as the module containing the view is imported the error is raised (and thus preventing us from having to do an extra SELECT every time the view is hit), but doing so would have required me to refactor the entire test suite, as during most tests it  is assumed that roles are not enabled. It isn't advisable to have roles enabled for all tests as currently in production they are disabled, so enabling them would mean that we are no longer testing conditions that mirror production.


aspidites:
Closing as won't fix. See the ticket for info.

PR #1898, opened 2015-11-19
Revert "Add allowed paths to secure blocks [PD-1613]"

tdphillips:
Reverts DirectEmployers/MyJobs#1883

Doesn't do what we want.


tdphillips:
This has a migration. I don't think we've though through reverting enough to merge this.

tdphillips:
My local environment isn't in a state to allow migrating backward so someone else will need to do so.

aspidites:
Visually inspected code before hand. If this doesn't pass tests, we have other problems.

PR #1900, opened 2015-11-20
Fix Topbar with respect to Roles

aspidites:
# Changes
- Added a few more convenience methods and refactored
- user management was missing from mobile views
- user management only shows up when role are enabled
- prm only shows up when prm is enabled
- if no features are enabled, Employers is disabled all together
- wrote tests for new template tag that I needed to support this
- wrote tests for new convenience methods

# Annotations
One complaint that I've heard often is that PRs aren't reviewed because people are either not familiar with the code or the amount of changes is daunting. I've annotated all my changes as an experiment to see if it hinders or improves the process. Please give feedback (negative or positive, but always constructive) with respect to this aspect. Note that the annotations are cleaner if you read them inline under "File changes"' rather than as part of "Conversation".

### DIFF
--- a/myjobs/decorators.py
+++ b/myjobs/decorators.py
@@ -227,7 +227,7 @@ def requires(*activities, **callbacks):
                 return access_callback(request)
             # the user should have at least the activities required by the view
             elif not bool(request.user.activities) or not set(activities).issubset(
-                    user.activities):
+                    request.user.activities):
                 return activity_callback(request)
             else:
                 return view_func(request, *args, **kwargs)
--- a/myjobs/decorators.py
+++ b/myjobs/decorators.py
@@ -222,12 +222,18 @@ def requires(*activities, **callbacks):
                     'name', flat=True).distinct())
 
             # company should have at least the access required by the view
-            if not bool(company.enabled_access) or not set(
-                    required_access).issubset(company.enabled_access):
-                return access_callback(request)
+            has_access = all([
+                bool(company.enabled_access),
+                set(required_access).issubset(company.enabled_access)])
+
             # the user should have at least the activities required by the view
-            elif not bool(request.user.activities) or not set(activities).issubset(
-                    request.user.activities):
+            has_activities = all([
+                bool(request.user.activities),
+                set(activities).issubset(requset.user.activities)])
+
+            if not has_acces:
+                return access_callback(request)
+            elif not has_activities:
                 return activity_callback(request)
             else:
                 return view_func(request, *args, **kwargs)
--- a/myjobs/decorators.py
+++ b/myjobs/decorators.py
@@ -215,11 +215,6 @@ def requires(*activities, **callbacks):
                 return view_func(request, *args, **kwargs)
 
             company = get_company_or_404(request)
-            # the app_access we have, determined by the current company
-            company_access = filter(bool, company.app_access.values_list(
-                'name', flat=True))
-            user_activities = filter(bool, request.user.roles.values_list(
-                'activities__name', flat=True))
 
             # the app_access we need, determined by the activities passed in
             required_access = filter(bool, AppAccess.objects.filter(
@@ -227,12 +222,12 @@ def requires(*activities, **callbacks):
                     'name', flat=True).distinct())
 
             # company should have at least the access required by the view
-            if not bool(company_access) or not set(required_access).issubset(
-                    company_access):
+            if not bool(company.enabled_access) or not set(
+                    required_access).issubset(company.enabled_access):
                 return access_callback(request)
             # the user should have at least the activities required by the view
-            elif not bool(user_activities) or not set(activities).issubset(
-                    user_activities):
+            elif not bool(request.user.activities) or not set(activities).issubset(
+                    user.activities):
                 return activity_callback(request)
             else:
                 return view_func(request, *args, **kwargs)
--- a/myjobs/decorators.py
+++ b/myjobs/decorators.py
@@ -229,9 +229,9 @@ def requires(*activities, **callbacks):
             # the user should have at least the activities required by the view
             has_activities = all([
                 bool(request.user.activities),
-                set(activities).issubset(requset.user.activities)])
+                set(activities).issubset(request.user.activities)])
 
-            if not has_acces:
+            if not has_access:
                 return access_callback(request)
             elif not has_activities:
                 return activity_callback(request)
--- a/myjobs/models.py
+++ b/myjobs/models.py
@@ -644,11 +644,7 @@ class User(AbstractBaseUser, PermissionsMixin):
 
     def can(self, company, *activity_names):
         """
-        Note: Until the activities feature is deployed, this method only
-        returns true when the company is a member company and the user is a
-        company user for that copmany.
-
-        Checks if a user may perform certain activities.
+        Checks if a user may perform certain activities for a company.
 
         Inputs:
             :company: The company who's role activities to check.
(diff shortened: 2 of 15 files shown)


aspidites:
Looks like the recent failures (mydashboard and friends during build #747) were transitive, as those particular tests pass locally. Did some PEP-8 changes to see if that resolves anything.

aspidites:
Just checked. These same tests are failing in QC but on on master.

myjobs/decorators.py
```diff
@@ -215,11 +215,6 @@ def wrap(request, *args, **kwargs):
                 return view_func(request, *args, **kwargs)
 
             company = get_company_or_404(request)
-            # the app_access we have, determined by the current company
```
aspidites: I moved `company_access` to `Company.enabled_access` and `user_activities` to `User.activities`.

myjobs/decorators.py
```diff
@@ -227,12 +222,18 @@ def wrap(request, *args, **kwargs):
                     'name', flat=True).distinct())
 
             # company should have at least the access required by the view
-            if not bool(company_access) or not set(required_access).issubset(
-                    company_access):
-                return access_callback(request)
+            has_access = all([
```
aspidites: I figured semantic names inside of boolean expressions would be easier to read than inlined expressions.

myjobs/models.py
```diff
@@ -670,17 +674,41 @@ def can(self, company, *activity_names):
...
         """
 
         if not company:
             return False
 
         if settings.ROLES_ENABLED:
-            return set(activity_names).issubset(
-                self.roles.filter(company=company).values_list(
-                    'activities__name', flat=True))
+            required_access = filter(bool, AppAccess.objects.filter(
```
aspidites: This ensures that `False` is returned when a user has an activity but the company doesn't have app access.

myjobs/models.py
```diff
@@ -670,17 +674,41 @@ def can(self, company, *activity_names):
...
+            # Company must have correct access and user must have correct
+            # activities
+            return all([
+                bool(company.enabled_access),
+                set(required_access).issubset(company.enabled_access),
+                bool(self.activities),
+                set(activity_names).issubset(self.activities)])
+
         else:
-            return company.member and company in self.company_set.all()
```
aspidites: This is really adhoc. Basically, the logic is such that when `settings.ROLES_ENABLED` is `False`, a user can never do anything with roles, if we are  checking for partner activities, we should defer to `Company.prm_access`, otherwise, we should check that the user is a company user for the company and that said company is a member.

The reason for not simply checking for prm_acces in the second case is that topbar, like other views, checks for activities, not app access.

myjobs/templatetags/common_tags.py
```diff
@@ -35,6 +35,12 @@ def completion_level(level):
 
     return get_completion(level)
 
+@register.assignment_tag
+def can(user, company, *activity_names):
+    """Template tag analog to `myjobs.User.can()` method."""
```
aspidites: Just a proxy to `User.can` for use in templates.

myjobs/templatetags/common_tags.py
```diff
@@ -71,7 +77,9 @@ def is_a_group_member(company, user, group):
     """
 
     if settings.ROLES_ENABLED:
-        return getattr(company, 'has_features', False)
+        return any([
```
aspidites: This allowed for a degrading topbar experience. That is, before, having PRM access would show the User Management item, even though the user might not have had access to roles. As such, the link would immediately return a 403. This way, they only ever see those links which are clickable.

myjobs/tests/setup.py
```diff
@@ -17,6 +17,7 @@ def setUp(self):
         self.ms_solr = Solr('http://127.0.0.1:8983/solr/seo')
         self.ms_solr.delete(q='*:*')
         setattr(settings, "PROJECT", 'myjobs')
+        settings.ROLES_ENABLED = False
```
aspidites: Disable roles in the base case, as we still need to ensure current functionality. Also, the tests aren't yet equipped to deal with roles, so they would all fail anyway.

myjobs/tests/test_activities.py
```diff
@@ -20,12 +20,12 @@
 from seo.tests.factories import CompanyFactory
 
 
-@override_settings(ROLES_ENABLED=True)
```
aspidites: Same thing. Decorating a subclass is pointless if the superclass is decorated, as the superclass would get called afterward, thus negating anything decorated in the subclass.

myreports/tests/test_activities.py
```diff
@@ -10,21 +10,22 @@
...
 from myjobs.tests.factories import (AppAccessFactory, RoleFactory, UserFactory,
                                     ActivityFactory)
 from mypartners.tests.test_views import TestClient
 from seo.tests.factories import CompanyFactory
+from myreports.tests.setup import MyReportsTestCase
 
 
-@override_settings(ROLES_ENABLED=True)
-class TestViewLevelActivities(TestCase):
+class TestViewLevelActivities(MyReportsTestCase):
```
aspidites: Probably not necessary, but it made sense for reports tests to derive from the same base class.

templates/includes/topbar-new.html
```diff
@@ -45,16 +45,18 @@ <h3 id="topbar-logo">My.jobs</h3>
                         </ul>
                     </li>
                     {% get_company_from_cookie as company %}
+                    {% can user company "read partner" as can_read_partner %}
```
aspidites: Mimicked the old top bar. Made sure that permissions were granular as far as which menu items to display.

templates/includes/topbar.html
```diff
@@ -45,16 +45,18 @@ <h3 id="topbar-logo">My.jobs</h3>
...
+                    {% can user company "read partner" as can_read_partner %}
+                    {% can user company "read role" as can_read_role %}
                     {% is_a_group_member company user "Employer" as group_member %}
                     {% get_company_name user as company_name %}
                     {% if group_member %}
                     <li class="{% active_tab 'employer-tools' %} has-drop">
                         <a>{% trans "Employers" %}</a>
                         <ul id="employer-apps" class="submenu">
-                            {% if company.prm_access %}
+                            {% if can_read_partner %}
```
aspidites: Granular permissions. In a later iteration, I might make the template tag return a list of activities for the company such that this would work:

```
{% user _can company as user_activities %}
{% if "read partner" in user_activities %}
{% if read role" in user_activities %}
```
Doing so would reduce the number of template tag calls.

lemniscate613:
Working my way through this. Looks good so far.

Excuse the n00b question:

Are Posted Jobs and Reports aspects of PRM? I imagine so, since when I remove PRM from a company, a user for that company no longer can see those items in the nav.

aspidites:
@lemniscate613 yes and no. Reports is it's own entity, but as we don't have any reports-specific activities, we simply ensure that you have the permissions needed to view what's being reported on -- in this particular case partners. Technically, this should be "read partner", "read contact", "read contactrecord", but that would require (with how things are implemented at the moment) two more variables and a longer conditional, which is why I didn't go that route. For 'Posted Jobs', this is the same, given there are no postajob activities. This will need to be updated once we get around to wrapping that app (which btw, is only used by maybe 4 companies). That we check for prm access for those is a historical artifact.

haliscooper:
Plan is to get franchises using the system, in which case we'll have many more companies quickly, we will make those changes as we work towards that.

koonsies:
I like the inline annotations. It can help clear up confusion and guide review of the code. :+1:

darrint:
:+1: on the inline annotations.

PR #1903, opened 2015-11-23
Update staff check

aspidites:
`staff_member_required` redirects to django admin, which is only desireable on seo. As such, I switched those occurrences in myjobs to my restrict_to_staff decorator. I also wrapped the rest of the unreleased reporting api's in the staff decorator.

This builds on top of #1900, so the only real changes are [these](https://github.com/DirectEmployers/MyJobs/compare/fix-topbar...update-staff-check?expand=1).


koonsies:
:+1:

PR #1917, opened 2015-11-30
Invitations are no longer automatically sent.

aspidites:
I could give a lot of  context, but as this is a sandbox, I don't much feel like writing a book that no one will read, so to speak. The short of it is that I am attempting to decouple saved searches from the Invitation model without explicitly having to rely on the use of the content types framework. 

I'm relying on the assumption that we only care about saved searches when initially creating the saved search, and thus don't care about any hard reference to it. As such, I've also removed the implicit sending of invitations so that I could manually pass in an instance to use in the templates.

@tdphillips, I'm looking to you to explicitly point out where this will fail. Registration tests pass and I was able to invite myself using the admin form using this code as is.


registration/models.py
```diff
@@ -184,12 +184,9 @@ def save(self, *args, **kwargs):
         else:
             self.invitee_email = self.invitee.email
 
-        if not self.pk:
-            self.send()
-
         super(Invitation, self).save(*args, **kwargs)
 
-    def send(self):
+    def send(self, instance=None):
```
tdphillips: I know what instance is right now. If I were to forget everything I know about invitations, I wouldn't know what kind of object this was.

mysearches/forms.py
```diff
@@ -275,9 +275,9 @@ def save(self, commit=True):
                 'invitee': instance.user,
                 'inviting_user': instance.created_by,
                 'inviting_company': instance.partner.owner,
-                'added_saved_search': instance,
```
tdphillips: Do these changes as presented precede the removal of this foreign key entirely?

tdphillips:
I don't see anything indicating that this would fail. Moving sending from a signal to an explicit action doesn't bother me - saved searches are explicit anyway.

registration/models.py
aspidites: Correct. I deliberately ignored documentation and coming up with something descriptive. In another iteration, I intended that instance could also be role or some other model that some how needs to be associated with the associated with the invitation when sending.

mysearches/forms.py
aspidites: I've been up since 1:30 doing homework, so I'm for some reason unable to parse the above. I will say that the reason 'added_saved_search' was removed here was because it's only purpose seemed to be to facilitate automatic sending of the invitation email. 

I do intend, once I'm done, to remove the foreign key relationship though, in case that's what you're getting at.

aspidites:
Going to the drawing board.

PR #1918, opened 2015-11-30
[PD-1974] Wrote helper method to invite a user.

aspidites:
# Overview
To allow for invites to be sent from anywhere in a simple fashion, I added a `send_invite` method to `User`.

# Testing
-  call `./manage.py shell`
- import `User` from `myjobs.models` and `Company` from `seo.models`
- assign a user and a compnay for which that user has a role (or assign a role to that user from that company
- call `user.send_invite` with an arbitrary email address (that you own), the company retrieved above, and either `True` if `settings.ROLES_ENABLED` is set to `False` or a valid role name
- You should receive an invitation email, and that user should now exist, assigned the appropriate role (or as a company user if roles aren't enabled)

# Future Work
- [PD-1972](https://directemployersfoundation.atlassian.net/browse/PD-1972) At the moment, the `Invitation` model is tied to saved search due to the original requirements of the feature. Ideally, we'd like the invitation system to be independent of any other models, but decoupling them at this point was more work that we were prepared to deal with in order to get Roles and Activities out the door
- [PD-1856](https://directemployersfoundation.atlassian.net/browse/PD-1856) The messaging here is really generic because the invitation email template doesn't take a parameter to explain the origin of the invite. Future template work should take care of this and allow targeted messaging such as "You've been invited to be an Admin for DirectEmployers"

# Screenshot of resulting email
2015-12-01-080731_1435x382_scrot

### DIFF
--- a/myjobs/models.py
+++ b/myjobs/models.py
@@ -710,6 +710,49 @@ class User(AbstractBaseUser, PermissionsMixin):
             else:
                 return is_company_user and company.member
 
+    def send_invite(self, email, company, role_name=None):
+        """
+        Sends an invitation to the `email` in regards to `role_name` for
+        `company`.
+
+        Inputs:
+            :email: The email address of the potential user being invited.
+            :company: The company inviting the user.
+            :role_name: The name of the role (optional) the user is to be
+            assigned.
+
+        Output:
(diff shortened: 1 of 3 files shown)


aspidites:
Note. The email template no longer mentions job boards.

PR #1923, opened 2015-12-02
Documentation sphinx update

aspidites:
Pushing @koonsies  work into qc so that I have a usable docs directory. Most of the work was already completed by an earlier PR to reduce circular dependencies. This just continues in that direction and experiments with writing documentation in Sphinx.


doc/source/conf.py
```diff
@@ -19,6 +19,8 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('../..'))
+os.environ['DJANGO_SETTINGS_MODULE'] = 'settings'
```
koonsies: I wonder if the failing tests have to do with this. I believe I had some issues with the settings environment variable in this call.

aspidites: Huh. I'll take a look. That said, I'm closing this, as I've subsumed it in some other work I'm doing (https://directemployers.github.io/MyJobs)

PR #1926, opened 2015-12-02
Remove requires decorator from process_email

tdphillips:
This view is used by SendGrid. request.user is an anonymous user.


tdphillips:
Currently working on prm email stuff. Leaving this as DNM until that's done.

PR #1927, opened 2015-12-02
Reverted a few changes that never made it to qc

aspidites:
@lemniscate613 , this is all the stuff I found in your branch that conflicted with QC (eg, came from feature branches that were never merged).


lemniscate613:
Thanks for this, but for some reason it broke my everything. I ended up just reverting the files that had nothing to do with manage users, and it worked.

PR #1928, opened 2015-12-03
Documentation Generation

aspidites:
# Description
**Proof of Concept**
- These changes allow us to generate documentation on http://directemployers.github.io/MyJobs 
- The resulting output* is **live** -- go look
- I updated the readme to include a link to the documentation as well as update the copyright.
# Rational
I have a laundry list of reasons for wanting to do this. It'd probably be best for me to put together a presentation for a meeting than try to enumerate everything here. The gist of it, though, is this:
sphinx > github wiki > google docs

# Testing
- As mentioned above, the generated documentation is live
- I'd rather you *didn't* regenerate the documents until after this is merged as doing so would clobber what's there and force old documentation**
- You *should* however, go ahead and generate documentation locally:
```python
pip install -r requirements
cd doc
make html
{firefox,chrome,chromium,opera} ./build/html/index.html
```
- You can also generate a pdf if you have the right tools installed:
```bash
make latexpdf
```
[Myjobs.pdf](https://github.com/DirectEmployers/MyJobs/files/50450/Myjobs.pdf)

# Future Work
- Convert remaining wiki docs to sphinx if the consensus is that this is better
- Experiment with our secret repositories to see about hosting internal documentation
- Finish formatting autodoc so that API documentation is generated for all modules, not just mypartners

*I generated that documentation from this branch, but the PR makes it so that documentation is generated from quality-control -- if you attempt to do so now, you will clobber what is currently there, so don't do so until after this is merged, please.

**though I suppose you could test the generation; merge, then regenerate; I'd rather that was handled by Jenkins though when something gets merged into qc


aspidites:
These tests pass locally, so I have no clue...

README.rst
```diff
@@ -20,15 +21,20 @@ needs. You may choose from:
...
+My.jobs is built on Python and Django, with heavy reliance on the
+DirectEmployers Foundation User Interface Framework
+(https://github.com/DirectEmployers/UI-Framework).
 
-Information about DirectEmployers Foundation can be found at http://directemployersfoundation.org
+Information about DirectEmployers Foundation can be found at
+http://directemployersfoundation.org
 
 
 We are hiring!
```
koonsies: While we're updating the README, we should fix the Hiring link. http://directemployersfoundation.org/careers is a dead link. Though I am not aware of another "Hiring" page to direct them to.

PR #1929, opened 2015-12-03
Update copyright year.

aspidites:
Just a sanity check. Wanted to see if all new prs would break, or if I did something some how.


aspidites:
Hmm. Ok Then.

PR #1930, opened 2015-12-03
Updated the FAQ and JavaScript Guide

aspidites:
# Description
This is an example of how updating documentation would work if we used Sphinx hosted on gh-pages. 

Benefits over a wiki
- Updating documentation is an active process. You can be assured that at least one person looked at your documentation changes before it became final
- Allows for outside contribution: If someone else wants to update our documentation, they need simply fork quality control (I'm using sphinx here as my changes haven't been merged yet) and make their changes. 

I should note that as github allows files to be edited inline, one could easily make a PR with just as little effort as updating the wiki:

2015-12-03-092318_1004x72_scrot
2015-12-03-092335_998x949_scrot

### DIFF
--- a/doc/source/developers/styleguides/javascript.rst
+++ b/doc/source/developers/styleguides/javascript.rst
@@ -1,5 +1,5 @@
 ===========================
-JavaSCript Codging Standard
+JavaScript Codging Standard
 ===========================
 
 Goals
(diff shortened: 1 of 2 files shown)


aspidites:
Sphinx is now in QC, so I'm pointing this elsewhere.

PR #1931, opened 2015-12-03
Stop trying to shadow the "partner" property [PD-1976]

tdphillips:
# Testing
Run tests. They pass.
If you want to test the old behavior, you'll have to copy the `test_attachments_attach` test from mypartners/tests/test_views.py to QC and run it. It should fail because there are no PRMAttachments.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -747,8 +747,12 @@ class PRMAttachment(models.Model):
             filename = 'unnamed_file'
 
         uid = uuid4()
-        path_addon = "mypartners/%s/%s/%s" % (self.partner.owner.pk,
-                                              self.partner.pk, uid)
+        if self.partner:
+            partner = self.partner.pk
+            owner = self.partner.owner.pk
+        else:
+            partner = owner = 'none'
+        path_addon = "mypartners/%s/%s/%s" % (owner, partner, uid)
         name = "%s/%s" % (path_addon, filename)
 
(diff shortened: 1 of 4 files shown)


mypartners/views.py
```diff
@@ -1307,7 +1306,7 @@ def process_email(request):
                 prm_attachment = PRMAttachment()
                 prm_attachment.attachment = attachment
                 prm_attachment.contact_record = record
-                setattr(prm_attachment, 'partner', contact.partner)
```
tdphillips: This is what was causing the original issue. Changing it to not try and shadow "partner" should fix it.

mypartners/models.py
```diff
@@ -747,8 +747,12 @@ def get_file_name(self, filename):
             filename = 'unnamed_file'
 
         uid = uuid4()
-        path_addon = "mypartners/%s/%s/%s" % (self.partner.owner.pk,
```
tdphillips: This is not currently an issue but should be done anyway as self.partner can be None.

mypartners/views.py
```diff
@@ -1163,7 +1163,6 @@ def prm_export(request):
 
 
 @csrf_exempt
-@requires("create partner", "create contact", "create communication record")
```
tdphillips: process_email is used by SendGrid, not through a UI by a user.

PR #1925, opened 2015-12-02
[PD-1836, PD-1946, PD-1959, PD-1924] User Overview and Listing Interface

lemniscate613:
# User Overview and Listing Interface

This PR represents _a lot_ of new and rewritten code.

# Highlights:

Style
* Conforms with Airbnb's style guide: https://github.com/airbnb/javascript
* Resolved ~1,000 ESLint warnings/errors (e.g. unused imports, unnamed functions, camel case variables, indentation, missing semicolons, etc.)

Organization
* Every component in its own file.
* Entry point is main.jsx
* Modularized validateEmail.js function.

URL routing with react-router: https://github.com/rackt/react-router
* Deep linking
* Back/next functionality: https://directemployersfoundation.atlassian.net/browse/PD-1946
* Required major code overhaul
* See <Router> component in main.jsx. Link components replace most a href tags.

ES6 style code.
* Const/let, map, etc.
* ES6 class definitions (see: http://babeljs.io/blog/2015/06/07/react-on-es6-plus/).
* Replace componentWillMount with constructor.
* Bind custom methods manually.

Views wrapped with Edwin's activity decorators.

Views protected by @restrict_to_staff() (for now)

In sync with @darrint's latest gulp file.js/package.json (https://github.com/DirectEmployers/MyJobs/pull/1922 ).

Respond.js pollyfill for IE8 compatibility. Thorough browser testing handled in future PR (PD-1945) .

Prebuilt Bootstrap3 components with react-bootstrap: https://react-bootstrap.github.io/ Only using Button component for now (but plan to use many others in future: https://github.com/DirectEmployers/MyJobs/wiki/JavaScript-Coding-Standard#library-policy )

Handle permissions gracefully. E.g. if user doesn't have edit_role activity, helpful error message.

Fix flicker. Call APIs more intelligently.

Bug resolved (thanks @koonsies!): https://directemployersfoundation.atlassian.net/browse/PD-1959

Updated topbar based on activities/roles (thanks @aspidites!).

Draft description in Overview.

Misc. UI changes (e.g. darken read only fields, like email address when editing user).

Updated webpack to play nicely with .jsx files.

# Review Guide:

```
cd gulp
npm install
gulp watch-tasks
```
Visit: http://127.0.0.1:8000/manage-users/

# Future Work in New Tickets:

Each of these will be in its own PR.

* Thorough browser / mobile testing: https://directemployersfoundation.atlassian.net/browse/PD-1945
[...]

PR #1933, opened 2015-12-04
PEP-8 For Manage Users Views

aspidites:
**note**: I usually link back to relevant changes while pointing directly to quality-control, but as the purpose of this PR is really pedagogical in nature, I wanted to facilitate that. Once #1925 gets merged, I'll close this and open a new PR against quality-control

# Description
This PR is largely concerned with correcting PEP-8 issues present in #1925. I have purposely ignored every urge to refactor the code except in cases where conforming to PEP-8 was improved by it. I have annotated my changes with an explanation for why I made the change, as well as a reference (when available) to what motivated the logic behind that change.

@lemniscate613 let me know if this is helpful. As I don't know what the extent of your Python knowledge is, I've written all comments with the assumption that you know nothing in the interest of being thorough. Please do not mistake that as deliberate attempts at condescension. 

# Future Work
I'll be heavily refactoring these views to:
- address some of the unidiomatic coding practices therein
- take advantage of some machinery that was introduced in other PRs to make dealing with activities and roles less cumbersome
- reduce redundancy in some of the logic
- decrease the payload size of the resulting JSON, which will speed up response times.


myjobs/views.py
```diff
@@ -19,7 +19,6 @@
 from django.template import RequestContext
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
-from django.contrib.admin.views.decorators import staff_member_required
```
aspidites: This import is not used. We use `restrict_to_staff` instead.

myjobs/views.py
```diff
@@ -35,7 +34,6 @@
 from registration.forms import RegistrationForm, CustomAuthForm
 from tasks import process_sendgrid_event
 from universal.helpers import get_company_or_404
-from seo.models import Company
```
aspidites: This import is not used. 
We typically get the company via `get_company(request)` or `get_company_or_404(request)`.

myjobs/views.py
```diff
@@ -157,12 +155,12 @@ def home(request):
                 if len(login_form.get_user().profileunits_set.all()) > 0:
                     has_units = True
 
-                response_data = {
+                ctx = {
```
aspidites: While we haven't entirely standardized on a name for the context returned for a view, we do use ctx quite often:

```bash
$ ack "ctx =" -h | wc -l
$ 52
```
Also, the shorter name allowed me to correct a few line length violations.

myjobs/views.py
```diff
@@ -613,7 +611,7 @@ def manage_users(request):
         }
 
     return render_to_response('manageusers/index.html', ctx,
-                                RequestContext(request))
+                              RequestContext(request))
```
aspidites: The [Indentation](https://www.python.org/dev/peps/pep-0008/#indentation) section of PEP 8 states:

> Continuation lines should align wrapped elements either vertically using Python's implicit line joining inside parentheses

Here, you were over by a single character.

myjobs/views.py
```diff
@@ -624,7 +622,8 @@ def api_get_activities(request):
     """
 
     activities = Activity.objects.all()
-    return HttpResponse(serializers.serialize("json", activities, fields=('name', 'description')))
+    return HttpResponse(serializers.serialize("json", activities,
```
aspidites: The [Maximum Line Length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) section states:
> Limit all lines to a maximum of 79 characters. 

Here, I wrapped the method call to abide.

myjobs/views.py
```diff
@@ -643,28 +642,36 @@ def api_get_roles(request):
...
-        available_activities = Activity.objects.filter(app_access__in=company.app_access.all())
-        response_data[role_id]['activities']['available'] = serializers.serialize("json", available_activities, fields=('name', 'description'))
+        available = serializers.serialize(
+            "json",
+            Activity.objects.filter(app_access__in=company.app_access.all()),
+            fields=('name', 'description'))
+        ctx[role_id]['activities']['available'] = available
         # Retrieve all activities assigned to this role
-        assigned_activities = role.activities.all()
-        response_data[role_id]['activities']['assigned'] = serializers.serialize("json", assigned_activities, fields=('name', 'description'))
```
aspidites: Wrapping on brackets is a bit ugly:
```python
response_data[role_id]['activities'][
    'assigned'] = serializerrs.serialize("json", assigned_activities, fields=('name', 'description'))
```
So I opted to move the serialization into the declaration.

myjobs/views.py
```diff
@@ -643,28 +642,36 @@ def api_get_roles(request):
...
-        response_data[role_id]['users'] = {}
-        response_data[role_id]['users']['assigned'] = serializers.serialize("json", users_assigned, fields=('email'))
+        users_assigned = serializers.serialize(
+            "json", users_assigned, fields=('email',))
+        ctx[role_id]['users'] = {}
+        ctx[role_id]['users']['assigned'] = users_assigned
 
         # Retrieve users that can be assigned to this role
-        # This is simply a list of all users already assigned to roles associated with this company
+        # This is simply a list of all users already assigned to roles
```
aspidites: Technically, per PEP 8, we should be wrapping comments to 72 characters, but I don't think we really follow that.

myjobs/views.py
```diff
@@ -764,23 +784,25 @@ def api_create_role(request):
...
+            ctx["success"] = "false"
+            ctx["message"] = "Another role with this name already exists."
+            return HttpResponse(json.dumps(ctx),
+                                content_type="application/json")
 
         activity_ids = []
         if request.POST.getlist("assigned_activities[]", ""):
             activities = request.POST.getlist("assigned_activities[]", "")
             # Create list of activity_ids from names
-            for i, activity in enumerate(activities):
```
aspidites: My linter warned me of the unused variable `i`. 
Basically, the only reason you are looping here is to do something with each activity, and you never index into it. The only time I'd recommend even using the index would be to map one list over another, and even then I'd probably use a comprehension.

myjobs/views.py
```diff
@@ -793,12 +815,13 @@ def api_create_role(request):
 
         # Add role to relevant users
         if users:
-            for i, user in enumerate(users):
+            for user in users:
                 user_object = User.objects.filter(email=user)
-                user_object[0].roles.add( new_role.id )
```
aspidites: The [Pet Peeves](https://www.python.org/dev/peps/pep-0008/#pet-peeves) section of PEP 8 states:
> Avoid extraneous whitespace in the following situations: 
> - Immediately inside parentheses, brackets or braces.

myjobs/views.py
```diff
@@ -818,47 +841,54 @@ def api_edit_role(request, role_id=0):
...
         company = get_company_or_404(request)
 
         # Check if the company the user is associated with manages this role
         role = Role.objects.get(pk=role_id)
         if role.company.id != company.id:
-            response_data["success"] = "false"
-            response_data["message"] = "The company you are associated with does not manage this role."
-            return HttpResponse(json.dumps(response_data), content_type="application/json")
+            ctx["success"] = "false"
+            ctx["message"] = ("The company you are associated with does not "
```
aspidites: The [Maximum line length](https://www.python.org/dev/peps/pep-0008/#maximum-line-length) section states:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses,
> brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. 
> These should be used in preference to using a backslash for line continuation

As such, I took advantage of the fact that parenthesis allow for implicit line wrapping and strings on consecutive lines are automatically concatenated. I'm not happy with this, but it was the simplest thing to do without restructuring the entire function.

myjobs/views.py
```diff
@@ -1000,42 +1035,44 @@ def api_get_specific_user(request, user_id=0):
...
     current_companys_roles = Role.objects.filter(company=company)
     ## List user's roles
-    roles_assigned_to_this_user = user[0].roles.all()
+    user_roles = user[0].roles.all()
     ## Overlap?
-    if bool(set( current_companys_roles ) & set( roles_assigned_to_this_user)) == "False":
-        response_data["success"] = "false"
-        response_data["message"] = "You do not have permission to view this user"
-        return HttpResponse(json.dumps(response_data), content_type="application/json")
+    if not set(current_companys_roles).intersection(user_roles):
```
aspidites: The line was too long and wrapping an if statement gets to be really ugly. As such, I opted to express the same thing you were expressing more concisely. I also used the `intersection` method over the overloaded `&` operator as I think the intent is clearer.

PR #1934, opened 2015-12-04
Fix so build fails on webpack errors.

darrint:
This was causing gulp to not fail on certain webpack specific errors.


aspidites:
Confirmed by pulling locally and checking $?.

PR #1938, opened 2015-12-07
Use separate SendGrid accounts for QC, Staging, and Production [PD-1470]

JLMcLaugh:
We're now using separate email accounts for QC, Staging, and production in my.jobs and redirects. Any other environments will default to use QC. 

The old SendGrid login should no longer be used, and we may add additional accounts to separate jenkins or microsites in the future.


deploy/settings_dseo_prod.py
```diff
@@ -0,0 +1,129 @@
...
+BROKER_HOST = '204.236.236.123'
+BROKER_PORT = 5672
+BROKER_USER = 'celery'
+BROKER_VHOST = 'dseo-vhost'
+
+CELERY_DEFAULT_EXCHANGE = 'tasks'
+CELERY_DEFAULT_EXCHANGE_TYPE = 'topic'
+CELERY_DEFAULT_ROUTING_KEY = 'dseo.default'
+
+EMAIL_HOST_USER = PRODUCTION_EMAIL_HOST_USER
```
JLMcLaugh: These 2 lines are new.

aspidites:
No way to test this locally, I presume? @mklauber  can you take a look at this?

tdphillips:
Will any of these be distributed? I'm working on email and sending said email without an account would prove interesting.

tdphillips:
@mklauber Does Jenkins have what it needs to successfully deploy this?

PR #1940, opened 2015-12-07
Revert "PEP-8 For User Management"

aspidites:
I apparently broke stuff. leaving it to the author to fix PEP-8.


myjobs/views.py
```diff
@@ -1007,6 +968,10 @@ def api_get_users(request):
         # Email
         ctx[user.id]["email"] = user.email
 
+        # Roles
```
lemniscate613: Front end needs this info. Many of the other changes look okay.

PR #1941, opened 2015-12-07
fix history, a dependency of react-router

lemniscate613:
http://stackoverflow.com/questions/34128823/warning-history-pushstate-is-deprecated-use-push-instead

### DIFF
--- a/gulp/package.json
+++ b/gulp/package.json
@@ -30,6 +30,7 @@
     "gulp-if": "^2.0.0",
     "gulp-jasmine": "^2.2.1",
     "gulp-util": "^3.0.7",
+    "history": "1.13.1",
     "react": "^0.14.3",
     "react-autosuggest": "^2.1.1",
     "react-bootstrap": "^0.28.0",
--- a/myjobs/views.py
+++ b/myjobs/views.py
@@ -768,11 +768,12 @@ def api_create_role(request):
             return HttpResponse(json.dumps(ctx), content_type="application/json")
 
         activity_ids = []
+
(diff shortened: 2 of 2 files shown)


myjobs/views.py
```diff
@@ -768,11 +768,12 @@ def api_create_role(request):
             return HttpResponse(json.dumps(ctx), content_type="application/json")
 
         activity_ids = []
+
         if request.POST.getlist("assigned_activities[]", ""):
             activities = request.POST.getlist("assigned_activities[]", "")
             # Create list of activity_ids from names
             for activity in enumerate(activities):
-                activity_object = Activity.objects.filter(name=activity)
+                activity_object = Activity.objects.filter(name=activity[1])
```
koonsies: The variable name here is "activity_object", but it is being assigned a queryset. Do you mean to only retrieve one Activity here? If so, it should read as either Activity.objects.filter(name=activity[1]).first() or Activity.objects.get(name=activity[1])

gulp/package.json
```diff
@@ -30,6 +30,7 @@
     "gulp-if": "^2.0.0",
     "gulp-jasmine": "^2.2.1",
     "gulp-util": "^3.0.7",
+    "history": "1.13.1",
```
koonsies: @darrint  any objections to this include? if not, this is ready for merge imo

darrint: It's fine, especially since this is just pinning a version.

Before we get to production I'd like to implement something like node.js shrinkwrap.

PR #1946, opened 2015-12-08
Allow use of django-compressor on qc.

mklauber:


aspidites:
Looks fine to me. @tdphillips you've used django compressor before. Can you confirm those variable names are correct?

deploy/settings.dseo_qc_admin.py
```diff
@@ -4,6 +4,8 @@
 from dseo_celery import *
 from secrets import REDIRECT_QC, REDIRECT_STAGING, ARCHIVE_STAGING
 
+COMPRESS_ENABLED = True
+COMPRESS_OFFLINE_MANIFEST = 'manifest.json'
```
tdphillips: Unless it's being overridden somewhere, this should be the default.

PR #1948, opened 2015-12-09
Use django-compressor for gulp files.

mklauber:
Is this everywhere we refer to the compiled files from gulp?

### DIFF
--- a/templates/manageusers/index.html
+++ b/templates/manageusers/index.html
@@ -1,5 +1,6 @@
 {% extends "base-react.html" %}
 {% load compress %}
+{% load staticfiles %}
 
 {% block site-title %}
     <title>My.jobs - {{ company.name }} - Manage Users</title>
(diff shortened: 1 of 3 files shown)


templates/manageusers/index.html
```diff
@@ -34,9 +35,9 @@
 SteelToe: https://github.com/jclem/steeltoe -->
 <!-- <script src="{{ STATIC_URL }}steeltoe.js"></script> -->
 
-
+{% compress js %}
 <script src="{{ STATIC_URL }}bundle/vendor.js"></script>
```
tdphillips: Won't this need to be the {%static%} tag?

aspidites:
Seems fine, pending @tdphillips's comment.

templates/manageusers/index.html
mklauber: It's probably a good idea to use it, but running locally, it didn't need it.  I'll update it anyway, since {% static %} is a good habit imo.

PR #1949, opened 2015-12-09
Add text-only to saved search [PD-1870]

tdphillips:
# Testing
Create a partner saved search with this button checked.
2015-12-09-113755_646x815_scrot

View that saved search.
2015-12-09-113813_679x138_scrot

Press the send button.
2015-12-09-113827_359x254_scrot

If you're using a client that prettifies the email (like Gmail does), click this dropdown and then select "Message text garbled?". The link being on the same line as other text in this screenshot and the next is a bug and has been fixed since these screenshots were taken.
2015-12-09-114003_794x351_scrot

Ta-dah! The en-dashes are being added somewhere between Exchange and Gmail. If sent directly to a Gmail account, you'll see an en-dash first but a normal dash on the following screenshot. If sent to my Exchange account and then forwarded to Gmail, I get this.
2015-12-09-114029_1182x492_scrot


mysearches/models.py
```diff
@@ -80,6 +80,8 @@ class SavedSearch(models.Model):
     # Custom messages were created for PartnerSavedSearches
     custom_message = models.TextField(max_length=300, blank=True, null=True)
 
+    text_only = models.BooleanField(default=False, blank=True)
```
tdphillips: I have no doubt that this will be promoted to normal saved searches in the future. YAGNI but this is better than the alternative of adding it to PSS and then doing umpteen migrations to fix it later.

aspidites:
I'm not sure that "Text Only" is obvious for the user. I wonder if we need help text or a blurb in the Tips box to explain what this option is doing.

aspidites:
I'd say we should fix the en-dash before merging.

mypartners/tests/test_views.py
```diff
@@ -962,7 +966,28 @@ def test_create_new_saved_search(self):
...
+    def test_create_new_saved_search(self):
+        """
+        Test initial partner saved search invitations as HTML.
+        """
+        self._create_new_saved_search()
+        body = mail.outbox.pop().body
+        self.assertTrue(BeautifulSoup(body, 'html.parser').find(),
+                        "This saved search email contains no html but should")
+
+    def test_create_text_only_saved_search(self):
```
aspidites: Wait..if we are testing create text only saved search, shouldn't we be asserting that there aren't any html tags rather than throwing the html parser at it?

mysearches/forms.py
```diff
@@ -31,6 +31,12 @@ def __init__(self, *args, **kwargs):
         choices = make_choices(self.user)
         self.fields["email"] = ChoiceField(widget=Select(), choices=choices,
                                            initial=choices[0][0])
+
+        initial = kwargs.get("instance")
+        # TODO: Rework text_only overrides when this is promoted to SavedSearch
+        text_only_kwargs = {'widget': HiddenInput(),
+                            'initial': initial.text_only if initial else False}
```
aspidites: Initial takes a string *or* a boolean?

mysearches/forms.py
```diff
@@ -56,6 +62,9 @@ def clean(self):
         cleaned_data = self.cleaned_data
         url = cleaned_data.get('url')
 
+        cleaned_data['text_only'] = self.instance.text_only
+        self._errors.pop('text_only', None)
```
aspidites: Why do we have to remove text_only related error messages?

mypartners/tests/test_views.py
tdphillips: So I should check for < and > instead?

universal/helpers.py
```diff
@@ -185,8 +189,94 @@ def add_pagination(request, object_list, per_page=None):
     return pagination
 
 
-def send_email(email_body, email_type=settings.GENERIC,
-               recipients=None, site=None, headers=None, **kwargs):
+class HTMLTextExtractor(HTMLParser):
+    """
```
aspidites: Can we get an example of input and output?

mysearches/forms.py
tdphillips: Inertia. That's how feed was handled.

I'm improving this mess right now.

mypartners/tests/test_views.py
aspidites: NVM. I read the assertion now. I see that find() on plain text returns a None object, which is good enough for me, so +1. Sorry about that.

tdphillips:
Changed "Text only" on the form to "Send as a plain text email"

solr/management/commands/build_myjobs_solr_schema.py
```diff
@@ -155,7 +155,8 @@ def handle(self, *args, **options):
                     # so far.
                     if field_type == 'OneToOneField' or \
                             any(s in field.attname
-                                for s in ['password', 'deactivate_type']):
+                                for s in ['password', 'deactivate_type',
```
aspidites: if you move the any up a line, you can get rid of the `\`

universal/helpers.py
```diff
@@ -185,8 +186,112 @@ def add_pagination(request, object_list, per_page=None):
...
+    Sample output:
+    '''
+    \n\nI'm a saved search!\n\nfoobar.com\n\nI can include \tlinks
+    '''
+
+    Caveats:
+    - We remove whitespace at the beginning and end of each line but,
+        as demonstrated, we did not remove the tab that was inside the anchor.
+    - This started with two endlines. I'm banking on end users not caring but
+        it can be fixed if desired.
```
koonsies: This mentions using two endlines, but I'm seeing two \n being added to the beginning of the text return. Is this intentional? I may be missing something here.

tdphillips: I should have called it a newline, which is what \n is.

koonsies: Yup. The wording just confused me as to the intent. So, whatever the HTML input, the output should always have two newlines at the beginning?

universal/helpers.py
```diff
@@ -185,8 +186,112 @@ def add_pagination(request, object_list, per_page=None):
...
+    def handle_starttag(self, tag, attrs):
+        """
+        Extracts hrefs from anchors and inserts them into our output list
+        if they are not present.
+        """
+        if tag == 'a':
+            attrs = dict(attrs)
+            href = attrs.get('href')
+            if href and href not in self.extracted:
+                if not self.last_entry.endswith('\n\n'):
```
koonsies: Ah, I see where the beginning newlines comes from. I'm assuming that since the first tag doesn't have has a last_entry, thus no newlines preceding it, it appends two newlines at the start. Is this desired?

tdphillips: I don't know how I didn't catch that. I can fix it but it's not really breaking anything.

koonsies: Might be something that gets called out in the future. Better to fix now than be annoyed later.

PR #1950, opened 2015-12-09
[PD-2018] Unicode on custom pages

aspidites:
# Description
**HOTFIX**
This converts the CustomPage table in mysql to utf-8 so that unicode can be added to the content of custom pages. 

There is no backward migration because we shouldn't have a reason to convert the table back to swedish_ci (I vaguely remember @JLMcLaugh telling me to not worry about a backward migration the last time I had to do this). 

I've included screenshots in the jira ticket for the curious.

### DIFF
--- a/seo/migrations/0030_unicode_custom_pages.py
+++ b/seo/migrations/0030_unicode_custom_pages.py
@@ -1,0 +1,453 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        """Change seo_custompage table to utf-8"""
+
+        # CustomPage inherits Django's FlatPage
+        db.execute("ALTER TABLE django_flatpage "
+                   "CONVERT TO CHARACTER SET utf8 "
+                   "COLLATE utf8_general_ci")
+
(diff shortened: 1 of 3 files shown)


seo/tests/test_views.py
```diff
@@ -2629,6 +2629,19 @@ def test_flatpage(self):
         self.assertEqual(response.status_code, 200)
         self.assertNotEqual(secret_code, -1)
 
+    def test_unicode(self):
```
tdphillips: If this fails, you might need to add another annoying line to seo/tests/setup

aspidites: On it. Thanks @tdphillips .

seo/migrations/0030_unicode_custom_pages.py
```diff
@@ -0,0 +1,448 @@
...
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        """Change seo_custompage table to utf-8"""
+
+        db.execute("ALTER TABLE seo_custompage "
```
tdphillips: FlatPage has title and content fields. Do those need to be unicode-capable as well?

aspidites: The unit tests checks against the content field explicitly.

aspidites:
Reverted my last commit, which was an experiment to see if subclassed models updated the inherited fields when updating collation -- they don't.

PR #1952, opened 2015-12-09
Moved migration to make hotfix happy

aspidites:
**Hotfix**
Because of #1950 going against master, this migration needs to come afterward, otherwise once qc propagates forward, master will complain because of duplicate names.


aspidites:
NVM. Looks like my local master was out of sync. prm_access is gone.

PR #1954, opened 2015-12-09
[PD-1955] Add Saved Search List Widget

koonsies:
## Goal

To add saved search list secure blocks widget for use with cross site verify logic. This PR also represents part of the investigation into the state of cross site verification.

## Testing
NOTE: This currently only works for Docker environments due to limitations of cross-site functionality on regular localhost. Outside of docker, if you can setup port forwarding so that your secure instance resolves to secure.my.jobs rather than secure.my.jobs:8000, go for it.

Next, head to SeoSite and make secure.my.jobs a parent of my.jobs

Following that, head to admin -> myblocks -> Saved Search List Widget Block -> Add

Leave the defaults for everything except for element id, which is `saved-search-list`
- 0 for Span and Offset

Navigate to my.jobs and click "search". The list will be on the right if you..
* Are logged in
* Have saved searches


myblocks/models.py
```diff
@@ -398,13 +398,32 @@ def context(self, request, **kwargs):
         return {
             'user': user,
             'search': search,
-            'success': success_email,
-            'is_pss': hasattr(search, 'partnersavedsearch'),
```
aspidites: Do we not care about identifying a saved search as a partner saved search anymore?

koonsies: Sorry, I should have annotated. I added a property to the saved_search base model for retrieving that info

myjobs/cross_site_verify.py
```diff
@@ -77,8 +77,10 @@ def xrw_ok(xrw):
 
 def get_site(domain):
     sites = SeoSite.objects.filter(domain=domain)
-    if len(sites) != 1:
-        return None
+    if len(sites) == 0:
```
aspidites: Are we meaning to raise an exception here? Would that not throw an HTTP 500 on production?

myblocks/models.py
aspidites: Ah, OK. I Quick grep suggests that there are a few templates that use this, but I don't see those templates updated here.

koonsies: This context method applies only to the saved search widget used in secure blocks API.

myjobs/cross_site_verify.py
koonsies: Yup. The exception is caught and handled by the calling function and returns a 404. There's an integration test that captures this bit.

myblocks/models.py
aspidites: +1

myjobs/cross_site_verify.py
aspidites: Ah ok.

myjobs/cross_site_verify.py
```diff
@@ -44,7 +44,7 @@ def guess_child_domain(host, origin, referer):
     raise DomainRelationshipException('no-child-info')
 
 
-def fail():
+def return_404():
     return HttpResponse(status=404, content="not found")
```
tdphillips: I know this wasn't added in this PR but I think the reason we're returning an HttpResponse rather than raising an Http404 should be documented.

koonsies: Good call... But I don't know. @darrint?

darrint: Best I can remember is either it was needed for the decorator (unlikely) or I was preparing to show more detailed error messages in a related response. (Which doesn't make sense here anyway.) I'm not married to this code so if an exception works and is more idomatic then let's use that.

tdphillips: I assumed that it might have reduced payload size as raising an exception would render our default error page but I didn't want to put words into your mouth.

aspidites: I'd be curious to see the difference. Ironic that less code would produce a larger response.

aspidites: Well, considering that removing this was reversed, I assume it's safe to say we at least have enough info to at least note why it exists now?

koonsies: Yup. Good call, I'll put a comment in.

lemniscate613:
I started reviewing this, but I'm waiting to get microsites set up with the new Docker environment.

templates/myblocks/blocks/savedsearcheslistwidget.html
```diff
@@ -0,0 +1,16 @@
+{% if user.is_authenticated and saved_searches.count %}
+<div class="direct_rightColBox direct-action-box">
+<div id="de-myjobs-widget">
+    <strong>Searches</strong>
+    <ul>
+        {% for search in saved_searches|slice:":5" %}
+        <li><a href="{{search.url}}">{{search.label}}</a> - {{search.last_sent|default:"Not yet sent"}}</li>
```
aspidites: We use "Never" elsewhere:

2015-12-16-093106_640x112_scrot

aspidites:
@koonsies , following your instructions, I get:
2015-12-16-093714_956x1056_scrot

It's trivial enough to enter these values, but either:
- the instructions should be updated to note these required fields
- if these fields are meant to be automatically populated, they are not.

aspidites:
Should probably also note that this is for www.my.jobs, not secure.my.jobs, so PROJECT in settings needs to be 'dseo' to test this and domain=www.my.jobs should be passed as a query parameter (or hosts file should be updated)

aspidites:
I'm not getting desired results after following your instructions above:
2015-12-16-094714_1006x614_scrot

I'm pointing to a local solr instance, but I would suspect that the list of saved searches should work independently of job search results. 

As you can see, I have saved searches:
2015-12-16-094859_644x445_scrot

koonsies:
Closing pull request pending resolution of qc/staging redirect issues.

default_settings.py
```diff
@@ -645,3 +645,7 @@
 EXCLUDED_VIEW_SOURCE_CACHE_KEY = 'excluded_view_sources'
 
 CUSTOM_EXCLUSION_CACHE_KEY = 'custom_excluded_view_sources'
+
+ENV_URL_PREFIXES = ['qc', 'staging']
+
```
koonsies: Addtion of URL prefixes to settings files. In case we expose additional environments in the future

default_settings.py
```diff
@@ -645,3 +645,7 @@
...
+RESTRICT_SECURE_BLOCKS_TO_STAFF = True
```
koonsies: Flag for whether or not secure blocks will display for non staff users

myjobs/cross_site_verify.py
```diff
@@ -19,13 +19,18 @@ def extract_hostname(url):
     if url is None:
         return None
     else:
-        return urlparse(url).hostname
+        return remove_test_prefix(urlparse(url).hostname)
+
+
+def remove_test_prefix(url):
+    url_split = url.split('.', 1)
+    return url_split[1] if url_split[0] in settings.ENV_URL_PREFIXES else url
```
koonsies: Remove URL prefix for determining site relationships.

templates/job_listing.html
```diff
@@ -72,6 +72,7 @@ <h3 class="direct_highlightedText">
     <a href="#top" class="direct_mobileJumpLink">{% blocktrans %}Return to top{% endblocktrans %}</a>
 
 {% endif %}
+<div data-secure-block-id="test-saved-search-list"></div>
```
koonsies: Added for this to be testable in QC. This will be an empty div for all members other than staff users

koonsies:
closing to fix tests

koonsies:
fixed tests, restriction logic

lemniscate613:
:+1:

PR #1955, opened 2015-12-09
Use the jenkins settings file.

mklauber:


aspidites:
But why? I mean, what problem does this solve?

mklauber:
I need to use a settings file to compress the files, if I'm going to compress them on the jenkins server and push them to s3.  Whichever settings file I use, needs to have `COMPRESS_ENABLED` set to `True`.  Previously, I was using the myjobs_staging.py settings file, since it already had the setting.  

However, In looking at things, I think the issue is the manifest is not being pushed to the webservers, not that the manifest is being based on old files in s3.

PR #1957, opened 2015-12-10
Meta-Documentation

aspidites:
- started writing a styleguide for documentation
- increased depth of tables of contents
- added an internal table of contents to the javascript styleguide with backlinks

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -7,7 +7,7 @@ Welcome to My.jobs's documentation!
 ===================================
 
 .. toctree::
-  :maxdepth: 3
+  :maxdepth: 2
 
   developers/index
   faq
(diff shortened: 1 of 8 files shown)


doc/source/index.rst
```diff
@@ -7,7 +7,7 @@ Welcome to My.jobs's documentation!
 ===================================
 
 .. toctree::
-  :maxdepth: 2
+  :maxdepth: 3
```
koonsies: I'm not sure about making the depth changes. To my eyes, it makes the table of contents a bit overwhelming. This will be especially true as we add more modules.

aspidites: Are you saying to uniformly keep depth at 2, that no depth should have changed, or that 3 is too much  for the main table of contents only?

koonsies: My biggest issue is with the main ToC. That being said, it was made a bit better when I switched to the classic format. The linespacing of the current template is a bit large.

aspidites:
@koonsies  how about now? : 
2015-12-11-131405_1916x1056_scrot

PR #1962, opened 2015-12-11
Fix import_jobs cycles.

mklauber:


import_jobs/__init__.py
```diff
@@ -19,10 +19,10 @@
 from xmlparse import DEv2JobFeed
 from seo.helpers import slices, create_businessunit
 from seo.models import BusinessUnit, Company
-import tasks
 from transform import hr_xml_to_json, make_redirect
 import re
 
+from .solr import add_jobs, chunk, delete_by_guid
```
aspidites: I'd rather we didn't use relative imports.

import_jobs/__init__.py
```diff
@@ -34,7 +34,7 @@
 FEED_FILE_PREFIX = "dseo_feed_"
 
 
-def update_job_source(guid, buid, name, clear_cache=False):
+def update_job_source(guid, buid, name):
```
aspidites: Why did we remove clear_cache?

tasks.py
```diff
@@ -777,6 +777,8 @@ def task_clear_bu_cache(buid, **kwargs):
 def task_update_solr(jsid, **kwargs):
     try:
         import_jobs.update_solr(jsid, **kwargs)
+        if kwargs['clear_cache']:
```
aspidites: If clear_cache isn't found this will throw an error, no? Rather if 'clear_cache' in kwargs or 'if kwargs.get('clear_cache', False)

postajob/models.py
```diff
@@ -17,6 +17,8 @@
 from location_data import countries, all_regions, country_list, state_list
```
aspidites: To this whole file: YESS!!!

import_jobs/__init__.py
```diff
@@ -19,10 +19,10 @@
 from xmlparse import DEv2JobFeed
 from seo.helpers import slices, create_businessunit
```
aspidites: Would it make sense to move the seo-specific stuff into import_jobs/seo.py?

redirect/actions.py
```diff
@@ -1,4 +1,88 @@
-import redirect.helpers
```
tdphillips: Was this seriously causing issues with imports?

aspidites:
:+1:

PR #1964, opened 2015-12-14
Removed military stuff from non-military pages.

aspidites:
Basically we were loading extra data that wasn't being used. No new tests because this is a template-only change. The only way I can think of to test what this is doing is to render every template and ensure that the json file is only used on the military template, but that seems pointless and would necessitate that any new template would require an update of that test.

### DIFF
--- a/mysearches/tests/local/careers.html
+++ b/mysearches/tests/local/careers.html
@@ -94,8 +94,6 @@ function goalClick(goalURL, the_url){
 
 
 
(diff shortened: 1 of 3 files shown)


mysearches/tests/local/careers.html
```diff
@@ -94,8 +94,6 @@
 
 
 
-<script src="//d2e48ltfsb5exy.cloudfront.net/content_ms/files/military_family_loc.json"></script>
```
tdphillips: This (and jobs.html) are just local copies of pages on production used in testing.

PR #1965, opened 2015-12-14
[PD-1856] Custom invitation reasons

aspidites:
# Description
At the moment, we send send specific invitation emails depending on whether or not the invitation is associated with a saved search or a group. This PR abstracts that functionality from the template and moves it to the backend, in addition to allowing for a custom reason to be provided in the email. 

# Key Changes
- `registration.models.Invitation.send` now takes an optional `reason` parameter
- `myjobs.models.User.send_invite` now takes an optional `reason` parameter
- `registration.models.Invitation` no longer automatically sends an email on creation
- Conditional messages inside of the invitation email template has been moved to the `send` method.
- New tests to cover each of these different template use cases. Note that I didn't give the tests a custom failure message as the default was more informative than anything I could have come up with (I tried).

# Testing
You can either:
- use the UI to generate the different invitation emails:
    - create a saved search
    - with ROLES_ENABLED set to False, create a company user
    - ~~with ROLES_ENABLED add a user to a role [1]~~
- generate the three conditions from the the './manage.py shell' prompt and call user.send_invite:
    - with a user and company, call user.send_invite
        - with and without a role
        - with and without a reason
    - create an invitation with and without a reason:
        - attach a saved search and call send
        - attach a group and call send

~~[1] I don't remember if the User Management frontend allows inviting of users yet, so this may not yet be possible from a UI~~

### DIFF
--- a/myjobs/models.py
+++ b/myjobs/models.py
@@ -719,7 +719,13 @@ class User(AbstractBaseUser, PermissionsMixin):
             :email: The email address of the potential user being invited.
             :company: The company inviting the user.
             :role_name: The name of the role (optional) the user is to be
-            assigned.
+                        assigned.
+            :reason: The readon for the invite, included in the email body. If
+                     a ```role_name``` is provided but reason is not, then the
+                     recipient will be notified that they are being invited to
+                     assume that role for the ```company``` which was provided.
+                     If neither is provided, a generic invitation email is sent
+                     instead.
 
         Output:
             The invited user if sucessful, otherwise None.
@@ -742,13 +748,10 @@ class User(AbstractBaseUser, PermissionsMixin):
         invitation = Invitation.objects.create(
             inviting_user=self, inviting_company=company, invitee=user)
 
-        if not reason:
-            if role_name:
-                reason = "as a(n) %s for %s." % (role_name, company)
-            else:
-                reason = "."
+        if not reason and role_name:
+            reason = "as a(n) %s for %s" % (role_name, company)
 
-        invitation.send(reason)
+        invitation.send(reason + ".")
 
         if role_name:
             if settings.ROLES_ENABLED:
--- a/myjobs/models.py
+++ b/myjobs/models.py
@@ -710,7 +710,7 @@ class User(AbstractBaseUser, PermissionsMixin):
             else:
                 return is_company_user and company.member
 
-    def send_invite(self, email, company, role_name=None):
+    def send_invite(self, email, company, role_name=None, reason=""):
         """
         Sends an invitation to the `email` in regards to `role_name` for
         `company`.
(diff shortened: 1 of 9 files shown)


koonsies:
On the footnote: The user UI portion does not properly send emails. Only the admin/command line methods will work in this instance.

registration/models.py
```diff
@@ -184,20 +184,41 @@ def save(self, *args, **kwargs):
...
+        activiation_profile, _ = ActivationProfile.objects.get_or_create(
+            user=self.invitee, email=self.invitee.email)
+
+        if not reason:
+            if self.added_saved_search:
+                reason = ("in order to begin receiving their "
+                                     "available job opportunities on a "
+                                     "regular basis.")
+            elif self.added_permission:
+                reason = ("in order to help administer their "
```
koonsies: There needs to be a space here before "in". It runs together with the end of the reason, otherwise.

registration/models.py
```diff
@@ -184,20 +184,41 @@ def save(self, *args, **kwargs):
...
+                                     "recruitment and outreach tools.")
+
+        if activiation_profile.activation_key_expired():
+            activiation_profile.reset_activation()
+            ap = ActivationProfile.objects.get(pk=activiation_profile.pk)
 
         context = {'invitation': self,
-                   'activation_key': ap.activation_key}
+                   'activation_key': activiation_profile.activation_key,
+                   'reason': reason + "."}
```
koonsies: The period is already present at the end of the reason. This results in "outreach tools.."

aspidites: > On the footnote: The user UI portion does not properly send emails. Only the admin/command line methods will work in this instance.

Do what now?

registration/models.py
aspidites: Fixed this by updating the template instead.

registration/models.py
aspidites: Ah. I'll update the instructions then.

koonsies:
Tests nooooooo

aspidites:
Templates are stupid...

templates/registration/invitation_email.html
```diff
@@ -6,42 +6,18 @@
...
-            administer {{ invitation.inviting_company.name }}'s recruitment
-            and outreach tools.
-        </p>
+    <p>
+        {% if inviting_user_name %}
+            {{ inviting_user_name }}
+        {% else %}
+            {{ invitation.inviting_user.email }}
+        {% endif %}
+        of {{ invitation.inviting_company.name }} has invited you to join My.jobs{{ reason }}
```
koonsies: The space is missing here again. It seems the most commit removed it?

myjobs/models.py
```diff
@@ -739,12 +745,18 @@ def send_invite(self, email, company, role_name=None):
...
             email=email, send_email=False, in_reserve=True)
         Invitation = get_model('registration', 'Invitation')
-        Invitation.objects.create(
+        invitation = Invitation.objects.create(
             inviting_user=self, inviting_company=company, invitee=user)
 
+        if not reason and role_name:
+            reason = "as a(n) %s for %s" % (role_name, company)
+
+        invitation.send(reason + ".")
```
koonsies: I'm still getting double periods at the end of emails. Is this the cause?

Dear Josevius Kooie,
Josevius Kooie of Verizon has invited you to join My.jobsin order to help administer their recruitment and outreach tools..

You can log in here.

Sincerely,
My.jobs on behalf of Josevius Kooie  


P.S., ignore the name..

aspidites: @koonsies, false alarm, right?

koonsies:
Yup, false alarm. :+1: I found no further issues with this PR

PR #1967, opened 2015-12-14
[PD-1958]  tags interface

darrint:
This implements David's prototype tags interface without drag and drop. I have added autocomplete as that was intended to be implemented, although it was not a part of the prototype.

The CSS is lifted almost verbatim from the prototype. The behavior is implemented as React components. There is a working python API behind this component.

Also present is a new common directory scheme for common JS code. I have written documentation for all the new code in common/. I did not experiment with any kind of doc tooling.

David's original ticket with some screenshots, etc. is here: https://directemployersfoundation.atlassian.net/browse/PD-1428

From @aspidites comment below:

- [ ] ~~there are times when a tag doesn't show up when filtering despite it showing up pre-filter~~ pushing to PD-2040
- [ ] ~~if I start typing a tag that doesn't exist, rather than clear on enter, the erroneous tag remains (vs. other behavior pertaining to create tag permissions)~~ pushing to PD-2040
- [x] add ticket for drag and drop: PD-2048
- [x] placeholders
- [x] already added tags aren't highlighted when clicked
- [x] the input box doesn't span the entire add box
- [x] tags aren't colored when they should be

Future work:

* [[PD-2039]](https://directemployersfoundation.atlassian.net/browse/PD-2039) what to do about partner tags and the contacts datasource
* [[PD-2040]](https://directemployersfoundation.atlassian.net/browse/PD-2040) improve autocomplete with regard to "hidden" items
* [[PD-2048]](https://directemployersfoundation.atlassian.net/browse/PD-2048) add drag and drop


koonsies:
Interface available by navigating as follows:
* Use gulp to build essential JS files
* Run ./manage.py reset_dynamic_data
* Navigate to https://secure.my.jobs/reports/view/dynamicoverview
* Click PRM
* Click Contacts
* Click Unaggregated
* Click Contact CSV

Tags interface should appear

Not sure if I am out of the loop, but I had trouble locating this.

koonsies:
I am unclear as to the goal of this ticket. I understand that drag and drop is not a function included in this PR, but should this module otherwise be functional? In testing, I cannot get more than one of my tags to appear.

darrint:
I get lots of tags. I suspect that there is something odd about the tags you have. We can investigate in person when you have time.

darrint:
After looking at what @koonsies saw, I concluded that the code in this PR is working as designed. His questions are valid and I've created two new tickets to start looking at them.

gulp/src/common/array.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * Return a new array with the results of calling inputFn with calls to
+ * sepFn between.
+ *
+ * sepFn: function(item from input, index in new array)
+ * inputFn: function(last item from input, index in new array)
+ * input: array of items
+ */
+export function intersperse(sepFn, inputFn, input) {
```
aspidites: This isn't being used.

darrint: I wrote it, used it, then realized I really needed flatMap. Since it has tests I kept it for now. I'll probably kill this whole file in favor of just making lodash available to our code.

gulp/src/common/ui/tags/AndOrContainer.js
```diff
@@ -0,0 +1,65 @@
...
+import {flatMap} from 'common/array';
+import {TagBrowser} from './TagBrowser';
+
+
+/**
+ * Box of collected items with a search box for adding more.
+ */
+export function AndOrContainer(props) {
+  const {andGroup, addTag, removeTag, getHints} = props;
+  const content = flatMap((orGroup, index) => [
```
aspidites: It appears we are flat mapping a list of singletons, which begs the question: Why not just map over individual items?

aspidites:
Thanks for the instructions @koonsies

aspidites:
Side note: At what point is mobile meant to work?

aspidites:
I'm not sure I'd consider this usable yet:
- certain manipulations of tags makes tags disappear
- there are times when a tag doesn't show up when filtering despite it showing up pre-filter
- if I start typing a tag that doesn't exist, rather than clear on enter, the erroneous tag remains [1]
- You mention that drag-n-drop isn't done,  but there doesn't seem to be any future work covering it, which essentially makes the top half of this widget no-op since you have to manually type tags anyway
- correct me if I'm wrong, but I thought I vaguely remembered double

Stuff different from the original prototype and not mentioned above (but not sure if it's really a focus yet:
- "Type Item and hit Enter" label missing, so it's not obvious I'm supposed to type to add tags
- already added tags aren't highlighted when clicked
- the input box doesn't span the entire add box
- tags aren't colored when they should be

Prototype for reference:
2015-12-16-145425_632x475_scrot

The prototype still runes if you migrate seo back to 18 and visit /reports/view/andor

[1] If you go to tag management or any other tag widget on PRM, you'll see that invalid tags are removed if you don't have "create tag" permissions.

darrint:
I've addressed these concerns either by making changes or setting up new tickets. See description above.

tdphillips:
I'm not seeing anything related to tags.

2015-12-22-082152_671x468_scrot

tdphillips:
The previous appears to be an issue with compressor that we're looking into at the moment.

darrint:
@tdphillips and I looked into this and it turned out to be related to compression settings in his development environment. We don't fully understand exactly what happened but I'm willing to say it doesn't affect this ticket.

tdphillips:
I'm betting it's an issue with cached templates. After restarting runserver while debugging, this works whether we've got one compress block or two. This should never crop up in production as Apache gets restarted on deployment.

PR #1972, opened 2015-12-15
Comma conditionals for Job Listings (PD-1919)

misterjonscott:
Found a suitable set of conditions in the myblocks implementation of
the searchresult.html template and used it to add the comma logic to
the current detail page.


aspidites:
I'm not pointing to solr. There a screenshot demonstrating how this changes output?

templates/job_detail.html
```diff
@@ -32,20 +32,25 @@
...
-          <span itemprop="name">{{the_job.company}}</span>
+            <span itemprop="name">{{the_job.company}}</span>
           </span>
           <span class="direct_highlightedText">
             <span itemprop="title">{{ the_job.title|cut:"?"|cut:"~" }}</span> in 
             <span itemprop="jobLocation" itemscope itemtype="http://schema.org/Place">
               <span itemprop="address" itemscope itemtype="http://schema.org/PostalAddress">
-                {% if the_job.city != None %}
-                <span itemprop="addressLocality">{{ the_job.city }}</span>
+                {% if the_job.state_short %}
```
JLMcLaugh: We should keep using the_job.state. Switching to state_short is outside the scope of the ticket.

templates/myblocks/blocks/searchresult.html
```diff
@@ -18,14 +18,14 @@ <h4 itemprop="title" {% if featured %}style="background-image:url('//d2e48ltfsb5
                             {% endif %}
                             <span class="hiringPlace" itemscope itemtype="http://schema.org/Place">
                                 <span itemprop="address" itemscope itemtype="http://schema.org/PostalAddress">
-                                    {% if job.state_short %}
+                                    {% if job.state %}
```
JLMcLaugh: This shouldn't be changed either.

aspidites:
Need a regression test. It's a template, but it should be trivial enough to render that template with various contexts and ensure the comma appears where it needs to.

templates/job_detail.html
haliscooper: Correct me if I'm wrong, but I don't see where the state ever gets written out in this logic...

haliscooper: > Correct me if I'm wrong, but I don't see where the state ever gets written out in this logic...
> And a span open is left hanging from line 42?

Ignore, git cuts it off...

JLMcLaugh:
@aspidites I agree.

misterjonscott:
I've updated based on comments and added a test.

aspidites:
@misterjonscott  once this is opened, you should probably resync with quality control, as you're diff is showing changes from other developers which are already in QC.

PR #1973, opened 2015-12-16
Apparently we still can't assign admins from a UI.

aspidites:
Adding more faqs. At some point some of this information will be converted into a guide, but I'd rather all the pieces line up before I waste time doing that.


lemniscate613:
"I think I'm having an existential crisis."

Even though the answer should have been, "take a deep breath and make yourself a sam'ich," I'm going to go ahead and merge this.

:+1:

PR #1975, opened 2015-12-16
[PD-1919] Adding comma to locations.

misterjonscott:
Adding commas to location string in job detail templates, and tests to verify that they're being rendered appropriately.

### DIFF
--- a/seo/tests/test_views.py
+++ b/seo/tests/test_views.py
@@ -17,7 +17,6 @@ from django.contrib.redirects.models import Redirect
 from django.template import Template, Context
 from django.template import RequestContext as TemplateContext
 from django.test.client import RequestFactory
-from django.test.utils import override_settings
 from django.utils.http import urlquote
 from django.core.urlresolvers import reverse
 
@@ -1720,29 +1719,26 @@ class SeoViewsTestCase(DirectSEOTestCase):
         Test to make sure job location is printed with commas, appropriately.
         """
         SeoSite.objects.get(id=1).delete()
-        ats = factories.ATSSourceCodeFactory.build()
-        ats.save()
-        gac = factories.GACampaignFactory.build()
-        gac.save()
-        site = factories.SeoSiteFactory.build(
+        ats = factories.ATSSourceCodeFactory()
+        gac = factories.GACampaignFactory()
+        site = factories.SeoSiteFactory(
             google_analytics_campaigns=gac,
             view_sources=factories.ViewSourceFactory(id=1),
             id=1)
-        site.save()
         site.ats_source_codes.add(ats),
         site.special_commitments.add(factories.SpecialCommitmentFactory(id=1))
-        site.save()
-        view_source = factories.ViewSourceFactory.build()
-        view_source.save()
-        special_commitment = factories.SpecialCommitmentFactory.build()
-        special_commitment.save()
+        view_source = factories.ViewSourceFactory()
+        special_commitment = factories.SpecialCommitmentFactory()
         
         # Job lookup by guid, check for full location.
         resp = self.client.get(
             u'/indianapolis-in/retail-associate-розничная-ассоциированных/11111111111111111111111111111111/job/',
             HTTP_HOST='buckconsultants.jobs')
         self.assertEqual(resp.status_code, 200)
-        self.assertContains(resp, 'itemtype="http://schema.org/PostalAddress"><span itemprop="addressLocality">Indianapolis</span>, <span itemprop="addressRegion">Indiana</span><meta itemprop="addressCountry" content="United States')
+        self.assertContains(resp, 'itemtype="http://schema.org/PostalAddress">'\
+                                  '<span itemprop="addressLocality">Indianapolis</span>, '\
+                                  '<span itemprop="addressRegion">Indiana</span>'\
+                                  '<meta itemprop="addressCountry" content="United States')
         
         # Job lookup by guid, check for empty state case.
         self.conn.delete(q="*:*")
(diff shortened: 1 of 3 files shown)


seo/tests/test_views.py
```diff
@@ -1715,6 +1715,81 @@ def test_title_boost(self):
         test_search_not(self)
         test_title_boost(self)
 
+    def test_job_detail_heading(self):
+        """
+        Test to make sure job location is printed with commas, appropriately.
+        """
+        SeoSite.objects.get(id=1).delete()
+        ats = factories.ATSSourceCodeFactory.build()
```
aspidites: The save lines after *Factory.build() are redundant. Actually, build is redundant, as it's the default build strategy, so *Factory() is sufficient. In this case: `ats = factories.ATSSOurceCodeFactory()`

seo/tests/test_views.py
```diff
@@ -1715,6 +1715,81 @@ def test_title_boost(self):
...
+        ats.save()
+        gac = factories.GACampaignFactory.build()
+        gac.save()
+        site = factories.SeoSiteFactory.build(
+            google_analytics_campaigns=gac,
+            view_sources=factories.ViewSourceFactory(id=1),
+            id=1)
+        site.save()
+        site.ats_source_codes.add(ats),
+        site.special_commitments.add(factories.SpecialCommitmentFactory(id=1))
```
aspidites: `save()` is implicit after an `add()` operation, thus redundant.

misterjonscott:
Updated with shortened lines, and changes based on @aspidites feedback.

seo/tests/test_views.py
```diff
@@ -1715,13 +1714,89 @@ def test_title_boost(self):
...
+        view_source = factories.ViewSourceFactory()
+        special_commitment = factories.SpecialCommitmentFactory()
+        
+        # Job lookup by guid, check for full location.
+        resp = self.client.get(
+            u'/indianapolis-in/retail-associate-розничная-ассоциированных/11111111111111111111111111111111/job/',
+            HTTP_HOST='buckconsultants.jobs')
+        self.assertEqual(resp.status_code, 200)
+        self.assertContains(resp, 'itemtype="http://schema.org/PostalAddress">'\
```
tdphillips: You probably need to compare against `resp.content`

JLMcLaugh: assertContains takes the response as the first argument, not a string.

misterjonscott: I'm confused about what's being requested, here.  Are there changes required?  This was tested.

seo/tests/test_views.py
```diff
@@ -1715,13 +1714,89 @@ def test_title_boost(self):
...
+                                  '<span itemprop="addressCountry">United States</span></span>')
+
     def test_job_detail(self):
         """
         Test that job detail pages return no server errors and that objects
         all work.
 
         """
-        SeoSite.objects.get(id=1).delete()
```
JLMcLaugh: This line needs to go back in.

misterjonscott: I've added it back.

seo/tests/test_views.py
```diff
@@ -1759,6 +1834,7 @@ def test_job_detail(self):
             u'/indianapolis-in/retail-associate-розничная-ассоциированных/11111111111111111111111111111111/job/',
             HTTP_HOST='buckconsultants.jobs')
         self.assertEqual(resp.status_code, 200)
+        self.assertContains(resp, 'Indianapolis, Indiana, United States')
```
JLMcLaugh: I think this is matching the title, not the heading we're looking for. assertContains doesn't ignore html.

misterjonscott: I removed this line, as it's the one we added to the test for response code, and it's not part of the test for location formatting.

seo/tests/test_views.py
aspidites: Same feedback I gave Daniel: I unit test should be testing a single unit of code. That there are three different scenarios we are testing here necessitates three different tests. I'm ok with testing a precondition though.

seo/tests/test_views.py
```diff
@@ -1715,6 +1714,83 @@ def test_title_boost(self):
...
+        view_source = factories.ViewSourceFactory()
+        special_commitment = factories.SpecialCommitmentFactory()
+        
+        # Job lookup by guid, check for full location.
+        resp = self.client.get(
+            u'/indianapolis-in/retail-associate-розничная-ассоциированных/11111111111111111111111111111111/job/',
+            HTTP_HOST='buckconsultants.jobs')
+        self.assertEqual(resp.status_code, 200)
+        self.assertContains(resp, 'itemtype="http://schema.org/PostalAddress">'\
+                                  '<span itemprop="addressLocality">Indianapolis</span>, '\
```
aspidites: The `\`s are redundant. Strings within parens are automatically concatenated. Eg:

```python
# this
('foo bar '
  'baz buzz')
# becomes this
('foo bar baz buzz')
```

aspidites:
Maybe I commented on the wrong ticket before? Thought I mentioned unit tests being independent, but I don't see that comment anymore. Oh well. I was just mentioning that different scenarios should have different test cases. As it stands now, if the empty city/state case fails, we have to get passed the others in order to iteratively fix/test it. Further, if the one test case fails, it isn't obvious which of the cases failed until you examine the test case.

misterjonscott:
Added messages to subsection tests within heading test, to make review of failures easier.

misterjonscott:
The change from None to '' didn't help.  We've been unable to see what's happening differently in Jenkins' build vs local dev.  Still working.

misterjonscott:
Indiana is still being rendered into the page, in the Jenkins build, where it's not in the local build.  That's why this test is breaking...  Unsure why that would happen.

seo/tests/test_views.py
```diff
@@ -1715,6 +1715,89 @@ def test_title_boost(self):
...
+        }
+        solr_docs_copy[0].update(kwargs1)
+        self.conn.add(solr_docs_copy)
+        resp = self.client.get(
+            u'/indianapolis-in/retail-associate-розничная-ассоциированных/11111111111111111111111111111111/job/',
+            HTTP_HOST='buckconsultants.jobs')
+        # import ipdb; ipdb.set_trace()
+        self.assertEqual(resp.status_code, 200, msg="Not receiving 200 response in test for no state")
+        self.assertContains(resp, 'itemtype="http://schema.org/PostalAddress">'
+                                  '<span itemprop="addressLocality">Indiandapolis</span>'
```
tdphillips: Should this contain a comma? According to Jenkins, the response contains a comma.

aspidites:
What ended up being the fix?

tdphillips:
Removing the tests. :/

aspidites:
That's... Less than satisfactory

PR #1978, opened 2015-12-18
Tweak docker docs.

darrint:
@lemniscate613 did a run without me looking over his shoulder. Here are a few details that didn't make the merge which happened while I was out sick.


lemniscate613:
With the caveat that I've run through these instructions as they've evolved, not fresh from start to finish:
:+1:

PR #1979, opened 2015-12-18
[PD-2017] Added military_family-loc.json to military template.

aspidites:
A previous pr removed this static file from seo_base. Apparently it was never in the military template. I'm not sure where it's supposed to go, so I assume this is sufficient. 

@tdphillips, mind double checking my use of `{% static %}`. Never used it before, so I'm not 100% I've gotten the syntax correct.

### DIFF
--- a/templates/seo/military.html
+++ b/templates/seo/military.html
@@ -1,5 +1,6 @@
 {% extends "seo_base.html" %}
 {% load i18n %}
+{% load staticfiles %}
 {% block directseo_searchbox %}{% endblock %}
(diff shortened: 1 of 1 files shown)


templates/seo/military.html
```diff
@@ -8,6 +8,7 @@
 {% endblock %}
 
 {% block extra-js %}
+<script src="{% static "military_family_loc.json" %}"></script>
```
tdphillips: You need to `{% load staticfiles %}` I believe.

PR #1980, opened 2015-12-21
Add deployment files to .gitignore

tdphillips:


.gitignore
```diff
@@ -70,3 +69,10 @@ settings.py
 
 # Documentation
 doc/build
+
+# Deployment
```
tdphillips: These are added when we collect static files.

.gitignore
```diff
@@ -3,8 +3,7 @@
 *.db
 *.flv
 *.wmv
-static/*.gz
```
tdphillips: These two were woefully inadequate - gz files are being generated more than one level deep in the static directory.

I checked. We have no gz files in the repo currently.

koonsies:
Not sure what is up with the failing check here.

tdphillips:
Jenkins timed out contacting pypi.

PR #1981, opened 2015-12-21
Misc Changes to Docker, Secure Blocks

koonsies:
This is a small PR.

I had not intended on lumping the changes to Docker with the changes to secure blocks, but it happened that way.

## Docker

Fixed/Add functions restartsecure and restartmicrosites. These functions will stop and remove any running containers for either and subsequently restart their respective container.

## Secure Blocks

Modified the logic for determining the url of the parent site to no longer affix/remove the qc/staging prefix, but rather modify https/http based on the environment. This is due to the fact that QC and Staging do not use HTTPS.

I also added the exception/authentication error from cross site verify to the response. I may remove this in a later PR, but it will help as I attempt to get cross site verify running on QC.

### DIFF
--- a/dk2/dockerdev.sh
+++ b/dk2/dockerdev.sh
@@ -104,9 +104,15 @@ backgroundstop() {
 }
 
 restartsecure() {
-    docker stop secure.my.jobs
-    docker rm secure.my.jobs
-    runsecure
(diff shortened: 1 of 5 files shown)


lemniscate613:
Works for me. :+1:

PR #1982, opened 2015-12-21
Modification to Saved Search, Tweak for QC/Staging Communication

koonsies:
Another tweak to see if I can get QC to behave properly. Reinstated qc/staging prefix modifications along with https/http resolution logic.


lemniscate613:
:dancer:

PR #1983, opened 2015-12-21
Update unsubscribe links, add tests [PD-1865]

tdphillips:
Email being sent (Note the "Unsubscribe" link up at the top - just a happy little accident). Content was removed for ease of screenshotting.
2015-12-21-165014_465x295_scrot

Resultant page.
2015-12-21-165305_420x155_scrot


mysearches/tests/test_forms.py
```diff
@@ -147,17 +147,6 @@ def test_user_only_linked_to_contact_after_pss_created(self):
         contact = Contact.objects.get(id=contact.id)
         self.assertTrue(contact.user)
 
-    def test_initial_email_has_unsubscription_options(self):
```
tdphillips: This test was moved (with modification) to the test file added elsewhere in this PR.

mysearches/urls.py
```diff
@@ -3,19 +3,23 @@
...
-    url(r'^view/more-results/$', 'more_feed_results', name='more_feed_results'),
+    url(r'^view/more-results/$', 'more_feed_results',
+        name='more_feed_results'),
     url(r'^view/validate-url/$', 'validate_url', name='validate_url'),
     url(r'^view/save-digest/$', 'save_digest_form', name='save_digest_form'),
     url(r'^view/edit/$', 'edit_search', name='edit_search'),
     url(r'^view/edit$', 'edit_search', name='edit_search'),
     url(r'^view/save/$', 'save_search_form', name='save_search_form'),
     url(r'^view/unsubscribe$', 'unsubscribe', name='unsubscribe'),
+    url(r'^view/unsubscribe/confirm$', 'unsubscribe_confirmation',
```
tdphillips: I might have swapped these (confirmation on view/unsubscribe/, actual unsubscribe on view/unsubscribe/<insert_foo_here>) but that would probably break existing emails.

templates/mysearches/email_base.html
```diff
@@ -54,9 +54,6 @@
                     </td>
                     <td colspan=1 style="padding: 8px; line-height: 20px; border-top: 1px solid #ddd;">
                         {% block footer-right %}
-                        <a href="https://secure.my.jobs{% url 'unsubscribe' %}?id={{ saved_search.id }}&verify={{saved_search.user.user_guid}}">{% trans 'Deactivate this saved search' %}</a><br>
```
tdphillips: The four templates that use this template override this section anyway (except when they didn't) so I went ahead and removed the default options.

tdphillips:
@fezick I'm a little curious how you think we should do the confirmation page. I've seen links, dropdowns, and buttons.

lemniscate613:
I think simple is best here. You mentioned buttons. Maybe something like this:

```
Unsubscribe:

This Search | All Searches | Everything
```

If we were selling hard to job seekers, I'd suggest we use ideas from this roundup: http://blog.hubspot.com/marketing/email-unsubscribe-pages

Specifically, we could take JCrew's approach. Something like:

unsub-jcrew

koonsies:
Is this still WIP? Or should I review it now?

tdphillips:
WIP.

tdphillips:
Buttons added.

2015-12-28-090243_388x281_scrot
2015-12-28-090256_609x158_scrot

koonsies:
Everything seems to work properly, with two slight caveats.

A 404 is returned if a user tries to unsubscribe to a email they have already unsubscribed for.. This could be replaced with a notification that they are already unsubscribed (may ease confusion). 

Also, a 500 server error is thrown if the id is omitted from the querystring (https://secure.my.jobs/saved-search/view/unsubscribe)

The 404 is obviously working as intended, and the 500 is a edge case. Up to you whether these should be addressed. 

Aside from that, :+1:

tdphillips:
The 500 should only arise if someone is messing with urls. Turning it into a 404 or redirecting to somewhere appropriate might be a good idea.

As for unsubscribing to an already-unsubscribed saved search, I'll see what I can do.

tdphillips:
The 500 should now be a 200 (if on the confirmation url) or 404 (if on the actual unsubscribe url).

If you unsubscribe from an already-unsubscribed search, we treat it like a normal unsubscription but we don't update the database.

Would those be adequate?

aspidites:
I like everything except "Everything". While I know what everything means, does a user?

tdphillips:
Would "All MyJobs Communication" or something similar be adequate?

aspidites:
@tdphillips  perfect!

tdphillips:
Button text changed. I didn't make new screenshots as they're otherwise identical.

I did fix the tablet issue, however - the buttons were in a span8 but didn't really need to be as there's no sidebar. They're now in a span12.

aspidites:
:+1, just waiting on tests and confirmation that none of the UI people have further comments about @lemniscate613  comments. @fezick , @misterjonscott?

lemniscate613:
I'd also change "I've changed my mind" to say simply, "Cancel."

fezick:
I agree on just saying "Cancel", otherwise the button approach looks good.

tdphillips:
Button text changed to "Cancel"

PR #1984, opened 2015-12-21
Hopefully the last PR for Cross Site

koonsies:
## Overview
I forgot a period.. I added that period (well, I had to change a bit about how the whole statement was structured.. but I really just added a period)

## The Problem
My url contained no period

## The solution
I added a period

## Testing
See if the URL has a period


lemniscate613:
It could be a little larger (it's hard to see), but in general you did a good job with this period.

:+1:

PR #1990, opened 2015-12-28
Create Home Page Template for Cross Site

koonsies:
## Overview

I am creating a new home page template for the purposes of testing and developing secure blocks widgets. This template will not be tied to any existing configurations, but will be a copy of the home_page_listing.html home page. This allows testing of secure blocks widgets without including non-functional hooks on existing templates.

## Testing

The new template contains hooks for both saved_search_list and saved_search, but only saved_search_list exists in QC. To verify that the list hook works, change the template in my.jobs configuration to use CS_TEST_home_page_listing.html .. If you have not done so already, do the steps below as well

### Addtl. Cross Site Testing Steps (from old PR)
NOTE: This currently only works for Docker environments due to limitations of cross-site functionality on regular localhost. Outside of docker, if you can setup port forwarding so that your secure instance resolves to secure.my.jobs rather than secure.my.jobs:8000, go for it.

Next, head to SeoSite and make secure.my.jobs a parent of my.jobs

Following that, head to admin -> myblocks -> Saved Search List Widget Block -> Add

Leave the defaults for everything except for element id, which is `saved-search-list`
- 0 for Span and Offset

Navigate to my.jobs and click "search". The list will be on the right if you..
* Are logged in
* Have saved searches

### DIFF
--- a/templates/home_page/CS_TEST_home_page_listing.html
+++ b/templates/home_page/CS_TEST_home_page_listing.html
@@ -1,0 +1,82 @@
+{% extends "seo_base.html" %}
+{% load i18n %}
+{% load seo_extras %}
+{% load cache %}
+{% cache 600 joblisting %}
+
+{% block direct_extraHeaderContent %}
+<link rel="stylesheet" href="/style/def.ui.dotjobs.results.css" type="text/css">
+{% endblock direct_extraHeaderContent %}
+
+{% block publisher %}
+    {% if site_config.publisher and "network" not in site_tags %}
+        <link href="https://plus.google.com/{{ site_config.publisher }}" rel="publisher" />
+    {% else %}
+        <link href="https://plus.google.com/100406566063094579991" rel="publisher" />
+    {% endif %}
+{% endblock %}
+
+{% block directseo_title %}
+    {{site_title}} 
+{% endblock directseo_title %}
+
+{% block rss_feed %}
+<link rel="alternate" type="application/rss+xml" title="Jobs - {{site_title}}" href="{{request.build_absolute_uri}}feed/rss">
+{% endblock rss_feed %}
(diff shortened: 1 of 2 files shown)


templates/home_page/CS_TEST_home_page_listing.html
```diff
@@ -0,0 +1,82 @@
...
+
+{% block directseo_right_hand_column %}
+{% if default_jobs or featured_jobs %}
+<a name="filters"></a>  
+<div id="direct_shareDiv" class="direct_rightColBox direct_shareSolo" role="menu">
+    <h3>{% trans "Share" %}</h3>
+    {% include 'includes/add_this.html' %}<!--temp rollback of my.jobs sharing [JPS 10.4.12]-->
+</div>
+<a href="#top" class="direct_mobileJumpLink">{% blocktrans %}Return to top{% endblocktrans %}</a>
+{% if "network" in site_tags or site_config.show_saved_search_widget %}
```
JLMcLaugh: If these are the only 4 lines changing, we should add an empty block to home_page_listing.html and have this template extend off of that.

koonsies: I can definitely see where you're coming from. This introduces a good amount of redundant code. That being said, this template will only exist temporarily and only for widget testing and is to be removed in a subsequent PR. Adding a block on home_page_listing actually increases the footprint of this code and introduces the possibility of more dead code that would have to be removed. Beyond that, I'm worried about use cases in the future where we may want to display widgets in other areas of the page. In this case, I feel it's clear cut on what code to remove and it provides more flexibility for development.

koonsies: If I'm mistaken, let me know. Otherwise, Darrin is ready to merge.

PR #1991, opened 2015-12-28
GitHub doc update

koonsies:
Addition of a GitHub page to documentation, along with Guidance Requested steps.


doc/source/developers/contributing/git_procedures.rst
```diff
@@ -0,0 +1,45 @@
...
+GitHub Procedures
+=================
+
+Creating a PR
+=============
+
+When creating a PR on GitHub, be sure to add the JIRA ticket number in the
+title. This allows JIRA to track the progress of the ticket and keep the
+reporter/watches updated. Titles should be in the following format:
+**[PD-9999] GitHub Ticket Description**
```
tdphillips: If we're changing this, it needs to be reflected here as well.

https://github.com/DirectEmployers/private-docs/blob/bc8cad47ea34199dc3677442516837d55b775382/developer/workflow_style_guide.rst#git-and-version-control

koonsies: Hmm.. I hadn't noticed that the existing Git documentation was in private docs. Perhaps I should just open a PR to those instead of the public docs. The question is.. Do we want to keep it that way? There's nothing about Git guidelines/style guides that needs to be private.

tdphillips: All documents from Google Docs were placed there just to get them out of Docs. Not all of them are necessarily secret but some of them may or may not have contained sensitive info.

PR #1992, opened 2015-12-28
[PD-2051]  Pin NPM Dependencies

darrint:
# The Fix

`npm shrinkwrap` takes a snapshot of a working node_modules directory tree and assures that the same package versions will be used in production and other developers' machines.

My reading of the npm docs suggests that no changes are needed in deployment. Just the presence of the `npm-shrinkwrap.json` file is enough to change the behavior of `npm install`. With the file present it will simply reproduce the shrinkwrap.

If you are a developer and need to change the dependencies see the new `README.md` file in `gulp/`.

# Testing

I suggest that we walk though all the current React apps.

# Future

Diffs within `npm-shrinkwrap.json` should help us gauge the impact of future JS dependency changes.

### DIFF
--- a/gulp/README.md
+++ b/gulp/README.md
@@ -1,0 +1,19 @@
+# Updating Dependencies
+
+If you update dependencies in `package.json`, take care to make sure the
+`npm-shrinkwrap.json` file is consistent.
+
+To add or upgrade a package use:
+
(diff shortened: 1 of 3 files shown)


lemniscate613:
:+1:

PR #1994, opened 2015-12-29
Testing concept to handle company names dynamically

misterjonscott:
Testing the simplest way to convey the company name from Django to React, for use in the NUO component's title area.

### DIFF
--- a/gulp/src/manageusers/main.jsx
+++ b/gulp/src/manageusers/main.jsx
@@ -27,6 +27,7 @@ export class App extends React.Component {
       usersTableRows: [],
       callRolesAPI: this.callRolesAPI,
       callUsersAPI: this.callUsersAPI,
+      company: companyName,
     };
     this.callActivitiesAPI = this.callActivitiesAPI.bind(this);
     this.callRolesAPI = this.callRolesAPI.bind(this);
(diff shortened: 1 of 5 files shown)


misterjonscott:
@koonsies and @lemniscate613 should check this out.

koonsies:
Seems to be an elegant way to handle the symptom that led to the creation of the ticket. We may want to eventually expand out into more of a base header component, but I could see this working in the interim. I believe @aspidites had done something similar a while back

aspidites:
Except now you have to do the same thing for reporting and user management as company name is hard-coded there as well. I think the proper solution is to consolidate the entry templates for each app and move the row containing the company name up a level from the container react component to the template.

aspidites:
And @koonsies is right, I experimented with this method right before going on vacation.

misterjonscott:
@aspidites Moments after submitting this PR, @haliscooper suggested we need to redesign this interface, anyway, but the approach to transferring the data to React would be the same.  I'm not sure those other pages you've mentioned are in the scope of this ticket, but if they should be added, I'm happy to proceed with that.

haliscooper:
Yeah, gotta tackle those other two pages/groups of pages, for now

misterjonscott:
@haliscooper I added the dynamic company name to the User Management section, but it's not present in Reporting, yet.  I talked with @darrint about this, and he can reference this ticket when he's ready to add this detail.  I've also moved this variable declaration to the React base template, so it can be used by any React templates, ongoing.

PR #1996, opened 2015-12-30
Quick regex experiment to fix specific ~230 PEP errors

lemniscate613:
Putting this up to see Jenkins report. No feedback necessary.


doc/source/conf.py
```diff
@@ -68,13 +68,13 @@
 
 # The language for content autogenerated by Sphinx. Refer to documentation
 # for a list of supported languages.
-#language = None
+# language = None
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
-#today = ''
+# Today = ''
```
tdphillips: Why was the case altered here and at least six other locations?

PR #1999, opened 2016-01-05
fix jslint error, companyName is coming in from the base template

lemniscate613:
The dynamic companyName feature worked, but JSLint was pitching a fit:
```
"companyName" is not defined
```


misterjonscott:
Worked with Daniel on this, and this looks to be the best and most appropriate solution.

PR #2002, opened 2016-01-06
Import jobs documentation

mklauber:


doc/source/developers/index.rst
```diff
@@ -13,4 +13,4 @@ Developer Documentation
   styleguides/index
   testing/index
   contributing/index
-  etl_imports
```
tdphillips: Should this file be removed as well? It's currently at developers/etl_imports as well as developers/import_jobs/etl_imports

aspidites:
The colons at heading level are probably not necessary. Looks good otherwise. Can't confirm or deny quality of document as I'm not familiar with the process, but it looks rather organized overall.

doc/source/developers/index.rst
tdphillips: :arrow_up:

PR #2003, opened 2016-01-06
Stop double-unquoting [PD-2060]

tdphillips:
With a bit of hand-testing, I determined that some Django magic happens earlier in request processing which unquotes the thing we're unquoting. That results in "%2523" getting unquoted all the way to "#" rather than stopping at "%23". Since octothorpes have meaning as far as urls go, this is bad.


tdphillips:
Opened for testing - Authorize.net is failing locally for me and I want to know if it's just my issue.

PR #2006, opened 2016-01-06
[PD-1383] New Datasources

darrint:
This PR creates working reports for the Partners and Communication Types report types.

To test:

    manage.py reset_dynamic_data
    cd gulp/
    npm install
    gulp watch-tasks

Go to http://.../reports/view/dynamicoverview

Working:

* There should be one reporting type: PRM
* All three report types: Comm Records, Contacts, and Partners should work.
* All filter fields in the report types should work and provide meaningful help.

Not working:

* The report pages are still hideous and do not conform with our standards.

Review Guide:

* The bulk of this PR is the addition of the new partners and communication records report types.
* Moved the datasources to their own directory since each one is fairly large.
* There is some duplication between the datasource classes but the methods are subtly different between them. (IMHO further factoring would lead to very abstract code and inflexibility when creating a datasource for data that is not a Django model.)
* I did factor out a common DataSourceJsonDriver. It is responsible for adapting the datasources to the api views and was extremely similar.

- [x] Forward "Usability Nits" to @fezick for new tickets.
- [x] Replace `get_datasource_json_driver` with a plain old dictionary.
- [x] Use `abc` or otherwise document the datasource protocol.
- [x] Don't shadow `filter`.
- [x] Rename `$abbrev_qs` vars to be more descriptive.
- [x] Eliminate need for `'pass'` in `filter_key_types`.
- [x] Include `filter_key_types` in documentation.
- [x] Consider `unicodecsv` over `csv`.
- [x] pep8
- [x] download report: 'NoneType' object has no attribute 'strftime''
- [x] Clarify unit test docs.
- [x] Detect unicode in JoinFormatter fallback.
- [x] Expose tag `hex_color` for formatting later.


aspidites:
# Usability Nits
- can't filter partners by name
- it's not clear to me whether or not it is intended that "Run report' do anything at this point
- Despite that, I apparently have a generated report, but clicking on it gives me an attribute error, stating that 'NoneType' object has no attribute 'strftime''
- Navigation is still broken (ie, non-existent), which makes it **very** tedious to QC this thing
- In the contact report, there is a field named "Partners" that's a freeform text - invalid entries aren't automatically cleared (I mentioned this a while ago, and showed that it was possible to do so, since I do it with the tag widget site-wide for PRM)
- the forementioned issues with the date input are still present
- similar issue with state - without the dropdown, it isn't clear whether I should be typing "IN" or "Indiana" for instance, or whether "indiana" or "in" would fail, for instance.

I'll comment inline about code once I've studied this a bit more.

aspidites:
I'll comment on this though:
> There is some duplication between the datasource classes but the methods are subtly different between them. (IMHO further factoring would lead to very abstract code and inflexibility when creating a datasource for data that is not a Django model.)

I find this hard to believe, considering that the point of abstract code is to introduce flexibility. Concrete, hard-coded idioms, on the other hand, are by definition, inflexible.

gulp/src/common/ui/tags/AndOrContainer.js
```diff
@@ -56,6 +56,7 @@ export class AndOrContainer extends Component {
         highlights={highlights}/>
       {content}
       <AddNewBox
+        id="add-new-box"
```
aspidites: Do we want ID's here? I presume all "add_new_boxes" should act similar. If there is more than one instance of them, nothing about this id is unique enough to guarantee that this exact input will be selected.

myreports/datasources/__init__.py
```diff
@@ -0,0 +1,13 @@
+from myreports.datasources.contacts import ContactsDataSource
+from myreports.datasources.partners import PartnersDataSource
+from myreports.datasources.comm_records import CommRecordsDataSource
+from myreports.datasources.util import DataSourceJsonDriver
+
+
+def get_datasource_json_driver(datasource_name):
```
aspidites: Any reason to not use a dictionary here?

myreports/datasources/comm_records.py
```diff
@@ -0,0 +1,231 @@
...
+    Contact, Status, Location, Tag, ContactRecord, Partner)
+
+from myreports.datasources.util import filter_date_range
+
+from universal.helpers import dict_identity
+
+from django.db.models import Q
+
+
+class CommRecordsDataSource(object):
```
aspidites: The datasources all subclass object, which tells me nothing about how this class is supposed to work without reading the entirety of the class. Documentation would help, but I suspect that leveraging `abc`s or `NotImplementedError`s would help out a lot as well.

myreports/datasources/comm_records.py
```diff
@@ -0,0 +1,231 @@
...
+    def run(self, company, filter, order):
```
aspidites: We're shadowing the `filter` function, and we shouldn't.

aspidites:
None of this documented, which makes reviewing the code rather difficult.

myreports/datasources/comm_records.py
```diff
@@ -0,0 +1,231 @@
...
+                'key': t['name'],
+                'display': t['name'],
+                'hexColor': t['hex_color'],
+            } for t in names_qs]
+
+    def help_communication_type(self, company, filter, partial):
+        """Get help for the communication type field."""
+        comm_records_qs = self.filtered_query_set(company, filter)
+
+        ct_qs = (
```
aspidites: These variable names are very non-descriptive.

myreports/datasources/contacts.py
```diff
@@ -197,6 +113,15 @@ def __init__(self, date=None, locations=None, tags=None, partner=None):
         self.tags = tags
         self.partner = partner
 
+    @classmethod
+    def filter_key_types(self):
+        return {
+            'date': 'date_range',
+            'locations': 'pass',
```
aspidites: What does pass mean, and if it's what I think it is, would None not be a better fit, in particular because it serializes to "null"?

myreports/views.py
```diff
@@ -22,8 +22,9 @@
 from universal.decorators import has_access
 
 from myreports.datasources import get_datasource_json_driver
-from myreports.report_configuration import (
-    ReportConfiguration, ColumnConfiguration)
+
+from cStringIO import StringIO
+import csv
```
tdphillips: You may or may not want unicodecsv

gulp/src/common/ui/tags/AndOrContainer.js
darrint: the aria-* attributes want id's.

myreports/datasources/__init__.py
darrint: Warm fuzzy feeling. I'm not ready to commit to it working that way. Probably I shouldn't worry about it as it's only called in two places.

darrint:
Closing PR to address feedback.

myreports/column_formats.py
```diff
@@ -24,8 +28,12 @@ def __init__(self, between, inner_formatter=None):
             self.inner_formatter = inner_formatter
 
     def format(self, values):
-        return self.between.join(
-            self.inner_formatter.format(v) for v in values)
+        if isinstance(values, Iterable) and not isinstance(values, str):
```
aspidites: We're using Python 2, so this will fail on unicode:

```python
In [1]: isinstance(u"", str)
Out [1]: False
```

myreports/datasources/comm_records.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class CommRecordsDataSource(DataSource):
+    def run(self, company, filter_spec, order):
+        qs_filtered = self.filtered_query_set(company, filter_spec)
+        qs_ordered = qs_filtered.order_by(*order)
+        qs_distinct = qs_ordered.distinct()
+        return [self.extract_record(r) for r in qs_distinct]
+
+    def extract_record(self, record):
```
aspidites: is `extract_record` only used for exporting to CSV and similar, or is the dictionary used on the front-end? If so, I worry that you're leaving out color information from tags.

myreports/datasources/comm_records.py
```diff
@@ -0,0 +1,235 @@
...
+            'last_action_time': record.last_action_time,
+            'length': record.length,
+            'location': record.location,
+            'notes': record.notes,
+            'partner': record.partner.name,
+            'subject': record.subject,
+            'tags': [t.name for t in record.tags.all()],
+        }
+
+    def extract_user_email(self, user):
```
aspidites: Is there rhyme or reason to when you write another extract method? For instance, I would assume `extract_record` to be the end-all-be-all, but `extract_user_email` exists. That would then lead me to believe it was a per field basis, but I don't see any other extract methods. I'm thus led to believe that its rather arbitrary and adhoc. Is that so?

myreports/datasources/comm_records.py
```diff
@@ -0,0 +1,235 @@
...
+        qs_company = ContactRecord.objects.filter(partner__owner=company)
+        qs_live = (
+            qs_company
+            .filter(approval_status__code__iexact=Status.APPROVED)
+            .filter(archived_on__isnull=True))
+        qs_filtered = filter_spec.filter_query_set(qs_live)
+        return qs_filtered
+
+
+@dict_identity
```
aspidites: Still not in love with this `dict_identity` thing. I'll defer to the conversation we had on IRC about isinstance + a map or comprehension being sufficient and easier to grock than a decorator. 

Kudos for being all mathy, though (identity).

myreports/datasources/comm_records.py
darrint: Arbitrary. If I caught myself repeating the function across multiple datasources (which is inevitable) I would move it to util. In fact, I expect that to be common in the near future.

myreports/datasources/comm_records.py
darrint: It is export only. However, I should add the hexcolor attribute. Formats other than csv may want it.

myreports/datasources/base.py
```diff
@@ -0,0 +1,127 @@
+from abc import ABCMeta, abstractmethod
+
+
+class DataSource:
```
aspidites: Is it intentional that this isn't a new-style class (subclassing object)?

aspidites:
Thanks for updating the documentation.

myreports/datasources/base.py
darrint: Yes.

https://docs.python.org/2/library/abc.html

I'm following their examples. They don't say why. Maybe subclassing object is either in conflict with or redundant with declaring ABCMeta as a metaclass.

aspidites: Aah. Strange considering object is a superclass, not a metaclass. Seems legit though. so :+1:  on that front.

PR #2007, opened 2016-01-07
[PD-1901] job count cache

mklauber:
Ready for merge, imo.  See my second comment asking if anyone else can think of a reason why that would need to be delayed.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -286,7 +286,7 @@ def process_user_events(email):
     logs = EmailLog.objects.filter(email=email).order_by('-received')
     newest_log = logs[0]
 
-    filter_by_event = lambda x, num=None: [log for log in logs[:num]
+    filter_by_event = lambda x, num = None: [log for log in logs[:num]
                                            if log.event in x]
 
     max_errors = 3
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -788,6 +788,7 @@ def task_update_solr(jsid, **kwargs):
 def task_etl_to_solr(guid, buid, name):
     try:
         import_jobs.update_job_source(guid, buid, name)
+        BusinessUnit.clear_cache(int(buid))
```
mklauber: This is the new functionality

tasks.py
```diff
@@ -798,7 +799,7 @@ def task_etl_to_solr(guid, buid, name):
 def task_priority_etl_to_solr(guid, buid, name):
     try:
         import_jobs.update_job_source(guid, buid, name)
-        task_clear_bu_cache.delay(buid=int(buid), countdown=1500)
+        BusinessUnit.clear_cache(int(buid))
```
mklauber: Any reason to delay this further, rather than running immediately after indexing?

tasks.py
aspidites: so buid for import_jobs.update_job_source needs to be a string and for BusinessUnit.clear_cache needs to be an integer?

tasks.py
aspidites: I tried to git blame to see when that delay was added, but this file has moved around enough that it was hard for me to tell.

PR #2019, opened 2016-01-11
Fix bug with webpack and bootstrap-react

lemniscate613:
The bootstrap-react package was trying to load its own version of react and caused an error.


aspidites:
Any reason to not rely on the same version of React everywhere? I'm mostly concerned as we've had issues where Django admin expects one version of JQuery, most of secure relies on another, and the few remaining pages rely on yet another, which as you can imagine, made for some fun to find bugs.

darrint:
@aspidites No reason at all. This PR is a workaround to get back to that. @lemniscate613 and I are considering removing react-bootstrap as a dependency. This is one in a string of ugly problems it has caused.

PR #2020, opened 2016-01-11
[PD-2081] Invitation Context

aspidites:
First, note how this is going into a feature branch. That is intentional. The idea is that I wanted a generic way to generate different messaging for different contexts. Unfortunately, a message isn't always enough for certain contexts, so I've written this function such that it returns a context to be used by a template instead. 

To test this, simply open a shell, import invitation_context from universal.helpers and call it, passing an instance of a supported type as the sole argument (str, ,mysearches.models.SavedSearch, myjobs.modelsRole). 

In another ticket, I'll be using this to rewrite functionality that creates and sends invitation emails.


tdphillips:
Tests need to be updated with the preceding space. Otherwise, it appears to be good to go.

aspidites:
Tests are passing now.

PR #2025, opened 2016-01-13
Return app_access in /manage-users/api/roles/ API

lemniscate613:
The /manage-users/api/roles/ API returns, among other things, available and assigned activities. For each, the front-end needs the name and description, but it also needs app_access. This enables us to break up the activities by app for use in the accordion:

screen shot 2016-01-13 at 9 15 22 am

# Test

Run tests:
`dkm test myjobs.tests.test_manage_users`

Visit locally:
https://secure.my.jobs/manage-users/#/role/add?action=Add


koonsies:
Problems found in the edit role form in IE8. Accordion object do not properly collapse. As this has nothing to do with this PR, I am merging anyway.

PR #2026, opened 2016-01-13
[PD-2082] Update User.send_invite

aspidites:
# Description
I originally planned to turn `user.send_invite(email, company, role_name)` to `user.create_invitation(email, company).send(role)` before realizing that the original was doing more than just those to operations. In particular, if I went to the latter, a single call would then turn into about four or five. Given that the base `send` already has a lot of side effects within it (creating a user, activation profile, and communication record), I'm not as compelled to make the `send_invite` method pure. 

As such, the only thing this PR is doing is taking advantage of my `invitation_context` helper function any time we need to render an invitation email.


aspidites:
Closing while I wait for the other PR to be merged.

PR #2029, opened 2016-01-13
Update to Docs for Docker w/ Multiple VMs

koonsies:
A brief blurb following the investigation by me and @darrint. The steps included will allow an OSX user to set up a Windows VM to use in conjunction with the docker-machine VM used by docker on OSX.

P.S. I like hamburgers. :hamburger:


aspidites:
What about this is OSX-specific? It makes me thing separate instructions exist for other hosts, yet those instructions don't exist.

koonsies:
It may not be that different. The only difference is that docker runs in a VM on OSX, but not on Linux. If these steps are relevant to Linux distros, I can remove that tag. @darrint?

aspidites:
I suspect that from [here](https://github.com/DirectEmployers/MyJobs/pull/2029/files#diff-8f85d3f6ac2f2d4feb4ecf77523f41c9R230) to the end is the most relevant part of this doc.

lemniscate613:
:hamburger:

darrint:
I can confirm that I use the --natdnshostresolver1 on my IE VMs on Linux.

aspidites:
Clarification on my last comment: If the only thing that is mac specific is the lines before 230, then perhaps the rest of it should be moved out to a subjection so that it doesn't appear as though this only works on Mac. Alternatively, make this article exclusively about the natdnshostresolver as that seems to be the crucial bit (unless I've missed something), which would allow you to remove the (Mac) marker.

koonsies:
I updated the doc a bit. I didn't remove the extra information yet. I believe it can be useful in lieu of another doc with info on VirtualBox. Might save someone some time.

aspidites:
:+1:

PR #2031, opened 2016-01-13
Hide "Canada" if a province is available [ST-2153]

tdphillips:
Better diff:

https://github.com/DirectEmployers/MyJobs/compare/add-comma-to-thing...canada-location-fix?expand=1

P.S. I like hamburgers. :hamburger:


tdphillips:
The test fix regarding SNI hasn't left QC yet. Two tests are probably going to fail.

PR #2032, opened 2016-01-13
[PD-1941] Added more common links

aspidites:
Added more common links to the django admin interface:
2016-01-13-151736_516x292_scrot


tdphillips:
AJ gave this a :+1: in the ticket.

PR #2034, opened 2016-01-14
Constrain roles by current company

lemniscate613:
# Overview

This PR fixes a bug where, if a user belonged to multiple companies, all of their roles would display in various points in the UI. For example:

## Before

screen shot 2016-01-14 at 11 49 01 am

## After

screen shot 2016-01-14 at 11 49 40 am

## Before
screen shot 2016-01-14 at 11 49 22 am

## After
screen shot 2016-01-14 at 11 49 31 am

# Test

Run Django tests: `dkm test myjobs.tests.test_manage_users`

Build JS and start server:
`
dkgg build
dk runsecure
`

### DIFF
--- a/myjobs/tests/test_manage_users.py
+++ b/myjobs/tests/test_manage_users.py
@@ -396,10 +396,7 @@ class ManageUsersTests(MyJobsBase):
         # uses to make the request. Therefore, the API should only return two
         # roles
         self.assertEqual(len(roles_assigned),2)
-
-
-
-
+        
     def test_get_users(self):
         """
         Tests that the Users API returns the proper data in the proper form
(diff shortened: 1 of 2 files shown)


aspidites:
You say to run tests, but there aren't any new tests. As this was a backend change, it should be easy enough to write a tests that ensures the API isn't returning redundant information.

lemniscate613:
@aspidites I didn't include tests initially because I can see the argument both for and against. FYI, I didn't mean to indicate the PR had tests, but more, "run these, this PR won't break existing tests."

Anyway, I just added two that make sure the following views won't return roles related to companies not making the request:

* api_get_users
* api_get_specific_users

misterjonscott:
Pure genius.

PR #2035, opened 2016-01-14
[PD-1982][PD-2082] Model-Agnostic Invitations

aspidites:
The short of it is that with this PR, invitations are no longer directly tied to any models. I left the foreign keys that were there in place in order for existing invitations to work. Half of this code has already been reviewed (the bits about invitation_context), so it shouldn't be too difficult to look over.

To test, create a partner saved search or use the user management tool to invite yourself to the system. You should see that not only did you get an invitation, but that invitation is contextually relevant. For partner saved searches, you should see a custom message in the email if you provided one while creating it.


myjobs/models.py
```diff
@@ -749,34 +750,20 @@ def send_invite(self, email, company, role_name=None, reason=""):
...
-            else:
-                role_name_string = role_name
-            reason = "as a(n) %s for %s" % (role_name_string, company)
-
-        invitation.send(reason + ".")
-
+        context_object = ""
         if role_name:
             if settings.ROLES_ENABLED:
-                # role_name could be an array like ["Admin", "PRM User"]
```
aspidites: Talked to @fezick  and @JLMcLaugh: adding multiple roles should be a rare case, so it's ok to send multiple emails.

myjobs/models.py
```diff
@@ -974,3 +961,10 @@ def remove_activity(self, name):
             activity = None
 
         return activity
+
+
+@invitation_context.register(Role)
```
aspidites: This is the bit responsible for making sure the message for roles is formatted correctly.

PR #2037, opened 2016-01-15
User management box fix mulitple users

lemniscate613:
Different behavior on local and QC. Saving user with multiple roles was causing error in send_invite() helper function.

### DIFF
--- a/myjobs/models.py
+++ b/myjobs/models.py
@@ -720,7 +720,7 @@ class User(AbstractBaseUser, PermissionsMixin):
             :email: The email address of the potential user being invited.
             :company: The company inviting the user.
             :role_name: The name of the role (optional) the user is to be
-                        assigned.
+                        assigned. Can be either a string or array of strings.
             :reason: The readon for the invite, included in the email body. If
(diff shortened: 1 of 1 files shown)


myjobs/models.py
```diff
@@ -759,9 +759,14 @@ def send_invite(self, email, company, role_name=None, reason=""):
 
         if role_name:
             if settings.ROLES_ENABLED:
-                assigned_role = Role.objects.get(
-                    company=company, name=role_name)
-                user.roles.add(assigned_role)
+                # role_name could be an array like ["Admin", "PRM User"]
+                if isinstance(role_name, list):
+                    for role in role_name:
+                        user.roles.add(role)
```
tdphillips: If role_name is actually a list of strings, you'll need to grab the roles from the database.

koonsies: This can be done with user.roles.add(*roles)

lemniscate613: Thank you @tdphillips.

misterjonscott:
Provocative and entertaining.

PR #2038, opened 2016-01-15
User Management - redirect to users page

lemniscate613:
Previously, if you successfully edited a preexisting user, it would tell you such. This PR redirects you back to the users page.


misterjonscott:
Exquisite.  A masterpiece.

PR #2039, opened 2016-01-15
Allow filtering users on secondary email [PD-2053]

tdphillips:
This search would normally be empty.
2016-01-15-104244_1088x223_scrot

9001. Steps to test:
15. Open admin
300. Click "User"
2. Search for a secondary email
65535. ...
1337. Profit


tdphillips:
We don't have many admin tests. Might as well start now.

tdphillips:
There's an endemic issue with our tests involving settings.SITE. If you run this test by itself (or myprofile tests on their own), for instance, you'll run into it. As a whole, tests are "okay".

PR #2045, opened 2016-01-19
Hide "Canada" if a province is available (master) [ST-2153]

tdphillips:
@JLMcLaugh, looks like I actually based this off of master. There shouldn't be any issues.


tdphillips:
If this isn't reviewed within a week, it's going to production anyway.

Closing so it doesn't get merged before review.

PR #2048, opened 2016-01-20
[PD-1628] Convert saved search widget

koonsies:
## Overview

This PR covers the initial implementation of the saved search widget for secure blocks. Documentation is included. A lot of what I would write here is located in the secureblocks.rst file below.

## Testing

Most of the testing steps are found in the "Testing a Secure Block Widget" doc section. For ease of testing, ensure the following:

* my.jobs is a child of secure.my.jobs
* the home page of the production configuration for my.jobs is set to CS_TEST_homepage_listing.html
* a SavedSearchWidgetBlock is created with the element_id "saved-search" (recreate any old widgets to get the newest default template
* do not use the search button to execute a new search. This will load a different template than CS_TEST_homepage_listing.html and thus not display the widget. Instead, type in any desired q variables.
 * Example url: www.my.jobs?q=Awesome%20jobs (NOT www.my.jobs/jobs?q=Awesome%20jobs)

## Future work

* Create integration plan for adding saved search to standard home page templates

* Convert/Add CORS functionality to Saved Search API (PD-2066)


myblocks/fixtures/login_page.json
```diff
@@ -131,4 +131,22 @@
...
+        "name": "Test Saved Search",
+        "content_type": ["myblocks", "savedsearchwidgetblock"],
+        "offset": 0,
+        "updated": "2015-03-18T04:00:00Z"
+    }
+},
+{
+    "pk": 7,
+    "model": "myblocks.savedsearchwidgetblock",
+    "fields": {}
```
koonsies: adding savedsearchwidgetblock entry into test json

myblocks/models.py
```diff
@@ -383,7 +391,43 @@ def submit_btn_name(self):
...
+        """
+        Setup context of secure block by appending any kwargs to context
+        dict.
+
+        :param request:
+        :param kwargs: jQuery Data attributes provided from calling site
+        :return: kwargs dict to serve as context dictionary
+
+        """
+        return kwargs
```
koonsies: added this simple overwrite to the base class to be clear that this is functionality that should be present in all child classes

myblocks/models.py
```diff
@@ -383,7 +391,43 @@ def submit_btn_name(self):
...
+        Render template then append all required js tags (if applicable)
+
+        """
+        rendered_template = super(SecureBlock,
+                                  self).render_for_ajax(request,
+                                                        params)
+        static_string = '<script src="%s"></script>'
+        required_js = [static_string % js for js in self.required_js()]
+        return '%s %s' % (rendered_template, ''.join(required_js))
+
```
koonsies: add whatever is in required_js to the rendered template

myblocks/tests/test_secure_blocks.py
```diff
@@ -71,6 +70,24 @@ def test_secure_blocks_hiding_properly(self):
...
+        body = '{"blocks": {"test_saved_search": {}}}'
+        resp = self.make_sb_request(body)
+        self.assertEqual(resp.status_code, 200,
+                         msg="Expected 200, got %s. User was not able to "
+                             "retrieve blocks for test" % resp.status_code)
+        for js in savedsearch.required_js():
+            self.assertContains(resp, js,
+                                msg_prefix="Did not find %s in response,"
+                                           "missing required js" % js)
+
```
koonsies: didn't change a lot of python code here.. this verifies that the required_js inclusion is working

myblocks/views.py
```diff
@@ -80,8 +81,8 @@ def set_page(self, request):
 
 
 # The django csrf exemption should stay first in this list.
-@restrict_to_staff()
 @django_csrf_exempt
+@restrict_to_staff()
```
koonsies: it is worth noting that while restrict_to_staff is turned on, you cannot test the unauthenticated versions of the widgets. to test unauthenticated saved search widget, remove this tag temporarily

myblocks/views.py
```diff
@@ -113,7 +116,11 @@ def secure_blocks(request):
         if block is None:
             logger.warn("Failed block lookup: %s", element_id)
         else:
-            rendered = block.render_for_ajax(request, blocks[element_id])
-            response[element_id] = rendered
+            try:
+                block = block.cast()
```
koonsies: cast sets the class of block to the proper subclass.. yay, blocks weirdness!

static/saved-search.js
```diff
@@ -0,0 +1,120 @@
...
+  // and variables supplied by myjobs are persistent, so the experience
+  // for logged in users continues to be the same.
+  var error = error || "Something went wrong!"
+  $(blocks_widget_div).data("error", error);
+  reload_widget(remove_error_flag);
+}
+
+function remove_error_flag() {
+  $(blocks_widget_div).data("error", "");
+}
```
koonsies: callback function to be executed in ajax promise resolution

static/secure-block.168-30.js
```diff
@@ -54,19 +55,40 @@ function jsonp_secure_block(secure_block_url, data) {
...
-    secure_block(dashboard_url, request).fail(function(xhr, text, error) {
+    secure_block(saved_dashboard_url, request).fail(function(xhr, text, error) {
         console.error("dashboard fail: ", xhr, text, error);
     }).done(function(data, text, xhr) {
         $.each(data, function(key, value) {
         $("[data-secure-block-id=" + key + "]").html(value);
         });
+        if (typeof callback === "function") {
+          callback();
+        }
```
koonsies: checks if callback was passed in and executes it at the end of promise resolution

static/secure-block.168-30.js
```diff
@@ -54,19 +55,40 @@ function jsonp_secure_block(secure_block_url, data) {
   });
 }
 
-function load_secure_blocks(dashboard_url) {
-  var request = {};
-  $("*[data-secure-block-id]").each(function(i, block) {
-    element_id = $(block).data('secure-block-id');
-    request[element_id] = $(block).data();
-  });
```
koonsies: This code was refactored to be used in both load_secure_blocks and reload_secure_block

templates/myblocks/blocks/savedsearchwidget.html
```diff
@@ -1,3 +1,63 @@
...
+error - True/False based on if there was a form error
+success - True/False based on if an email had just been created
+user - Member's User object if logged in
+search - Existing search object if user already has this saved search
+
+REQUIRED TAGS:
+saved-search-form
+saved-search-btn
+ss-btn-link
+{% endcomment %}
```
koonsies: Attempt at clarity with comments and tags.

myblocks/views.py
```diff
@@ -113,7 +116,11 @@ def secure_blocks(request):
...
+                rendered = block.render_for_ajax(request, blocks[element_id])
+                response[element_id] = rendered
+            except Exception as ex:
```
tdphillips: Is it unknown what exception types may be used?

koonsies: Good question. It really depends. If the context or render_for_ajax functions throw exceptions, they could be anything, especially as more secure blocks widgets are created. The try/catch is here to ensure that one broken widget doesn't prevent all widgets from displaying, but still prints some useful information. (On that thought, perhaps I should convert this into a print_exc...?)

But as far as being able to use specific exception handling here.. I can't think of a good way to define them.

static/saved-search.js
```diff
@@ -0,0 +1,120 @@
...
+  // be a problem
+  if (typeof reload_secure_block !== "undefined"){
+    var element_id = $(blocks_widget_div).data('secure-block-id')
+    reload_secure_block(element_id, callback);
+  }
+}
+
+
+function create_saved_search() {
+  debugger;
```
tdphillips: Intentional?

koonsies: Erm... Yeesss?

aspidites:
Haven't gone through this yet, but I appreciate the level of detail involved in putting this together.

doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -32,4 +32,102 @@ call is made to the parent of the microsite (or itself, if it has no parent) to
 retrieve all secure blocks divs. The request is subject to the cross site
 verification step above.
 
-.. autofunction:: myblocks.views.secure_blocks
\ No newline at end of file
+.. autofunction:: myblocks.views.secure_blocks
```
aspidites: The paragraph before this one reads:

> The value of that key is tied to secure block...

Should that either be "tied to secure blocks" or "tied to a secure block"?

doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -32,4 +32,102 @@ call is made to the parent of the microsite (or itself, if it has no parent) to
...
+    :members:
+
+Testing a Secure Block Widget
+------------------------------
+
+To test a secure block widget, do the following:
+
+* Navigate to Django Admin
+
+* If parent site relationship is not set up, do the following:
```
aspidites: These are numbered steps. so an enumeration may be more fitting. you can use `#.` to keep from having to type numbers.

doc/source/developers/myblocks/secureblocks.rst
koonsies: Good call. I updated the wording to be more clear. Thanks for the input.

PR #2054, opened 2016-01-21
Speed up docker workflow osx

darrint:
I need someone on OSX to try this:

`dkgg watch-tasks watch`

Note how long it takes to build.

Save a few `.js` files. Note how long these builds take. Usually the first build takes a little longer than subsequent builds once `watch` gets caches warmed up.

`dk initscratchdata`

This is a one time thing. It will never be needed again unless you roast your docker storage somehow.

`dk optimizenode`

`dkgg watch-tasks watch`

Note build time.

Save a few `.js` files. If this worked then builds should be an order of magnitude faster.


.gitignore
```diff
@@ -21,6 +21,7 @@ static/redirect.css
 
 # Gulp stuff
 gulp/node_modules/
+gulp/node_modules
```
aspidites: Was it not enough to just remove the slash? Seems weird to have to ignore both

dk2/dockerdev.sh
```diff
@@ -42,6 +43,11 @@ initmysqldata() {
     docker create --name mysqldata darrint/mysqldata true
 }
 
+initscratchdata() {
+    docker build -t darrint/scratchdata scratchdata
```
aspidites: Is this going to work with your name being referenced directly?

koonsies:
A note about the steps listed on the PR.. You want to do the test on a separate branch first (probably quality-control) THEN checkout this branch and move forward. Otherwise, you receive the following:

Error response from daemon: no such id: scratchdata
  0.13s user 0.01s system 82% cpu 0.169 total

dk2/dockerdev.sh
darrint: Yes. In the future "darrint" should change to the hostname of a server where we can push and pull filesystem images.

.gitignore
darrint: Didn't think that through. Git didn't ignore the symlink (which doesn't resolve locally) without the /.

misterjonscott:
I'm seeing 'test' complete in ~3s, 'dev-bundle' in ~7s and 'lint' in ~11s, with <1s benefit after running 'optimizenode'

Discussed with @darrint and ready to assist with further testing.

darrint:
Rats.

darrint:
The command given at the end of the doc is approaching acceptable performance. In the future I can try some other approaches.

PR #2055, opened 2016-01-21
Update redirect to use cloudfront [PD-1988]

tdphillips:
Tested in environments resembling qc and production. Both function. Not sure if we want to start writing tests specific to static files.

### DIFF
--- a/redirect/models.py
+++ b/redirect/models.py
@@ -20,7 +20,7 @@ CHOICE_LIST = tuple((choice, choice) for choice in CHOICES)
 
 
 def make_yesno_image(view_source):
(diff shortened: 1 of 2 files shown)


redirect/models.py
```diff
@@ -18,6 +19,19 @@
 CHOICE_LIST = tuple((choice, choice) for choice in CHOICES)
 
 
+def make_yesno_image(view_source):
+    tag = '<img src="{img}" alt={alt}>'
```
koonsies: Why no quotes around "{alt}"?

tdphillips: You would have to ask past-me that. I'm just moving stuff around.

PR #2062, opened 2016-01-25
Make activities page dynamic

lemniscate613:
# Overview

The Activities view will expand as we add more app access types (currently we have two -- PRM and User Management).

# Review

```
dkgg build
```
View https://secure.my.jobs/manage-users/#/activities

I updated a test in this PR, so:

```
dkm test myjobs.tests.test_manage_users.ManageUsersTests
```

# Before Merging

I need to finish these before merging:

1. Pull in Troy's changes so all tests pass

2. I'm adding a test to ensure a user doesn't see activities related to apps for which they don't have access.

3. I'm refactoring database queries, for example with selected_related (thank you @JLMcLaugh).


tdphillips:
We've got underscore in our repo. Is groupBy not appropriate?

http://underscorejs.org/#groupBy

darrint:
https://lodash.com/docs#groupBy is what's available.

Something like this should work

    import collection from 'lodash-compat/collection';
    
    console.log(collection.groupBy([1, 2, 3, 4, 5], i => i % 2));

lemniscate613:
Actually both are available -- lodash with our new build system and underscore in static: https://github.com/DirectEmployers/MyJobs/blob/quality-control/static/underscore-1.3.1.min.js Their latest is 1.8.3.

aspidites:
Your instructions assume everyone is using docker.

registration/tests/test_views.py
```diff
@@ -46,9 +47,8 @@ def setUp(self):
 
         # Update the only existing site so we're working with
         # my.jobs for historical/aesthetic purposes.
-        site = SeoSite.objects.get()
-        site.domain = 'my.jobs'
-        site.save()
+        settings.SITE.domain = 'my.jobs'
```
JLMcLaugh: Changing values in settings.py directly tends to cause problems with other tests. Check out the django.tests.utils.override_settings decorator. We use it here and in a few other tests:
https://github.com/DirectEmployers/MyJobs/blob/dfa61eb203ab5de88122c44ff8bdce8734087e51/myjobs/tests/test_autoserialize.py#L21

tdphillips: We're not updating settings, we're updating a model instance that was shoved into settings.

aspidites: What Troy said. settings.SITE is just a handle to a model instance. We're updating that instance's domain to avoid having to refactor a bunch of tests. It's not pretty, but it's functional. If we really wanted to eliminate such hackery, I suspect we would have to refactor the entire seo test suite.

In other words, using `@override_settings` or `with self.settings()` won't help here.

koonsies: I was about to rebuttal and say that changing the state of an object referenced in settings is just like changing settings itself.. then I realized we're talking about tests, and in tests all DB related changes are automatically rolled back. It's always a pleasure working with people smarter than me!

koonsies:
It seems unrelated.. is the test that is failing here a transient error??

koonsies:
I tested the UI and it is functional.

tdphillips:
The failing test is one of the differences between QC and master. It's fixed in a branch I've got locally but I don't know how long it will take to get that up.

lemniscate613:
RE: lodash/underscore and groupBy(), I'll try to use helper methods like this in the future, but it doesn't seem essential to refactor this PR since it's already been tested. I added a comment to the code so it could be refactored in the future if need be.

RE: @tdphillips's comment on the failing test be okay and @koonsies's test, it sounds like this is good to merge...?

tdphillips:
That is not what I said. Merging broken tests is never a good idea.

tdphillips:
> I'll try to use helper methods like this in the future, but it doesn't seem essential to refactor this PR since it's already been tested.

Why should we ever refactor in that case if the current code is tested?

aspidites:
> RE: lodash/underscore and groupBy(), I'll try to use helper methods like this in the future, but it doesn't seem essential to refactor this PR since it's already been tested. 

1. We have plenty of code in production that technically passes tests but could do with a refactor. Project vision continued to move forward, and we thus never found ourselves with enough time to go back and refactor. That code has since  bit rotten or suffers tremendously speed wise. As an example [this](https://github.com/DirectEmployers/MyJobs/blob/quality-control/myreports/helpers.py#L18) little gem was a dirty hack to get around frequently changing requirements in production. I *still* want to get rid of that hack and do something more robust, but that will never happen. [Here](https://github.com/DirectEmployers/MyJobs/blob/quality-control/mypartners/helpers.py#L370)'s another example to prove that it's not just a one time thing.
2. If simply passing tests were enough to consider that something should be merged, we could just merge our own code as soon as they passes. Case in point: PR's aren't simply about resolving a ticket and passing tests. They are about receiving implementable feedback and, you guessed it, **implementing** said feedback. 
3. I'm certainly not the boss, and anyone (or everyone) can feel free to disagree with me, but there seems to be a rash of "we'll fix it later, here's a comment/jira ticket" remarks as of late, whci h Ithink is affecting the quality of our code. Personally, I'd like to see that stop. The only time I think spawning a new ticket is a reasonable course of action for a refactor or improvement is when a) that refactor would take enough time to considerably raise the point value of a ticket and b) when stuff is  broken in production NOW, effecting users, and a proper solution would take considerably more time to vet than a quick fix.

OK, I'm done ranting. Carry on.

lemniscate613:
RE: the failing test. I just chatted with @tdphillips, and the plan is to wait for him to merge his fix into QC. Then, I'll pull the latest QC into this branch and all should tests should pass.

RE: groupby() I can refactor if that's what makes sense. My vote is that it doesn't, but that's one vote. I'm prioritizing the other part of this task (another PR). It's about 80% of the work, and I was asked to complete it this week. I use ```TODO```'s as a mental note to change something, if it time allows. 

I see your points @aspidites. RE: 1, I want to raise awareness that, when deciding where to invest time, aesthetic satisfaction in code is only one part -- functionality, readability, reliability, business value, usefulness in the future, etc. are also important. 2 is a straw man. I'm going to try to use helper methods from lodash where appropriate in the future. 3 is a larger point.

gulp/src/manageusers/main.jsx
```diff
@@ -47,56 +49,41 @@ export class App extends React.Component {
...
-      // Create unique array of appIDs present
-      const appIDs = [];
-      for (let i = 0; i < parsedResults.length; i++) {
-        if (parsedResults.hasOwnProperty(i)) {
-          if (appIDs.indexOf(parsedResults[i].fields.app_access) === -1) {
-            appIDs.push(parsedResults[i].fields.app_access);
-          }
+      // Create an array of unique app_access_names
+      const appAccessNames = [];
+      for (const i in results) {
```
tdphillips: This should be a code smell. If the data were serialized a bit differently, we wouldn't need to iterate over our data len(results) + (len(results) * len(apps)) times.

gulp/src/manageusers/main.jsx
```diff
@@ -47,56 +49,41 @@ export class App extends React.Component {
...
-
-      // Parse API JSON response
-      const parsedResults = JSON.parse(results);
-
-      // Sort activities by app_access
-      parsedResults.sort(function sortActivities(a, b) {
-        return a.fields.app_access - b.fields.app_access;
-      });
+      // Create an array of tables, each a list of activities of a
+      // particular app_access
```
JLMcLaugh: Where are we filtering by company app access? Restricted activities shouldn't even be making out of the backend.

lemniscate613: Good point. I'll add that now.

aspidites:
re point 2: Then your original argument is appeal to tradition.

PR #2064, opened 2016-01-26
[PD-1384]  presentation layer

darrint:
~~Guidance: see comments inline.~~ Thanks.

## Review Guide

Here are some starting points:

* `myreports/presentation/base.py` has documentation.
* Note the library being used to write Excel spreadsheets. `requirements.txt`
* Previously report names were unused by the API. This will be addressed in PD-1952. With this change the backend uses the name field and the UI sets a hard coded name. `WizardPageFilter.js`

### DIFF
--- a/gulp/src/reporting/wizard/WizardPageFilter.js
+++ b/gulp/src/reporting/wizard/WizardPageFilter.js
@@ -10,6 +10,13 @@ import {WizardFilterCityState} from './WizardFilterCityState';
 import {SearchInput} from 'common/ui/SearchInput';
 
 export class WizardPageFilter extends Component {
+  constructor() {
+    super();
+    this.state = {
+      reportName: 'Report Name',
+    };
+  }
+
   componentDidMount() {
     this.updateState();
   }
@@ -71,6 +78,7 @@ export class WizardPageFilter extends Component {
 
   render() {
     const {reportConfig} = this.props;
+    const {reportName} = this.state;
 
     const rows = [];
     reportConfig.filters.forEach(col => {
@@ -150,7 +158,7 @@ export class WizardPageFilter extends Component {
         <hr/>
         {this.renderRow('', 'submit',
           <Button
-            onClick={() => reportConfig.run()}>
+            onClick={() => reportConfig.run(reportName)}>
             Run Report
           </Button>)}
 
--- a/myreports/presentation/base.py
+++ b/myreports/presentation/base.py
@@ -2,4 +2,48 @@ from abc import ABCMeta
 
 
 class Presentation:
+    """Base class for Reporting presentation types.
+
+    Implementations are responsible for selecting the content type and
+    filename extension which the user's browser will see at download time,
+    as well as writing bytes of the document to an output stream.
+    """
     ___metaclass__ = ABCMeta
+
+    def content_type(self):
+        """Content type as seen by the user's browser.
+
+        i.e. 'application/json', 'text/csv', etc.
+        """
+        pass
+
+    def filename_extension(self):
+        """Filename extension seen by the user at download time.
+
+        i.e. 'csv', 'xls', etc.
+        """
+        pass
+
+    def write_presentation(self, values, records, output):
+        """Write bytes representing the downloaded artifact.
+
+        values - Ordered list of columns to appear in the download.
+            i.e. ['a', 'c', 'b']
+        records - List of dicts containing data to appear in the download.
+            i.e. [
+            {'a': '1', 'b': '2', 'c', '3'},
+            {'a': '4', 'b': '5', 'c', '6'},
+        ],
+        output - has method output.write(bytes) which accepts bytes for
+            download
+
+        Example:
+
+            def write_presentation(self, values, records, output):
+                output.write('|'.join(values))
+                output.write('\\n')
+                for record in records:
+                    output.write("|".join(record[v] for v in values]))
+                    output.write('\\n')
+        """
+        pass
--- a/myreports/presentation/base.py
+++ b/myreports/presentation/base.py
@@ -1,0 +1,5 @@
+from abc import ABCMeta
+
+
+class Presentation:
+    ___metaclass__ = ABCMeta
--- a/myreports/presentation/base.py
+++ b/myreports/presentation/base.py
@@ -1,28 +1,22 @@
-from abc import ABCMeta
-
-
-class Presentation:
+class Presentation(object):
     """Base class for Reporting presentation types.
 
     Implementations are responsible for selecting the content type and
     filename extension which the user's browser will see at download time,
     as well as writing bytes of the document to an output stream.
     """
-    ___metaclass__ = ABCMeta
 
-    def content_type(self):
-        """Content type as seen by the user's browser.
+    """Content type as seen by the user's browser.
 
-        i.e. 'application/json', 'text/csv', etc.
-        """
-        pass
+    i.e. 'application/json', 'text/csv', etc.
+    """
+    content_type = ''
 
-    def filename_extension(self):
-        """Filename extension seen by the user at download time.
+    """Filename extension seen by the user at download time.
 
-        i.e. 'csv', 'xls', etc.
-        """
-        pass
+    i.e. 'csv', 'xls', etc.
+    """
+    filename_extension = ''
 
     def write_presentation(self, values, records, output):
         """Write bytes representing the downloaded artifact.
--- a/myreports/presentation/disposition.py
+++ b/myreports/presentation/disposition.py
@@ -5,5 +5,13 @@ filename_bad_chars = re.compile(r'\W')
 
 
 def get_content_disposition(report_name, extension):
+    """Build the contents of a Content-Disposition header.
+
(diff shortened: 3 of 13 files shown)


gulp/src/reporting/wizard/WizardPageFilter.js
```diff
@@ -10,6 +10,13 @@ import {WizardFilterCityState} from './WizardFilterCityState';
 import {SearchInput} from 'common/ui/SearchInput';
 
 export class WizardPageFilter extends Component {
+  constructor() {
+    super();
+    this.state = {
+      reportName: 'Report Name',
```
darrint: PD-1952 will add controls and default behavior for naming reports.

myreports/presentation/disposition.py
```diff
@@ -0,0 +1,17 @@
+import re
+
+
+filename_bad_chars = re.compile(r'\W')
```
darrint: Guidance: Is this being naive? See below where all characters matching this regex are replaced by underscores.

myreports/presentation/base.py
```diff
@@ -0,0 +1,49 @@
...
+        pass
+
+    def filename_extension(self):
+        """Filename extension seen by the user at download time.
+
+        i.e. 'csv', 'xls', etc.
+        """
+        pass
+
+    def write_presentation(self, values, records, output):
```
darrint: I expect this interface will change when I implement reports that summarize data with charts.

requirements.txt
```diff
@@ -54,3 +54,4 @@ django-compressor==1.4
 Markdown==2.4
 xlrd==0.9.3
 Sphinx==1.3.3
+XlsxWriter==0.8.4
```
darrint: Guidance: Any gripes with using this library?

myreports/tests/test_presentations.py
```diff
@@ -0,0 +1,78 @@
...
+        values = ['A', 'B']
+        records = [
+            {'A': 1, 'B': 'b'},
+            {'A': 2, 'B': 'bb'},
+        ]
+        self.xlsx.write_presentation(values, records, output)
+        sheet_entry = 'xl/worksheets/sheet1.xml'
+        sheet_content = ZipFile(output).open(sheet_entry).read()
+        root = fromstring(sheet_content)
+        cells = root.findall('.//ssml:t', self.ns)
```
darrint: For test purposes I'm just opening the xlsx as a zip file, looking at the worksheet, and verifying that the expected data is present.

myreports/presentation/disposition.py
tdphillips: Right or wrong, this is what we do for PRM attachments.

requirements.txt
tdphillips: I was thinking xlrd had write functionality but that's false.
:+1:

myreports/tests/setup.py
```diff
@@ -197,11 +197,18 @@ def create_full_fixture():
...
-        id=2, presentation_type="Maybe Inactive")
+        id=2, description="Maybe Inactive")
     pre_csv = PresentationTypeFactory.create(
-        id=3, presentation_type="Unformatted CSV")
+        id=3, presentation_type="csv", description="Unformatted CSV")
+    pre_xlsx = PresentationTypeFactory.create(
+        id=4, presentation_type="xlsx", description="Excel xlsx")
+    pre_json = PresentationTypeFactory.create(
+        id=5,
+        is_active=False,
```
darrint: Off by default. (This script isn't meant to be run in production anyway. Setting this record inactive keeps JSON Passthrough from showing up as a user visible option in the UI in development.

myreports/tests/test_views.py
```diff
@@ -631,8 +668,47 @@ def test_dynamic_report_with_filter(self):
...
+        self.assertEqual(200, resp.status_code)
+        self.assertEqual(
+            {'reports': [
+                {'id': report_id, 'name': 'The Report'},
+            ]},
+            json.loads(resp.content))
+
+        resp = self.client.get(reverse('download_dynamic_report'),
+                               {'id': report_id})
+        self.assertEquals(200, resp.status_code)
```
darrint: This test is end-to-end. It is checking that xlsx works at all. It is as much a test of test configuration setup as it is actual code. I'm relying on the more focused test in test_presentation for making sure data is plumbed through into the file.

myreports/views.py
```diff
@@ -614,16 +615,17 @@ def download_dynamic_report(request):
     values = report_configuration.get_header()
     records = [report_configuration.format_record(r) for r in report.python]
 
-    response = HttpResponse(content_type='text/csv')
-    content_disposition = "attachment; filename=%s-%s.csv"
-    response['Content-Disposition'] = content_disposition % (
-        report.name.replace(' ', '_'), report.pk)
+    presentation_driver = (
+        report.report_presentation.presentation_type.presentation_type)
+    presentation = presentation_drivers[presentation_driver]
```
darrint: Here is where we switch presentation formats.

myreports/views.py
```diff
@@ -614,16 +615,17 @@ def download_dynamic_report(request):
...
+    response = HttpResponse(content_type=presentation.content_type())
+    disposition = get_content_disposition(
+        report.name,
```
darrint: This is where report name matters. Before I was using report.pk. (see above) I didn't want to write `get_content_disposition` and continue passing it report.pk.

myreports/presentation/xlsx.py
```diff
@@ -0,0 +1,23 @@
+from xlsxwriter import Workbook
+from myreports.presentation.base import Presentation
+
+
+class Xlsx(Presentation):
+    def content_type(self):
+        return (
+            'application/vnd.openxmlformats-' +
+            'officedocument.spreadsheetml.sheet')
```
darrint: Someday it may be possible to produce a document smaller than its mime type.

myreports/tests/test_presentations.py
```diff
@@ -0,0 +1,78 @@
...
+
+class TestCsv(TestCase):
+    csv = Csv()
+
+    def test_presentation(self):
+        """Happy path, puts data in csv columns, etc."""
+        expected = """A,B\r
+1,b\r
+2,bb\r
+"""
```
misterjonscott: Should the indentation be consistent, here?

myreports/tests/test_presentations.py
```diff
@@ -0,0 +1,78 @@
...
+        records = [
+            {'A': 1, 'B': 'b'},
+            {'A': 2, 'B': 'bb'},
+        ]
+        self.assert_content(expected, values, records)
+
+    def test_presentation_unicode(self):
+        """Run a unicode character through the presentation."""
+        expected = """A\r
+aa\xe2\x80\x99zz\r
```
misterjonscott: And here?  Or is this github messing up your work?

myreports/tests/test_presentations.py
tdphillips: If the indentation is done visually, there would be lots of spaces in the middle of the string. I might have done the continuation a bit differently but this would work.

misterjonscott:
This is a pretty big PR, but everything looks good to me.

aspidites:
Commenting for posterity, though we spoke in person:

If we aren't frequently instantiating these presentation objects, then I'm ok with what we have here. If we are, I think I'd want a friendlier interface. Two things come to mind:

First, if `mime_type` and friends are properties, they should look like it. Consider:
```python
class XML(Presentation):
    @property
    def mime_type(self):
        return "text/xml"
```

Client code can now do this:
```python
print xml.mime_type
```
Instead of:
```python
print xml.mime_type()
```
It's only too characters, but it makes all the difference to an API feeling pythonic in my opinion.

Second, if all the behavior within these presentation objects is contingent on static properties, then it might be better to simply have class attributes, like this:

```python
class XML(Presentation):
    mime_type = "text/xml"
```
The benefit here is that creating new presentation types is declarative, which is more evocative of meaning (since I know that a presentation has a mime type, rather than having to find out if mime_type() *does* something. As an added bonus, creating new presentation types requires far less boiler plate. In particular, with the property style above, subclasses would have to continue wrapping properties with a decorator, or you'd have to overwrite the __new__ method to make sure that the correct things were turned into properties, which could be messy.

The only thing that would stop the class attribute style above from being  viable would be if the methods on Presentation are dependent upon certain values changing over time, which at a cursory glance doesn't seem to be the case.

koonsies:
UI is working. I was able to generate a report in both csv and excel.

myreports/presentation/xlsx.py
koonsies: the + is unnecessary here. Obviously not a big deal, just a note.

aspidites:
Records working for each type except when invariants about the data don't hold. As those invariants *should* hold, I'm not making that this PR's problem. Namely, a contact record in the past could have a null contact, which this api doesn't account for. As contacts should now never be null (and instead  be archived), the proper thing to do (spoke with @haliscooper and @darrint), is to create a ticket which fixes the bad data.

koonsies:
Looks good to me, but merge conflicts

PR #2065, opened 2016-01-26
[PD-2118] Update Unit Tests with Roles Enabled [Contains Bug Fixes]

aspidites:
Note: I plan to completely eradicate `settings.ROLES_ENABLED` in a separate PR. The goal of this ticket in particular was to make sure that all (not simply those relating to roles) tests pass when `settings.ROLES_ENABLED` is true. 

~~Also note that this work isn't finished, but as I can't run tests locally, the only alternative I had beyond spending days figuring out how my local environment was different than Jenkins was to simply run tests on Jenkins.~~

While updating these tests, I've found that we already have a few bugs that are live in production that might have been caught were this done first.

This is ready to go!


mydashboard/tests/test_views.py
```diff
@@ -177,25 +176,25 @@ def test_filters(self):
         self.candidate_user.save()
         update_solr_task(settings.TEST_SOLR_INSTANCE)
 
-        country_str = 'http://testserver/candidates/view?company=1&amp;location={country}'
-        country_filter_str = '<a class="applied-filter" href="http://testserver/candidates/view?company=1"><span>&#10006;</span> {country_long}</a><br>'
-        region_str = 'http://testserver/candidates/view?company=1&amp;location={country}-{region}'
-        region_filter_str = '<a class="applied-filter" href="http://testserver/candidates/view?company=1&amp;location={country}"><span>&#10006;</span> {region}, {country}</a>'
-        city_str = 'http://testserver/candidates/view?company=1&amp;location={country}-{region}-{city}'
-        city_filter_str = '<a class="applied-filter" href="http://testserver/candidates/view?company=1&amp;location={country}-{region}"><span>&#10006;</span> {city}, {region}, {country}</a>'
+        country_str = 'http://testserver/candidates/view?company={company}&amp;location={country}'
```
aspidites: Because we were hard-coding the company ids. I know the lines are long, but they were before I started.

myjobs/models.py
```diff
@@ -229,7 +229,10 @@ def is_group_member(self, user, group):
         Outputs:
         :is_member: Boolean representing the user's membership status
         """
-        return user.groups.filter(name=group).count() >= 1
+        if settings.ROLES_ENABLED:
```
aspidites: Bugfix. Without it, topbar acts funny. Don't remember how.

myjobs/templatetags/common_tags.py
```diff
@@ -77,7 +77,8 @@ def is_a_group_member(company, user, group):
     """
 
     if settings.ROLES_ENABLED:
-        return user.roles.exists()
+        # deleted users don't have a PK
```
aspidites: The delete user page renders a template which calls this template tag. Given the user doesn't exist anymore, we shouldn't try to access properties on it.

myjobs/tests/setup.py
```diff
@@ -1,23 +1,128 @@
+from importlib import import_module
```
aspidites: Just moved the test client from test_views to setup.

myjobs/tests/test_models.py
```diff
@@ -284,39 +253,21 @@ def test_send_invite_method(self):
...
-            self.assertFalse(
-                user.can(self.company, 'create user'),
-                "User shouldn't be able to 'create user', but can.")
-
-            user = admin_user.send_invite(
-                'regular@joe.com', self.company, True)
-            self.assertTrue(
-                user.can(self.company, 'create user'),
-                "User should be able to 'create user', but can't")
+        user = self.user.send_invite('regular@joe.com', self.company)
```
aspidites: Roles are enabled and in theory, company admin is going away, so I decided to not  bother with manually updating the setting in a context manager.

myreports/tests/test_activities.py
```diff
@@ -25,17 +25,9 @@ class TestViewLevelActivities(MyReportsTestCase):
 
     def setUp(self):
         super(TestViewLevelActivities, self).setUp()
-        settings.ROLES_ENABLED = True
```
aspidites: All this was moved to setup.

PR #2066, opened 2016-01-26
[PD-2122] Add a link to PRM in invitation email [hotfix]

aspidites:
This adds a link to the invitation email to PRM when a user gets sufficient privileges to at least click that link.

This email template that I worked so hard to remove  conditional logic from is starting to regress back to its original form, which makes me sad.


aspidites:
Can't send this hotfix to master because the implementation that allows it isn't represented in master.

PR #2069, opened 2016-01-26
[PD-2122] Hard coded link to PRM in invitation email [hotfix]

aspidites:
This is NOT identical or equivalent to what I opened against QC. This hard-codes a link to PRM. If a user is assigned to a role that doesn't include the "read partner" activity, this link will blow up in their face, resulting in a 403.


misterjonscott:
Simple and effective.  :+1:

PR #2070, opened 2016-01-26
Make password do unsecret things on the activation page [PD-2124]

tdphillips:
Exactly what it says on the tin.

### DIFF
--- a/myjobs/tests/setup.py
+++ b/myjobs/tests/setup.py
@@ -7,6 +7,7 @@ from django.core.urlresolvers import clear_url_caches
 from django.test import TestCase
 
 from myjobs.tests.helpers import return_file
+from seo.models import SeoSite
 
 
 class MyJobsBase(TestCase):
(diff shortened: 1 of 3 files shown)


myjobs/tests/setup.py
```diff
@@ -28,6 +29,7 @@ def setUp(self):
 
         self.patcher = patch('urllib2.urlopen', return_file())
         self.mock_urlopen = self.patcher.start()
+        settings.SITE = SeoSite.objects.first()
```
tdphillips: Registration tests were failing in isolation due to the absence of this.

myjobs/views.py
```diff
@@ -1167,7 +1167,8 @@ def api_create_user(request):
                                 content_type="application/json")
 
         # Create User
-        new_user, created = User.objects.create_user(email=user_email)
+        new_user, created = User.objects.create_user(email=user_email,
+                                                     in_reserve=True)
```
tdphillips: The fix.

registration/tests/test_views.py
```diff
@@ -47,9 +48,8 @@ def setUp(self):
 
         # Update the only existing site so we're working with
         # my.jobs for historical/aesthetic purposes.
-        site = SeoSite.objects.get()
```
tdphillips: settings.SITE wasn't picking up these changes as it's effectively cached.

PR #2071, opened 2016-01-26
Make password do unsecret things on the activation page [PD-2124]

tdphillips:
Backport hotfix to QC.


fezick:
Can we please get this merged? It is blocking QC work on PD-1972.

aspidites:
merge conflicts. @tdphillips  I'm hijacking this since you're out.

aspidites:
So, I borked your branch and restored it from your master hotfix @tdphillips, really sorry about that. I do not, however, think this fix is appropriate for QC. While messing with tests, I discovered that the user creation api is duplicating work done by the invitation model, so it may be better to refactor it such that it takes advantage of existing functionality. In particular, rather than only using send_invitation when a user exists, we should be unconditionally sending it. The Invitation model, as you know, has provisions for ensuring that the email the invitation is being sent to is either attached to a user, or gets attached to a user.

myjobs/views.py
```diff
@@ -1121,65 +1121,37 @@ def api_create_user(request):
...
 
-            ctx["success"] = "true"
+            ctx["message"] = "User already exists. Role invitation email sent."
+        else:
             ctx["message"] = "User created. Invitation email sent."
-            return HttpResponse(json.dumps(ctx),
-                                content_type="application/json")
 
-        ctx["success"] = "false"
-        ctx["message"] = "User not created."
```
tdphillips: We first try getting the user. If the user doesn't exist, we call User.objects.create_user. If that returns a False (the user already existed), we do this. That will never happen.

tdphillips:
If you see registration test changes, they should be mostly self-contained in the "Stop overriding self.user in registration tests" commit. It does exactly what it says. User states were weird and this brings them in line with having common configuration for tests in setUp()

PR #2073, opened 2016-01-26
Hotfix remove staff restriction in toolbar for user management link

lemniscate613:
We were hiding the User Management link in the toolbar, but now that it's live this restriction is no longer needed.

The other hotfix for this issue didn't include all instances of the restriction: https://github.com/DirectEmployers/MyJobs/pull/2067


fezick:
Merging despite failed tests, as they are result of the jenkins bug.

fezick:
@JLMcLaugh, this needs to be deployed first thing in the morning.

PR #2077, opened 2016-02-01
Add newrelic browser monitoring javascript for 1% of requests.

mklauber:


seo/templatetags/seo_extras.py
```diff
@@ -24,6 +25,17 @@
...
+        ;NREUM.info={beacon:"bam.nr-data.net",errorBeacon:"bam.nr-data.net",licenseKey:"49b0eca3f4",applicationID:"480149",sa:1,agent:"js-agent.newrelic.com/nr-768.min.js"}
+        </script>"""
+
+    # Only use the script tag on
```
tdphillips: Unfinished comment?

seo/templatetags/seo_extras.py
```diff
@@ -24,6 +25,17 @@
 register = template.Library()
 
 
+@register.simple_tag
+def newrelic_browser_monitoring():
```
JLMcLaugh: Can we set this up as a static file?

mklauber: The script?  I suppose so.

seo/templatetags/seo_extras.py
mklauber: Fixed.

seo/templatetags/seo_extras.py
mklauber: Do you want it served by the CDN, or just the script tag put in a static file, so that it doesn't clutter the clutter the source code?  e.g. can/should I just read it from a file and stick the string in there.  b/c I can add a script tag to an external file on a CDN, but I don't see a huge reason why we should do that in this case.

aspidites:
Spoke with @JLMcLaugh. As @mklauber is out today, I'm moving the JS to static files and loading it from there. As there still seems to be some concerns that @mklauber is raising in this PR I'll leave it open until I have the ability to speak with him. In the case that Jason and I missed something obvious, this will give us a way to punt and go with Matt's original implementation.

aspidites:
Have a PR up. Feel free to use that or just update yours and close it.

seo/templatetags/seo_extras.py
mklauber: Putting this here for posterity.

Based on the response to these questions, loading that file after the page loads isn't supported and  wouldn't provide accurate timing info for other javascript files.  
https://discuss.newrelic.com/t/loading-new-relic-browser-snippet-via-requirejs/33249
https://discuss.newrelic.com/t/using-copy-paste-method-with-existing-applications/30896

PR #2079, opened 2016-02-02
Added padding declaration to UL to fix menu alignment

misterjonscott:
:hamburger:

### DIFF
--- a/static/custom.css
+++ b/static/custom.css
@@ -2998,6 +2998,10 @@ em.pendingText {
     z-index: 9999 !important;
 }
 
+.submenu-dropdown > ul {
+    padding: 0;
+}
+
 /*Firefox fix for homepage alignment issues*/
 @-moz-document url-prefix() {
     #title {
--- a/templates/base-react.html
+++ b/templates/base-react.html
@@ -36,7 +36,9 @@
   https://github.com/scottjehl/Respond/tree/master/cross-domain
   http://lipstickonthepig.com/wordpress/using-respond-js-when-your-site-css-is-on-a-cdn-or-subdomain/
   -->
+  {% compress css %}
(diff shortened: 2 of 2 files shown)


misterjonscott:
# Changes
I added a padding declaration to remove padding from the UL element in the topmenu.  The styles are a bit scattered, right now, so I would like to dive deeper into that, soon, but this is a solution for the ticket at hand.

# Testing
You'll need to have multiple companies on your account, and reference [this ticket](https://directemployersfoundation.atlassian.net/browse/PD-2102) to see the previous error versus this fix.

tdphillips:
custom.css appears to only be used by base-react.html. How does this look on non-react pages?

misterjonscott:
Fortunately, @tdphillips , the problem only appears on the Users page, which is the only page that seems to be using base-react template right now.  This will only affect that page.

aspidites:
Begs the question if User Mangement is the only one using bsae-react, then why do we even have base-react?

aspidites:
We need to version this file so that changes are picked up downstream. Eg. rename it "custom.173-2.css"

misterjonscott:
My latest understanding was that all 3 new React apps were to use base-react, but that is not currently the case.  If this style, or that template, were adopted later, this style would, at worst, override a 0 padding, and, at best, fix future problems.  That said, our styles do need to be combined and cleaned up.  Hoping to get to that, soon.

misterjonscott:
@aspidites I've added a wrapper to has the filenames, as that's the new process.  Thanks for pointing this out!

aspidites:
Even better.

lemniscate613:
:+1: 

`base-react.html` could be renamed to something like `base-bootstrap3.html` as that's why it exists. The plan is to have all new projects (e.g. all of the React apps) using Bootstrap 3.

aspidites:
Tests are failing because compress isn't loaded in this template. [Here's](https://github.com/DirectEmployers/MyJobs/blob/quality-control/templates/redirect/base.html#L3) an example.

PR #2085, opened 2016-02-04
paragraph inside a paragraph is not what we want.

aspidites:
link_text has a \<p\> tag, so this needs to be moved outside of the other \<p\> tags.


aspidites:
@JLMcLaugh probably, but that required more work and it was a hot fix. Eg. first thing that worked. Also means that `{{ link_text }}` becomes a conditional. Can I submit that as a separate thing?

JLMcLaugh:
I didn't notice the conditional in the link_text function. I'm fine keeping it like this.

PR #2088, opened 2016-02-04
[PD-2148] First stab at company access request form.

aspidites:
# Description
- User facing form to request access to a company. The company is *not* checked to exist (intentional)

# Known Issues
- ~~This thing is ugly. If you reference the jira ticket, you'll see that I'm awaiting feedback as to whether I should spend time sprucing it up, or if I should only be concerned with functionality~~ See PD-2151

# Testing
1. Visit /request-company-access
2. Enter a company name
3. Note that you get an access code

There is a test that ensures that that code is valid, but if you'd like to prove it to yourself, you can do something like this in a console:
```
>>> from myjobs.models import CompanyAccessRequest
>>> a = CompanyAccessRequest.objects.last()
>>> print a.check_access_code("<code from form here>")
```
If the above returns False, we have a problem. Otherwise, we are good to go.


aspidites:
Talked with @misterjonscott and he'll be helping me sort out the css issues. Leaving this open though as It's still be nice to get feedback on the base functionality.

myjobs/views.py
```diff
@@ -1288,3 +1290,21 @@ def api_delete_user(request, user_id=0):
...
+            _, access_code = AccessRequest.objects.create_request(
+                company_name, request.user)
+            ctx['access_code'] = access_code
+    else:
+        form = AccessRequestForm()
+
+    ctx['form'] = form
+
+    return render_to_response(
+        '../templates/myjobs/request_access.html', ctx,
```
tdphillips: Can't everything before "myjobs" be left out?

aspidites: Sure can. Vim only auto completes when it can find the path in question. I forgot to get rid of the redundant bits after completing the path. 

Thanks for catching this. I'd have been sad if this made it through.

aspidites:
Closing this until @misterjonscott  is done with his form-error-hightlight work, as styles will conflict otherwise.

aspidites:
Or it'll just stay open. Meh. Talked to @misterjonscott  and he doesn't mind swinging back around to pretty this up. I did at least update the template to use bootstrap 3 with his help, so it shouldn't be *too* bad.

aspidites:
Missed a requirement.

PR #2089, opened 2016-02-05
moved presentation to template.

aspidites:
When I first wrote this, I was in a hurry as it was meant to be a hotfix, so I deliberately mixed presentation with logic. This corrects that.


lemniscate613:
:+1:

PR #2093, opened 2016-02-08
PD-2028

mklauber:
Created an importRecord model and added code to create and delete records at appropriate times.

We create records as part of the task execution process.  If a task retries, multiple records will be created.  

We delete records as a daily task run via celerybeat.  It by default deletes all tasks older than 31 days.


aspidites:
The initial migration seems...off.

import_jobs/models.py
```diff
@@ -0,0 +1,10 @@
+from django.db import models
+
+class ImportRecord(models.Model):
+    buid = models.IntegerField(db_index=True)
```
aspidites: Buid has no relation to Seo.BusinessUnit?

tasks.py
```diff
@@ -779,8 +780,10 @@ def task_update_solr(jsid, **kwargs):
         import_jobs.update_solr(jsid, **kwargs)
         if kwargs.get('clear_cache', False):
             task_clear_bu_cache.delay(buid=int(jsid), countdown=1500)
+            ImportRecord(buid=int(jsid), success=True).save()
     except Exception as e:
```
aspidites: Not a fan of blanket exceptions.

import_jobs/models.py
mklauber: Since this is a record table, I didn't want it tied to anything else.  I didn't want modifying a seo.BusinessUnit record to change the recorded data here.  e.g. Deleting a business unit would remove the records from the ImportRecord table b/c of On-Delete-Cascade constraints.

tasks.py
mklauber: In this case, we kind of want it.  Regardless of what broke, we want to retry it.  Could be an issue with services being down, with the server losing network connectivity, or a variety of other unknown causes.  If it retries 3 times and fails, it'll get formally failed, and the exception messages are logged.

mklauber:
The initial migration is empty, e.g. has no tables/models.  I did that so it's easier to delete the table entirely.  Rather than having to hack the DB to run `DROP...` commands, then faking back to 1 then migrating to latest, you can just migrate to 1 then remigrate to latest.

tdphillips:
That's what "./manage.py migrate zero" should do.

mklauber:
Oh, wasn't aware of that.

mklauber:
Fixed.

tasks.py
aspidites: Talked in person - :+1:

PR #2094, opened 2016-02-08
[PD-2149] Create jira ticket when new access request is submitted

aspidites:
This builds on top of #2088 by ensuring that a ticket is created when the request is submitted. I ended up writing a new functioon (that doubles as a task) to facilitate creating a new ticket, as the original wasn't well documented, did some extra email type things, and couldn't be run as a task. I couldn't simply wrap a task around it as that would have caused merge conflicts anyway. I suspect that this new task could deprecate that one (`log_to_jira`), but I didn't want to risk breaking old functionality to find out.

# Testing
As before:
- head to /request-company-access
- enter a company name and submit the form
- head over to [Jira](https://directemployersfoundation.atlassian.net/projects/MEMBERSUP/issues/MEMBERSUP-810?filter=allopenissues) in the MEMBERSUP project and note that a new ticket should have been created.


tasks.py
```diff
@@ -16,6 +16,7 @@
 from celery.task import task
 
 from django.conf import settings
+from secrets import options, housekeeping_auth
```
tdphillips: I suspect housekeeping_auth is jiratools-specific at the moment. If we're using this, @mklauber will have to add it - last I checked, I didn't have shell access to Jira.

aspidites: I'm not sure the orginal auth is used at all. If we can confirm, I can probably just have auth updated in the settings file.

aspidites:
Gave creds to @mklauber  waiting for those to be updated before I rerun tests.

tasks.py
lemniscate613: @aspidites I temporarily commented out the `from secrets import options, housekeeping_auth` line in tasks.py on my feature branch as it was throwing the following with my Docker setup:

```
  File "/MyJobs/import_jobs/../tasks.py", line 19, in <module>
    from secrets import options, housekeeping_auth
ImportError: cannot import name housekeeping_auth
```

aspidites: You need to update your secrets file. You can get the correct values from Jason Sole

PR #2096, opened 2016-02-09
[PD-2150] update invites for user management

lemniscate613:
Users should only receive invite emails for the _new_ roles added to their account.

# Review

Confirm previous functionality still works:

```
dkm test myjobs.tests.test_manage_users.ManageUsersTests
```

Each of the following should result in _only one invite email_:

1. __Edit your user__ (which already has roles like Admin and PRM). Add one role.

2. __Edit a role__ not assigned to you. Add your user.

3. __Add a role.__ Select your user.


myjobs/views.py
```diff
@@ -1041,6 +1047,9 @@ def api_get_specific_user(request, user_id=0):
     """
     ctx = {}
 
+    from django.db import connection
```
tdphillips: Debugging stuff here and line 1099.

lemniscate613: Thank you. Just removed it: https://github.com/DirectEmployers/MyJobs/commit/3ab2d667f15d43a7d846e863db09b5be70956ab4

tdphillips:
Conflicts.

myjobs/views.py
```diff
@@ -992,7 +995,10 @@ def api_edit_role(request, role_id=0):
         for user in existing_users.exclude(pk__in=new_users.values("pk")):
             user.roles.remove(role_id)
         for user in new_users.exclude(pk__in=existing_users.values("pk")):
+            # Add role to user
```
aspidites: I don't think these comments are adding anything. For instance, `user.roles.add(role_id)` tells me that we are adding a role to a user.

I really appreciate your tenacity for ensuring that things are commented, but comments for comment sake make things more difficult to read. 

If you're wondering when I decide something is worth commenting, it's when I'm doing tricky things. A prime example of something that might need to be commented would be the custom SQL statement that @darrint  has been messing with over the last week or so.

myjobs/views.py
```diff
@@ -1213,9 +1221,8 @@ def api_create_user(request):
 
         # Send one invitation per new role. This will handle creating the user
         # if one doesn't exist.
-        if request.user.can(company, "create user"):
-            for role in set(new_roles).difference(existing_roles):
-                request.user.send_invite(user_email, company, role.name)
+        for role in set(new_roles).difference(existing_roles):
```
aspidites: Yay, set theory!

myjobs/views.py
```diff
@@ -1258,7 +1264,7 @@ def api_edit_user(request, user_id=0):
         # 1. List current company's roles
         company_roles = Role.objects.filter(company=company)
         # 2. List user's roles
-        roles_assigned_to_user = user[0].roles.all()
+        roles_assigned_to_user = user.roles.all()
 
         # 3. Overlap?
```
aspidites: This comment tells me nothing. If it's a question you have, it should be asked in the PR. If it's meant to say "check for overlap" it's redundant, as we are doing set operations, which by definition gaurantee uniqueness.

myjobs/views.py
```diff
@@ -1258,7 +1264,7 @@ def api_edit_user(request, user_id=0):
...
         if bool(set(company_roles) & set(roles_assigned_to_user)) == "False":
```
aspidites: This will never be true. You're converting the left-hand side to a boolean then checking it against a string. While you could remove the quotes to fix this, more pythonic would be something like this:

```if set(company_roles) & set(roles_assigned_to_user):```

That said, I've mentioned this before, but in Python, I prefer descriptive method names to overloaded operators:
```if set(company_roles).intersection(roles_assigned_to_user):```

As a bonus, you don't have to explicitly cast roles_assigned_to_user since intersection takes an iterable, not necessarily a set.

lemniscate613: Nice catch with the string. The point about ```intersection()``` is interesting, and it led me to ```disjoint()``.

PR #2097, opened 2016-02-09
[PD-2066] New Saved Search API

koonsies:
## Overview

This PR covers the creation of a new Saved Search API that uses both CORS and JSONP. Unlike the original saved seach API, this API takes into consideration whether the user is logged in to the secure parent of the given microsite.

## Testing

To test the new API, follow the same instructions as the original secure blocks widget. The testing steps can be found at http://directemployers.github.io/MyJobs/developers/myblocks/secureblocks.html 
(NOTE: Docs are currently outdated, this should be fixed soon. Otherwise, run docs locally to find the testing steps)


myblocks/tests/test_secure_blocks.py
```diff
@@ -88,13 +90,32 @@ def test_saved_search_includes_required_js(self):
...
-        url = reverse('secure_blocks')
-        return self.client.post(
-            url, body,
-            HTTP_HOST='jobs.example.com',
-            HTTP_ORIGIN=http_origin,
-            HTTP_X_REQUESTED_WITH='XMLHttpRequest',
-            content_type="application/json",
-        )
+
+def make_cors_request(client, url, json_data,
```
koonsies: Mostly a rename of the make_sb_request with some abstraction for reuse.

myblocks/tests/test_secure_blocks.py
```diff
@@ -88,13 +90,32 @@ def test_saved_search_includes_required_js(self):
...
+    return client.post(
+        url, json_data,
+        HTTP_HOST='jobs.example.com',
+        HTTP_ORIGIN=http_origin,
+        HTTP_X_REQUESTED_WITH='XMLHttpRequest',
+        content_type="application/json",
+    )
+
+def make_jsonp_request(client, url, json_data,
+                       http_referer='http://jobs.example.com'):
```
koonsies: Gotta test that JSONP also

mysearches/tests/test_api.py
```diff
@@ -0,0 +1,362 @@
...
+        #DELETE BELOW LINE WHEN DEPLOYED TO PRODUCTION
+        self.assertEqual(response.status_code, 302)
+        # UNCOMMENT BELOW THIS LINE WHEN ACTIVATED IN PRODUCTION
+        #
+        # self.assertEqual(response.status_code, 200)
+        # response_msg = json.loads(response.content)
+        # self.assertTrue(response_msg['user_created'],
+        #                  msg="Expected user to be created, but it was not")
+        # self.assertTrue(response_msg['search_activated'],
+        #                  msg="Expected search to be activated, but it was not")
```
koonsies: Not sure if there is a more elegant way of doing this.

mysearches/views.py
```diff
@@ -398,3 +406,144 @@ def send_saved_search(request):
...
+
+def add_or_activate_saved_search(user, url):
+    """
+    Attempt to add a new saved search to a user, or it is already exists,
+    make sure it is active.
+    :param user: User recieving search
+    :param url: URL for search
+    :return: The new or activated search
+
+    """
```
koonsies: A good amount of this code is copied from the original API, with the necessary changes made to bring this in line with CORS/authentication logic.

static/saved-search.js
```diff
@@ -2,17 +2,19 @@
...
 //global iables for interacting with secure blocks template assigned values
 var email_input;
 
+$('#saved-search-email').bind('keypress', function(e) {
+	if(e.keyCode==13){
+    e.preventDefault();
+    blocks_widget_div = $(this).closest("[data-secure-block-id]");
+    save_search();
+	}
+});
```
koonsies: Bind "enter" key while the focus is on the email input.

static/saved-search.js
```diff
@@ -21,7 +23,7 @@ $('#saved-search-btn').click(function(e) {
 });
 
 //borrowed from gulp/src/util/validateEmail.js
-function validateEmail(email) {
+function validate_email(email) {
   var re = /(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])/;
```
tdphillips: Given the troubles we just had yesterday, this needs to be updated as well.

koonsies: Good call. I will update this as soon as his code is merged.

static/saved-search.js
aspidites: I'd maybe inline this comment. I don't think months from now I'll remember that keycode 13 is Enter.

mysearches/views.py
```diff
@@ -398,3 +406,144 @@ def send_saved_search(request):
...
+def user_creation_retrieval(auth_user, email):
+    """
+    Retrieve user account for adding saved searches. If the email belongs
+    to a user and that user is not logged in, error is raised.
+
+    :param auth_user: Currently logged in user
+    :param email: Email attempting to be added
+    :return: New or existing user account tied to provided email
+    """
+    if not auth_user:
```
aspidites: Just curious: When/how do we call this falsey values? Feels like java-style null checks, but maybe that's unavoidable?

mysearches/views.py
```diff
@@ -398,3 +406,144 @@ def send_saved_search(request):
...
+    :return: value for key or none
+    """
+    request_body = {}
+    try:
+        # if request.body exists and is a json object, load it
+        request_body = json.loads(request.body)
+    except ValueError:
+        pass
+
+    value = (request.GET.get(key) or
```
aspidites: If you leverage the second argument of get you won't need the outer parentheses or the `or`s.

tdphillips: It's a little frowned upon but given that we're checking both dicts anyway, we could just look in request.REQUEST. That's deprecated in 1.7, however.

tdphillips: Doing a bit of spelunking in the code, we may not want to follow my suggestion.

mysearches/views.py
koonsies: In the way I am using this function, auth_user should never be null.. It'll be generic unauthenticated user instance, or the instance of the current logged in user. This check is just here just "in case" a None is passed in.

mysearches/views.py
koonsies: What would that look like? request.GET.get(key, request.POST.get(key, body.get(key, None)))?

aspidites: Yeah, though the last None is redundant:
```python
value = request.GET.get(key, request.POST.get(key, request_body.get(key)))
```

If the nesting makes you sad:
```python
value  = request.GET.get(
    key, request.POST.get(
        key, request_body.get(key)))
```

koonsies: Hmm. I definitely see where you're coming from, but I don't find that it increases readability or reduces processing. Thanks for the input, though!

koonsies: On a side note, your point made me reevaluate how Python resolves "or" relationships. In this event, it goes the lazy route and only processes until it hits a "True" resolution. I had a feeling that is how it works, but I went ahead and tested to make sure.

```
In [14]: def true1(): print 'true1'; return True;

In [15]: def false1(): print 'false1'; return False;

In [16]: def false2(): print 'false2'; return False;

In [17]: true1() or false1() or false2()
true1
Out[17]: True

In [18]: false1() or true1() or false2()
false1
true1
Out[18]: True
```

PR #2101, opened 2016-02-10
[PD-2152, PD-2148] Staff-Facing Access Request Form

aspidites:
Preperation
```sh
pip install -r ./requirements.txt --upgrade
```

For Docker users:
```sh
dk rebuilddev
```

# Testing
- visit /request-company-access
  - submit the form and take note of the access code
- visit /admin/myjobs/companyaccessrequest
  - click on the request you just submitted
  - choose a company (that you don't have access to) and enter your access code
- visit / (or refresh)
  - you should see a new company in your drop down list


mydashboard/admin.py
```diff
@@ -1,86 +0,0 @@
-from django.contrib import admin
```
aspidites: Not using CompanyUser anymore, so no need for company user admin.

dseo_urls.py
```diff
@@ -63,6 +64,7 @@
 )
 
 urlpatterns += patterns('',
+    url(r'^ajax_select/', include(ajax_select_urls)),
```
aspidites: Needed for the autocomplete widget.

myjobs/admin.py
```diff
@@ -57,8 +65,54 @@ class FAQAdmin(admin.ModelAdmin):
     list_display = ['question', 'is_visible']
     search_fields = ['question', ]
 
+
+def company_name(company):
```
aspidites: Moved over from the mydashboard admin module and refactored.

myjobs/models.py
```diff
@@ -914,7 +907,8 @@ class CompanyAccessRequestManager(models.Manager):
     def create_request(self, company_name, requested_by):
         access_request = self.create(
             company_name=company_name, requested_by=requested_by)
-        access_code = get_random_string(8)
+        # generate a hexadecimal access code
+        access_code = get_random_string(8, allowed_chars='ABCDEF1234567890')
```
aspidites: Part of PD-2148 - we only wanted valid hexadecimal characters for codes. No lower case, no ambiguous letters (O vs 0 for instance).

registration/tests/test_views.py
```diff
@@ -206,21 +206,6 @@ def test_saved_search_invitation_message(self):
         self.assertIn("in order to begin receiving their available job "
                       "opportunities on a regular basis", email.body)
 
-    def test_permission_invitation_message(self):
```
aspidites: We aren't using groups for permissions any more.

seo/lookups.py
```diff
@@ -0,0 +1,10 @@
+from ajax_select import register, LookupChannel
+from seo.models import Company
+
+@register('companies')
```
aspidites: Used for autocomplete.

universal/tests/test_helpers.py
```diff
@@ -24,19 +24,6 @@ def test_saved_search_invitation_context(self):
             "initial_search_email": saved_search.initial_email(send=False),
             "text_only": saved_search.text_only})
 
-    def test_group_invitation_context(self):
```
aspidites: Again, we don't use group contexts anymore. I can get tests to pass but it seemed pointless to test something that wasn't a reality anymore.

koonsies:
I don't recall from the conversations during scrum, but is the CSS a consideration here?

screen shot 2016-02-11 at 9 43 02 am

aspidites:
@koonsies 

1. This is for the staff-facing form, not the user-facing form, which was merged previously and where the above comes from
2. See #2088 where I reference styling issues. Namely they'll be fixed in another ticket, and part of what you're seeing is because @misterjonscott's changes to form-error-highlight haven't hit QC. With his changes, that monstrosity you've captured doesn't exist.

myjobs/tests/test_forms.py
```diff
@@ -38,3 +42,38 @@ def test_password_form(self):
...
+        being updated and a new admin being assigned.
+        """
+
+        # required for django admin during tests, apparently
+        self.user.is_superuser = True
+        self.user.save()
+
+        # give us a handle to the unhashed access code
+        access_code = '1234ABCD'
+        user = UserFactory(email="requestuser@example.com")
```
koonsies: small note here, may want to make this requesting_user or something like that. In this scope we have self.user and user. May cause confusion as some point. That being said (and unrelated to your code) it's kinda sad that MyJobsBase has self.user since it is to generic.

myjobs/tests/test_forms.py
```diff
@@ -38,3 +42,38 @@ def test_password_form(self):
...
+            "company": self.company.pk,
+            "verification_code": access_code
+        }
+        admin_url = "%s%s/" % (
+            reverse("admin:myjobs_companyaccessrequest_changelist"),
+            request.pk)
+        self.client.post(path=admin_url, data=data)
+
+        # object is cached so we need to refetch it
+        request = CompanyAccessRequest.objects.get(pk=request.pk)
```
koonsies: "request", while it's not ambiguous in this context, may still be a confusing term since it could be easily confused for a http request.

myjobs/tests/test_forms.py
aspidites: MyJobsBase's self.user is intentionally generic, as various tests need different levels of users but ALL tests need at least a basic user to which to assign roles. The alternative would have been having multipole users - self.normal_user, self.staff_user, self.superuser -- but that would have been problematic for different reasons.

aspidites: `user` is now `requesting_user`.

koonsies: I suppose I can agree with that. Though, by default, it is a staff_user. I'd argue that if we almost always want a staff_user, it's better to name it as such and let something create a normal_user and sign it into the client, rather than depend on context.

myjobs/tests/test_forms.py
aspidites: I wasn't worried about it since:
1. It's used immediately after it's declared and ever used again
2. It's in a unit test.

I'll update anyway.

koonsies: Yeah, it definitely wasn't a big deal, just a observation. Thanks!

mypartners/admin.py
```diff
@@ -4,7 +4,19 @@
...
+    """
+    Returns a string ot be used next to each company that displays how many
+    admins belong to that company, or a warning if there aren't any.
+
+    """
+    template = "{name} ({count} users){warning}"
+    count = company.admins.count()
+    warning = "" if count else " **Might be a dupliate**"
+
+    return template.format(name=company.name, count=count, warning=warning)
```
koonsies: This seems to be a more elegant solution to what is found in myjobs/admin. Could it be imported there, or at least copied there?

koonsies:
:+1: All I found were nit-picks. Clever overloading of the admin pages.

aspidites:
@koonsies:

- fixed typo in company_user_name
- renamed it to format_company_name
- deleted company_name because it wasn't actually being used
- used `format_company_name` in seo/lookups.py

End result:
2016-02-12-085041_634x517_scrot

requirements.txt
```diff
@@ -55,3 +55,4 @@ Markdown==2.4
 xlrd==0.9.3
 Sphinx==1.3.3
 XlsxWriter==0.8.4
+django-ajax-selects==1.4.2
```
JLMcLaugh: So now we have 3 apps installed for foreign keys, m2m select many, and m2m select. Can this be used to replace FSM and django extensions at some point?

aspidites: Just messed with the FSM bit, which proved problematic since it's inheriting directly from the built-in filtered multi select (the thing that gives you two boxes instead of one). Should be doable for foreign key autocompletes though, assuming that's the only reason we use django extensions.

aspidites: @JLMcLaugh 

Now we'll be [back to 2](https://github.com/DirectEmployers/MyJobs/pull/2107), which is no worse off than before  I started on this ticket.

PR #2102, opened 2016-02-10
Disable excessive session saving on redirect

tdphillips:
Tested on staging.

### DIFF
--- a/redirect_settings.py
+++ b/redirect_settings.py
@@ -28,3 +28,5 @@ MIDDLEWARE_CLASSES = global_settings.MIDDLEWARE_CLASSES + (
     'redirect.middleware.ExcludedViewSourceMiddleware',
     'redirect.middleware.MyJobsRedirectMiddleware',
 )
+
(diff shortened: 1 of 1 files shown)


koonsies:
This seems very simple, but I'm unsure on the context. What was happening that provokes this change?

tdphillips:
We were saving lots of sessions on redirect. Before the projects were merged (before you started), we were making ~10 commits per hour. After the merge, we inherited this setting. It shot up to ~80k per hour.

koonsies:
Ah, thank you for the explanation. Has there been any documentation made on the redirects project? I realize that I know very little about it. If so, could you direct me to it? If not, would you mind giving me a 10 minute crash course on it? It'll help me build some context on the project.

tdphillips:
You could think of it like our very own tinyurl service. We take long links to job pages and turn them into something shorter. When someone clicks on a my.jobs link, we can add query parameters that various applicant tracking systems use to attribute the source of a given view. Then we do a permanent redirect to the job itself.

PR #2106, opened 2016-02-11
[PD-2161] Added 'display_name' to activities.

aspidites:
# Testing
1. Migrate forward (`./manage.py migrate myjobs`)
2. Visit `/admin/myjobs/activity/` and have at it

# Changes
- Activity now has a "display_name" field
- Best way to access this field is "Activity.get_display_name()", as it will conveniently default to Activity.name when it's blank
- Added Activities to admin with restrictions:
  - only the display name should be editable
  - user should not be able to create or delete activities
- Updated the views @lemniscate613 wrote to query for display_name instead of name
- Activity names are no longer mangled
  - instead, they are ordered by noun
  - when comparing implementations, you should notice that the order is the same, but that names remain untouched
  - also note that this implementation is using display_name instead of name

# Caveats
- Custom display names will break name mangling we do on the frontend to change "foo bar" to "bar - foo". ~~I've noted this in the ticket and am awaiting feedback on how to proceed there (preferably in a subtask), as changing that behavior has implications for the frontend code.~~ Apparently we are OK with being forced to give every activity a display name to  cope with name mangling.
- I didn't add any unit tests as this was quite literally me adding a field to a model. I can add a test to ensure that `get_display_name` returns the name when display_name is blank, but that could would look exactly like the method itself, so I'm not sure if it's really testing anything other than the fact that I wrote the code the same twice


gulp/src/manageusers/formatActivityName.js
```diff
@@ -1,14 +0,0 @@
-// Formats an activity name. Assume verb is always 'create', 'read', 'update',
```
aspidites: We aren't mangling names anymore.

gulp/src/manageusers/reverseFormatActivityName.js
```diff
@@ -1,15 +0,0 @@
-// Reverses what formatActivityName() does.
```
aspidites: If we aren't mangling, then there is no need to unmangle.

gulp/src/manageusers/sortActivities.js
```diff
@@ -44,22 +42,22 @@ import {formatActivityName} from './formatActivityName';
 // ]
 //
 // Output:
-// Same format as above, but activity names like "update role" become "role - update"
+// Same format as above, but activities are ordered by the action being
+// performed. For instance, All "create" actions would appear before "update"
+// actions.
+export function sortActivities(appAccessGroups) {
```
aspidites: @lemniscate613 

See [_.groupBy](https://lodash.com/docs#groupBy), [_.flatten](https://lodash.com/docs#flatten), and [_.values](https://lodash.com/docs#values). Essentially:
- groupBy returns an object whose keys are the groups and values are arrays
- I use values to get an array of arrays, the groups from before without there "tags"
- I use flatten to give me a single array rather than one for every group

In this case, groups would be "create", "read", "update", and "delete". If we use custom display names, I'd expect that we follow a convention where a display name starts with a verb, so "add", "remove", etc.

gulp/src/manageusers/sortActivities.js
```diff
@@ -0,0 +1,63 @@
...
+// Output:
+// Same format as above, but activities are ordered by the action being
+// performed. For instance, All "create" actions would appear before "update"
+// actions.
+export function sortActivities(appAccessGroups) {
+  return appAccessGroups.map( appAccessGroup => {
+    const orderByVerb = activities => {
+      // group by the fisrt word in the activity name, then flatten those
+      // groups into a single array
+      return _.chain(activities).groupBy(activity => {
```
aspidites: @lemniscate613 

See [_.groupBy](https://lodash.com/docs#groupBy), [_.flatten](https://lodash.com/docs#flatten), [_.chain](https://lodash.com/docs#chain), and [_.values](https://lodash.com/docs#values). 

Essentially:
- groupBy returns an object whose keys are the groups and values are arrays
- I use values to get an array of arrays, the groups from before without there "tags"
- I use flatten to give me a single array rather than one for every group

In this case, groups would be "create", "read", "update", and "delete". If we use custom display names, I'd expect that we follow a convention where a display name starts with a verb, so "add", "remove", etc.

Chain makes things (IMHO) easier to read. Without it, I'd have had something like this:
```javascript
return _.flatten(_.values(_.groupBy(activities, activity => {
    return activity.name.split(' ').shift();
})));
```
Which you'd have to read right to left since the most interesting bit is the groupBy. As written, you can just read left to right.

gulp/src/manageusers/sortActivities.js
```diff
@@ -44,22 +41,23 @@ import {formatActivityName} from './formatActivityName';
...
-// Same format as above, but activity names like "update role" become "role - update"
+// Same format as above, but activities are ordered by the object being
+// manipulated. For instance, All "contact" actions would appear before
+// "partner" actions.
+
+export function sortActivities(appAccessGroups) {
+  const orderByNoun = activities => {
+    // group by the last word in the activity name, then flatten those groups
+    // into a single array
+    return _.chain(activities).groupBy(activity => {
```
aspidites: @lemniscate613 

See [_.groupBy](https://lodash.com/docs#groupBy), [_.flatten](https://lodash.com/docs#flatten), [_.chain](https://lodash.com/docs#chain), and [_.values](https://lodash.com/docs#values). 

Essentially:
- groupBy returns an object whose keys are the groups and values are arrays
- I use values to get an array of arrays, the groups from before without there "tags"
- I use flatten to give me a single array rather than one for every group

In this case, groups would be "create", "read", "update", and "delete". If we use custom display names, I'd expect that we follow a convention where a display name starts with a verb, so "add", "remove", etc.

Chain makes things (IMHO) easier to read. Without it, I'd have had something like this:
```javascript
return _.flatten(_.values(_.groupBy(activities, activity => {
    return activity.name.split(' ').shift();
})));
```
Which you'd have to read right to left since the most interesting bit is the groupBy. As written, you can just read left to right.

gulp/src/manageusers/Role.jsx
```diff
@@ -158,12 +157,9 @@ class Role extends React.Component {
     });
 
     // Format assigned activities
-    let formattedAssignedActivities = [];
-    if (assignedActivities) {
```
aspidites: The if check is redundant as mapping over an empty list just returns an empty list.

JLMcLaugh:
Merge conflicts

aspidites:
Feel free to hijack this and fix up the tests if you don't want to wait until I'm back from vacation.

myjobs/models.py
```diff
@@ -881,14 +881,23 @@ class Activity(models.Model):
     """
     An activity represents an individual task that can be performed by a
     user.
+
     """
+    class Meta:
+        verbose_name_plural = "Activities"
+
     name = models.CharField(max_length=50, unique=True)
+    display_name = models.CharField(max_length=50, blank=True)
```
lemniscate613: :+1: Looks good. When I thought about this earlier the idea was to avoid creating a migration. I think this makes the most sense though.

tdphillips: @aspidites wasn't at the sprint meeting.

I think @JLMcLaugh had some ideas regarding using \_\_unicode\_\_() for this.

gulp/src/manageusers/sortActivities.js
lemniscate613: I haven't been receiving mentions like this in my email, and I only stumbled upon this. Thank you for taking the time to give the feedback. I just updated my Github settings so I should get mentions now.

PR #2107, opened 2016-02-12
[PD-2167] Replace Django Extensions with Ajax Select

aspidites:
# Description
django-ajax-select can be used outside of django admin, which was not the case for extensions. This PR replaces all occurences of django-extensions with django-ajax-select. 

# Testing
1. Remove django extensions
```pip uninstall django_extensions```
2. Install django-ajax-select
```pip install -r requirements
3. Visit admin and check that places where you expect there to be auto complete are still there.
To the best of my knowledge, that includes the following admin pages:
  - Seo -> SeoSite
  - Seo -> Query Redirects
  - MyPartners -> Nonuser Outreach Domains


tdphillips:
Deletions > additions. :+1:

JLMcLaugh:
The failing tests check admin pages based on PROJECT_APPS. Do we need to remove something?

PR #2108, opened 2016-02-12
Remove deps

aspidites:
Just some house cleaning. We don't use this dependencies, so I removed it.


templates/includes/error-highlight.html
```diff
@@ -1,4 +1,3 @@
-{% load widget_tweaks %}
```
JLMcLaugh: @tdphillips Could you check this if you haven't? I think you're the one who added it.

tdphillips: We don't appear to use anything contained in widget_tweaks in this file.

PR #2109, opened 2016-02-12
Pd 2107 - update posting approval form to standard

misterjonscott:
# Changes
Made changes to the template, styles, and form, to adopt new standards.  These changes affect the look of microsites' view of this page.

# Notes for testing
You need to view a microsite, on local dev, and be a user with that microsite.  Beyond that, that company must have the feature enabled, the user must have access to that feature, and the microsite must by linked to the company.  

http://californiahotel.jobs/posting/job/add/
is a good example, if you've added californiahotel.jobs to your hosts file.

NOTE: You will need to login to the site, first, at californiahotel.jobs/admin


templates/includes/error-highlight.html
```diff
@@ -4,59 +4,54 @@
...
-                <span class="help-block">{{ field.help_text|safe }}</span>
-            {% endif %}
-        {% endif %}
-    </div>
-{% endif %}
-
-{% if field.errors %}
-    </span>
+        <div class="row">
+            <div class="col-xs-12 col-md-4 {% if is_boolean %}checkbox-label{% else %}initial{% endif %} {{ field.label|slugify }}-label">
```
tdphillips: These classes don't exist in Bootstrap 2.2.1, which is still in use on most of our pages.

misterjonscott: Hmmm...  That's a pretty big problem and oversight, here.  Without this responsive framework, we can't properly update this form, but it uses form-error-highlight...  Wondering if we can just include BS3 throughout the site, and support both for a transitional period.

misterjonscott: Alternatively, maybe we duplicate the error templates and call the new version when transitioning a form? Thoughts?

tdphillips: @fezick, any thoughts? I would prefer not having two templates as that's why we currently have two topbars.

aspidites: What @tdphillips  said. I got away with using bootstrap 3 for my changes because of react-base, but that only lives on secure, not microsites.

misterjonscott:
Added transitional templates that will be used only while the remaining tickets for other forms are being updated, at which time we'll remove the BS2.0-based templates.

Also added some supporting styles to retain base appearance in microsites, shimming for differences between BS2.0 and BS3.*

misterjonscott:
@darrint I've made updates to keep the BS3 local for this piece, as well as adding form highlighting on error.

JLMcLaugh:
Don't we still need to manually version static files outside of gulp?

misterjonscott:
@JLMcLaugh If they're wrapped in compress tags, they get hashed based on their contents.  I talked to @mklauber about this.

JLMcLaugh:
These don't seem to be wrapped in compress tags, and I don't see any versioning on S3.

misterjonscott:
Added versioning to JS and CSS where appropriate.  I don't see any instances of unwrapped custom.css, as @tdphillips mentioned...  I've also replaced the styles for the nav menu in DSEO, and added functionality to the template to act as an active indicator.

tdphillips:
https://github.com/DirectEmployers/MyJobs/blob/quality-control/templates/base-react.html

tdphillips:
I was on master, where one instance was not wrapped. Sorry for the confusion.

misterjonscott:
I've added a PDF, with the form field visual guidelines for testing, to the original ticket.

lemniscate613:
_So_ much better. The old version's columns contained high quantities of goofiness:

screen shot 2016-02-19 at 2 02 35 pm

__Here are some observations:__

Upon error, text fields have what appears the correct color (F00), but .error-text is the old brownish red (900):

screen shot 2016-02-19 at 2 01 17 pm

The errors are missing the exclamation mark from the standard:

screen shot 2016-02-19 at 1 30 44 pm

The error messages end with periods here, but in the design standard they don't. What's the accepted way generally? Not sure, but it reminds me of writing a bulleted list -- each a short phrase. In that case, my opinion is to:
1. ensure errors are always simple as opposed to sentences (unless its a message in the new expanded "Details" section
2. lose the capitalization
3. lose the periods. That way, they are similar to point 2 here: http://www.oxforddictionaries.com/us/words/bullet-points-american

It'd be helpful to have a 3rd tab for "Post/Add a Job," that's designed to show it's active:

screen shot 2016-02-19 at 1 35 43 pm

If I don't list an Application Method, I get an error, but the relevant text box isn't colored red:

screen shot 2016-02-19 at 1 34 08 pm

PR #2113, opened 2016-02-15
Fix object reference

koonsies:
I originally had the site object inserted directly into the URL instead of the domain of the site object. This corrects that.


koonsies:
@lemniscate613

PR #2119, opened 2016-02-18
[PD-2165] User management - update activity display name

lemniscate613:
Since the front end only knows about Activity.display_name, that's what it sends to the view. Therefore, the view needs to look up the Activity by `display_name` and not `name`.

### DIFF
--- a/gulp/src/manageusers/Role.jsx
+++ b/gulp/src/manageusers/Role.jsx
@@ -122,7 +122,7 @@ class Role extends React.Component {
       const tempRef = activity.app_access_name.replace(/\s/g, '');
       const selected = refs.activities.refs[tempRef].state.assignedActivities;
       _.forOwn(selected, function loopThroughEachSelectedActivity(item) {
-        assignedActivities.push(item);
+        assignedActivities.push(item.name);
       });
     });
     // User must select AT LEAST ONE activity
--- a/gulp/src/manageusers/Role.jsx
+++ b/gulp/src/manageusers/Role.jsx
@@ -122,7 +122,7 @@ class Role extends React.Component {
       const tempRef = activity.app_access_name.replace(/\s/g, '');
       const selected = refs.activities.refs[tempRef].state.assignedActivities;
       _.forOwn(selected, function loopThroughEachSelectedActivity(item) {
-        assignedActivities.push(item.name);
+        assignedActivities.push(item.id);
       });
     });
     // User must select AT LEAST ONE activity
--- a/myjobs/tests/test_manage_users.py
+++ b/myjobs/tests/test_manage_users.py
@@ -87,9 +87,11 @@ class ManageUsersTests(MyJobsBase):
         """
         Tests creating a role
         """
+        an_activity_id = self.role.activities.first().id
+
         data_to_post = {}
         data_to_post['role_name'] = "TEST ROLE"
-        data_to_post['assigned_activities[]'] = ['read role']
+        data_to_post['assigned_activities[]'] = [an_activity_id]
         response = self.client.post(reverse('api_create_role'), data_to_post)
         output = json.loads(response.content)
         self.assertEqual(output["success"], "true")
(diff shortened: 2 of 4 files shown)


myjobs/views.py
```diff
@@ -871,7 +871,7 @@ def api_create_role(request):
             activities = request.POST.getlist("assigned_activities[]", "")
             # Create list of activity_ids from names
             for activity in enumerate(activities):
-                activity_object = Activity.objects.get(name=activity[1])
+                activity_object = Activity.objects.get(display_name=activity[1])
```
JLMcLaugh: We have an id. Whey are we querying by an un-indexed name field?

JLMcLaugh: Let's wait on merging this. I'd like to go work with @lemniscate613 on removing this loop, changing the query, and possibly switching to using __unicode__ method instead of a separate display name method.

lemniscate613: Okay, closing the PR for now.

gulp/src/manageusers/Role.jsx
```diff
@@ -122,7 +122,7 @@ class Role extends React.Component {
       const tempRef = activity.app_access_name.replace(/\s/g, '');
       const selected = refs.activities.refs[tempRef].state.assignedActivities;
       _.forOwn(selected, function loopThroughEachSelectedActivity(item) {
-        assignedActivities.push(item);
+        assignedActivities.push(item.id);
```
lemniscate613: Send PK instead of display_name, so the view can access it directly without a lookup.

myjobs/tests/test_manage_users.py
```diff
@@ -87,9 +87,11 @@ def test_create_role(self):
         """
         Tests creating a role
         """
+        an_activity_id = self.role.activities.first().id
+
         data_to_post = {}
         data_to_post['role_name'] = "TEST ROLE"
-        data_to_post['assigned_activities[]'] = ['read role']
+        data_to_post['assigned_activities[]'] = [an_activity_id]
         response = self.client.post(reverse('api_create_role'), data_to_post)
```
lemniscate613: The front end no longer sends activity names, but PKs.

myjobs/views.py
```diff
@@ -866,14 +866,8 @@ def api_create_role(request):
...
-
         if request.POST.getlist("assigned_activities[]", ""):
-            activities = request.POST.getlist("assigned_activities[]", "")
-            # Create list of activity_ids from names
-            for activity in enumerate(activities):
-                activity_object = Activity.objects.get(name=activity[1])
-                activity_id = activity_object.id
-                activity_ids.append(activity_id)
+            activity_ids = request.POST.getlist("assigned_activities[]", "")
         # At least one activity must be selected
```
lemniscate613: No need to do this lookup if we already have the activity PKs.

myjobs/views.py
```diff
@@ -955,20 +949,14 @@ def api_edit_role(request, role_id=0):
...
             ctx["message"] = "At least one activity must be assigned."
             return HttpResponse(json.dumps(ctx),
                                 content_type="application/json")
-        # Create list of activity_ids from names
-        activity_ids = []
-        for i, activity in enumerate(activities):
-            activity_object = Activity.objects.filter(name=activity)
-            activity_id = activity_object[0].id
-            activity_ids.append(activity_id)
         # INPUT - assigned_users
```
lemniscate613: Ditto.

aspidites:
Lookup by PK is a bad idea, as they could easily vary between production, qc, and staging. In fact, the whole reason for introducing display_name was so that we could reliably use name for queries as those would be fixed, unlike foreign keys.

myjobs/views.py
```diff
@@ -866,14 +866,8 @@ def api_create_role(request):
...
         activity_ids = []
-
-        if request.POST.getlist("assigned_activities[]", ""):
-            activities = request.POST.getlist("assigned_activities[]", "")
-            # Create list of activity_ids from names
-            for activity in enumerate(activities):
-                activity_object = Activity.objects.get(name=activity[1])
-                activity_id = activity_object.id
-                activity_ids.append(activity_id)
+        if request.POST.getlist("assigned_activities[]"):
```
JLMcLaugh: activity_ids = request.POST.getlist("assigned_activities[]", []) would turn 3 lines into 1.

PR #2120, opened 2016-02-18
Call correct _trackPageview method on secureMyJobs tracker [PD-2184]

JLMcLaugh:
This was never exactly right, but this line was probably failing harmlessly or after the secure.my.jobs-specific beacon was sent. Now, it looks like there are exceptions that keep that from happening, and GA is not tracking the secure.my.jobs beacon.

The next step should be moving from ga.js to analytics.js.


lemniscate613:
As far as the syntax of `_trackPageview` specifically and the eventual need to move to analytics.js:

:+1:

I haven't grokked our full use of GA yet.

PR #2126, opened 2016-02-22
Refined button class to work with Input buttons, and changed template…

misterjonscott:
… for form handling.


misterjonscott:
# Changes

- Changed the template to use the new (transitional) BS3 enabled template for form-error-highlight.
- Refined styles for button to allow Input type:submit buttons to be used as well.
- Added classes as needed.

# Testing Notes

The changes are purely visual, and should look like the updated screenshot, shown in the ticket (PD-2151)

PR #2142, opened 2016-02-24
Updated the way we handle publishing documentation.

aspidites:
# Differences
- we need to track the output of the build directory for subtrees to work
- as a bonus, that means that someone could open up build/html/index.html to browse generated documentation locally, without needing to install sphinx
- we only need the single Makefile and it lives in doc per usual
- the contents of the build directory are converted to a branch (gh-pages) and published
- make file doesn't rely on adhoc variables to decide what to copy over to the branch
- far less commands needed to publish
- less risk of leaving the various branches in a bad state since the local gh-pages branch is deleted immediately after running the appropriate command

# How to Build Documentation
For local docs, just:
```
cd doc
make html
```

For Publishing docs
```
make ghpages
# enter password

### CODE REVIEW

**Comment by koonsies** [comment]
Without having tried this, I'm reading this as.. A branch is opened with the built GH pages doc after you run the file. Do you then make a PR for the changes? I'm assuming "no", but I don't have a firm grasp on where the built documentation is stored.

**Comment by aspidites** [comment]
@koonsies not quite. Grasping it probably would be easier if you try yourself. You could do so by forking MyJobs into your personal account so that you would be publishing to koonsies/MyJobs@gh-pages if you're concerned about breaking things.

I'll try and explain what's happening anyway.

Before running make:

/doc
/mypartners
...

After running `make html`
/doc
/doc/build/html
/mypartners
...

After running `make ghpages`
/doc
/mypartners

So what `make ghpages` actually does is:
- run the html target, which creates artifacts in `build/html`
- using git subtree, turn the entire contents of `build/html` into it's on branch and push that
- delete the locally created gh-pages branch so we can do this again in the future

In short, it's the same as what we were doing before in spirit, but far less error prone and more clever (eg, we don't have to manually list all the files we want to be in the new branch by hand, since `git subtree` takes care of that for us. We just point to a subdirectory and say "yo, make that a branch".

**Comment by koonsies** [comment]
That makes sense. But in the end, the build will wind up being part of the MyJobs repo? I'm a little concerned that by no longer ignoring build on Git, we'll be constantly having built documentation trying to get added into commits on accident. Would it be better to create a new repo specifically for docs and have the subtree split populate -that- repo?

I may still be misunderstanding something, but thanks for the explanation.

**Comment by aspidites** [comment]
In practice though, you shouldn't ever be making build artifacts in the course of normal development, jenkins will.

@koonsies  if you want to make that your personal project, that's fine, but I won't personally be doing so. Also keep in mind that for that to work and have the same url as this does, you have to use submodules, which are significantly more complicated and introduce other complications.

**Comment by aspidites** [comment]
Alternatively you can add this to ".git/info/exclude"
```
doc/build
```

Which would ignore build artifacts locally, thus allowing you to build without committing new files unnecessarily.

aspidites:
Also note that without submodules building in a separate repo defeats one of the primary goals of having documentation living close to the code: it's less of a barrier to entry. With module references in documentation, a separate repo would also have to carry around the code its referencing. If not, we would end up having to copy paste module docstrings into rst files to get the same effect as we have now.

koonsies:
Oh, I definitely did not mean moving the source files to a separate repo, I meant potentially moving the build file branch to a separate repo (thus only build files existing in that repo, and the github pages pointing to that repo). The documentation source would still be close to the code and the module references would have no issues. 

In the current design, if I were to build locally in order to ensure my RST is displaying correctly, I would be inundated with file changes that aren't actually pertinent to the PR at hand. 

I see modifying the exclude file as an option, I am just curious if there is a more elegant option that didn't involve having to manually change the exclude file.

aspidites:
Feel free to open up a PR exploring that option. However, what you suggest
is a large undertaking that I'm not personally committed to spending my
time on. I'm also not interested in maintaining the current Makefile.

Thus, if this isn't seen as ideal, we simply need to understand that
documentation will be broken until someone comes up with something else.

koonsies:
I'm fine with leaving it as-is and seeing how it works out. If it becomes a stumbling block, we can reevaluate.

PR #2143, opened 2016-02-25
[PD-1952]  Restore Report Naming

darrint:
## Test

* Go to `/reports/views/dynamicoverview`
* Create a report. On the filter screen you can name the report.
* The default should be a timestamp in the same format as the production reporting system.
* If you leave the field blank you get an error.

## Note
* There is _no_ server side validation. This will be addressed in [PR-2193](https://directemployersfoundation.atlassian.net/browse/PD-2193).

### DIFF
--- a/gulp/package.json
+++ b/gulp/package.json
@@ -37,6 +37,7 @@
     "tinycolor2": "~1.3.0",
     "warning": "^2.1.0",
     "webpack": "^1.12.9",
-    "lodash-compat": "~3.10.1"
+    "lodash-compat": "~3.10.1",
+    "moment": "~2.11.2"
   }
 }
--- a/gulp/src/common/ui/ValidatedInput.jsx
+++ b/gulp/src/common/ui/ValidatedInput.jsx
@@ -51,8 +51,25 @@ export class ValidatedInput extends Component {
 }
 
 ValidatedInput.propTypes = {
+  /**
+   * This contains a controlled input. Value in the box goes here.
+   */
   value: PropTypes.string.isRequired,
+
+  /**
+   * Function to call with input change values.
+   */
   onValueChange: PropTypes.func.isRequired,
-  helpText: PropTypes.string.isRequired,
+
+  /**
+   * Help text to show alongside the input control.
+   */
+  helpText: PropTypes.string,
+
+  /**
+   * Array of strings to show as errors alongside the input control.
+   *
+   * If this prop is missing or empty the control will display as valid.
+   */
   errorTexts: PropTypes.array,
 };
--- a/gulp/src/common/ui/ValidatedInput.jsx
+++ b/gulp/src/common/ui/ValidatedInput.jsx
@@ -1,0 +1,58 @@
+import React, {PropTypes, Component} from 'react';
+import classnames from 'classnames';
+import {map} from 'lodash-compat/collection';
+
+
+/**
+ * Input field which shows help text and errors.
+ */
+export class ValidatedInput extends Component {
+  renderError(errorText, key) {
+    return <span key={key} className="error-text">{errorText}</span>;
+  }
+
+  renderErrors() {
+    const {errorTexts} = this.props;
+    if (!errorTexts) {
+      return '';
+    }
+
+    if (errorTexts.length === 1) {
+      return <div>{this.renderError(errorTexts[0])}</div>;
+    }
+
+    const errors = map(errorTexts, (errorText, i) =>
+        <li>{this.renderError(errorText, i)}</li>);
+
+    return <div><ul>{errors}</ul></div>;
+  }
+
+  render() {
+    const {value, helpText, errorTexts, onValueChange} = this.props;
+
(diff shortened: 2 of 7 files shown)


gulp/package.json
```diff
@@ -37,6 +37,7 @@
     "tinycolor2": "~1.3.0",
     "warning": "^2.1.0",
     "webpack": "^1.12.9",
-    "lodash-compat": "~3.10.1"
+    "lodash-compat": "~3.10.1",
+    "moment": "~2.11.2"
```
darrint: This appears to be the leading brand for javascript time handling libraries. I wouldn't bother but the native Date class lacks even a basic strftime.

Here is a google link to some [javascript date handling alternatives](https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=npm+date+handling). Now would be a good time to object to moment.js.

It's 13k minified which is a little large for what I'm _currently_ using it for. I'm getting a warm fuzzy about it being more capable than that though.

gulp/src/common/ui/ValidatedInput.jsx
```diff
@@ -0,0 +1,75 @@
+import React, {PropTypes, Component} from 'react';
+import classnames from 'classnames';
+import {map} from 'lodash-compat/collection';
+
+
+/**
+ * Input field which shows help text and errors.
+ */
+export class ValidatedInput extends Component {
+  renderError(errorText, key) {
```
darrint: Real UX folks:

I want to not make the class names configurable here if I can get away with it. What should the classNames really contain?

gulp/src/reporting/reportEngine.js
```diff
@@ -161,8 +168,26 @@ export class ReportConfiguration {
...
+  changeReportName(name) {
+    this.name = name;
+  }
+
+  getReportName() {
+    return this.name;
+  }
+
+  getReportNameError() {
+    if (!this.name) {
```
darrint: This is where the client side validation is happening.

gulp/src/reporting/spec/reportEngineSpec.js
```diff
@@ -158,8 +162,31 @@ describe('ReportConfiguration', () => {
...
+    spyOn(fakeComponent, 'newReportNote').and.callThrough();
+
+    config.changeReportName('bbb');
+    const result = await config.run();
+
+    expect(fakeApi.runReport).toHaveBeenCalledWith(2, 'bbb', {});
+    expect(result).toEqual(7);
+  }));
+
+  it('warns about invalid report names', promiseTest(async() => {
```
darrint: Here's the test for the client side validation.

gulp/src/reporting/wizard/WizardPageFilter.js
```diff
@@ -93,13 +102,21 @@ export class WizardPageFilter extends Component {
...
+    const {loading, reportConfig, reportName, reportNameError} = this.state;
 
     if (loading) {
       return <Loading/>;
     }
 
     const rows = [];
+    const errorTexts = reportNameError ? [reportNameError] : null;
+    rows.push(this.renderRow('Report Name', 'reportName',
+      <ValidatedInput
```
darrint: This is where we include the validated input component.

tdphillips:
Tested and works.

:+1: pending whoever is going to respond regarding class names.

gulp/package.json
aspidites: I'll object to this having already not been objected to. We raised a bit of a fuss when I introduced a dependency for something that literally wasn't possible without the dependency. We hand-rolled a readCookie funciton. Seems doing so for strftime should be possible without adding 13k. In fact, the function you want is self-contained in a [separate module](https://github.com/moment/moment/blob/develop/src/lib/moment/format.js).

gulp/src/common/ui/ValidatedInput.jsx
aspidites: In some places, you are importing classNames, while in others, it is classnames. Is that intentional? At the very least, it's confusing.

gulp/package.json
JLMcLaugh: It does seem like overkill if we're using it in 1 place to avoid ~10 lines of code and not worried about timezones. @misterjonscott or @lemniscate613, any thoughts? Any libraries that solve this and would also help with https://directemployersfoundation.atlassian.net/browse/PD-1971?

darrint: I think this was my source for picking moment.js:

http://stackoverflow.com/questions/85116/how-do-i-display-a-date-time-in-the-users-locale-format-and-time-offset

@aspidites moment isn't nice and modular like lodash. https://github.com/moment/moment/issues/2373

I'm going to push this down to python and figure out the default report name on the server side.

gulp/src/common/ui/ValidatedInput.jsx
darrint: The use of "classNames" in Tag.js is I think what you mean. That was a typo which I will fix.

PR #2144, opened 2016-02-25
[PD-2146] Logout Other tabs when main tab is logged out of

aspidites:
# Description
What it says on the tin. Basically, if you have two tabs logged into My.jobs and you click "log out" on one of them, the other should log out as well.

# Details
I implemented this by setting a timeout that checks for a "loggedout" cookie. If the cookie is present, I log you out of other tabs.

# Tangential Changes
- refactored readCookie into utils.js (can't use the one in gulp, and having gulp output what we want is a pain)
- anything that was using read_cookie or defining readCookie is using utils instead and loaded with compress
- fixed documentation heading


static/utils.js
```diff
@@ -0,0 +1,44 @@
...
+      timer = window.setInterval(function() {
+        // if we are logged out and not already on the home page
+        if (this.readCookie('loggedout') && window.location.pathname !== url) {
+          window.location.assign(url);
+        }
+      }, 1000);
+    } else {
+      window.clearInterval(timer);
+    }
+  },
```
koonsies: This could be handled with an onFocus method, which would prevent having constant timers running, as well as not causing a gang of tabs to reload at the same time. 

Below is an example
http://stackoverflow.com/a/13513980

koonsies:
Good call on the utils.js file. I was considering making a file such as this as well, as I have some redundant code in my javascript (namely, CORS/JSONP ajax requests). 

This seems to be a good solution.

static/utils.js
tdphillips: That's assuming we don't want the tabs to reload. If you log out and then back in before switching to the tab containing sensitive info, what happens? The cookie says you're logged in.

tdphillips: I'm not saying that wouldn't work, but our cookie would need to be "I am X and I'm logged in" whereas it's currently "I'm logged in"

aspidites: Why? I'm not, in this PR, concerned with browsers running on two different
machines (there's another ticket for that). You shouldn't be allowed to log
in as multiple users at once, so I'd think knowing who X is would be
irrelevant.

Unless I've misunderstood what the ticket is meant to accomplish.

tdphillips: Both of my comments are related and are a response to @koonsies' comment.

aspidites: Oh. My apologies. That probably would have been more obvious if u were at a
computer rather than viewing this in my email.

Maybe that's a hint to turn off work notifications on my phone?

static/utils.js
```diff
@@ -0,0 +1,44 @@
...
+var timer;
+
+var utils = {
+  /**
+   * Returns the value of the cookie. Returns null if that cookie is not found.
+   * cookie: The name of the cookie to retreive.
+   **/
+  readCookie: function(cookie) {
+    const nameEQ = cookie + '=';
+    const ca = document.cookie.split(';');
```
koonsies: Using let/const will not work in legacy browsers.

static/utils.js
koonsies: There is also the issue of the cookie being unset before all tabs are refreshed. I opened 6 tabs to Manage Users and logged out of one.. Two tabs refreshed but the remaining three did not.. I am assuming this is because those three had not executed their timer before the cookie was unset by the others.

koonsies: Oops. Ignore that. The above javascript error had prevented a few of those tabs from refreshing, not the cookie.

static/utils.js
aspidites: Yay copy/paste. Thanks for catching this.

PR #2145, opened 2016-02-26
Removal of test prefix trim

koonsies:
In my most recent PR, I attempted to remove specific handling for QC/Staging from the secure blocks logic.

Well.. I missed a spot.

This is the removal of the -REST- of the logic.


misterjonscott:
Looks legit.

PR #2146, opened 2016-02-26
Changed Radio logic to select logic and adjusted JS to match.

misterjonscott:
# Changes

This is a slight change to an already merged PR, to change the radio selects to a select input.

# Notes

There may be a better way to make this element render as a select, but I wasn't able to figure it out and wanted to get this finished by EOD.

### DIFF
--- a/postajob/forms.py
+++ b/postajob/forms.py
@@ -450,7 +450,7 @@ class ProductForm(RequestForm):
     job_limit_choices = [('unlimited', "Unlimited"),
                          ('specific', 'A Specific Number'), ]
     job_limit = CharField(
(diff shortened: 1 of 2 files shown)


lemniscate613:
Looking at this now...

tdphillips:
https://github.com/DirectEmployers/MyJobs/blob/PD-2108/static/postajob.173-18.js#L97

This is hiding the entire form now.

tdphillips:
The version of that javascript file wasn't updated.

lemniscate613:
:+1: 

I can merge this once @tdphillips confirms it's not flashing for him any more ( @misterjonscott thinks it might be a settings isssue...?).

PR #2149, opened 2016-03-02
[PD-2209] Allow use of confirm, and alert etc.

darrint:
We've been looking at a warning about the use of JS confirm for weeks. Since the code is shipping I'm removing the warning.

### DIFF
--- a/gulp/gulpfile.js
+++ b/gulp/gulpfile.js
@@ -186,6 +186,8 @@ function lintOptions() {
     rules: {
       "babel/object-curly-spacing": 1,
       "babel/no-await-in-loop": 2,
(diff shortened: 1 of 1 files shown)


aspidites:
@fezick  @JLMcLaugh, thoughts? I don't really care, but want to make sure before I go pushing buttons.

darrint:
Closing due to the fact that I don't know if we will be sticking with bootstrap in the future.

aspidites:
The effort of converting everything using bootstrap to foundation is surely higher than the effort to add a modal in place of a confirm. I count 17 references in the react apps to bootstrap. You even recently submitted a pr which made direct use of bootstrap class names.

PR #2150, opened 2016-03-02
[PD-2116] Removed CompanyUser from Postajob

aspidites:
# Description
I removed the two foreign keys pointing to a `ComapnyUser` as that model is being deprecated in favor of roles and activities in the future. I updated tests where it was necessary in addition to some helper methods. I intentionally ignored obvious refactoring as nothing seems to be *that* obvious in postajob. That is, I attempted to do what seemed to be rudimentary refactoring yesterday in one module only to find out that said refactoring broke 4 or 5 other modules. This is precisely the reason that I ignored the fact that, for instance, `user_has_access` is declared in multiple places. 

# Assumptions
- Anywhere that we were checking for company user, we were really checking for users assigned as admin for some company
- We will *eventually* want to replace the use of groups with roles including appropriate activities, but until said activities exists, this should be sufficient
- Postajobs is tested sufficiently enough that should code change which doesn't introduce new concepts, those same tests should be sufficient to prove that everything still works (at least until things hit QC)
- the company tied to created_by is the same as the one tied to owner, so owner + created_by as a regular user should be equivalent

# Testing
Uh, if you're lucky enough to have postajob running locally (by following the instructions in the [documentation](http://directemployers.github.io/MyJobs/developers/postajob/00_quickstart.html#posting-directly-to-a-microsite), then making sure everything works as it did before should be good enough. Otherwise, I assume we aren't really going to be testing  this properly until it's on QC pointing to a populated solr instance.


myjobs/models.py
```diff
@@ -766,6 +761,11 @@ def secondary_emails(self):
     secondary_emails.short_description = "secondary emails"
     secondary_emails.allow_tags = True
 
+    def make_purchased_microsite_admin(self):
```
aspidites: Consider this a stub method. Once activities are implemented, this will be unnecessary (as will `ADMIN_GROUP_NAME`).

postajob/management/commands/test_posting.py
```diff
@@ -116,12 +116,14 @@ def setup_objects(cls):
         cls.CREATION_ORDER.append(cls.admin_company_2)
 
         # Company Users:
-        cls.admin_company_user = CompanyUser.objects.create(
-            company=cls.admin_company, user=cls.admin)
-        cls.CREATION_ORDER.append(cls.admin_company_user)
-        cls.admin_company_user_2 = CompanyUser.objects.create(
-            company=cls.admin_company_2, user=cls.admin_2)
-        cls.CREATION_ORDER.append(cls.admin_company_user_2)
+        cls.admin_role = Role.objects.create(
```
aspidites: This whole file is not used at the moment. It's apparently part of some selenium tests, but hooks up to production instances of various services, so I'd much rather if we could get rid of it all together.

tdphillips: qc or staging instances. Without modifying the code, it bails if run against production.

postajob/tests/test_models.py
```diff
@@ -228,7 +228,12 @@ def test_purchased_product_jobs_remaining(self):
                              expected_num_jobs)
         self.assertFalse(self.purchased_product.can_post_more())
 
-    def test_invoice_send_invoice_email(self):
+    def test_invoice_sent_to_specific_email(self):
```
aspidites: This used to be one large test, so I wanted the different use cases differentiated (and commented).

postajob/tests/test_views.py
```diff
@@ -851,21 +853,9 @@ def test_offlinepurchase_add_without_company(self):
         self.assertIn(offline_purchase.redemption_uid,
                       response.content.decode('utf-8'))
 
-    @skip('Feature disabled for now.')
```
aspidites: This feature was removed more than a year ago, so no need to test it.

postajob/management/commands/test_posting.py
aspidites: Ah. I vaguely remember you implementing that. We still import production secrets. I can't even run this command.

aspidites:
Need to fix get_company_or_404...again...

PR #2152, opened 2016-03-03
Add messages, do PEP 8 [PD-2213]

tdphillips:
I left some of the messages alone as, while they don't include method names, they are fairly descriptive and should be unique if searching the codebase.

Testing example: /candidates/view/details/ should now contain a vaguely helpful error message.

I'm open to further suggestions.

All tests pass.

### DIFF
--- a/seo/admin.py
+++ b/seo/admin.py
@@ -479,7 +479,7 @@ class RowPermissionsAdmin(admin.ModelAdmin):
             raise PermissionDenied
 
         if obj is None:
-            raise Http404(u"{name}{transblock}{key} {noexist}".format(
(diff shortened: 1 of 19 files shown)


seo/admin.py
```diff
@@ -479,7 +479,7 @@ def change_view(self, request, object_id, extra_context=None):
             raise PermissionDenied
 
         if obj is None:
-            raise Http404(u"{name}{transblock}{key} {noexist}".format(
+            raise Http404(u"{name} {transblock} {key} {noexist}".format(
```
tdphillips: This is not how translation works. :cry:

PR #2154, opened 2016-03-03
[PD-2219] Remove Special Error Messaging

aspidites:
Before, you had to have your site misconfigured in a specific way to see helpful error messages (user facing). As we (correct me if I've misunderstood, @fezick) are now considering each of the ways in which a page may fail to load a configuration issue, it's easier to just 404. That said, in a debug environment (eg, local and QC), you'll get helpful error messages about what configuration needs to be done to allow access to the page which was 404ing. 

In the future, when activities are set up, there is always the opportunity to pass the @requires decorator a callback to give the user more useful error messages. That's all handled in a single decorator though, so reasoning about misconfigured companies should be way easier than it is today. I also went ahead and moved some decorators from urls.py to views.py, which should make debugging much easier.


aspidites:
Closing until my other one (which this is built on top of) is merged.

postajob/urls.py
```diff
@@ -24,67 +18,32 @@
...
-         'block': False},
-        name='deny_admin_request'),
-    url(r'^admin/request/block/(?P<pk>\d+)/$',
-        company_in_sitepackages(message_when_site_misconfigured(
-            feature='Requests are')(
-                views.process_admin_request)),
-        {'approve': False,
-         'block': True},
-        name='block_admin_request'),
+    url(r'^admin/request/approve/(?P<pk>\d+)/$', views.process_admin_request,
```
tdphillips: :+1:

PR #2157, opened 2016-03-07
[PD-2212]  Remove jQuery from User Management

darrint:
* Create a common API class useful from any app.
* Use it from Manage Users and Reporting.
* Remove use of `/* global $ */` in Manage Users.

We depend on a common behavior in jQuery implemented by it's param function starting in jQuery 1.4.

http://api.jquery.com/jquery.param/

This affects sending arrays and objects as form data. jQuery renames array keys to `field[]` and object keys to `field[key]`.

Using this I can use the same code for all browsers.

### DIFF
--- a/gulp/package.json
+++ b/gulp/package.json
@@ -37,6 +37,7 @@
     "tinycolor2": "~1.3.0",
     "warning": "^2.1.0",
     "webpack": "^1.12.9",
-    "lodash-compat": "~3.10.1"
+    "lodash-compat": "~3.10.1",
+    "jquery-param": "~0.2.0"
   }
 }
--- a/gulp/src/common/myjobs-api.js
+++ b/gulp/src/common/myjobs-api.js
@@ -1,23 +1,4 @@
-// This is needed for IE8. The fetch-polyfill module doesn't quite do enough.
-// based on:
-// http://stackoverflow.com/questions/24710503/how-do-i-post-urlencoded-form-data-with-http-in-angularjs
-function ancientSerialize(obj) {
-  const str = [];
-  for (const p in obj) {
-    if (obj.hasOwnProperty(p)) {
-      const type = typeof obj[p];
-
-      if (type === 'object' || type === 'array' || type === 'function') {
-        throw new Error(
-                  'Tried to serialize non primitive object. key: ' + p);
-      }
-      str.push(encodeURIComponent(p) + '=' + encodeURIComponent(obj[p]));
-    }
-  }
-  return str.join('&');
-}
-
-const hasFormData = !(window.FormData === undefined);
+import param from 'jquery-param';
 
 /**
  * Utility for making XHR calls from all our supported browsers.
@@ -27,10 +8,6 @@ const hasFormData = !(window.FormData === undefined);
 export class MyJobsApi {
   constructor(csrf) {
     this.csrf = csrf;
-  }
-
-  withCsrf(formData) {
-    return {...formData, csrfmiddlewaretoken: this.csrf};
   }
 
   checkStatus(response) {
@@ -47,16 +24,6 @@ export class MyJobsApi {
     return response.json();
   }
 
-  objectToFormData(obj) {
-      // No good way around this:
-    if (hasFormData) {
-      const formData = new FormData();
-      Object.keys(obj).forEach(k => formData.append(k, obj[k]));
-      return formData;
-    }
-    return ancientSerialize(obj);
-  }
-
   async get(url) {
     const response = await fetch(url, {
       method: 'GET',
@@ -65,17 +32,25 @@ export class MyJobsApi {
     return this.parseJSON(this.checkStatus(response));
   }
 
-  async post(url, data) {
-    const formData = this.objectToFormData(this.withCsrf(data));
+  async ajaxWithFormData(method, url, data) {
+    const formData = param(data);
     const response = await fetch(url, {
-      method: 'POST',
+      method: method,
       body: formData,
       credentials: 'include',
-          // No good way around this.
-      headers: hasFormData ? undefined : {
+      headers: {
         'Content-Type': 'application/x-www-form-urlencoded',
+        'X-CSRFToken': this.csrf,
       },
     });
     return this.parseJSON(this.checkStatus(response));
+  }
+
+  async post(url, data) {
+    return this.ajaxWithFormData('POST', url, data);
+  }
+
+  async delete(url, data) {
+    return this.ajaxWithFormData('DELETE', url, data);
   }
 }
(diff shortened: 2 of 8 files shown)


gulp/src/common/myjobs-api.js
```diff
@@ -0,0 +1,56 @@
...
+
+  async ajaxWithFormData(method, url, data) {
+    const formData = param(data);
+    const response = await fetch(url, {
+      method: method,
+      body: formData,
+      credentials: 'include',
+      headers: {
+        'Content-Type': 'application/x-www-form-urlencoded',
+        'X-CSRFToken': this.csrf,
```
darrint: This header works for all http methods in my testing. This eliminates the need to modify request bodies to include the csrf token.

gulp/src/manageusers/Role.jsx
```diff
@@ -83,10 +44,49 @@ class Role extends React.Component {
       activites: currentActivitiesObject,
     });
   }
-  handleSaveRoleClick() {
+  async initialApiLoad() {
```
darrint: This reads more clearly with whitespace differences hidden.

gulp/src/manageusers/Role.jsx
```diff
@@ -30,43 +27,7 @@ class Role extends React.Component {
...
-        this.setState({
-          apiResponseHelp: '',
-          roleName: '',
-          availableUsers: roleObject.available_users,
-          assignedUsers: [],
-          activities: activities,
-        });
-      }.bind(this));
-    }
+    this.initialApiLoad();
```
darrint: The linter is overzealous about use of setState in componentDidMount. It correctly allows it in callbacks but does not know that any code after an `await` is equivalent to code in a callback. So I move the api call to a separate method. The linter is also wants the new method moved below event handlers.

So this change is much smaller than it appears. It also reads more clearly with whitespace differences hidden. `?w=1`

gulp/src/common/myjobs-api.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  checkStatus(response) {
+    if (response.status === 200) {
+      return response;
+    }
+
+    const error = new Error(response.statusText);
+    error.response = response;
+    throw error;
```
darrint: Note that I'm treating failed requests as exceptions. See below for how the exceptions handlers look.

gulp/src/manageusers/Role.jsx
```diff
@@ -218,27 +216,18 @@ class Role extends React.Component {
...
-          this.setState({
-            apiResponseHelp: 'Role not deleted. Insufficient privileges.',
-          });
-        }
-      }.bind(this));
+    try {
+      await api.delete( '/manage-users/api/roles/delete/' + roleId + '/');
+      await this.props.callRolesAPI();
+      history.pushState(null, '/roles');
+    } catch (e) {
```
darrint: Here's how api error handling looks.

gulp/src/manageusers/Role.jsx
```diff
@@ -1,11 +1,8 @@
-/* global $ */
-
 import React from 'react';
 import _ from 'lodash-compat';
 import {Link} from 'react-router';
 import Button from 'react-bootstrap/lib/Button';
 
-import {getCsrf} from 'common/cookie';
```
darrint: Users of api are no longer concerned with csrf. This is encapsulated in the api class.

gulp/src/reporting/api.js
```diff
@@ -4,153 +4,87 @@
...
-          'Content-Type': 'application/x-www-form-urlencoded',
-        },
-      });
-      return this.parseJSON(this.checkStatus(response));
-    }
+  async postToReportingApi(url, data) {
+    return this.myJobsApi.post(url, data);
+  }
 
-    async listReports() {
```
darrint: The linter is not enforcing indentation level between classes and functions. I went ahead and fixed this file. This difference is best viewed with whitespace differences hidden.

gulp/package.json
```diff
@@ -37,6 +37,7 @@
     "tinycolor2": "~1.3.0",
     "warning": "^2.1.0",
     "webpack": "^1.12.9",
-    "lodash-compat": "~3.10.1"
+    "lodash-compat": "~3.10.1",
+    "jquery-param": "~0.2.0"
```
darrint: jquery-param provides jQuery like form data serialization that we want. It has no dependencies.

aspidites:
I thought there was a big push to get *rid* of jQuery? This seems to be adding jQuery-specific dependencies. Note, I don't mind jQuery, I would just like us to be consistent about our attitude toward one set of librires over another. If we are keeping jQuery, then I would think that fetch is redundant.

darrint:
What the library provides is equivalent to jQuery's param function. It is not jQuery. We're likely indirectly using `$.param` a lot in current ajax calls. Manage users definitely was. Having the myjobs-api.js calls work this way makes for fewer differences to worry about when calling other existing APIs.

https://www.npmjs.com/package/jquery-param

aspidites:
I see. Never mind then.

tdphillips:
What does this give us that we didn't have before?

gulp/src/manageusers/Role.jsx
```diff
@@ -198,19 +198,17 @@ class Role extends React.Component {
...
         });
       }
-    }.bind(this));
+    }
   }
-  handleDeleteRoleClick() {
-    const history = this.props.history;
+  async handleDeleteRoleClick() {
+    const {api, history} = this.props;
     // Temporary until I replace $.ajax jQuery with vanilla JS ES6 arrow function
```
JLMcLaugh: Is this comment obsolete now?

darrint:
@tdphillips 

* Details of making API requests such as csrf token handling are encapsulated in `class MyJobsApi` instead of copy-pasted everywhere we make requests.
* All browsers serialize request bodies the same way.
* The existing jQuery style parameter serialization is preserved.
* Some IE8 ugliness that was in `reporting/api.js` is gone. i.e. `ancientSerialize` is gone.
* No need to pass `/* global $ */` in from the window object.

gulp/src/manageusers/Role.jsx
darrint: Thanks.

gulp/src/manageusers/Role.jsx
aspidites: I would expect to see:
```javascript
await this.props.callDeleteAPI(roleID);
await this.props.callRolesAPI();
```
or this:
```javascript
await api.delete(...);
await api.get(...);
```
But not a mixture of both. Any particular reason for this asymmetry?

darrint: If you trace down this.props.callRolesAPI (and it's sibling, callUsersAPI) it's a fairly involved function that I would like to refactor. It is arguably in the wrong place as well. I started fixing that even but decided not to in order to keep this PR from getting too big. In this PR those functions stop using jQuery and I leave it at that.

I created this ticket to continue to refactor them: [PD-2228](https://directemployersfoundation.atlassian.net/browse/PD-2228)

PR #2160, opened 2016-03-08
[PD-2217] Create Production Implementation for Secure Blocks 

koonsies:
## Overview

This PR covers adding the secure blocks to the default templates. Whether the site uses secure blocks or the default widget rendering depends on the "use_secure_blocks" flag on a site's configuration.

Contains needed migration, doc update.

## Testing

This PR contains updated documentation for testing secure blocks. The only "new" step is making sure the child site's configuration has the "use secure blocks" flag checked.

### DIFF
--- a/doc/source/developers/myblocks/secureblocks.rst
+++ b/doc/source/developers/myblocks/secureblocks.rst
@@ -57,6 +57,12 @@ To test a secure block widget, do the following:
 
  * Save
 
+* Navigate to the site configuration for the child site
+
+ * Check the box labled 'Use secure blocks'
+
+ * Save
+
 * Locate the correct model under MyBlocks (ex. SavedSearchWidgetBlock)
 
 * Add new block
@@ -65,22 +71,26 @@ To test a secure block widget, do the following:
    automatically populated with a default template. This can be edited
    if needed.
 
-* If the element is not currently on any pages, add it with a
-  `<div data-secure-block-id="--YOUR ELEMENT ID--"></div>`
+ * The two widget element IDs included in microsites templates are..
 
- * This will only work on templates loaded from seo_base.html
+  * saved_search for saved search widget
 
- * Currently, all testing is done on templates/CS_TEST_homepage_listing.html
-   If you would like to edit and use this page for testing, ensure the
-   configuration for your child site is set with the above as it's
-   homepage
+  * sb_tools for toolbar/topbar
 
-* Widget will populate on child site if the relationship was set up properly
+  * If an element ID other than these two is used, it must also be included
+    in a template for display using the following format
 
-secure-block-xx-xx.js
----------------------
+   * `<div data-secure_block_id="--YOUR ELEMENT ID--"></div>`
 
-Secure Blocks relies on the file secure-block-xx-xx.js file located in the
+* This will only work on templates loaded from seo_base.html
+
+* Currently, all secure blocks widgets are wrapped in if/else statements
+  to ensure that they are only displayed if they are enabled for that site.
+
+secure-block.js
+---------------
+
+Secure Blocks relies on the file secure-block.js file located in the
 static directory. This file handles creating the ajax call based on the secure
 block ids as well as populating the respective divs with the response.
 
--- a/doc/source/developers/myblocks/secureblocks.rst
+++ b/doc/source/developers/myblocks/secureblocks.rst
@@ -78,7 +78,7 @@ To test a secure block widget, do the following:
 
   * saved_search for saved search widget
 
-  * sb_tools for toolbar/topbar
+  * sb_toolbar for toolbar/topbar
 
   * If an element ID other than these two is used, it must also be included
     in a template for display using the following format
--- a/doc/source/developers/myblocks/secureblocks.rst
+++ b/doc/source/developers/myblocks/secureblocks.rst
@@ -51,13 +51,13 @@ To test a secure block widget, do the following:
 
 * If parent site relationship is not set up, do the following:
 
- * Navigate to Company, edit desired child company (currently my.jobs)
+ * Navigate to SeoSite, edit desired child site (currently my.jobs)
 
  * Under "parent_site" drop down, select desired parent site (secure.my.jobs)
 
  * Save
 
-* Navigate to the site configuration for the child site
+* Navigate to the active site configuration for the child site
 
  * Check the box labled 'Use secure blocks'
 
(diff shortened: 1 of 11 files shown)


seo/tests/test_views.py
```diff
@@ -2592,12 +2596,13 @@ def test_saved_search_render(self):
                                HTTP_HOST='buckconsultants.jobs',
                                follow=True)
         self.assertContains(resp,
-                            '<div id="de-myjobs-widget"',
+                            'direct-saved-search-box',
```
koonsies: The original SS widget and the secure blocks SS widget use this div.. so use another.

doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -65,22 +71,26 @@ To test a secure block widget, do the following:
...
-   configuration for your child site is set with the above as it's
-   homepage
+* This will only work on templates loaded from seo_base.html
 
-* Widget will populate on child site if the relationship was set up properly
+* Currently, all secure blocks widgets are wrapped in if/else statements
+  to ensure that they are only displayed if they are enabled for that site.
 
-secure-block-xx-xx.js
----------------------
```
aspidites: Nice catch.

aspidites:
What's the relationship between secure blocks and whatever we were doing before, again? That is, does secure blocks replace functionality, or introduce it where it didn't exist before? I ask because I see where the templates are saying if, but not else, so I wonder which of the following is true, on a network microsite, for instance:
- if secure blocks are enabled, they will see both top bars
- if secure blocks are disabled, they will see just the old top bar
- if secure blocks are enabled, they will see the blocks variant and not the old one
- if secure blocks are disabled, they will see no top bar at all

I expect at least two of the above to be true, but don't know which two those would be.

koonsies:
Secure block will eventually create new functionality. Thus far, though, we have only replaced old functionality. In terms of widgets that replace functionality (like topbar, saved search), when secure blocks are turned on, you should ONLY see the secure block widget. When secure blocks is turned off, you should see the original widget. So..

if secure blocks are enabled, they will see the blocks variant and not the old one
if secure blocks are disabled, they will see just the old top bar

doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -57,6 +57,12 @@ To test a secure block widget, do the following:
 
  * Save
 
+* Navigate to the site configuration for the child site
```
aspidites: I'm actually concerned with line 54 (which you didn't change in this PR): 

I was messing with indiana.jobs when I realized I didn't know what company they were attached to, which further reminded me that I shouldn't be worried about companies here, but seo sites. That is, an `SeoSite`  has a `parent_site`, rather than a `Comapny` having a `parent_company`. Is this correct? If so, I think this bit of documentation should be updated to reflect that, otherwise new developers might go looking in the Company admin for something that isn't there.

doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -57,6 +57,12 @@ To test a secure block widget, do the following:
...
+
+ * Check the box labled 'Use secure blocks'
+
+ * Save
+
 * Locate the correct model under MyBlocks (ex. SavedSearchWidgetBlock)
```
aspidites: I didn't see a TopbarWidgetBlock, and clicking on  "ToolsWidgetBlock" throws an exception:
> ProgrammingError at /admin/myblocks/toolswidgetblock/

NVM, looks like you have to run migrations for both seo AND myblocks. 

In any case, it's not obvious to me that I should be using ToolsWidgetBlock for topbar.

doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -65,22 +71,26 @@ To test a secure block widget, do the following:
    automatically populated with a default template. This can be edited
```
aspidites: Offset and Span are also required.

doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -65,22 +71,26 @@ To test a secure block widget, do the following:
...
 
-* If the element is not currently on any pages, add it with a
-  `<div data-secure-block-id="--YOUR ELEMENT ID--"></div>`
+ * The two widget element IDs included in microsites templates are..
+
+  * saved_search for saved search widget
+
+  * sb_tools for toolbar/topbar
+
+  * If an element ID other than these two is used, it must also be included
```
aspidites: Does this mean that if I name a block "sb_tools" that it will automatically be included on any SeoSite which has secure blocks enabled? 

I guess what I'm struggling with is how the tools block I just made is being associated with my seo site.

aspidites:
@koonsies , I wasn't able to get this working locally, so I suspect I've glanced on some important detail in the documentation, or am misinterpreting it entirely. When you get a moment this afternoon, I can walk you through what I have, if you wouldn't mind, please.

doc/source/developers/myblocks/secureblocks.rst
koonsies: I will have to make that a bit more clear. The relationship resides on SeoSite as ibm-engineering.jobs is a child of ibm.jobs, but they are the same company.

doc/source/developers/myblocks/secureblocks.rst
koonsies: Oops! Missed the "migrations" tag. Thanks for adding that. 

Sorry for the rename confusion. I didn't call it "topbar widget" as it was stated early on that versions of the tools widget may be sidebar navigation, etc.

I'm not sure of the best way to note this change..

doc/source/developers/myblocks/secureblocks.rst
koonsies: Currently, yes. Adding a secure widget to a blocks page would be much easier and less.. hard-codie.. but as it stands, core templates have to have the element id anchors added to them. These anchors are along the lines of 

`<div data-secure_block_id='sb_tools'></div> `

Because that is hardcoded into job_listing.html, you are pretty much stuck using that ID for your widget or else it won't appear on the templates.

koonsies:
Sounds good!

I will walk through the set up with you whenever is convenient for you. That should help me refine my documentation.

Thanks for reviewing!

doc/source/developers/myblocks/secureblocks.rst
aspidites: Ah. Sounds like another instance of "naming things is hard". First toolbar, then "topbar", now "tools". Newcomers will probably not have legacy naming to worry about, so maybe it's fine? Only other thing I could suggest is having a section that runs through the available block somewhere, but that's more of user documentation which @jason' has said at least on one occasion isn't a priority.

PR #2162, opened 2016-03-08
Docs about gulp build

lemniscate613:


doc/source/developers/docker/use.rst
```diff
@@ -210,6 +210,18 @@ Add other microsites as needed.
 Go to http://secure.my.jobs. You should have to click through a security
 warning.
 
+Run Gulp
+=====================
```
aspidites: Too many "="'s. Should line up with the "p" in "Gulp" above.

doc/source/developers/docker/use.rst
```diff
@@ -210,6 +210,18 @@ Add other microsites as needed.
...
+
+Gulp processes our JavaScript applications.
+
+From the root:
+
+``dkgg build``
```
aspidites: I don't care if this gets changed or not, but double-colon signifies a code block. So instead of:

```
From the root:

``dkgg build``
```

You could do:
```
From the root::

dkgg build
```

doc/source/developers/docker/use.rst
lemniscate613: Updated.

PR #2163, opened 2016-03-09
[PD-2201, PD-2226, PD-2227] Implement Activities for use cases 2, 3, 7, and 8

aspidites:
# Description
This ports postajob to roles/activities for the mentioned use cases (which have descriptions in documentation as well). For the remaining use cases, postajob should behave as it did before. Errors encountered during those use cases, unless they somehow reference roles or activities (I don't know how that would happen) are considered out of scope. 

Once all eight use cases are ported, I plan to give the docs another once over, and perhaps retire/overhaul the quick start guide. I'm also contemplating making the "overview" documentation a glossary for postajob, since there is a lot of good information in there. In particular, it describes the relationships between the various models decently.

# Testing
Run migrations, then follow the instructions [here](https://github.com/DirectEmployers/MyJobs/pull/2163/files?short_path=a8009f3#diff-a8009f3689045301500b6b0a72d3c2e0) for each of the use cases. The reason I'm "passing the buck" as it were, instead of reiterating the requirements here, is that I'd like to see if what I've written as documentation is sufficient.


aspidites:
I also didn't document use cases 7 and 8, as those seem to be strictly related to user management, which is documented elsewhere.

tdphillips:
I worked on this. Someone else should review and potentially push the button.

koonsies:
Have no fear, I am reviewing this. I'll toss my thoughts down as I go.

First one, I haven't hit a place where logging in to the microsite by going to directemployers.jobs/admin wasn't sufficient. Is it actually necessary to set up a login page?

koonsies:
Second thought.. Is it me, or does the documentation start at Use Case 2? I don't see Use Case 1

tdphillips:
Logging in via /admin/ only works for staff, which most people won't be.

koonsies:
Oh, I gotcha. The documentation is under "developers", so I had assumed staff enabled.

aspidites:
@koonsies use case 1 is dependent on use case 3. As we were working in parallel, only use cases 2, 3, 7, and 8 are done. The others will follow in a separate PR. In other words, the original use cases aren't in a logical order, but most probably in the order @fezick first thought of them.

koonsies:
That's fine! Just wanted to make sure it wasn't an error. Now that you mention it, I remember that from the meeting.

Another thing.. I am having issues making the permissions have effect. I stripped myself of all posting permissions, but I am able to post/edit jobs as normal. Perhaps I have misconfigured something.

aspidites:
@koonsies as for the staff-enabled, I deliberately didn't mention that, since it encourages conflating user roles when testing. For instance, the bug @misterjonscott ran into with not being able to select a site to post on was caused specifically because he was a super user. Similar examples can be made for just staff user. I may later on add a footnote to the bottom, though, as that's less "in your face" than a dedicated paragraph or a note.

doc/source/developers/postajob/setup.rst
```diff
@@ -0,0 +1,302 @@
...
+>>> user = User.objects.get(email="dev@apps.directemployers.org")
+>>> user.roles.add(role)
+
+From the Django Admin and User Management Tool
+''''''''''''''''''''''''''''''''''''''''''''''
+
+#. select the appropriate company as the ``canonical_company`` for 
+   `your seo site`_.
+#. `enable posting access` for the company chosen above by adding "Posting" to
+   the "Chosen App-Level Access" box in teh "App-Level Access" section.
```
koonsies: teh

doc/source/developers/postajob/setup.rst
```diff
@@ -0,0 +1,302 @@
...
+
+  - :ref:`enable-marketplace`
+
+Views
+'''''
+
+=================================== ======================= ===================
+URL Path                            View Name               Required Activities
+=================================== ======================= ===================
+/posting/admin/products             admin_products          read product
```
koonsies: url reads /products, it's actually /product

doc/source/developers/postajob/setup.rst
```diff
@@ -0,0 +1,302 @@
...
+/posting/admin/product/add          ProductFormView         create product
```
koonsies: This may be unavoidable, but you can access the add product form without the "create product" permission. Once you try to submit the form, it errors.. But if possible, it's likely better to block access to the form at all (like with U.C. 2)

aspidites: Looking at this now.

aspidites: Thanks a lot for this, @koonsies . I corrected the mentioned and fixed a bug that's been sitting idle for a moment.

fezick:
What's holding this up from merger?

PR #2167, opened 2016-03-09
Secure Blocks Misc Cleanup

koonsies:
## Overview

This PR represents a bit of cleanup for the secure blocks implementation. A few of these changes came to my attention while reviewing previous PRs, but I did not want to delay the PRs by including these changes. 

The reasoning for each is annotated in the code.


doc/source/developers/myblocks/secureblocks.rst
```diff
@@ -166,4 +166,4 @@ block.
 
 The current anchor for the tools widget is as such:
 
-`<div data-secure_block_id="sb_toolbar" data-widget_type="toolswidget"></div>`
+`<div data-secure_block_id="sb_toolbar" data-widget_type="tools"></div>`
```
koonsies: Changing the widget type. Tools seemed easier to remember and less redundant.

myblocks/models.py
```diff
@@ -443,7 +443,7 @@ class ToolsWidgetBlock(SecureBlock):
 
     """
     # temporarily use the current topbar template
-    base_template = 'includes/topbar.html'
+    base_template = 'myblocks/blocks/toolswidget.html'
```
koonsies: Created new template specifically for secure blocks to prevent inclusion of JS in the database object.

myblocks/models.py
```diff
@@ -492,6 +492,13 @@ def get_cookies(self, request, **kwargs):
 
         return cookies
 
+    def required_js(self):
+        """
+        Return a list of all required javascript in URL format
+
+        """
+        return ['%ssecure-blocks/sb-tools.js' % settings.STATIC_URL]
```
koonsies: Tools widget JS is now external

myblocks/models.py
```diff
@@ -530,7 +537,7 @@ def required_js(self):
         Return a list of all required javascript in URL format
 
         """
-        return ['%ssaved-search.js' % settings.STATIC_URL]
+        return ['%ssecure-blocks/sb-saved-search.js' % settings.STATIC_URL]
```
koonsies: Moved JS to secure blocks folder for better organization

mysearches/views.py
```diff
@@ -509,6 +509,14 @@ def get_value_from_request(request, key):
...
     return value
 
+def remove_anchor_from_url(url):
+    """
+    Remove anchor tags from the provided URL. %23 is '#' url encoded.
+
+    """
+    if url.find('%23') != -1:
+        return url[:url.index('%23')]
+    return url
```
koonsies: Because I use HTTP_REFERER, the anchor tags are often included in the URL. This prevents a URL such as https://www.my.jobs/jobs/#1 by truncating anything after the anchor tag.

static/secure-blocks/sb-saved-search.js
```diff
@@ -48,7 +48,7 @@ function save_search() {
   // provided, attempts to get a user from the input and create a new user.
   // Otherwise, uses the currently provided user to create a saved search.
   email_input = $('#saved-search-email').val() || existing_user_email;
-  $(blocks_widget_div).data("current-input", email_input);
+  $(blocks_widget_div).data("current_input", email_input);
```
koonsies: I have swapped all data elements to using underscores. This prevents camelCase variables being sent to the API.

static/secure-blocks/sb-tools.js
```diff
@@ -0,0 +1,272 @@
+var readCookie = utils.readCookie;
```
koonsies: JS pulled from the topbar template, lightly modified to remove template tags.

templates/home_page/CS_TEST_home_page_listing.html
```diff
@@ -1,87 +0,0 @@
-{% extends "seo_base.html" %}
```
koonsies: This template existed to test the widgets. Now that we have included the blocks on the main templates, this is no longer needed.

templates/myblocks/blocks/savedsearchwidget.html
```diff
@@ -49,12 +49,12 @@
 
     {% elif search.is_active %}
     {# The block below is for the widget given that the saved search exists and is active #}
-        <p><b>This search is being sent to {{ search.email }}.</b></p>
+        <p class="confirmation-text"><b>This search is being sent to {{ search.email }}.</b></p>
```
koonsies: Adding classes to various places in the template to aid with future CSS manipulation.

templates/myblocks/blocks/toolswidget.html
```diff
@@ -0,0 +1,196 @@
+{% load i18n %}
```
koonsies: HTML portion of the topbar template

myblocks/models.py
aspidites: Should this be `tools.html` instead of `toolswidget.html` given you removed `widget` from the data attribute?

koonsies: Good catch. That was something that I was tossing around as well.. In terms of the data attribute, I didn't see ambiguity being a potential issue (using data-widget_type narrows the scope a good amount) but with a template, I was a bit worried tools.html was too generic. 

I welcome your thoughts on what I should do here.

static/secure-blocks/sb-saved-search.js
aspidites: Reference? I'm curious as to what part of our stack would convert "foo-bar" to "fooBar".

templates/myblocks/blocks/savedsearchwidget.html
aspidites: Does the rule about hyphens being replaced by underscores apply here? (confirmation-text vs confirmation_text)

static/secure-blocks/sb-saved-search.js
koonsies: Certainly! 

https://api.jquery.com/data/ <https://api.jquery.com/data/>

If you search for “camel” it’ll show a bit of the logic behind the data attributes. jQuery does this conversion to data attributes.

aspidites:
Does this toolswidget template only get used when secure widgets are enabled, or is it used for the top bar in the general case?

templates/myblocks/blocks/savedsearchwidget.html
koonsies: I more-or-less answered this above, but the hyphens aren't a problem for classes/IDs as they are all handled by javascript within the browser. Only the data attributes are an issue here as they are converted by jQuery to camelcase.

koonsies:
It is only used by the secure blocks template, though it is almost identical to the toolbar.html (not topbar-new.html).

myblocks/models.py
aspidites: I don't have a strong opinion one way or another, other than to say than having to remember when to use hyphens vs underscores. 

I'd probably have gone for a subdirectory named secure_widgets or something and tossed tools.html in that.

static/secure-blocks/sb-saved-search.js
aspidites: Ah. That's...pretty nasty. Mixed dashes makes me sad, but given that quirk, I can't blame you.

myblocks/models.py
koonsies: Ah, good call.

I pulled the static files into a folder for just secure_blocks, but I haven’t done that with templates. I can do that.

templates/myblocks/blocks/savedsearchwidget.html
aspidites: Yep, you did. Hyphens are preferred but underscores are used to get around jquery's name mangling.

static/secure-blocks/sb-saved-search.js
koonsies: I feel the same way. I had considered "converting" the camel case on the back end, but I wasn't sure it was any less dirty.

aspidites:
:+1: @koonsies , I think you need to pull qc again.

myblocks/models.py
aspidites: :+1:

PR #2168, opened 2016-03-10
Pd 2231

misterjonscott:
# Changes
I changed the reporting base template from base to base-react, for future component styling adherence.  This change required an update, in the style and layout, to Bootstrap 3, and some React changes.

# Testing

It should look like the screenshot attached
screenshot 2016-03-10 09 32 38
:


lemniscate613:
Looks good. :+1:  I'll play with it in a bit...

Seems like the begin/end date fields should both have labels.

misterjonscott:
@lemniscate613 This isn't a ticket to change the form, itself, but just to correct the formatting and bring it into BS3 for future updates.

templates/myreports/dynamicreports.html
```diff
@@ -9,19 +9,19 @@
 
 {% block meta-extra %}
 <link rel="stylesheet" href="{{STATIC_URL}}font-awesome.css">
-<link rel="stylesheet" href="{{STATIC_URL}}andor.171-14.css">
+<link rel="stylesheet" href="{{STATIC_URL}}andor.174-10.css">
```
aspidites: Since you are changing that file anyway, we should probably use compress instead of perpetually version bumping the css.

misterjonscott:
Thanks, @aspidites !

PR #2171, opened 2016-03-10
[PD-2229]  move js configs

darrint:
There are now several new config files in `gulp/` including the config for eslint. The gulpfile now only runs a production build and watches unit tests in development.

The technique I used for implementing `webpack-dev-server` is based on: http://webpack.github.io/docs/webpack-dev-server.html#combining-with-an-existing-server


gulp/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
```
aspidites: Appreciate this. Means my local editor can pick up on linting now. Also means I can manually run eslint from the command line on individual files.

gulp/src/reporting/spec/reportEngineSpec.js
```diff
@@ -44,6 +44,7 @@ describe('ReportFinder', () => {
     });
 
     it('can unsubscribe', () => {
+      finder.noteNewReport(22);
```
aspidites: Is this supposed to be part of this PR? Or is this part of the branch you said this was based on?

gulp/webpack.config.js
```diff
@@ -0,0 +1,58 @@
...
+    }),
+    // Factor common code in to vendor.js.
+    // This also establishes the parent relationship between the vendor
+    // and app chunks.
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'vendor',
+      filename: 'vendor.js',
+      minChunks: 2,
+    }),
+    // No idea if Dedupe and OccurenceOrder are actually doing anything.
```
aspidites: Then why not remove those lines and compare builds?

gulp/webpack.dev.config.js
```diff
@@ -0,0 +1,8 @@
+var prodConfig = require('./webpack.config.js');
+
+prodConfig.plugins = [];
+
+// Linux users not using docker can do without this.
```
aspidites: Reference as to why this is needed for docker users?

myjobs/templatetags/common_tags.py
```diff
@@ -15,6 +15,38 @@
 register = template.Library()
 
 
+@register.inclusion_tag("includes/js_bundle.html")
```
aspidites: Motivation for why we need this? What problem does this actually solve? I've read the docstring, but don't understand the motivating use case. Seems like a red flag to have something that is in a template that only works during development.

gulp/src/reporting/spec/reportEngineSpec.js
darrint: I added the line `random: true` to jasmon.json I found this was needed. If you feel strongly I can move this or not introduce the addition.

gulp/webpack.dev.config.js
darrint: https://github.com/webpack/webpack-dev-server/issues/143

inotify events which occur outside the container don't propagate inside the container.

At our project size on my machine I get incremental builds in 250 - 900ms. In the future that could get longer.

myjobs/templatetags/common_tags.py
darrint: This is the regular include. The js_bundle template tag _replaces_ the pair of script tags (vendor + app) that we put in our templates. See manageusers.html and `includes/js_bundle.html` below.

As a developer, when I want to include a webpack bundle, I always use js_bundle.

Here's a brain dump on my motivation and thought process behind this:

Motivation:

* A fast development cycle, edit-compile-test-debug, for both Linux and OSX users. That cycle is 5 secs for me and 25secs (aak!) for OSX before this PR. After this PR I see <1sec on Linux and I will test on OSX before I submit for merge.
* Easy to correctly set up Django templates for both production and development.

I can understand criticism that this is high impact to templates. Since I'm seeing <1sec compile time, there is some room to compromise even if it's slightly slower.

Pros of this solution:
* Webpack config is simple. "Mostly same as production, except no plugins."
* Details of how webpack is configured are encapsulated by the template tag.
* Easy to set the template up. There's only one way: `{% js_bundle "myapp" %}`
* Easy to eliminate compress in development.
* Easy to test the production configuration. (Comment out WEBPACK_... config var in dev_settings) and run gulp.
* If future use of hot module replacement requires even more impact to templates, I can encapsulate that in js_bundle.

Cons:
* More difficult to reason about what goes into the template with js_bundle.
* If using the reverse proxy, more https to fight with. (At least the app fails completely.)

The obvious compromise would be to make a template tag for just changing a url from static file to dev server. Unfortunately compress won't tolerate absolute urls pointing to the dev server. IIRC that was what pushed me to write js_bundle and `includes/js_bundle.html`.

I'm open to other ideas.

gulp/webpack.dev.config.js
aspidites: Ah.

myjobs/templatetags/common_tags.py
aspidites: Another con. In addition to `{% compress %}` and `{% static %}`, I now need to remember {% js_bundle %}, which as far as I can tell is only saving me remembering to say 'npm run dev' vs 'npm run prod'.

aspidites: Given that you have to edit the config file to get results anyway, I'm not convinced much is gained by introducing yet another special-purpose template tag. I even spent the last hour trying to figure out how to get rid of the {% can %} template tag I introduced a while back.

gulp/src/reporting/spec/reportEngineSpec.js
aspidites: Nah, it's fine.

aspidites:
I think I have an alternative which won't require a new template tag

aspidites:
Was in a hurry when I left that cryptic message yesterday. I'll be pulling this locally and trying out what I think should give you the flexibility you want without requiring a special template tag. I'll report back when I'm done.

misterjonscott:
Tested this branch and it's fantastic.  So far, everything's working significantly faster, which really boosts productivity over the ~30s wait times.

lemniscate613:
It worked for me last week, and it still works for me today. Very big improvement. Thanks @darrint  :+1:

aspidites:
Still don't like the tempalte tag, but the gulp directory is dense enough without a dedicated bootstrap template for testing purposes.

PR #2178, opened 2016-03-14
Make requires and User.can higher-order

aspidites:
I'm doing this as an untagged, small PR because I'd like feedback. I started working on Use Case 1 for postajob (hence the unfinished documentation) when I ran into this. Basically, for admin views, we'd like to say that any number of activities is required to see a view since from that view, any number of pages may be navigated to. Whereas the original requires decorator enforced that all specified activities were required to see the decorated view, this version allows that comparison to be changed. Here, we specify that at least one of the specified activities should be available before the  view is accessible.

### DIFF
--- a/myjobs/decorators.py
+++ b/myjobs/decorators.py
@@ -149,13 +149,21 @@ def requires(*activities, **callbacks):
                    check against.
     :callbacks: callbacks['activity_callback'] is a callable to be used as a
                 view response when the use isn't associated with the correct
-                subset of activities. callbacks['access_callback'] is a
-                callable to be used as a view response when the users's company
-                doesn't have the appropriate app access (as determined by the
-                passed in activities).
+                subset of activities.
+
+                callbacks['access_callback'] is a callable to be used as a view
+                response when the users's company doesn't have the appropriate
+                app access (as determined by the passed in activities).
 
                 Both callbacks take a `request` parameter which can be used to
                 do further processing before returning an alternate result.
+
+                callbacks['compare'] is the callable used to compare the
+                requested activities with those a user has access to. By
+                default, we check if a user can perform all activities.
+
+                This callback takes two iterables and returns a boolean
+                signifying whether or not the comparison was successful.
 
     Examples:
     Let's assume that the activities "create user", "read user", "update user",
@@ -163,7 +171,7 @@ def requires(*activities, **callbacks):
     further assume that a `modify_user` view exists. Finally, lets assume that
     the current user belongs to a company, `TestCompany`.
 
-    We might want to decorate that view as follows:
+    We might want to decorate that view as follows::
 
         @requires("read user", "update user")
         def modify_user(request):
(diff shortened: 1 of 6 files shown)


postajob/views.py
```diff
@@ -160,6 +160,9 @@ def purchasedjobs_overview(request, purchased_product, admin):
 @user_is_allowed()
 @error_when_company_missing_from_sitepackages(feature='Microsite Admin is')
 @error_when_site_misconfigured(feature='Microsite Admin is')
+@requires('read product', 'read request', 'read offline purchase',
```
aspidites: This is the use case which drove me adding this functionality.

myjobs/decorators.py
```diff
@@ -204,7 +222,9 @@ def modify_user(request):
     activity_callback = callbacks.get(
         'activity_callback', lambda request: MissingActivity())
     access_callback = callbacks.get(
-            'access_callback', lambda request: MissingAppAccess())
+        'access_callback', lambda request: MissingAppAccess())
+
+    compare = callbacks.get('compare', lambda x, y: set(x).issubset(y))
```
aspidites: This is the actual implementation of the new functionality.

koonsies:
This is a great solution to the issue at hand! I always love a good lambda.. 

That being said, I'm a bit torn on the API. As the `@requires` decorator is now, there is a great deal of flexibility in how a user of the decorator wants to handle the comparisons. That is a good thing. But we lose a bit of readability and introduce a situation where we'd see that lambda function copy-pasted in a number of places in the code. Could it, perhaps, be more elegant to introduce a "has_all" (working title) flag? Rather than providing the function in the decorator, you simply write..

`@requires("read user", "write user", has_all=True)`

Obviously, the name of the flag is up in the air, but I want to start the discussion on that approach. I see this as something that will be used fairly often, so I think there's something to be said about simplicity.

aspidites:
@koonsies  until the next time I need a different comparison. I also don't think it's any more cognitive load than the case than the `sort` function, especially since it's only the rare case that you would need to use a different compare function. 

The only places where we want to use the "one of" instead of "all of" approach is when we're on some landing page with multiple outlets with links we want to conditionally include. At the moment, that includes
- posting admin

If lambdas are that hard to read, we could just as easily write it as a function and reuse that (further eliminating the copy-pasting you mention).

```python
def at_least_one(left, right):
    """"Returns True if any of the items in the left is also in the right."""
    return bool(set(left).difference(right)\

@requires("read product", "read grouping", "read job", compare=one_of)
def some_view(request):
    ...
```

koonsies:
I kind of like where you're going with that. If we go that route, perhaps the compare methods should live near the requires declaration. Thereby making it easier for a developer to see what different compare methods are available when using the required decorator. Otherwise, I could see people reinventing the wheel later on down the road, not realizing that another compare method had already been written.

I'm definitely not married to this idea, spitballing ideas for you to think about!

aspidites:
Made it a separate function just above requires with it's own fancy docstring.

koonsies:
:+1: I'll delay merge in case anyone else would like to review, but I like it.

aspidites:
@koonsies  couldn't bundle compare functions with decorator because of circular dependency:

myjobs.decorators -> myjobs.models -> myjobs.decorators

Where arrows signify imports. myjobs.models needs decorators for the compare functions. As such, I just relocated them to universal decorators, which was being imported by models already, and probably less of a misnomer. I noted where the compare functions reside in the documentation for requires and User.can.

@tdphillips I fixed the broken tests by doing the above.

aspidites:
Wasn't being merged any time soon and my work has diverged enough that maintaining multiple branches is annoying.

PR #2180, opened 2016-03-15
Make auth code stand out on company access request form [PD-2223]

tdphillips:
thing

All tests pass.

### DIFF
--- a/static/custom.css
+++ b/static/custom.css
@@ -4502,3 +4502,9 @@ a.btn.pull-right {
 @media(min-width:1200px){
 
 }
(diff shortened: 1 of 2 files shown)


static/custom.css
```diff
@@ -4502,3 +4502,9 @@ a.btn.pull-right {
 @media(min-width:1200px){
 
 }
+#super-important-code-dont-lose-me {
```
tdphillips: Too much?

aspidites: probably.

PR #2182, opened 2016-03-16
Fix compress in deployment.

darrint:
Some cut and paste is still needed. This works with compress for me locally.


default_settings.py
```diff
@@ -271,6 +271,7 @@
     'django.core.context_processors.request',
     'myjobs.context_processors.current_site_info',
     'myjobs.context_processors.absolute_url',
+    'myjobs.context_processors.webpack_dev_setting',
```
tdphillips: I know it's a small function but are you okay with it being run on every request on a microsite and every secure request that doesn't end up including vendor.js?

aspidites: Until we've decided that react apps will live on microsites, this probably makes more sense in the myjobs config.

PR #2183, opened 2016-03-17
[PD-2196] Wrap Title Searches in Quotes

koonsies:
Big PR Here!

Add logic to wrap quotes around title search terms selected from the autocomplete drop down. So that titles with special characters like '/' and '&' work properly.

There's a small issue with the autocomplete not registering the characters immediately following an escaped character, but after discussion this was marked as "will not fix". If someone wants further information, I'd be glad to give it.

:hamburger:


tdphillips:
The file version wasn't updated.

koonsies:
Any idea why it is showing the careers.html file as being completely changed? I modified one character (changed the 4 to 5). I even checked out the file from quality control and made the change again. Same result.

tdphillips:
I'm not sure why most of the file is shown as having been modified but it's a local copy of a job search that's only used in tests. I haven't been modifying it when changing file versions

koonsies:
Alright. I removed the changes to the test templates.

aspidites:
>  If someone wants further information, I'd be glad to give it.
I'll bite.

templates/seo_base.html
```diff
@@ -55,7 +55,7 @@
 
     {% block ga_js %}{% include "ga.html"%}{% endblock %}
 
-    <script src="{{ STATIC_URL }}search-ac.161-4.js" type="text/javascript"></script>
+    <script src="{{ STATIC_URL }}search-ac.161-5.js" type="text/javascript"></script>
```
aspidites: Probably need to just use compress. Just confirmed that seo runs the compress command, though no templates are using them yet. Better late than never.

For the uninitiated:
```
{% load compress %}
{% load staticfiles %}

{% compress js %}
<script src="{% static "search.js" %}" type="text/javascript"></script>
{% endcompress %}

koonsies:
Good call.

I was considering using compress.. but then I didn't. It is now properly compressed!

As far as the additional info on the bug.. 

First, the bug.. I described the bug in the ticket [here] (https://directemployersfoundation.atlassian.net/browse/PD-2196?focusedCommentId=203910&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-203910)

I had done some early research into what was happening, but it appears that there is some overzealous escaping on special characters in the autocomplete function of solr. A title like this.. CORPORATE DIRECTOR/LEADER & DUDE would be parsed into solr query as such "title(CORPORATE) and title(DIRECTOR\\\\\\/) and title(LEADER) and title(\\\\\\&) and title(DUDE)".. 

That is not an exact output, of course, but it represents what I noticed as potentially being the problem. For some reason special characters, that should be escaped with one set of backslashes ( \\ ) are being escaped three times.

aspidites:
Ah. Didn't @darrint work on a ticket when he first started working here that converted search terms to solr queries? Mention that to say maybe that would offer  better hints as to why special characters are being escaped as such, rather than it residing in solr itself. Please correct me if I've misunderstood when that feature is used.

Thanks for the context and history lesson.

aspidites:
To be clear, I see where @haliscooper  suggested not to worry about tackling the root of the problem. I just wanted to narrow down the search in case it was tackled in the future.

aspidites:
I'm ok with this. Just had to keep updating the branch before I can merge (or anyone else).

PR #2187, opened 2016-03-18
Fix utils

aspidites:
Some linting and declaring functions outside of the original object creation. Only errors  are because i'm using var instead of let or const, which I am doing because this file isn't being handled by gulp.


koonsies:
:+1: Tested and still works as expected

aspidites:
Bleh. Need to do more thorough testing. I'm sure this works in the perfect use case, but something about what I'm doing is causing me problems locally.

aspidites:
Nvm. It's just on one of my branches, which means I need to fix the branch.

aspidites:
OK. located the issues I was having locally. Definitely not due to this pr.

PR #2188, opened 2016-03-18
Highlander - There can be only one

aspidites:
This PR deprecates both the secure wigdets topbar and the react topbar. Had some help from @koonsies , so someone else will have to push the button after it's reviewed. 

# Testing
- mess around with the topbar in PRM
- mess around with the topbar in User Management
- mess around with the topbar on a microsite with secure blocks enabled

Bonus points, as the js executed with topbar is now external to topbar itself, which means that page loads should theoretically be faster.

I annotated a few lines to give insight into the solution.


templates/base-react.html
```diff
@@ -82,7 +82,7 @@
 
 
       {% block topbar %}
-      {% include "includes/topbar-new.html" %}
+      {% include "includes/topbar.html" with topbar_class=" topbar-new" pop_menu_class="col-md-4" %}
```
aspidites: For react apps, the structure is identical, but we are using different class names.

templates/base.html
```diff
@@ -30,6 +30,7 @@
 
     {% compress js %}
       <script type="text/javascript" src="{% static "utils.js" %}"></script>
+      <script type="text/javascript" src="{{ STATIC_URL }}custom.166-06.js{{ gz }}"></script>
```
aspidites: Topbar needs jquery, which is in custom.js.

templates/base.html
```diff
@@ -3,7 +3,6 @@
 {% load common_tags %}
 {% load compress %}
 {% load staticfiles %}
-{% gz as gz %}
```
aspidites: If you look at gz (in common tags), you'll see that it always returns an empty string. Just getting rid of this piece by piece. Thank you, IE11.

templates/includes/topbar.html
```diff
@@ -1,13 +1,14 @@
...
+{% load staticfiles %}
+{% load compress %}
 {% if request.GET.callback %}
 <link href="{{ STATIC_URL }}topbar.directseo.165-10.css" rel="stylesheet" type="text/css">
 {% else %}
 <link href="{{ STATIC_URL }}topbar.myjobs.165-10.css" rel="stylesheet" type="text/css">
 {% endif %}
 
-<div class="topbar" data-dropdown="dropdown" role="menubar">
+<div class="topbar{{ topbar_class }}" data-dropdown="dropdown" role="menubar">
```
aspidites: For react apps, we pass "topbar-new" here, for everything else, it's an empty string.

templates/includes/topbar.html
```diff
@@ -92,7 +93,7 @@ <h3 id="topbar-logo">My.jobs</h3>
                     <ul id="nav">
                         <li>
                             <a class="main-nav"></a>
-                            <ul class="span4" id="pop-menu">
+                            <ul class="{% firstof pop_menu_class "span4" %}" id="pop-menu">
```
aspidites: If pop_menu_class isn't specified, we return span4. Eg, try to grab a Bootstrap 3 class, return a Bootstrap 2 class by default.

templates/includes/topbar.html
```diff
@@ -188,311 +189,12 @@ <h3 id="topbar-logo">My.jobs</h3>
         </div>
     </div>
 </div>
-
```
aspidites: This was all moved to an external javascript file.

templates/base-react.html
koonsies: I've never included a template with injected context variables. This is a pretty cool trick.

PR #2190, opened 2016-03-21
[PD-2230] Convert Myprofile Unit Pages to React

lemniscate613:
Summary
=======

This PR leaves https://secure.my.jobs/profile/view/ (mostly) as is, but it moves the Add / Edit profile unit pages to React. [2174](https://directemployersfoundation.atlassian.net/browse/PD-2174) is a related ticket and addresses the API.

Bonus
=====

Along the way we:

* prove Django forms can be replaced with a combination of JSON (see @JLMcLaugh's [fork](https://github.com/DirectEmployers/django-remote-forms) of django-remote-forms) and a more configurable, dynamic list of React components.
* implement five React Components (reusable in other apps) for the most common form fields (TextField, DateTime, Select, CheckBox, and TextArea -- located in `/gulp/src/common/ui`) which follow @misterjonscott's new form design standard. These are the result of conversations* between me, @darrint, @aspidites, and @misterjonscott. All Components use the HTML elements you'd expect, except Select, which is built from divs/lists and allows for styling:

screen shot 2016-03-21 at 3 49 26 pm
* implement FieldWrapper component to enforce specific design and layout of labels, help text, and error text.

Diagram:

screen shot 2016-03-21 at 4 28 53 pm

Example:

screen shot 2016-03-21 at 3 59 16 pm

Review
======

Build with your favorite method, and visit: https://secure.my.jobs/profile/view/

Footnotes
========

* Notes from that conversation:

0041907d-d266-40df-824e-3dc0b643cd7f


myprofile/tests/test_views.py
```diff
@@ -124,6 +124,17 @@ def test_handle_form_post_existing_valid(self):
...
 
+    def test_handle_form_json_serialize_get(self):
+        """When an ajax requests wants to GET json, serialize the form."""
+        resp = self.client.get(reverse('handle_form'),
+                               HTTP_ACCEPT='application/json',
+                               data={'module': 'Name'})
+        self.assertEquals(200, resp.status_code)
+        self.assertIn('application/json', resp['content-type'])
+        data = json.loads(resp.content)
+        self.assertEquals(3, len(data['fields']))
```
tdphillips: Should we also be testing that these fields are what we think they are?

lemniscate613: I can certainly do that if you think it's a good idea.

gulp/src/common/ui/Select.jsx
```diff
@@ -0,0 +1,151 @@
+import React from 'react';
```
tdphillips: I don't know how closely this is intended to replicate HTML select inputs. I'm unable to dismiss the dropdown by clicking again on the arrow.

There's also a (potentially unintended?) side effect here. Country selects make the page very long and potentially introduces scroll bars that weren't in place before, leading to distracting left/right movements of page contents when it is clicked or dismissed.

gulp/src/common/ui/Datetime.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
tdphillips: The example date we provide as placeholder text doesn't match what we're saving. Following the placeholder, I entered "01/01/2001". When editing that unit, I saw "2001-01-01".

tdphillips:
We appear to have lost delete confirmations.

lemniscate613:
@tdphillips very good catch. Will add.

gulp/src/common/ui/Select.jsx
lemniscate613: Very good points again. Let me look into them. TBC.

gulp/src/myprofile/main.jsx
```diff
@@ -0,0 +1,290 @@
...
+    });
+  }
+  handleDelete() {
+    if (confirm('Are you sure you want to delete this item?') === false) {
+      return;
+    }
+
+    const {formContents} = this.state;
+
+    $.ajax({
```
darrint: Need to import and use the api class instead of these now.

gulp/src/common/ui/CheckBox.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
```
darrint: Need comments here to describe what this component does.

gulp/src/common/ui/CheckBox.jsx
```diff
@@ -0,0 +1,33 @@
...
+      required={required}
+      hidden={isHidden}
+      defaultChecked={initial}
+      onChange={onChange}
+      />
+  );
+}
+
+CheckBox.propTypes = {
+  onChange: React.PropTypes.func.isRequired,
```
darrint: Need comments on these describing what each prop means.

myprofile/tests/test_views.py
darrint: It is a good idea.

gulp/src/myprofile/main.jsx
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    formContents[fieldID] = value;
+
+    this.setState({
+      formContents: formContents,
+    });
+  }
+  handleDelete() {
+    if (confirm('Are you sure you want to delete this item?') === false) {
```
darrint: We need a modal.

lemniscate613: Just chatted with @darrint about this. I played with a React/Bootstrap3 modal awhile ago and had issues in IE8. He played with one also, but I think he ran into similar issues (https://github.com/DirectEmployers/MyJobs/commit/5966eadd7ce263c023b2f91930720ec89415c6f9 ). User Management uses plain old `confirm`s. Let's make update these (and any others I'm forgetting) a separate ticket.

darrint: [PD-2257](https://directemployersfoundation.atlassian.net/browse/PD-2257)

gulp/src/common/ui/Datetime.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+function Datetime(props) {
+  const {name, onChange, required, maxLength, initial, isHidden, placeholder} = props;
+  return (
+    <input
+      type="datetime"
```
aspidites: According to Mozilla, this input type [is not supported](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime). It also doesn't seem to follow the spec outlined a few sprints back (where Date widgets would take a form similar to how they do in PRM).

gulp/src/common/ui/Select.jsx
```diff
@@ -0,0 +1,151 @@
...
+    this.openSelectMenu = this.openSelectMenu.bind(this);
+    this.closeSelectMenu = this.closeSelectMenu.bind(this);
+  }
+  onInputKeyDown(event) {
+    const {choices, name} = this.props;
+    const {keySelectedIndex, selectDropped} = this.state;
+
+    let killEvent = false;
+
+    if (selectDropped && event.key === 'ArrowDown') {
```
aspidites: Switch/case would be less verbose here.

gulp/src/common/ui/Select.jsx
```diff
@@ -0,0 +1,151 @@
...
+  }
+  render() {
+    const {choices, name} = this.props;
+    const {keySelectedIndex, selectDropped, currentValue} = this.state;
+
+    let dropdown;
+    let dropdownItems = [];
+    if (selectDropped) {
+      dropdownItems = choices.map((item, index)=> {
+        let active = '';
```
aspidites: You can use the same trick you used for required here.

gulp/src/myprofile/main.jsx
```diff
@@ -0,0 +1,290 @@
...
+      },
+    });
+  }
+  processForm(apiResponse) {
+    if (apiResponse) {
+      // TODO This could be abstracted further for reuse throughout all
+      // React / Django forms
+      const profileUnits = apiResponse.ordered_fields.map( (profileUnitName, index) => {
+        const profileUnit = apiResponse.fields[profileUnitName];
+        function wrap(child) {
```
aspidites: I think I was expecting all of these cases to be reusable.

aspidites:
I'm concerned by how much work main is doing. In particular, there's a processForm method which at any given point in type can return a different component. I was expecting these components to be individual top level elements such that I could actually invoke them elsewhere (otherwise, why are we making components). I also thought I remembered hearing at the meeting that a switch case to return a particular component type was undesirable, yet that's exactly what's being done in main.

gulp/src/common/ui/Datetime.jsx
lemniscate613: You're right, but I used `datetime` because I don't like Chrome's built in selector for `datetime-local` (the clickable bits are way too small):

screen shot 2016-03-22 at 11 18 43 am

I don't think we have a nice implementation of Datetime yet (unlike Select), and by "not supported" `datetime` safely falls back to a text field. I could make the advanced, homemade Datetime GUI as part of this ticket if @JLMcLaugh thinks it makes sense here. I was thinking that'd be a different ticket.

tdphillips: I don't think any profile units use datetimes. If we're going to use some kind of date input that's dubiously supported, it seems like "date" would be a better choice.

darrint:
The new components in `common/ui` can be invoked from anywhere. The switch case in process form is rendering them dynamically specifically for myprofiles. I object if that's the _only_ way to use the standard components, which seemed to be on the table at some point in the discussion.

Are you wanting to see a higher level component usable for more or less any Django form?

myprofile/tests/test_views.py
lemniscate613: Thanks @tdphillips Added more detail to this test: https://github.com/DirectEmployers/MyJobs/pull/2190/commits/9a58d1bf53aa56f3ad7b856342c964a430343329

lemniscate613:
@tdphillips Added delete confirmation and hiding button when adding new module: https://github.com/DirectEmployers/MyJobs/pull/2190/commits/780747eefb80f027531e0e5655a903456b64ffeb

myprofile/tests/test_views.py
tdphillips: That's not quite what I had in mind. You're grabbing the data for a Name profile unit. We know what fields should be present (given name, family name, and primary).

gulp/src/myprofile/main.jsx
lemniscate613: Done: https://github.com/DirectEmployers/MyJobs/pull/2190/commits/5b26e4669a998cfc6fcd8022c187a61af3da7c22

gulp/src/common/myjobs-api.js
```diff
@@ -28,6 +28,9 @@ export class MyJobsApi {
     const response = await fetch(url, {
       method: 'GET',
       credentials: 'include',
+      headers: {
+        'Accept': 'application/json',
```
darrint: I checked the other apps to make sure this header doesn't break anything. I'd prefer to keep this as the only way we send requests until there's a good reason not to.

gulp/src/common/ui/CheckBox.jsx
lemniscate613: Good idea: https://github.com/DirectEmployers/MyJobs/pull/2190/commits/d252b6c7a787de2804c1651021a0074828ed3d35

gulp/src/common/ui/CheckBox.jsx
lemniscate613: Here ya go: https://github.com/DirectEmployers/MyJobs/pull/2190/commits/10e21b1b5cc0e09d41d999118f446a75557f357c

gulp/src/common/ui/Select.jsx
JLMcLaugh: Is there a simple way to do that using both selectDropped and event.key?

lemniscate613: There's this ( http://stackoverflow.com/a/13207939 ), but it's less readable. Maybe something else?

myprofile/tests/test_views.py
lemniscate613: @tdphillips I see what you mean, and I updated it a bit more: https://github.com/DirectEmployers/MyJobs/commit/bb237838d1cdf28b72e76eca46c334dd417030ea

In general, though, I'm thinking API tests should focus on the size and shape of responses, and less their content. In other words, what does the front end absolutely need? For example, what pieces does the front-end refer to explicitly that would break if it wasn't there? We could theoretically change the model fields and their names, and from the frontend's perspective it shouldn't break anything.

If the *content* of the response matters (e.g. like a billing page where we *need* X, Y, Z) I'd argue we test that in some separate group of tests.

gulp/src/common/ui/Datetime.jsx
lemniscate613: @tdphillips The placeholder is coming to us directly from Django. The old Django forms must have been translating its form, or else the jQuery plugin was, before it was displayed to the user. This can be sorted out with this ticket: https://directemployersfoundation.atlassian.net/browse/PD-1971

gulp/src/common/ui/Select.jsx
lemniscate613: I'm going to leave this as is for this PR.

gulp/src/common/ui/Datetime.jsx
lemniscate613: This can be sorted out with this ticket: https://directemployersfoundation.atlassian.net/browse/PD-1971

PR #2191, opened 2016-03-22
[PD-2240] Parameterized top bar links

aspidites:
# Changes
- Got rid of the `toolbar` view, as it's not been used in quite a while. Microsites rely on `topbar`.
- moved hard-coded topbar links out of the topbar template into a template_tag (couldn't move it to the view since we don't always render the topbar using a view)
- simplified the external js logic used for topbar
- reorganized mobile version of topbar to more closely resemble the desktop version
- got rid of the back button in mobile view (use your browser's back button instead)
- removed postajob-specific links as we're not using postajob on secure.my.jobs, and I don't think we have any member companies using it


myjobs/views.py
```diff
@@ -468,50 +468,6 @@ def unsubscribe_all(request, user=None):
                               context_instance=RequestContext(request))
 
 
-def toolbar(request):
```
tdphillips: There are tests for this view.

tdphillips:
On super small resolutions (this is as small as my browser allows), this arrow is offset.
2016-03-22-105549_363x194_scrot

Top is production. Bottom is this branch.
2016-03-22-105445_400x150_scrot

aspidites:
@tdphillips  thanks, looking into it. Keeping this open in case I get other feedback, though.

aspidites:
Folded into my other PR.

PR #2193, opened 2016-03-23
Postajob UC 1

aspidites:
This mostly consists of use case 1 for postajob (see committed documentation) in addition to some things I needed to fix along the way to get postajob acting somewhat sanely. I also folded some of the long-standing PRs that I had open at one point throughout this sprint that weren't receiving any kind of commentary.

Added:
- activities  context processor, which will make it easier to do template-level activity checks*
- documentation for use case 1
- at_least_one and every helper functions to be used with universal.decorators.requires and User.can * 
- unified topbar (if my other PR doesn't get merged)
- admins property to Company for convenience
- can_modify helper for postajob to reduce boilier-plate -- not sure if this will stick around, but it's better than what I was doing
- postajob tests now have HTTP_HOST set by default so that we don't have to pass it in every time we call the test client
- proper activity checks for all UC 1 views -- conditionally hiding interface elements where appropriate instead of wrapping the entire view
- I rearranged utils.js so that it wasn't declared recursively *

Removed:
- quickstart guide as it was ineffective and inaccurate
- parts of the faq referencing company user
- MissingAppAccess, as it was deemed more appropriate to 404 when an app isn't available
- removed can_access_site since the check was causing more headaches than help
- toolbar view, as it was duplicating topbar
- mypartners template tags as they weren't being used 
- postajob decorators as they were getting in the way
- css declarations from psotajob forms -- moved to template so that we could use compress
- mydashboard templates ,as they weren't being used in any views
- unused mypartners templates (that were only around for the aforementioned template tags)
- all myjobs-specific postajob templates
-  while companyuser wasn't being used in postajob anymore, I made sure it didn't exist anywhere outside of seo.models *

Changed:
Last go around, @fezick had the following concerns:
- UC3 - needs revision in PD-2237 - posting form can be seen, but posting is not allowed
- UC8 - needs revision in PD-2237 - screen can still be seen

These should both be addressed by this PR

* Each of these was a smaller PR that didn't receive much attention. Rather than continue to have to maintain a lot of smaller branches while I awaited feedback, I folded those into this one. As I work on UC's 4, 5, and 6, if this one isn't reviewed by the time I'm done, I'll be folding this into that PR as well.


aspidites:
Few more wrinkles to iron out.

aspidites:
Tests are passing locally for me.

mklauber:
Not sure what to tell you.  Per #2195 The QC branch by itself will pass tests.  So something in your PR is causing failures.  I've looked through and I can't see anything, but I'll keep looking.

aspidites:
Don't want this rushed out the door, so I'm going to close until after the meeting.

PR #2197, opened 2016-03-24
Postajob UC 4, 5, and 6

aspidites:
Use cases explained in documentation and original ticket. Biggest caveat is that postajob has undefined behavior when you belong to more than one company. In other words, *don't* try to attempt to assume multiple personas with the same account. You should be fine if you unassign roles and reassign them as you're changing personas, or better yet, use two distinct email addresses for the two personas you are wishing to assume.

Size of this PR should go down once #2193 is merged. See [here](https://github.com/DirectEmployers/MyJobs/compare/pd-2237...pd-2237-2).


aspidites:
Don't want this rushed out the door, so I'm going to close until after the meeting

PR #2201, opened 2016-03-29
[PD-2235] Add Analytics to Job Searches

koonsies:
## Overview

This ticket is fairly simple. The goal is to add into javascript any facet data that is not retrievable from the URL of the job search. This includes BUIDs, default facets and featured facets.

## Testing

There are three templates that display this code.

- home_page_listing.html
- myblocks/blocks/searchresult.html (blocks, requires creation of Search Result block)
- job_listing.html

Testing Steps

- Go to any page that involves a job search.
- Open browser console
- View the object "analytics_info". It is in the global namespace


templates/home_page/home_page_listing.html
```diff
@@ -73,6 +74,14 @@ <h3 class="direct_highlightedText">{% trans "Jobs" %}</h3>
     </div>
     {% endif %}
 {% endblock directseo_micrositecarousel %}  
-
+{% block extra-js %}
+<script type="text/javascript">
+    var analytics_info = {
```
aspidites: If we add more info, this line will have to be updated. Alterantively, in `search_views.py`, you can do this instead:
```python
data_dict['analytics_info'] = get_analytics_info()
```

Which then allows you to simplify this code as:
```javascript
var analytics_info = {{ analytics_info |safe }}
```

koonsies: Yup, that's simpler.

myblocks/models.py
```diff
@@ -611,8 +614,18 @@ def context(self, request, **kwargs):
             'site_config': context_tools.get_site_config(request),
             'site_tags': settings.SITE_TAGS,
             'title_term': context_tools.get_title_term(request),
+            'analytics_info': json.dumps({
```
aspidites: Any reason for this to not use `get_analytics_info` or vice versa?

koonsies: Good question. I was torn on this decision.. Since this method exists in a models.py, I was concerned with potential circular imports. What are your thoughts?

aspidites: Valid concern. Give me a second to poke around.

aspidites: I'd move the function to seo.helpers myself. myblocks.models already imports it, and I can't imagine it doing much harm in seo.models.

aspidites: Deleted last comment because seo.helpers imports seo.models. universal.helpers seems the only safe place, but I'll leave it for now.

PR #2202, opened 2016-03-29
[PD-2273] Improve Custom Select Component

lemniscate613:
# Summary

This PR address two ideas from @tdphillips ( https://github.com/DirectEmployers/MyJobs/pull/2190#discussion_r56892665 ).

It also fixes ( https://github.com/DirectEmployers/MyJobs/commit/378598d01badea8edcf38b2f54ead63813f61157 ) an uncaught bug with a merged PR: https://github.com/DirectEmployers/MyJobs/pull/2190

# Review
## 1. Access React version
* Visit locally: https://secure.my.jobs/profile/view/
* Create an Address profile unit if you don't already have one
* Get the `ID` for an Address unit by clicking the edit button. For example **279727**:
```
https://secure.my.jobs/profile/api?id=279727&module=Address
```
* Use that ID to visit the React version (we aren't linking to the React version directly):
```
https://secure.my.jobs/profile/view/module/#/279727?module=Address
```
## 2. Click select
You should see a scrollable list like:

screen shot 2016-03-29 at 1 23 41 pm

## 3. Click the tiny arrow on the right

It should collapse.

## 4. Confirm bug fix

Edit a profile unit that has required fields. Don't edit anything -- just save the form. You should not see any errors about, "field required."


koonsies:
:+1: works as described. Though there are a few field issues I discussed with Daniel, such as the country drop down having an option called "Country"

PR #2203, opened 2016-03-29
Add cache servers

mklauber:


JLMcLaugh:
I don't see any changes to the microsite cache. I don't think 2 extra nodes are going to cut it either. Eviction rate is ~70% of new items. We should be aiming for sessions that are stored for at least 2 weeks and ~2 hours for the microsites cache. I think we're around 5 days and 15 minutes.

PR #2205, opened 2016-03-29
[PD-2175] Redirect Registration to /profile/view/

lemniscate613:
After a successful registration, simply redirect user to /profile/view/ (as opposed to showing user an alternate form).

While completing https://directemployersfoundation.atlassian.net/browse/PD-2273 (and scrubbing references to `account-page-2`) I found it also used upon successful activation. This PR removes that instance as well.


static/account-creation.173-22.js
```diff
@@ -91,21 +91,9 @@ $(document).on("click", "button#register", function(e) {
...
-                $("label[for=id_name-primary]").hide()
-                $("#titleRow").hide( 'slide',{direction: 'left'},250 );
-                $("#topbar-login").fadeOut(250);
-                setTimeout(function(){
-                    $("#account-page-2").show('slide',{direction: 'right'},250);
-                }, 250);
-                $("#gravatar").append(gravatar_url);
-                clearForm("form#registration-form");
-                $(".newUserEmail").html(user_email);
+                window.location.assign('/profile/view/');
```
aspidites: I like seeing red. Should javascript be doing anything at all if we're handling the redirect on the server? If not, that's even more red we can utilize.

PR #2207, opened 2016-03-30
[pd-2279] Make select component api export a controlled API.

darrint:
Controlled vs. Uncontrolled is in the sense of React form controls. Controlled is turning out to be what we want. I'm updating just the select control at this time.

### DIFF
--- a/gulp/src/myprofile/main.jsx
+++ b/gulp/src/myprofile/main.jsx
@@ -101,11 +101,20 @@ class Module extends React.Component {
       this.setState({
         apiResponse: apiResponse,
       });
+      this.updateFormContents(apiReponse);
     } else {
       window.location.assign('/profile/view/');
     }
   }
+  updateFormContents(apiResponse) {
+    const {formContents} = this.state;
+    apiResponse.ordered_fields.forEach( profileUnitName => {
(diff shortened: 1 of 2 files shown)


gulp/src/myprofile/main.jsx
```diff
@@ -163,12 +172,17 @@ class Module extends React.Component {
               />
           );
         case 'select':
-          const initial = find(profileUnit.choices, function findValueOfInitialItem(c) {return c.value === profileUnit.initial;});
+          const selected = formContents[profileUnitName];
```
darrint: This control is different in sending the current value to the component vs an initial value. Therefore we use `formContents`.

gulp/src/myprofile/main.jsx
```diff
@@ -101,11 +101,20 @@ class Module extends React.Component {
       this.setState({
         apiResponse: apiResponse,
       });
+      this.updateFormContents(apiReponse);
     } else {
       window.location.assign('/profile/view/');
     }
   }
+  updateFormContents(apiResponse) {
```
darrint: Since at least one control needs formContents to be populated as soon as the api completes, initialize all the fields to `thisProfileUnit.initial`.

PR #2208, opened 2016-03-30
[PD-2175] redirect users to central profile update functionality

lemniscate613:


myjobs/views.py
```diff
@@ -232,7 +229,6 @@ def home(request):
 def contact_faq(request):
     """
     Grabs FAQ and orders them by alphabetical order on question.
-
```
lemniscate613: I think my text editor added/removed all of these carriage returns...

PR #2210, opened 2016-03-31
[PD-2274] Add SB Error Reporting

koonsies:
## Overview

When there is an error loading a single secure block, there is no message back to the browser indicating that. Instead, the error is printed to the server log and left alone. This makes troubleshooting more cumbersome and potentially spams the logs.

This bit of code will return the exception message is settings.DEBUG is set to true, or will otherwise simply state that the block failed to load.

## Testing

Go to a secure block in the database and enter an invalid tag into the template area. I added a random {% else %} somewhere in the code and this triggered a template render error when the block was requested.

To view the error, check the network activity on the  page and view the "response" object from the call to /secure-blocks/. There will be a attribute called "errors" which will contain the info about the erroneous block.

### DIFF
--- a/myblocks/views.py
+++ b/myblocks/views.py
@@ -108,12 +108,12 @@ def secure_blocks(request):
                     request.GET))
         raise SuspiciousOperation(message)
 
-    response = {'cookies':[]}
+    response = {'cookies':[], 'errors': {}}
 
     for element_id in blocks:
         block = Block.objects.filter(element_id=element_id).first()
         if block is None:
-            logger.warn("Failed block lookup: %s", element_id)
+            response['errors'][element_id] = "Failed block lookup"
         else:
             try:
                 block = block.cast()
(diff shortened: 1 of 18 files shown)


myblocks/views.py
```diff
@@ -108,12 +108,12 @@ def secure_blocks(request):
...
         raise SuspiciousOperation(message)
 
-    response = {'cookies':[]}
+    response = {'cookies':[], 'errors': {}}
 
     for element_id in blocks:
         block = Block.objects.filter(element_id=element_id).first()
         if block is None:
-            logger.warn("Failed block lookup: %s", element_id)
+            response['errors'][element_id] = "Failed block lookup"
```
aspidites: If you use get here instead of filter + first, you'd get an exception, which you're already checking for anyway. In addition, you have DEBUG logic down there which you could take advantage of.

koonsies: I like this idea. This change will add a bit to code simplicity, but obfuscate the exception a bit. All the rest being the same, the person reading an exception on a non-DEBUG site would not know if there was an exception processing the block, or the block was non-existent/misnamed.

PR #2211, opened 2016-03-31
[PD-2175] redirect users to central profile update functionality

lemniscate613:


static/account-creation.173-22.js
```diff
@@ -198,23 +185,10 @@ $(document).on("click", "button.activation-login", function(e) {
                 }
             } else {
                 // perform the visual transition to page 2
```
tdphillips: This comment is no longer accurate.

myjobs/views.py
```diff
@@ -133,13 +133,9 @@ def home(request):
...
                     password=registration_form.cleaned_data['password1'])
                 expire_login(request, user_cache)
-                # pass in gravatar url once user is logged in. Image generated
-                # on AJAX success
-                html = render_to_response('includes/account-page-2.html',
-                                          data_dict, RequestContext(request))
-                data = {'gravatar_url': new_user.get_gravatar_url(size=100),
-                        'html': html.content}
-                response = HttpResponse(json.dumps(data))
+                ctx = {}
```
aspidites: Why are we still returning a json response? I'd expect a lot more red, given what this view is meant to do.

PR #2214, opened 2016-04-01
Add timeout to loggedout cookie [PD-2283]

tdphillips:
# TESTING

## Before switching to this branch

Log out of your account

Go to /accounts/password/reset/

Note the redirect :sob: 

## After switching to this branch

Do the same thing

Note the lack of a redirect :boom: 

## But there's more code here

Log out of your account

Walk away for some amount of time greater than three seconds

Come back

Go to /accounts/password/reset/

Note the lack of a redirect :boom:

### DIFF
--- a/registration/tests/test_views.py
+++ b/registration/tests/test_views.py
@@ -75,9 +75,6 @@ class RegistrationViewTests(MyJobsBase):
             response.cookies.get("loggedout"),
             "Expected the loggedout cookie to not be set, but it was")
 
-
-
-
     def test_valid_activation(self):
         """
         Test that the ``activate`` view properly handles a valid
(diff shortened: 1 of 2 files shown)


aspidites:
Looks good initially. I need to get my machine in a state were i can switch branches before taking for a test spin (don't let that stop others from looking though).

lemniscate613:
This is use for @darrint @misterjonscott and probably @koonsies : I'm not able to use this branch unless I add this to myjobs/context_processors:

```python
def webpack_dev_setting(request):
    """
    Used in templates that need to use webpack-dev-server in development.

    Don't set the setting in default_settings.py as it should remain off in
    production.

    i.e. Docker machine users probably want this:
    WEBPACK_DEV_SERVER_BASE_URL = 'http://secure.my.jobs:8080'

    Linux users probably want something like this:
    WEBPACK_DEV_SERVER_BASE_URL = 'http://10.10.12.999:8080'

    Linux users need an IP reachable by Modern IE VMs etc.
    """
    return {
        'wp_base_url': settings.WEBPACK_DEV_SERVER_BASE_URL,
    }
```

PR #2216, opened 2016-04-01
Add company roles to Pd 2236

JLMcLaugh:
Displays company user data on the company admin page with a new TableWidget.

I also made the prm saved search field lazy load to match the behavior of the Job Source ID and Featured On fields.


aspidites:
Tests.

aspidites:
Failing for new company form since those won't have a pk, let alone roles.

PR #2223, opened 2016-04-05
[PD-2261]  Replace Reporting Setup Walkthrough with Dropdowns

darrint:
This task includes these things:

1. Rework the models so that reports are based on DataTypes and are independent of PresentationType.
2. Create a new API for finding usable DataTypes.
3. Use the new UX design.
4. Remove the old wizard-like screens and API.
5. Add a placeholder for exporting reports.

I punted on history support for the choices in the new setup config bar. It can be done but this PR is already way too big.

## Test

1. `python manage.py migrate myreports`
2. `python manage.py reset_dynamic_data`
3. Go to https://secure.my.jobs/reports/view/dynamicoverview/
4. Run some reports.

## Notes

Clicking on a report in the report list takes you to a placeholder with a list of presentation types. Click one of those to download.


gulp/src/common/ui/Select.jsx
```diff
@@ -122,7 +122,7 @@ class Select extends React.Component {
       <div className="select-element-outer" tabIndex="0" onBlur={this.closeSelectMenu} onKeyDown={e => this.onInputKeyDown(e)}>
         <div className="select-element-input" onClick={selectAction}>
           <div className="select-element-chosen-container">
-            <span className="select-element-chosen">{value}</span>
+            <span title={value} className="select-element-chosen">{value}</span>
```
darrint: Current default data is too wide for the control. A hover is useful.

gulp/src/reporting/DataTypeSelectBar.jsx
```diff
@@ -0,0 +1,85 @@
+import React, {Component, PropTypes} from 'react';
+import Select from 'common/ui/Select';
+import {lookupByValue} from 'common/array';
+
+export default class DataTypeSelectBar extends Component {
```
darrint: This component replaces the earlier pages of the report type wizard.

gulp/src/reporting/ExportReport.jsx
```diff
@@ -0,0 +1,50 @@
+import React, {Component, PropTypes} from 'react';
+import {Loading} from 'common/ui/Loading';
+import {map} from 'lodash-compat/collection';
+
+export default class ExportReport extends Component {
```
darrint: This is a placeholder for where a downloaded report can later be customized.

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -0,0 +1,248 @@
...
+import classnames from 'classnames';
+import {WizardFilterDateRange} from './wizard/WizardFilterDateRange';
+import {WizardFilterSearchDropdown} from './wizard/WizardFilterSearchDropdown';
+import {WizardFilterTags} from './wizard/WizardFilterTags';
+import {WizardFilterCityState} from './wizard/WizardFilterCityState';
+import FieldWrapper from 'common/ui/FieldWrapper';
+import {SelectElementController} from 'reporting/SelectElementController';
+import DataTypeSelectBar from 'reporting/DataTypeSelectBar';
+
+export default class SetUpReport extends Component {
```
darrint: This is renamed from WizardPageFilter.

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -0,0 +1,248 @@
...
+        (...choices) => this.onMenuChanged(...choices));
+    this.loadData();
+  }
+
+  componentWillUnmount() {
+    const {reportFinder} = this.props;
+    reportFinder.unsubscribeToMenuChoices(this.menuCallbackRef);
+  }
+
+  onIntentionChange(reportingType) {
```
darrint: This function and two below handle changes to the user's changes to the report setup bar.

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -0,0 +1,248 @@
...
+    const {reportingType, dataType} = this.state;
+    this.buildReportConfig(reportingType, reportType, dataType);
+  }
+
+  onDataSetChange(dataType) {
+    const {reportingType, reportType} = this.state;
+    this.buildReportConfig(reportingType, reportType, dataType);
+  }
+
+  onMenuChanged(reportingTypes, reportTypes, dataTypes,
```
darrint: This callback is for _after_ the api calls behind the report set up bar have completed successfully and we know what to show.

gulp/src/reporting/api.js
```diff
@@ -24,58 +24,44 @@ class Api {
     return response.reports;
   }
 
-  async getReportingTypes() {
-    const promise = this.postToReportingApi('/reports/api/reporting_types', {});
-    return (await promise).reporting_type;
-  }
-
-  async getReportTypes(reportingTypeId) {
+  async getSetUpMenuChoices(reportingType, reportType, dataType) {
```
darrint: This replaces the walkthrough api from before.

It's proving difficult to implement a history mechanism in the set up report UI so I expect to rework this some.

gulp/src/reporting/api.js
```diff
@@ -24,58 +24,44 @@ class Api {
...
+      report_type: reportType,
+      data_type: dataType,
     };
-    const promise = this.postToReportingApi('/reports/api/report_types', formData);
-    return (await promise).report_type;
+    return this.postToReportingApi(
+      '/reports/api/select_data_type_api', formData);
+  }
+
+  async getExportOptions(reportId) {
```
darrint: Other new API. This serves the placeholder for export. It will expand to cover field order, sorting, etc.

gulp/src/reporting/reportEngine.js
```diff
@@ -9,49 +9,49 @@ import {mapValues, assign} from 'lodash-compat/object';
...
-      onErrorsChanged) {
-    const filters = await this.api.getFilters(rpId);
-    const name = await this.api.getDefaultReportName(rpId);
-    return await this.configBuilder.build(
-      name.name, rpId, filters.filters,
-      reportId => this.noteNewReport(reportId),
-      onNameChanged,
-      onFilterChange,
-      onErrorsChanged);
+  async buildReportConfiguration(reportingType, reportType, dataType,
```
darrint: 1. Make API call.
2. Set up ReportConfiguration which has the business logic.
3. Call back React components with results.

myreports/migrations/0011_auto__del_field_reportpresentation_configuration__add_field_reporttype.py
```diff
@@ -0,0 +1,508 @@
...
+                      self.gf('django.db.models.fields.related.ForeignKey')(to=orm['myreports.Configuration']),
+                      keep_default=False)
+
+        # Deleting field 'ReportTypeDataTypes.configuration'
+        db.delete_column(u'myreports_reporttypedatatypes', 'configuration_id')
+
+
+        # User chose to not deal with backwards NULL issues for 'DynamicReport.report_presentation'
+        # Better than nothing
+        #raise RuntimeError("Cannot reverse this migration. 'DynamicReport.report_presentation' and its values cannot be restored.")
```
darrint: I commented this out. South didn't give me a lot of options for dealing with this.

myreports/models.py
```diff
@@ -173,10 +175,78 @@ class DataType(models.Model):
     objects = DataTypeManager()
 
 
+class ReportTypeDataTypesManager(models.Manager):
+    def build_choices(
```
darrint: Most of the business logic behind menu choices and default choices is here.

aspidites:
> Rework the models so that reports are based on DataTypes and are independent of PresentationType.

Did this change go through @haliscooper? I seem to remember him being pretty adamant about the ERD design, specifically with respect to the foreign key relationships.

> I punted on history support for the choices in the new setup config bar. It can be done but this PR is already way too big.

Is there a ticket for this 'punt'?

gulp/src/common/array.js
```diff
@@ -30,3 +32,25 @@ export function intersperse(sepFn, inputFn, input) {
...
+ *      {value, '2', display: 'red'},
+ *   ]
+ *
+ * If multiple entries have the same value, only the first will be found.
+ *
+ * value: entry should have a value key equal to this.
+ * returns: the matching object or an blank object if none is found:
+ *   {value: '', display: ''}
+ */
+export function lookupByValue(objects, value) {
```
aspidites: This isn't very generic (thus I'm wondering if it belongs in common). Reasons for saying so:
- it assumes an object of shape {value:..., display:...}, thus if I wanted to use it on an array of objects that had values but where display didn't make sense, I can't reuse it
- the default return value, given what I mentioned before, could be changing the type of the fields -- if my values are integers, this would silently convert them to strings in the case where a result isn't found.
- while it is aptly named, this function is only useful in the case where I have a collection of objects with a value key. The moment i want to look up by display instead, for instance, this is useless to me.

What I'd have done:
```javascript
const value = _.find(intentionChoices, o => 
  o.value === intentionValue) || {value: "", display: ""}
```

If I found myself repeating the pattern, I might have written the following generic function:
```javascript
const lookup = (collection, predicate, default) =>
  _.find(collection, predicate) || default
```
To then use as follows:
```javascript
const default = {value: '', display: ''};

// using a lambda
const value = lookup(intentionChoices, o) =>
  o.value === intentionValue, default);

// using the _.matchesProperty shortcut
const  value = lookup(intentionChoices, ['value', intentionValue], default);
```
While you don't have to use this exact formulation, it does exhibit a few advantages:
- it will work for *any* collection, with predictable results
- switching between it and vanilla _.find is trival
- I know exactly what will be returned as a default in every case
- as a result of the previous point, the types of keys will only change if my default value forces it to do so

If you still prefer your version, I'd suggest that it wasn't put into common, as it's not very reusable in its current form, as discussed.

gulp/src/reporting/DataTypeSelectBar.jsx
```diff
@@ -0,0 +1,85 @@
...
+      onIntentionChange,
+      categoryChoices,
+      categoryValue,
+      onCategoryChange,
+      dataSetChoices,
+      dataSetValue,
+      onDataSetChange,
+    } = this.props;
+
+    return (
```
aspidites: Undecided on how I feel about the large number of props required to instantiate this component. I do like that the render function looks mostly to be data-driven. That is, there doesn't seem to be a lot of logic other than a key/value lookup, so kudos for that.

gulp/src/reporting/ExportReport.jsx
```diff
@@ -0,0 +1,50 @@
...
+  constructor() {
+    super();
+    this.state = {
+      loading: true,
+      options: null,
```
aspidites: Would this be better represented as an empty array or empty object? Nulls kinda scare me.

myreports/models.py
```diff
@@ -173,10 +175,78 @@ class DataType(models.Model):
...
+            self, user, reporting_type_name, report_type_name, data_type_name):
+
+        if not user:
```
aspidites: Wouldn't you have to explicitly pass None here for this exception to be raised? It also wouldn't catch the case where a user exists, but isn't logged in.

myreports/models.py
```diff
@@ -173,10 +175,78 @@ class DataType(models.Model):
...
+            raise SuspiciousOperation("No user provided.")
+
+        reporting_types = self.get_choices_from_model(
+            ReportingType.objects.active_for_user(user),
+            None, None, 'reporting_type')
```
aspidites: Having to explicitly pass in None twice raises a red flag for me. You also seem to be unconditionally passing the model type. As such, i'd probably move it before the other two attributes and have those default to None. Alternatively, I'd pass 'reporting_type' with it's key and omit the Nones, which would at least document what it means in this case.

myreports/models.py
```diff
@@ -173,10 +175,78 @@ class DataType(models.Model):
...
+            ReportType.objects.active_for_reporting_type(
+                selected_reporting_type),
+            None, None, 'report_type')
+        selected_report_type = self.select_best(
+            report_types, 'report_type', report_type_name)
+
+        data_types = self.get_choices_from_model(
+            DataType.objects.active_for_report_type(selected_report_type),
+            None, None, 'data_type')
+        selected_data_type = self.select_best(
```
aspidites: `select_best` isn't documented. In fact, non of the Manager methods are.

myreports/tests/setup.py
```diff
@@ -68,11 +64,11 @@ def create_full_fixture():
     ReportingType.objects.all().delete()
     rit_prm = ReportingTypeFactory.create(
```
aspidites: What's the significance of the 'rit' prefix vs the 'rt' prefix?

myreports/tests/setup.py
```diff
@@ -577,59 +560,78 @@ def create_full_fixture():
         configuration=con_comm_count,
         multi_value_expansion=False)
 
+    ReportTypeDataTypes.objects.all().delete()
+    ReportTypeDataTypesFactory.create(
```
aspidites: That you have to manually specify the ids for every factory instance is probably an indication that it needs to be a `Sequence` instead.

myreports/tests/test_activities.py
```diff
@@ -143,44 +140,24 @@ def test_dynamicoverview(self):
             "dynamicoverview",
             "read partner", "read contact", "read communication record")
 
-    def test_reporting_types_api(self):
+    def test_select_data_type_api(self):
         """
```
aspidites: Yay comments!

myreports/tests/test_models.py
```diff
@@ -57,7 +57,7 @@ def test_active_report_type_by_reporting_type(self):
                         .active_for_reporting_type(reporting_type))
         names = set(t.report_type for t in report_types)
         expected_names = set([
-            'Partners', 'Communication Records', 'Contacts'])
+            'partners', 'communication-records', 'contacts'])
```
aspidites: For future reference, Python has set literals: `{'partners', 'communication-records', 'contacts'}`.

myreports/tests/test_models.py
```diff
@@ -69,28 +69,171 @@ def test_active_data_types(self):
         expected_names = set(['unaggregated'])
         self.assertEqual(expected_names, names)
 
+    def test_first_active_report_data(self):
+        """Avoid different kinds of inactive report/data types."""
+
+        report_type = ReportType.objects.get(id=2)
```
aspidites: Didn't you name these things? Seems you could grab them by name rather than running another db query.

myreports/tests/test_models.py
```diff
@@ -69,28 +69,171 @@ def test_active_data_types(self):
...
+        self.assertEquals(
+            choices['report_types'][0],
+            choices['selected_report_type'])
+        self.assertEquals([], list(choices['data_types']))
+        self.assertEquals(None, choices['selected_data_type'])
+
+    def test_build_choices_select_reporting_type(self):
+        choices = ReportTypeDataTypes.objects.build_choices(
+            self.user, 'prm', None, None)
+        self.assertEquals(2, len(choices['reporting_types']))
```
aspidites: If you compared results using [assertDictEqual](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertDictEqual), you could save yourself a few asserts, since you'd then be comparing an expected result with an actual result, rather than checking against properties piece meal. That is, asserting that [1, 2] is equal to [1, 2] implicitly checks that they have a length of 2, that the item at index 0 is 1, etc.

myreports/models.py
```diff
@@ -173,10 +175,78 @@ class DataType(models.Model):
...
+            'reporting_types': reporting_types,
+            'selected_reporting_type': selected_reporting_type,
+            'report_types': report_types,
+            'selected_report_type': selected_report_type,
+            'data_types': data_types,
+            'selected_data_type': selected_data_type,
+        }
+
+    def get_choices_from_model(
+            self, model_qs, filter_field, filter_value, name_field):
```
tdphillips: If this were something like `(self, model_qs, name_field, **kwargs)`, the conditional could be removed and you could return `model_qs.filter(**kwargs)`.

I'm also not seeing name_field being used here.

myreports/tests/test_views.py
```diff
@@ -411,61 +413,112 @@ def setUp(self):
...
-        resp = self.client.post(reverse('presentation_types_api'),
-                                data={'report_type_id': '2',
-                                      'data_type_id': '3'})
-        data = json.loads(resp.content)['report_presentation']
-        names = set(v['name'] for v in data.values())
-        expected_names = set(["Contact CSV", "Contact Excel Spreadsheet"])
-        self.assertEquals(expected_names, names)
+    def test_select_data_type_api(self):
+        """Test that we get useful report setup menu choices."""
+        self.maxDiff = 10000
```
aspidites: See, you're already doing something similar (see my comment about assertDictEqual). :+1: .

myreports/views.py
```diff
@@ -371,92 +370,119 @@ def dynamicoverview(request):
...
-                    .active_for_reporting_type(reporting_type_id))
-
-    def entry(report_type):
-        return (str(report_type.id),
-                {'name': report_type.report_type,
-                 'description': report_type.description})
-
-    data = {'report_type':
-            dict(entry(rt) for rt in report_types)}
+    reporting_type = request.POST.get('reporting_type', None)
```
aspidites: Doesn't `.get()` imply None?

aspidites:
Note to everyone else reviewing this: Don't forget to `gulp build`.

aspidites:
Feedback.

- Right side doesn't seem to be aligning. I'd either expect the frame for the intention selection to line up with the widgets below, or the widgets within the intention selection to line up with those widgets:
2016-04-08-092314_652x366_scrot
- I can't tell the difference between an intention being broken and options not being available. For instance, compliance reporting is not implemented, but there was nothing to say "No reports available" for this intention type. If such messaging clutters up the interface, I'd expect that the option wouldn't be there at all.
- contents of the mutliselect are clipped, and theres no hover text or scrolling to allow me to reveal the rest of the text:
2016-04-08-092722_622x314_scrot
- which brings me to my next concern: there is a lot of empty space to the left of each component.
- If i click run report, but i'm scrolled far enough downward, there is no indication that the button actually worked. I can scroll up and see a link in the sidebar.
- I'm not sure it makes sense to see a "Run Report" button when there's no report to run (such as before selecting an intention)
- You can still manipulate report data after your session expires. This isn't caught by the logout behavior i wrote with a timer because it's simply cookie expiration that happens, rather than an explicit call to the logout view. We have a ticket with respect to stale data in User Management, so I suspect we will also be concerned with similar in reporting (even if you can't generate a report with that stale data).

TL;DR - The stuff in this comment doesn't need to be fixed in this PR, but we need to track it. My coding feedback I *would* like to be addressed in this same PR.

aspidites:
Almost forgot - I like the new approach. Actually feels like a single page app. Multiple locations will be scary (we never addressed it with original reporting), but that's for another day. Good stuff, @darrint  and @misterjonscott .

myreports/tests/setup.py
darrint: rit -> report-**ing** type
rt -> report type

myreports/tests/test_models.py
darrint: Those names are local to the setup method.

myreports/tests/test_models.py
darrint: If I did this it was because I only wanted to check a subset of the data. I'll take another look though.

myreports/tests/test_models.py
aspidites: You could make them properties then `self.report_type`. Or a more descriptive name.

gulp/src/common/array.js
```diff
@@ -30,3 +32,25 @@ export function intersperse(sepFn, inputFn, input) {
 export function flatMap(fn, input) {
```
darrint: I will remove this in PD-2192. (The change is committed locally.) In that branch the function is unused. If I do remove it here I'll end up creating a merge conflict.

gulp/src/common/array.js
darrint: I'd like to keep it as it's useful for nearly every Select and some other controls as well.

<Select ... value={lookupByValue(choices, selected).display} .../> is needed in nearly every app.

I'll reconsider the name though.

myreports/models.py
darrint: I took out the whole method. You are right, it wasn't doing much.

lemniscate613:
So much good work. Awesome job.

At least one comment: after selecting an item in Select, it seems that item should show in the Select. Currently it disappears.

aspidites:
Counter to the [mock-up](https://projects.invisionapp.com/share/JC5PNPKP4#/screens/130777992), you can only select contacts/partners with a multiselect. In the mock-up, there is a select box which toggles between the tag component and the multiselect component. Further, all contacts seem to be selectable from the drop-down.

It's unclear to me whether this deviation is an evolution of the design or an oversight. Clarification on how we expect partner and contact selection to work would be great, thanks.

tdphillips:
How many of the comments made on this ticket could/should be made into actual comments in our code?

Barring that, I'd really really really like to start seeing comments in new Reporting functionality.

darrint:
Added [PD-2035](https://directemployersfoundation.atlassian.net/browse/PD-2305) for tracking browser history.

darrint:
@haliscooper and @fezick both approved the ERD change several weeks ago.

myreports/tests/setup.py
darrint: Documented this a comments in a line above.

myreports/tests/setup.py
darrint: Ids are gone.

myreports/tests/test_models.py
darrint: Also, some of the values in the returned dict are still query sets.

darrint:
Regarding "counter to the mock-up", this is a design change. We expect to add the functionality you describe after initial release.  [PD-2307](https://directemployersfoundation.atlassian.net/browse/PD-2307)

misterjonscott:
I've made tickets to respond to Edwin's comments:
PD-2308 - Wrap setup fields to correct alignment with filter controls
PD-2309 - Disallow config options that would result in an empty report
PD-2310 - Replace side-by-side multiselect with tag interface.
PD-2311 - Make Run Report 'inactive' reset interface on Run

darrint:
Also added PD-2308, PD-2309, PD-2310, PD-2311. These address Edwin's earlier visual critiques.

myreports/tests/test_models.py
aspidites: But you know what should be in those querysets, which can be checked for equality, right?

aspidites:
> @haliscooper and @fezick both approved the ERD change several weeks ago.
Conversation? I'm sure they did, but I've neither seen the thread that motivates the change (which would help me understand why they are there and what I should be looking for), nor have I seen a comment from either of them to that effect.

myreports/management/commands/reset_dynamic_data.py
```diff
@@ -13,3 +14,8 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         create_full_fixture()
+        # mark compliance inactive
```
aspidites: Isn't this where myreports/tets/setup.py comes in? I mean, we are defining it as active to begin with when we shouldn't. I understand that later down the road this would be the way to go since we'd have data that we wouldn't want to nuke every release, but doing it here seems to defeat the purpose of having a management command that creates these different model instances as we want them from the get-go.

aspidites:
I have other feedback, but am putting that in a notebook as it technically pertains to the platform as a whole rather than this particular PR (eg, as I'm learning how the pieces fit together I'm finding things, but those pieces have been in motion for months now).

myreports/management/commands/reset_dynamic_data.py
tdphillips: Ditto.

myreports/migrations/0011_auto__del_field_reportpresentation_configuration__add_field_reporttype.py
aspidites: We should delete things rather than commenting them out.

PR #2224, opened 2016-04-05
[PD-2254] Wrap exact facets in quotes

koonsies:
## Overview

Search facets that involve Lucene reserve words (such as "and" and "or") would cause facets to fail. This PR involves wrapping facet terms in quotes to prevent this, as well as adding a test for the country Andorra (AND) to ensure this problem remains fixed

## Testing

To test, first execute the following line in Django shell
`import os, import_jobs; import_jobs.update_solr(36086)`

Using quality control, navigate to www.my.jobs/and/jobs .. There should be no jobs listed

Using this branch, navigate to the same URL .. There should be jobs listed. 

I also wrapped MOC, state, and city in quotes (as per instructions in the ticket) but these facets had no yet caused issues with reserve words. This change is more of a precaution.


seo/helpers.py
```diff
@@ -563,13 +563,13 @@ def filter_sqs_by_location(sqs, location_slug):
     for k, v in loc.items():
         if v:
             if k == 'country_short':
-                sqs = sqs.narrow('country_short_exact:(%s)' % v.upper())
+                sqs = sqs.narrow('country_short_exact:("%s")' % v.upper())
             elif k == 'state':
                 if v != 'none':
-                    sqs = sqs.narrow("state_slug:(%s)" % v)
+                    sqs = sqs.narrow('state_slug:("%s")' % v)
```
JLMcLaugh: Are these slug fields exact strings?

koonsies: They are not. I had made this change when I was considering where "OR"s or "AND"s could be found. I'm doing a bit of reading now on the impact of the quotes. I can remove these, though, as state and city didn't present any issues. I'm mostly trying to figure out why the quotes are good for the "_exact" fields but not the others.

JLMcLaugh: If dashes are treated as whitespace on indexing and unquoted searches, they would match. If search terms preserve dashes when quoted, it would no longer match, for instance.

PR #2225, opened 2016-04-06
Remove double periods in invitation emails [PD-2097]

tdphillips:


tdphillips:
2016-04-06-101049_483x52_scrot

PR #2226, opened 2016-04-06
[PD-1954] API endpoints for NUO

aspidites:
This adds 3 endpoints for NUO:
- add inbox
- update inbox
- delete inbox

I messed with the UI just enough to get the respective APIs testable. There are still some fun oddities such as the UI not refreshing when you add an inbox that I intentionally did not deal with. As this app isn't in line with the work that @misterjonscott has done, it didn't seem worth going to the trouble of modifying the UI when it'll  be completely different once this is closer to production ready anyway. I did go ahead and write JS and Python tests, though.


myjobs/migrations/0018_more_nuo_activities.py
```diff
@@ -0,0 +1,515 @@
...
+from south.v2 import DataMigration
+from django.db import models
+from myjobs.models import Activity, AppAccess
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        nuo = AppAccess.objects.get(name='NUO')
+
+        nuo_activities = [
```
tdphillips: Are you planning on addming more to this? As there's only one item in the list, you could do away with the for loop later on.

aspidites: I did and ended up not needing anything else. I'll fix this. Thanks.

PR #2228, opened 2016-04-07
[PD-2147] Company name now properly displays on all react apps

aspidites:
Before, we only showed the company name on user management and nonuser outreach. It's now on those, new reporting, and any future react app that derives from react base. If no company is passed into the template before rendering, the row is not displayed. I also made it so that the company name has the correct weight.

### DIFF
--- a/templates/base-react.html
+++ b/templates/base-react.html
@@ -3,128 +3,75 @@
 {% load common_tags %}
 {% load compress %}
 {% load staticfiles %}
-{% gz as gz %}
 
 {% localtime on %}
 <!DOCTYPE html>
 <html lang="en">
-<head>
-  {% block site-title %}
-  <title>My.jobs</title>
-  <meta name="title" content="My.jobs">
-  <meta charset= "utf-8">
-  {% endblock %}
+    <head>
+        {% block site-title %}
+            <title>My.jobs</title>
+            <meta name="title" content="My.jobs">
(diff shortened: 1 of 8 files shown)


lemniscate613:
Looks good as code. I'll pull it down locally in a bit.

templates/base-react.html
```diff
@@ -3,222 +3,164 @@
...
+                                        <br>
+                                        {{ message_info.message.body|safe }}
+                                        <div class="clear"></div>
+                                        <button class="btn pull-right mymessage-read-{{ message_info.message.id }}-{{ user.id }}">{{ message_info.message.btn_text }}</button>
+                                        <div class="clear"></div>
+                                    </div>
+                                {% endfor %}
+                            {% endif %}
+                        {% endblock%}
+                        <div class="contanier-fluid">
```
lemniscate613: Should be `container-fluid`, which alters style a bit.

aspidites: Yikes. Thanks

aspidites:
As @lemniscate613 container-fluid was messing with layout a bit. I opted to simply remove it as, of the 3 react apps, only reporting was using it. I figure @lemniscate613  or @misterjonscott can re-add it after evaluating the styles as a whole on some future date if need be. Our mobile support isn't exactly stellar anyway.

PR #2231, opened 2016-04-11
Add autofocus to lots of things [PD-1487]

tdphillips:
Some of this (reporting) is unnecessary. I did the work before confirming that I didn't need to bother, however, and was told I could just leave it in if it wasn't breaking anything.

I didn't touch our Select component as adding "autofocus" to a div seems wrong.

Testing:
Load up a page with forms. Examine the various inputs to ensure something is selected.


gulp/src/reporting/wizard/WizardPageFilter.jsx
```diff
@@ -68,6 +68,13 @@ export class WizardPageFilter extends Component {
 
     const rows = [];
     const errorTexts = reportNameError ? [reportNameError] : [];
+
+    // If the current list of filters does not contain a "city", we
+    // need to autofocus on the report name.
+    const autoFocusOnName = reportConfig.filters.filter(
+      function hasCity(element) {
```
aspidites: If you use an arrow function, you won't need to name it: `reportConfig.filters.filter( e => e.filter === 'city')`

lemniscate613:
This *is* a lot of things. 👍

PR #2233, opened 2016-04-13
[PD-2295, PD-2296] Re-enabled location for communication records in reporting

aspidites:
This PR re-enables the location filter for communication record:

2016-04-14-112401_1916x1056_scrot

I don't know what the implications of these changes are. Filtering seems to work, but I don't know if this should be handled differently. After much digging, it turned out contact location was already being defined, but was being ignored because we were excluding filter-only fields from search results. Afterward, I was getting a key error due to this particular field not having a FORMAT, whatever that means. 

I'm pointing this at @darrint's branch to illuminate the minimality of the change. If that branch gets merged first, i'll rebase onto QC and republish.


aspidites:
So apparently autocomplete breaks after a single completion. As I didn't touch that functionality, I'm not even sure how to fix it.

aspidites:
Also, while I see that tests are failing, I can't tell if its because of false assumptions made by the original code (and thus the tests), or because my fix is addressing the issue in the wrong way. Leaving this open until @darrint has a chance to peek at it.

aspidites:
Spoke with @darrint  in person. Confirmed that the tests need to be updated and the logic is correct. Rationale: The format attribute is meant to convey how the field should appear when exported (eg, to csv). As a column which is meant to be used for filtering only isn't formatted (as it won't show up in the resulting export), this format property should in fact, be None, which is exactly the result of dict().get(key) when a key isn't found.

What determines whether a column (in the reporting sense) is filter only or not is generally whether that column exists on the model being reported on. So for instance, a communication record has a contact_name, so that would appear in the output, but filtering by a communication records location is useful, despite not being a field in the comm records table (contact is a foreign key, location is a foreign key on contact, and city/state are fields on that location model/table).

myreports/models.py
```diff
@@ -259,7 +259,6 @@ def build_configuration(self):
             for cm in (
                 self.configurationcolumn_set
                 .filter(is_active=True)
-                .exclude(filter_only=True)
```
JLMcLaugh: Are we using filter_only for anything now? Are we assuming fields that aren't displayed won't have a format?

darrint: We aren't using it now AFAIK. Fields that aren't displayed won't have a format. The whole concept of `filter_only` is under-developed in this code.

I haven't had the mental bandwidth to think about how I want to handle `ColumnConfiguration.format` on these columns.

aspidites: We are using filter_only for contact locations, which is why I removed it to begin with. Without this, you'd never see the contact location column, despite it being defined. The alternative would be to make contact location filter_only=False, but I can only suspect that it was declared as filter_only intentionally, which is why I went this direction.

darrint: @aspidites I take that back. You are right. We do need `filter_only` to match the current behavior of the communication records report. I see a Location field in the output but it doesn't pertain to the city/state we are filtering on here.

I don't know exactly what @JLMcLaugh means by "fields that aren't displayed." If a field can appear in the report, it needs a format. Otherwise no. But this `filter_only` behavior is clearly undertested.

JLMcLaugh: I was just guessing at what filter_only was actually for. I assumed it was a field that we use to search on, but didn't want to display in the resulting report.

aspidites: I think that was the original intent, but I can't think of a case where we
would filter on something but not show it in a report. If new reporting
(unlike old reporting) is meant to exclude that location field, then it'd
make sense. Unfortunately Hal is home sick at the moment.

darrint: Looking at the current communication record report, I see a city/state filter available. I don't see a column where they appear in the report, which is what leads me to believe `filter_only` is needed.

haliscooper: Tons of cases where you would filter on something and not have it as a return in a report.  All aggregate reports, of course, and many raw as well.  Both for incompatibility with the results (aggregates) and for user experience reasons (If I report on Indiana, a column full of Indianas is redundant, I may show cities, but wouldn't show state, ESPECIALLY in a canned report, we look unpolished and unplanned in a situation like that.
For example, let's say we have a canned "communication records by state" report, user chooses indiana...

aspidites: @haliscooper -- i think the root question here is whether or not it make sense for contact location to be filter_only in the context of communication records report. If so, I think it's safe to say that my implementation is correct and barring any other issues, this is safe to merge. If not, then I would need to update the column type, though I still think in this instance we don't want to ignore filter_only columns, as the whole point is to determine which columns should be displayed for...filtering.

aspidites:
Need to point this to QC

myreports/models.py
JLMcLaugh: Trying to clarify for @haliscooper based on a conversation with @darrint:
filter_only was never implemented as it was intended. We were excluding filter_only columns from both the filter options and the resulting report. We currently leave columns out of the report by leaving the column's display format blank. This change will fix the filter options, but also allow the user to include filter_only fields in their report. I'd like to address that in a separate ticket if we can since it was an existing bug. It doesn't look like we have any columns we're using it for now.

@darrint @aspidites Is all that correct?

PR #2234, opened 2016-04-14
Assign a template when creating a login block.

aspidites:
When creating the login block, i wasn't assigning a default template. Admin seems to do some funkiness to accomplish this. I didn't look at how it was done there, but did go ahead and make sure the default template was assigned when generating login blocks as a result of ruining `enable_marketplace`. 

~~Tomorrow morning I'll write a regression test to ensure that the command creates a login block with  a proper template.~~ (Done)

### DIFF
--- a/postajob/helpers.py
+++ b/postajob/helpers.py
@@ -33,6 +33,22 @@ def can_modify(user, company, kwargs, update_activity, create_activity):
     else:
         return user.can(company, create_activity)
 
+
+def create_login_block(company, site):
+    page = Page.objects.create(
+        name="%s Login Page" % company.name, page_type=Page.LOGIN)
+    page.sites.add(site)
+    row = Row.objects.create()
+    row_order = RowOrder.objects.create(row=row, page=page, order=0)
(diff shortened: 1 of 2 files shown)


postajob/helpers.py
```diff
@@ -33,6 +33,22 @@ def can_modify(user, company, kwargs, update_activity, create_activity):
...
         return user.can(company, create_activity)
 
+
+def create_login_block(company, site):
+    page = Page.objects.create(
+        name="%s Login Page" % company.name, page_type=Page.LOGIN)
+    page.sites.add(site)
+    row = Row.objects.create()
+    row_order = RowOrder.objects.create(row=row, page=page, order=0)
+    template = open("templates/" + LoginBlock.base_template, "r").read()
```
aspidites: This is the crucial line.

tdphillips: Could we use myblocks.models.raw_base_template here? I'm betting some wsgi settings on the server might change the working path so this might be more complicated.

aspidites: That's way better. Didn't even know that was a thing. Thanks.

PR #2235, opened 2016-04-15
[PD-2147] Re-enabled content block for request company access request page.

aspidites:
# Description
I forgot that I was using the react-base template for the company access request page (so that I could have access to bootstrap 3). Thus, attempting to visit qc.secure.my.jobs/request-company-access results in a blank page. This fixes that.

# Testing
On this branch, you should see that visiting /request-company-access actually shows up now. I also added a proper header and breadcrumb div. Furthermore, the real react apps should still function (/reports/view/dynamicreporting, /manage-users, /prm/view/nonuseroutreach). No tests written  as it's all template work.

### DIFF
--- a/templates/base-react.html
+++ b/templates/base-react.html
@@ -72,6 +72,7 @@
                                 {% endfor %}
                             {% endif %}
                         {% endblock%}
+                        {% block content %}
                         <div>
                             <div class="row">
                                 <div class="col-sm-12">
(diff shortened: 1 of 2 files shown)


aspidites:
Also note that I have a public server available if you're wanting to test this without pulling the branch. Ping me in IRC and I'll give you the IP.

PR #2236, opened 2016-04-15
[PD-2192] Implement new tags control

darrint:
This replaces the original Tags prototype with the new `TagAnd` component.

There is also a `SelectByNameOrTag` component present in this PR but I'm not using it at this time. I decided not to delete it as it is styled and more or less functions. We will need it in the future when we allow "foreign" tag filtering. (i.e. filter contacts by partner tags)

## Test 

1. Set up a report .
2. Play with the tags interface. Also play with the two-pane selects.
3. Run some reports and verify that the results are sane.

## Review

This is a from scratch rewrite of the tags interface based on design by @misterjonscott .

There are no back end changes except for tweaking a template to remove a stylesheet link.


gulp/src/common/array.js
```diff
@@ -27,13 +27,6 @@ export function intersperse(sepFn, inputFn, input) {
 }
 
 /**
- * Like map but the function returns an array of items to concatenate.
- */
-export function flatMap(fn, input) {
```
darrint: @aspidites noticed this was redundant with `lodash`.

gulp/src/common/ui/TextField.jsx
```diff
@@ -4,17 +4,18 @@ import React from 'react';
...
       type="text"
+      autoComplete="off"
       id={name}
       name={name}
       className=""
       maxLength={maxLength}
       required={required}
       hidden={isHidden}
-      defaultValue={initial}
+      value={value}
```
darrint: @lemniscate613 beware of merge conflicts here.

lemniscate613:
good-guy-greg

koonsies:
Though this is merged, I noticed a small issue and wanted to see if anyone else had this.. When you highlight a tag, it illuminates a bit on hover.. If you click the 'x' to remove one of those tags from the selected list, it returns to the bottom box with the "hover" coloration still active. So.. The button will be lighter in the lower box until you hover over it and off of it again.

PR #2241, opened 2016-04-18
Update to newer nginx-proxy.

darrint:
@lemniscate613 bumped into this.

I upgraded and repatched our version of nginx proxy as this was fixed upstream.


lemniscate613:
Reviewing...

PR #2243, opened 2016-04-19
Impersonation backend [PD-2300]

tdphillips:
There's no UI for submitting these requests. I'm leaving it for PD-2302.

From a shell:
`SecondPartyAccessRequest.objects.create(account_owner=..., second_party=..., site=..., reason=...)`

This results in notifications being sent to the account owner:

This is shown on most pages.
2016-04-19-150133_992x205_scrot

This is what you see when looking at messages via the inbox.
2016-04-19-145845_629x339_scrot

The email being sent contains the same text.
2016-04-19-150035_698x202_scrot

Upon approving, the account owner sees the following:
2016-04-19-150155_784x102_scrot

The second party will then receive two notifications:

This one will be red if the request was denied.
2016-04-19-150951_988x171_scrot

There would normally be a link here but I was testing with localhost. Gmail stripped out the link. Rest assured one appears if you're not being dumb like that.
2016-04-19-150934_686x163_scrot

Upon clicking on that link, we come full circle and find out the account owner hasn't marked their original message as having been read.
2016-04-19-151016_1055x581_scrot

When hovering over the dropdown in the topbar, you see a handy dandy link that you've never seen before.
2016-04-19-152138_305x214_scrot


tdphillips:
While my naming isn't exactly all over the place, I've caught myself using two or three different verbs for "accepted" and "rejected" - which would we prefer? I think they all feel right.

In messages to the account owner, I use "allow" and "deny" for actions that the user can take.

In messages to the impersonator, I use "approved" and "rejected" to display the account owner's decision.

aspidites:
Haven't dug into the code yet, but preliminary feedback from working on my ticket:

- I think HttpResponseForbidden might be more appropriate than HttpResponse with a status code of 403
- I'd have to look again, but if there are multiple different scenarios where we'd want to raise a 403, I think it'd be helpful in debugging if those different scenarios were covered by different subclasses -- the subclasses provide enough information to tell you why the view is failing without leaking information to the end user
- Katie or someone else in marketing might be able to help in choosing wording for messages. 
- Do we need the quotes around the reason in the message box?

tdphillips:
We don't necessarily need the quotes around the reason. I wanted to differentiate it from our own text somehow.

aspidites:
> We don't necessarily need the quotes around the reason. I wanted to differentiate it from our own text somehow.

Ah. @misterjonscott, are quotes usually the best way to set apart different textual values like this? Indentation?

I'm fine if it stays how it is. Mostly curious.

koonsies:
merge conflicts here

tdphillips:
I'm unable to get a topbar on a microsite despite not overriding wide header, using a template that has a wide header, and doing site tag shenanigans locally.

koonsies:
Is "use secure blocks" disabled in the company config you are accessing?

tdphillips:
It is.

PR #2245, opened 2016-04-20
[PD-2297] Reporting Export Screen

darrint:
**Guidance Requested:** I'm putting this out early in hopes of streamlining the merge later. Please review it as if it were presented for merge.

**Note:** I don't recall every talking about where in the data models we put user visible names for columns *as they appear in the report*. Do we need a new ticket for that?

## Test:

1. Run a report.
2. Export it. Behold the new export options screen.
3. Play with the options to look for problems. Everything should work. It should look and behave as designed by @misterjonscott.

### DIFF
--- a/gulp/npm-shrinkwrap.json
+++ b/gulp/npm-shrinkwrap.json
@@ -1458,6 +1458,9 @@
     "react-prop-types": {
       "version": "0.3.0"
     },
+    "react-reorder": {
+      "version": "2.0.0"
+    },
     "react-router": {
       "version": "1.0.3"
     },
--- a/gulp/package.json
+++ b/gulp/package.json
@@ -38,6 +38,7 @@
     "react-bootstrap": "^0.28.0",
     "react-dom": "^0.14.3",
     "react-filtered-multiselect": "^0.4.2",
+    "react-reorder": "^2.0.0",
     "react-router": "^1.0.0",
     "tinycolor2": "~1.3.0",
     "warning": "^2.1.0",
--- a/gulp/src/reporting/ExportReport.jsx
+++ b/gulp/src/reporting/ExportReport.jsx
@@ -1,12 +1,20 @@
 import React, {Component, PropTypes} from 'react';
 import {Loading} from 'common/ui/Loading';
-import {map} from 'lodash-compat/collection';
+import SortableField from './SortableField';
+import Select from 'common/ui/Select';
+import Reorder from 'react-reorder';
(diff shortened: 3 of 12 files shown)


koonsies:
This isn't related to this PR, but what is the point of the "AND" sections in tags? Do they process differently if you select one tag, then select another tag in one of the ANDed sections vs just putting all the tags in the first input?

koonsies:
Few other general notes. Sorry if these don't belong here, I haven't looked as reports in a while.

When to click "Run Report", there's no notification informing you that something happened. There should probably be something that says "Report run, view it in the sidebar on the right" or something like that.

Report has both date_time and created_on but both have a simple date. Neither has a timestamp. (ex. 6/1/2015) Is date_time supposed to contain a timestamp?

aspidites:
> This isn't related to this PR, but what is the point of the "AND" sections in tags? Do they process differently if you select one tag, then select another tag in one of the ANDed sections vs just putting all the tags in the first input?

Within a box, the items are ored together, between boxes, they are anded. So:
2016-04-20-160459_460x226_scrot

Would translate to:
(Women OR Female OR Veteran) AND Minority

Which in turn would match any of these:
"Women Minority", "Female Minority", "Veteran Minority"

Yay cartesian products. Obviously more permutations if the second box had more items in it.

koonsies:
Ah. Thanks! That makes sense now. I'm sure this will receive a lot of questions, but this module isn't intended for untrained people.

aspidites:
> Report has both date_time and created_on but both have a simple date. Neither has a timestamp. (ex. 6/1/2015) Is date_time supposed to contain a timestamp?

In the db it does. In the interface, well, that answer has changed numerous times. Compare:
- https://secure.my.jobs/reports/view/overview
- https://secure.my.jobs/prm/view/records?partner=16803
- https://secure.my.jobs/prm/view/records/details?partner=16803&page=1

Note that the above links will only work if you are at least a PRM User for DirectEmployers. Point being that at various points in the interface (When displaying the exact same data) we have either opted for or against using a time stamp. As such, I don't really know that I could ever confidently answer "Should we be showing time stamps". If the question is instead "Do we?", then I'd answer "it depends", and if the question is "can we?", then the answer is "yes". 

Not to hijack @darrint's PR.

aspidites:
@koonsies with respect to training, that's probably where the tip box comes in, though from a design point I don't know if it makes sense. I foresee training videos a la user management.

JLMcLaugh:
@darrint Did you add react-reorder to package.json?

PR #2246, opened 2016-04-20
[PD-2301] Added impersonation information to logging.

aspidites:
# Description
When a contact log entry is created (except for saved searches), we should be logging if they were created as a result of an interaction initiated by an impersonator. That is, if user A impersonates user B and creates a phone record, then while the record will state that it was created by B, it should _also_ state that A was impersonating B in the process of being created.

# Testing
There isn't yet a UI for the impersonation stuff, so you'd need to manually create the objects necessary to begin an impersonation session, then manually create the objects that would be logged (ContactRecord instance) and inspect the log. That said, there is a test doing so, which means it should be safe to expect that if that test passes, this functionality is working properly. Alternatively, feel free to manually run lines analogous to what's in the provided test (using real models instead of factories) and you should be able to observe the results. 

On that note, there is a new migration, so you'll want to ``./manage.py migrate myjobs`` before poking around.


tdphillips:
I'll also note that work isn't entirely done on #2243. When that is complete, there will be less to look at here.

koonsies:
Merge conflicts, broked tests

aspidites:
Now that tests pass, I'll close this until @tdphillips work is done.

PR #2247, opened 2016-04-20
[PD-2295] Locations and Comm Rec Types

aspidites:
This is just #2233 pointing to QC instead of @darrint's branch, which was already merged.


aspidites:
I think we confirmed in the last ticket that this was ready to go, with the caveat that some concerns are overlapping. Decoupling that functionality will be handled in a different ticket that @JLMcLaugh made, which is assigned to @darrint .

koonsies:
When I merge quality control into this branch and run the code, I'm getting an error from the api/help call, as well as when I try to run the report.

`TypeError: __init__() got an unexpected keyword argument '' `

The report appears in the list after I refresh the page, but it is always empty. Is this happening to anyone else?

aspidites:
Runs fine for me. Did you rerun gulp build?

aspidites:
I will say I was getting that error a few commits back, so you may be running a stale branch.

koonsies:
Ah, that did the trick. I thought I had done that already.. It seems I was wrong.

PR #2248, opened 2016-04-20
[PD-2264] Adapt Old Report Preview to New Reporting

darrint:
With help from @aspidites and @JLMcLaugh this has come together more quickly than expected. Thank you.

## Test

1. Run a report
2. Click the preview link in the report list bar.
3. The current report should be highlighted in the preview and export screens.

The navigation works differently between the old and new systems. i.e. The report list bar has all the nav now.

## Review

~~**Guidance requested:** If `pd-2192--implement-tags-control` were merged this would be ready to review for merge. I'd like to get some early feedback to streamline merging this.~~

Most of the ugly hackery is in the new component `OldPreviewEmbed.jsx`. I've made minimal changes to `reporting.js` to support using the new api and rendering slightly differently shaped data.

As far as I can tell the preview numbers exactly match the numbers coming from the old system provided you are running the same query. The default query in the new system may not always match the default query in the old system. We may need a ticket for that?

Edit: Preview page has some style issues to be fixed in [PD-2317](https://directemployersfoundation.atlassian.net/browse/PD-2317)


koonsies:
Just to verify, this is not supposed to be a polished view, correct? Are we more-or-less ignoring CSS errors?

aspidites:
@koonsies  I'd been mentioning them. For the last such PR, new tickets were created in response for @misterjonscott to address later.

darrint:
We had created a ticket for @misterjonscott but forgot to note it here. See edited description above.

PR #2249, opened 2016-04-20
[PD-2315] Changed the FSM endpoint so that all sites are visible.

aspidites:
# Testing
If you view a business unit in the django admin, you'll see that the import field probably has two sites listed. If you do the same with this branch, you'll see that all sites are listed, which is desired behavior.

It's really a single line change, but my editor went ahead and stripped trailing whitespace.

### DIFF
--- a/seo/forms/admin_forms.py
+++ b/seo/forms/admin_forms.py
@@ -10,7 +10,7 @@ from django.core.urlresolvers import reverse_lazy
 from django.core.validators import EMPTY_VALUES
 from django.utils.decorators import method_decorator
 from django.utils.encoding import smart_text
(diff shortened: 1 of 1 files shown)


seo/forms/admin_forms.py
```diff
@@ -649,7 +649,7 @@ def __init__(self, *args, **kwargs):
         sites = SeoSite.objects.all()
         site_package_label = 'Show on these sites only'
         site_packages_widget = FSM(site_package_label,
-                                   reverse_lazy('site_fsm'), async=True)
+                                   reverse_lazy('site_admin_fsm'), async=True)
```
aspidites: This is the line that matters.

PR #2250, opened 2016-04-21
[PD-2325] Destination manipulation improvements

aspidites:
# Description
This adds a few conveniences in the django admin for staff when messing with destination manipulations. The parent ticket has a couple of screenshots if you're interested in seeing exactly what this allows.

This also adds a couple of lookup channels to be used with the ajax selects (which are used as autocomplete fields).

### DIFF
--- a/redirect/admin.py
+++ b/redirect/admin.py
@@ -1,6 +1,9 @@
+import unicodecsv as csv
+from cStringIO import StringIO
 from django.conf import settings
 from django.contrib import admin
 from django.db.models import Q
+from django.http import HttpResponse
 from django.utils.translation import ugettext_lazy as _
 
 from redirect.models import (CanonicalMicrosite, CompanyEmail,
@@ -148,7 +151,9 @@ class CustomExcludedViewSourceAdmin(admin.ModelAdmin):
     search_fields = ['buid', 'view_source']
 
 
+
 class DestinationManipulationAdmin(admin.ModelAdmin):
+    actions = ['export_as_csv']
     list_filter = ['action_type',
                    ('action', MultiSearchFilter),
                    BlankValue1ListFilter,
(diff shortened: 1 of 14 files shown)


redirect/admin.py
```diff
@@ -158,6 +166,40 @@ class DestinationManipulationAdmin(admin.ModelAdmin):
     list_display = ['buid', 'get_view_source_name', 'action_type',
                     'action', 'value_1', 'value_2']
 
+    def export_as_csv(self, request, queryset):
+        """Gives the ability to save the queryset as a csv"""
+
+        query = request.GET.get('q')
```
tdphillips: I know why this is doing what it's doing but it could use a few comments to that effect.

aspidites: Sure.

redirect/tests/factories.py
```diff
@@ -77,12 +77,12 @@ class Meta:
...
 class DestinationManipulationFactory(django.DjangoModelFactory):
     class Meta:
-        model = models.DestinationManipulation
+        model = 'redirect.DestinationManipulation'
 
     action_type = 1
     action = 'sourcecodetag'
     buid = 0
-    view_source = 10
+    view_source = factory.Sequence(lambda n: n)
```
tdphillips: Why is this a sequence?

redirect/admin.py
aspidites: @tdphillips, better? Too verbose?

redirect/tests/factories.py
aspidites: Irrelevant now, but it was a sequence because I needed to have a unique value for it (well, view_source, buid, and action_type together technically). Without it being a sequence, I was getting integrity errors by trying to batch create items. That said, I switched to using a for-loop and a range to specify this value because a lot of pre-existing tests were counting on view source being exactly 10.

redirect/admin.py
tdphillips: :+1:

PR #2253, opened 2016-04-21
[PD-1971 PD-2327 PD-2328] Date Selection Component, fix Select IE8 bug, make autofocus consistently a boolean

lemniscate613:
# Overview

Selecting a `date` field brings up a new date widget. User can edit date through the widget or by editing text.

screen shot 2016-04-21 at 11 41 54 am

# Update

* Fixed an issue with new modules (React's default PropTypes catches undefined, not nulls: https://github.com/DirectEmployers/MyJobs/pull/2253/commits/ed6a641c9c239e2c77077a3e23a775a46710119d ) that @aspidites @misterjonscott and @koonsies were having.
* Folded a closed PR ( https://github.com/DirectEmployers/MyJobs/pull/2254 ) into this current PR ( )

# Notes

* Works in IE8 🍔 
* Fixes Select bug in IE8 🍔 
* Uses date format which works with backend (django-remote-forms)
* Introduces node package: https://www.npmjs.com/package/calendar-base
* Introduces node package: https://www.npmjs.com/package/react-onclickout I started with react-onclickoutside, but it's not compatible with our stack.
* Year list defaults to past 50 years

PR #2254, opened 2016-04-21
[PD-2328] Make autofocus consistent

lemniscate613:
# Overview

In React components, we often have an "autofocus" value which designates if the field has focus on page load. Previously, it was sometimes a string sometimes a boolean. From now on, it's a boolean as that's how it's returned in `django-remote-forms`.

This PR depends on another PR: https://github.com/DirectEmployers/MyJobs/pull/2253

Therefore... all of those commits down there? Fahgettaboudit. The only one that matters: https://github.com/DirectEmployers/MyJobs/commit/3d70c3a247377376e301f843b0ffe9b5d7f7847c

# Good Point

yeah-dating-is-cool-funny-meme


aspidites:
Why does a PR that only fixes autofocus depend on one which introduces a new component?

tdphillips:
The other PR modifies autofocus as well. This appears to be built off of that one to reduce conflicts.

aspidites:
Then I'll sit on this one. I was hoping that as a smaller pr, it would be easy to merge, but I'm not keen on throwing the baby out with the bath water here.

PR #2257, opened 2016-04-22
[PD-2323] Only add roles, no kill-and-fill.

aspidites:
Without this, creating a user with an email address that already exists would override preexisting roles (which we want for edit, but not create). 

Long term, we'll probably want to disallow creating an existing user, but for now, this avoids the UI changes a proper solution would have required.

To test, add yourself as a user to a company you are already a member of with a role that you don't already have. You should see that you have that role in addition to the admin role you already had. On QC, you'll notice that you've then locked yourself out of a company instead.

### DIFF
--- a/myjobs/views.py
+++ b/myjobs/views.py
@@ -1133,6 +1133,7 @@ def api_create_user(request):
 
     Returns:
     :success:                   boolean
+
     """
     ctx = {'success': 'true'}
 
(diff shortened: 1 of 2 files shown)


myjobs/views.py
```diff
@@ -1160,9 +1161,7 @@ def api_create_user(request):
         # Does user already exist?
         if user:
             existing_roles = set(user.roles.filter(company=company))
-
-            # Update the user's roles, overwriting those for "company"
-            user.roles = user.roles.exclude(company=company) | new_roles
+            user.roles.add(*new_roles)
```
JLMcLaugh: Did the above test fail before this fix? send_invite should add the new roles. I don't even see where we replace the user's existing roles in the code. It's all adds as far as I can see.

PR #2258, opened 2016-04-25
[PD-2297] Export Screen

darrint:
## Test:

1. Run a report.
2. Export it. Behold the new export options screen.
3. Play with the options to look for problems. Everything should work. It should look and behave as designed by @misterjonscott. (Except for the fact that the column names are ugly)

**Note:** I don't recall every talking about where in the data models we put user visible names for columns *as they appear in the report*. This is mentioned in PD-2293.

## Review:

Some previous discussion happened in another PR: https://github.com/DirectEmployers/MyJobs/pull/2245

As far as I can tell in that discussion there wasn't actionable feedback except for a broken dependency. Please point out if I missed something from that PR.


aspidites:
>  don't recall every talking about where in the data models we put user visible names for columns as they appear in the report. 

Does `ConfigurationColumn`'s `alias` field not cover that?

aspidites:
> As far as I can tell in that discussion there wasn't actionable feedback 
[This](https://github.com/DirectEmployers/MyJobs/pull/2245#issuecomment-212583824) isn't actionable in its current state, but may be worth a ticket to track it's consideration. @fezick?

gulp/src/reporting/SortableField.jsx
```diff
@@ -0,0 +1,60 @@
...
+          onChange={sharedProps.onChange}
+          onMouseDown={e => {e.stopPropagation();}}
+          checked={item.checked}/>
+        {item.display}</label>
+      </div>
+    );
+  }
+}
+
+SortableField.propTypes = {
```
aspidites: These are really helpful (the comments). Specifically the shape method is the closes I'll ever get to type signatures.

myreports/helpers.py
```diff
@@ -138,3 +138,24 @@ def determine_user_type(user):
 
     if user.is_staff:
         return 'STAFF'
+
+
+def compare_records(order_by):
+    """Compare two dynamic reporting records by the given field."""
```
aspidites: What is a record here? Does record simply mean dictionary, or is it an instance of some custom class? If the latter, using dunder methods would allow for common syntax to be used, which would be less cognitive overhead for someone else reading code that uses this.

darrint:
I agree that `alias` is the right choice. I did some searching last week and I couldn't find anywhere that said officially that the display name should go there. @haliscooper?

@fezick I think we should make a separate ticket for the column display names in the linked feedback comment from @aspidites. It mentioned near the bottom of the list in PD-2293 so you will want to link it there. (`created_on` and `date_time` should be timestamps, not dates.)

aspidites:
Feedback:

- strange flickering issues (on chrome & firefox) - I may see if it's my compositor
- unchecking select all doesn't select none (check old reporting to see what I mean
- strange hovering text box in ie 8 (see screenshot below)
- choosing select all then clicking on another box is not allowed (i'd expect select all to be unchecked along with whatever i just clicked on)
- clicking select all aftward just unchecks whatever i had unchecked before I decided to click select all
- i didn't realize there were export format options until after I scrolled - I nearly missed it

Didn't actually get to check the output as another bug is preventing me from generating valid reports unless I modify the code. (None object has no attribute name when trying to grab record.contact.name). 

2016-04-26-155935_572x159_scrot

aspidites:
Looks clean and I do like that the output format is a select rather than yet another menu entry item.

darrint:
Not fixing:
* Weird rendering on @aspidites machine. Can't reproduce.
* Export options at the bottom per direction from @misterjonscott 

All other feedback should be addressed.

aspidites:
@darrint I'll be WM hopping some time this or next week, so I can revisit the flickering. I suspect it might be due to my compositor, but am not 100% sure yet.

aspidites:
All issues (including the weird flicker behavior) seem to be fixed. The only other thing I noticed was that I'm not able to click on an entire row to toggle the check box (check out old reporting to see what I mean). I mention it because it's a difference and not what I was expecting, coming from the old reporting (or anywhere else we have a checkbox, actually). Not sure what the modern stance from a design point of view is on this, though.

Be back shortly with results from actually checking the output.

aspidites:
Other observations:
- I can't drag the second item to the top of the list. I have to instead drag the first item to the second position
- column names aren't capitalized in the list, nor in the resulting output
- If i select a column to sort by and deselect it from the list, it disappears from the sort by select (expected). However, no fallback is selected (eg, the first column in the list. Instead, I get an empty select (not expected)
- Further, if I then re-enable the column, it's automatically selected -- I know why that's happening, but I'm not sure it's intuitive -- you decide. I do expect if the comment before this is addressed, this will become a non-issue, though

darrint:
Based on coversation with @aspidites @misterjonscott and me.

Not Fixing:
* Drag behavior. [constrained by Reorder component for now]
* Capitalization [separate ticket]

Will fix sort by select behavior.

aspidites:
Ticket being created for the checkbox behavior (the inability to click on an entire row to select/deselect anything).

PR #2259, opened 2016-04-25
[PD-2263] Reporting Improved Sidebar

darrint:
**Guidance Requested:** Seeking early feedback. This one is not based on quality-control.

## Test

1. Run a report.
2. Note sidebar spinner while running report.
3. Hover over reports to get a context menu for export and preview.


darrint:
Both tasks needed the same capability. @lemniscate613 did the research.

gulp/src/common/ui/ClickOutCompat.jsx
```diff
@@ -0,0 +1,20 @@
+import React, {PropTypes} from 'react';
+import ClickOutHandler from 'react-onclickout';
+
+export default function ClickOutCompat(props) {
+  const {children, onClickOut} = props;
+
+  if (window.addEventListener) {
```
aspidites: While `addEventListener` isn't present on IE 8, [attachEvent](http://stackoverflow.com/a/9769930) is.

gulp/src/reporting/DynamicReportApp.jsx
```diff
@@ -1,38 +1,62 @@
 import React, {PropTypes, Component} from 'react';
 import {ReportList} from './ReportList';
+import {remove} from 'lodash-compat/array';
+import {map} from 'lodash-compat/collection';
 
 // Props for this component come directly from the store state. (see main.js).
```
aspidites: is 'store state' still relevant, or is this from back when we were playing with redux?

gulp/src/common/ui/ClickOutCompat.jsx
darrint: It does appear to be an accurate indicator of whether OnClickOut will work.

aspidites: My point is that if the addEventListener doesn't exist, we should be falling back to attachEvent for IE 8, not ignoring the event all together.

darrint: It's a 3rd party component using addEventListerner (and who knows what else). We don't use the component if we're on ie8.

aspidites: Oh. Lovely :-\. Nothing to see here, moving along

aspidites:
Note to other reviewers -- be sure to run `npm install`.

aspidites:
Trying this a little differently. I'm writing these on the fly as I go through reporting.

- Much better create report button -- I even like that it matches the top bar
- there seems to be some extra padding between that button and the list of completed reports, or is that intentional? (see first screenshot)
- I like the pop up menu for menu items
- I know they are currently ordered as they were before -- would them being alphabetized make more or less sense?
- clicking on anything but the option button results in no action -- I wonder if we should make it do something by default (which is what current reporting does), or if we should change the cursor to only be a pointer when hovering over the icon?
- The pop up menu doesn't go away until after i click elsewhere, where  as I was expecting it to disappear as soon as my mouse left (given how large it is) -- is this a mobile consideration that the clicking needs to be deliberate?

Preview and Export pages are broken, but both of those are being handled in separate prs, so that shouldn't be an issue. That said, @darrint , this seems to be independent of your other PR, so if the issues above are addressed before those in your other PR, is there a reason to not go ahead and merge this one?

2016-04-28-092133_279x79_scrot

aspidites:
Other notes:
- Where did my other reports go? I have run more than 5 reports, but I only see the first 5. Note I said first, not last. As such, I have no way to preview or export reports that i just created if I've made more than this many.

aspidites:
Bug report; Trying to create a report with a name that is too long results in the following:
2016-04-28-093101_1920x1080_scrot

Also, I just created an absurd number of reports and am getting weird behavior. Can you stop by desk some time?

aspidites:
Looks like the create new report button works sporadically. I tried it earlier and it worked fine. After generating numerous reports, it then stopped working. Didn't see anything in the console to indicate errors. I can walk you through what i have, but I'm not sure of the precise steps to reproduce.

darrint:
I'll close this PR once I'm sure that all the feedback has been addressed or is copied to the new PR.

darrint:
Resolving earlier feedback per @misterjonscott 

* In the future clicking on a name will do something other than drop the menu. Likely it will go to the export screen. Will change the cursor style to not confuse the user though.
* New ticket for long report names: https://directemployersfoundation.atlassian.net/browse/PD-2370
* Other points are working as intended.

PR #2261, opened 2016-04-25
[PD-2343] Added location to partner Report

aspidites:
# Description
AJ noted while reviewing another of my tickets that the Partners report was missing the location filter.

# Testing
- Run `./manage.py reset_dynamic_data`
- Visit /reports/view/dynamicoverview
- Select "Partners Report" from the second select box. 
- You should see A City and State filter named "Contact Location"

### DIFF
--- a/myreports/tests/setup.py
+++ b/myreports/tests/setup.py
@@ -287,8 +287,17 @@ def create_full_fixture():
         configuration=con_part,
         multi_value_expansion=False)
     ConfigurationColumnFactory.create(
-        column_name="date",
         order=105,
+        column_name="locations",
+        filter_interface_type='city_state',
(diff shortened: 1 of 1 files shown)


koonsies:
When I export a partner report, the location field is blank. I see that this isn't part of this PR, just curious.

koonsies:
Otherwise, this appears to work at intended :+1:

aspidites:
> When I export a partner report, the location field is blank. I see that this isn't part of this PR, just curious.

That's intentional. Location is a filter_only configured column, which means that it's only used during filtering.

aspidites:
Also,  I think there's a separate ticket to handle export values (and properly dealing with filter_only types).

PR #2262, opened 2016-04-25
Pd 2317 - Make old Preview look acceptable in the new report builder.

misterjonscott:
Stylistic overrides to make old preview look ok in it's current home, but adapt to work in the new report builder, as well.


aspidites:
I don't know how close this is supposed to come with old reporting. In this version, there is no spacing between top three contacts:

2016-04-25-161806_690x276_scrot

But there is in the old reports view:

2016-04-25-161821_971x183_scrot

misterjonscott:
I made changes to respond to Edwin's observation.

aspidites:
Mobile is dumb, but fixing it would risk breaking old reporting.

PR #2263, opened 2016-04-25
[PD-2360] Removed duplicate config column for comm record reports.

aspidites:
What the title says. No regression tests because it was never determined to be a regression. That is, I asked if duplicate column configurations should be allowed, but never received a concrete response, so left that functionality as is. Thus, if duplicates are allowed, then it's not a bug and is simply human error.


koonsies:
:+1:

PR #2265, opened 2016-04-26
[PD-2362] Updated the way can works.

aspidites:
Up until now, the `User.can` method had been checking for app access, but not responding differently based on it. The reason for this was that we had, up until postajob, been using can in conjunction with requires, which checked for app level access automatically. Furthermore, until recently, the return types in either case (insufficient app access or missing activities) was the same. This PR, does a number of things to correct this:

- I added a custom manager/queryset so that you can check for required access on an activity queryset directly (not essential, but convenient).
- attempting to call `User.can` with a company with insufficient app-level access now raises an exception (Http404)
- due to the above, the explicit 404 introduced by an earlier commit is no longer necessary
- can now accepts another optional keyword argument, `check_access` - the alternative was to riddle code with try/excepts or `if company and 'Foo' in company.enabled_access`

# Testing
Hit these URLs for a microsite which doesn't have marketplace enabled (in app access):
- http://trustmarkcompanies.jobs/posting/all/
- http://trustmarkcompanies.jobs/posting/job/add/

If you don't want to add the domain to hosts, then `?domain=trustmarkcompanies.jobs` should work as well.

In either case, you should see a 404. Once you enable "MarketPlace" (see [documentation](https://github.com/DirectEmployers/MyJobs/blob/quality-control/doc/source/developers/postajob/setup.rst#enabling-the-marketplace) if you're unfamiliar with this), you should see 403s until you enable the correct activities for your user.

### DIFF
--- a/myjobs/models.py
+++ b/myjobs/models.py
@@ -5,10 +5,11 @@ import urllib
 import uuid
 
 from django.core.urlresolvers import reverse
-from django.db.models import Q, F
+from django.db.models import Q
 from django.db.models.loading import get_model
 from django.db.models.signals import pre_delete, post_save
 from django.dispatch import receiver
+from django.http import Http404
 from django.utils.crypto import get_random_string
 
 from impersonate.signals import session_begin, session_end
@@ -654,6 +655,9 @@ class User(AbstractBaseUser, PermissionsMixin):
         Output:
             A boolean signifying whether the provided actions may be performed.
 
+            If the company doesn't have sufficient app-level access, an
+            ``Http404`` exception is raised.
+
         Example:
 
             # given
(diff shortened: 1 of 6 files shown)


koonsies:
Refresh me.. Do I need to use a user other than my superuser to test this? I added www.facebook.jobs to my hosts (haven't used it previously) and was able to access both URLs. I verified and Facebook has no app accesses on my local.

koonsies:
:+1: It works as long as you are not a member of multiple companies.

aspidites:
To clarify @koonsies' last statement: **Postajob** works as long as you are not a member of multiple companies. The caveat is **not** limited to this ticket.

myjobs/models.py
```diff
@@ -685,16 +691,19 @@ def can(self, company, *activity_names, **kwargs):
             return False
 
         compare = kwargs.get('compare', every)
+        check_access = kwargs.get('check_access', True)
```
JLMcLaugh: Why use kwargs for compare and check_access instead of named arguments?

aspidites: Spoke in person: TL;DR - Python 2 has limited function signature rules. I can't have a named argument with a default value after a variable position argument list.

PR #2268, opened 2016-04-26
[PD-2328] Make autofocus consistent

lemniscate613:
# Overview

This used to be part of the -- now closed -- PR https://github.com/DirectEmployers/MyJobs/pull/2253

When determining if a form element should focus automatically on page load, we were using both booleans and strings. This PR makes them all bools. Why? This is how form data is returned to us by `django-remote-forms`.


aspidites:
Thank you!!

aspidites:
Autofocus isn't working on old reporting. BUT, @haliscooper  at one point said it was ok to ignore and users don't have an expectation for autofocus to work, so I'm leaving it be.

PR #2271, opened 2016-04-28
[PD-2352] Changed Data Source to Source

aspidites:
Exactly what it says. This is on the report generation page, not export.

@darrint there are a couple of tickets dedicated to fixing export names (column and values). For columns, what's the best place to change that? For values, I think you mentioned needing to come up with some other way to handle that, like some kind of format parameter or something.

### DIFF
--- a/myreports/tests/setup.py
+++ b/myreports/tests/setup.py
@@ -277,7 +277,7 @@ def create_full_fixture():
         order=103,
         output_format="text",
         filter_interface_type='search_select',
-        filter_interface_display='Data Source',
+        filter_interface_display='Source',
         configuration=con_part,
         multi_value_expansion=False)
     ConfigurationColumnFactory.create(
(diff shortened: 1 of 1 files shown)


darrint:
1. We need to officially decide that "Alias" in ColumnConfiguration is where the export display names go.
2. We need to plumb that name through to the export options api and download views.
3. Update all the ColumnConfiguration entries in setup.py to have names similar to the old system.

haliscooper:
Yes

aspidites:
@darrint 

1 & 2. I think it's safe to say that belongs in a separate ticket then (not that you would disagree, I'm just going on record and saying that because of this, I shouldn't be concerned with export display names here.
3. I can probably work on PD-2349 then. Sounds like export names, due to 1 and 2 would be tackled separately.

Is the above an accurate assessment about what I should be/am able to work on?

darrint:
@aspidites Yes.

PR #2272, opened 2016-04-28
Add action to User admin for requesting account access [PD-2302]

tdphillips:
Also adds some functionality I left out the first time around. I had to wrap django-impersonate's middleware to do it but we now disallow access to an account two hours after it starts, assuming the user hasn't relinquished control before then.

A superuser is selected in this screenshot.
2016-04-28-110310_324x313_scrot

As expected, the request failed.
2016-04-28-110328_381x403_scrot

Two normal users are selected.
2016-04-28-114740_326x375_scrot

Again, the request fails.
2016-04-28-114759_315x47_scrot

One normal user is selected.
2016-04-28-110351_329x368_scrot

We now see a form requesting the reason for access. This form is initially empty.
2016-04-28-110416_583x177_scrot

Submitting a blank form fails.
2016-04-28-111839_614x204_scrot

Submitting a valid form redirects back to the changelist. The original procedure now takes over and the account owner can approve or reject this request.


tdphillips:
The screenshots are identical using my last commit, with the exception that the green checks have been replaced with red dashes.

myjobs/admin.py
```diff
@@ -72,6 +75,27 @@ class UserAdmin(admin.ModelAdmin):
...
 
+    def request_account_access(self, request, queryset):
+        """
+        Determines if the "Request account access" action should be allowed to
+        proceed.
+
+        Succeeds if the staff member selects only one non-staff/super user to
+        request account access from.
+        """
+        if queryset.count() != 1:
```
tdphillips: ModelAdmin already handles the case where this is 0 so I've changed it from `!= 1` to `> 1`

myjobs/models.py
```diff
@@ -1040,6 +1040,10 @@ class SecondPartyAccessRequest(models.Model):
     # request was created on.
     site = models.ForeignKey('seo.SeoSite')
 
+    @property
+    def protocol(self):
+        return '' if self.site.domain.startswith('http') else 'http://'
```
tdphillips: Very few of our SeoSite instances already contain http[s]. They will probably never be used for this purpose, but... ¯\\\_(ツ)\_/¯

tdphillips: @JLMcLaugh, @fezick, what protocol should I default to here? Most seo sites don't have a protocol in the domain and I need to include one to fix the bug that @JLMcLaugh found. Defaulting to https would be nice but not all microsites support it.

aspidites:
I'm ready to merge this pending Jason or Jason's feedback. Discovered a couple of issues, but tickets are being created for them as they aren't related to the Admin or UI.

aspidites:
No response. I'm merging this. If the protocol becomes an issue, we can open up a new ticket...

JLMcLaugh:
I'm late to the party here, but http is fine. We force https where it's needed, and I don't see a practical way to store a preference for the request that wouldn't just confuse users.

PR #2273, opened 2016-04-29
[PD-2263] Reporting Improved Sidebar

darrint:
## Test

1. Run a report.
2. Note sidebar spinner while running report.
3. Hover over reports to get a context menu for export and preview.


darrint:
Need to change the cursor style to not confuse the user on report names.

aspidites:
Just transplanting these here so I remember what I found last time. Checked items are those that I know to have been addressed (will update shortly, as I'm only now reviewing the most recent changes):

- [ ] ~~there seems to be some extra padding between that button and the list of completed reports, or is that intentional? (see first screenshot)~~ intentional, leaving as is
- [ ] I know they are currently ordered as they were before -- would them being alphabetized make more or less sense?
- [ ] clicking on anything but the option button results in no action -- I wonder if we should make it do something by default (which is what current reporting does), or if we should change the cursor to only be a pointer when hovering over the icon?
- [ ] The pop up menu doesn't go away until after i click elsewhere, where  as I was expecting it to disappear as soon as my mouse left (given how large it is) -- is this a mobile consideration that the clicking needs to be deliberate?

aspidites:
From the other ticket:

> Other points are working as intended.

@darrint, could you elaborate?

aspidites:
New feedback: 

Strange wrapping on mobile/table:
2016-05-03-084320_572x1056_scrot

Other than this and my above two concerns (about ordering of items in the list and the on exit behavior), everything looks good. It may be that after @darrint  clarifies, those two are dealt with already as well.

darrint:
Thanks for making checkboxes. That helps.

> Other points are working as intended.

That meant that (based on a review with @misterjonscott) I believed "everything else" was working correctly. I'm assuming we just need to resolve the remaining checkboxes.

To reiterate what @aspidites said above there are three pieces of feedback to address: 2 comments with checkboxes and a third comment about wrapping. @misterjonscott can you give me some direction on these?

misterjonscott:
> I know they are currently ordered as they were before -- would them being alphabetized make more or less sense?

I think it makes sense, as the user, for my most recent reports to be at the top of the list, and older reports moving further and further from view.  In the prototype, we've also got a js-based search/filter at the top, that will help users find reports that aren't recent.

> The pop up menu doesn't go away until after i click elsewhere, where as I was expecting it to disappear as soon as my mouse left (given how large it is) -- is this a mobile consideration that the clicking needs to be deliberate?

This sort of menu typically requires a blur action to dismiss.  Think of it like a context menu in OSX or Windows.  The window stays visible until you've done something to dismiss it.  This also serves to prevent the menu from disappearing when users accidentally fumble the mouse, after clicking to activate the menu.

> Strange wrapping on mobile/table:

Unless I'm misunderstanding this, what we're seeing is the BS3 responsive framework moving the sidebar to the bottom of the page, when the viewport is too narrow.  This is how the sidebar is generally handled in responsive design, but also how we're handling it elsewhere.

aspidites:
> I think it makes sense, as the user, for my most recent reports to be at the top of the list, and older reports moving further and further from view. In the prototype, we've also got a js-based search/filter at the top, that will help users find reports that aren't recent.

Sorry, I should have been clearer here. I meant that in this screenshot:
2016-05-03-094752_337x240_scrot

While the order in old reports is Preview, then Export, I was wondering if Export coming before Preview made more sense. Before we didn't have words on this view, so the order didn't really matter. That we now have text is why I raised this issue.

> This sort of menu typically requires a blur action to dismiss. Think of it like a context menu in OSX or Windows. The window stays visible until you've done something to dismiss it. This also serves to prevent the menu from disappearing when users accidentally fumble the mouse, after clicking to activate the menu.

K. I marked this as resolved above.

> This is how the sidebar is generally handled in responsive design, but also how we're handling it elsewhere.

I was aware of this, but assumed other cases were a matter of "We don't care about how things look on mobile". I brought it up here because my impression was that we wanted to rectify that. I'd have expected the menu to be full width in this case. See PRM for example:

2016-05-03-095135_476x1056_scrot


Basically, our current handling (For bootsrap 3 apps) is consistent, but I wonder if it's consistently broken? Alternatively, is it the case that by changing some css attributes somewhere, we could unilaterally fix all occurences? If so, I'd be fine punting this as it's not something we have to do for every single invocation of the sidebar.

misterjonscott:
@aspidites 

The menu won't have enough contents, even in the foreseeable future, to concern ourselves with the order, but if we were to, I'd recommend putting the listing in order of popularity.  Most commonly used items nearer to the top.  Again, the list isn't large enough to justify that as a priority and I'm not sure how we'd know what's most popular, right now.

Regarding the mobile view, I adopted existing grid usage to make the sidebar 6 columns wide, but simply changing the class on that sidebar from col-xs-6 to col-xs-12 will make it full width when the viewport is small.  We could easily make that change, here, or as a sweeping change, later.  I'm open to either.

aspidites:
@misterjonscott  both of those responses make sense. Doesn't seem like anything but a simple class change, so I'll merge this for now. I'll see if @fezick  has a preference and create a ticket if he does, which I'll probably work on myself if it's that trivial to fix. Thanks.

PR #2275, opened 2016-04-29
[PD-2369] Eliminate mutation on fields list in ExportReport.

darrint:
Found that @aspidites dragging bug was not specific to him but due to a bug I introduced. This change protects us from any mutations done in the reorder component and better follows React's state management practices.

## Test

1. Export a report.
2. Drag fields up and down. Make sure to drag a field to the first position.

## Review

There shouldn't be any direct mutation of entities given to React state. They should be copied and replaced.

### DIFF
--- a/gulp/src/reporting/ExportReport.jsx
+++ b/gulp/src/reporting/ExportReport.jsx
@@ -26,15 +26,18 @@ export default class ExportReport extends Component {
   }
 
   onReorder(event, item, index, newIndex) {
+    console.log('onReorder', item, index, newIndex);
     const {fieldsSelected} = this.state;
-    const removed = fieldsSelected.splice(index, 1);
-    fieldsSelected.splice(newIndex, 0, ...removed);
-    this.setState({fieldsSelected});
+    const newFieldsSelected = [...fieldsSelected];
+    const removed = newFieldsSelected.splice(index, 1);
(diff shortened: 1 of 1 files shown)


gulp/src/reporting/ExportReport.jsx
```diff
@@ -26,38 +26,49 @@ export default class ExportReport extends Component {
   }
 
   onReorder(event, item, index, newIndex) {
+    console.log('onReorder', item, index, newIndex);
```
aspidites: This shouldn't be here.

gulp/src/reporting/ExportReport.jsx
```diff
@@ -26,38 +26,49 @@ export default class ExportReport extends Component {
...
     const {fieldsSelected} = this.state;
-    const removed = fieldsSelected.splice(index, 1);
-    fieldsSelected.splice(newIndex, 0, ...removed);
-    this.setState({fieldsSelected});
+    const newFieldsSelected = [...fieldsSelected];
+    const removed = newFieldsSelected.splice(index, 1);
```
aspidites: Splice mutates, slice doesn't.

darrint: Line 31 takes a copy. I'm operating on the copy.

PR #2279, opened 2016-05-02
Display names are correct for filtering and export screen.

aspidites:
# Screenshots
2016-05-02-133522_616x356_scrot
2016-05-02-133532_306x307_scrot

# Testing
Just pull this branch and run `manage.py reset_dynamic_data`. Visiting the preview or export screen should yield something similar to the above screenshots.

This is *not* dealing with the column names in csv/excel output. That will be handled by a separate ticket.


aspidites:
I know about the failure. Will address it after I look through the other PRs.

PR #2282, opened 2016-05-04
Add times to two dates [PD-2359]

tdphillips:
Magic.

Shows up as something like this: `04/07/2016 06:32:00 PM`


aspidites:
Rerunning tests.

PR #2283, opened 2016-05-04
[PD-2305] Browser History

darrint:
## Review

* Several components had to be updated to be controlled. (Wasn't needed at this point, but will be for clone. Keeping.)
* Certain user interactions push or replace history entries.
* **New**: I replaced the default index redirect one I wrote that figures out the menu entries before redirecting. This gives me a place to explicitly replace bad or missing report type selections. It also clears up some bad behavior that only showed up on Firefox.
* ~~I stuff a lot of things into the query string. Most are reasonable. I have my doubts about the current filter though.~~ Took this out. Old reporting doesn't try to solve this.
* The mapping from the API to the components was messy. API was sending `{key: ..., display: ...}` and the components wanted `{value: ..., display: ...}`. The mapping wasn't being handled consistently. I went ahead and fixed the API.
* `reportEngine.js` is supper messy right now. I have clear ideas of how I want to fix it but I don't want to attempt it until DEAM activities calm down.

## Test

* Manipulate, and switch between reports.
* Switching between reports should create useful history.
* Manipulating filters should not create new history entries. (i.e. you shouldn't have to back 3 times through the last 3 edits you made to the current report.)
* All components should work with history **except** date range, which I've left alone.
* **New**: If you edit a report, then switch to an export page, the come back, your work is gone. Same as old reporting.


aspidites:
> reportEngine.js is supper messy right now. I have clear ideas of how I want to fix it but I don't want to attempt it until DEAM activities calm down.

Are these catalogued somewhere? Would be easier/faster to fix these things in the future if they are catalogued. I also don't know what is a valid comment as a lot of things I might mention you may have already considered, others you may not have.

aspidites:
> All components should work with history except date range, which I've left alone.

Why?

aspidites:
Sometimes, going back in history leaves me with an invalid state. In this case, none of my previously selected report(ing) types were retained:

2016-05-04-140307_700x190_scrot

darrint:
>> ... date range

> Why?

Because I think @lemniscate613 is still working on it.

darrint:
> Sometimes, going back in history leaves me with an invalid state.

Can you copy paste the URL in your browser when you are in that bad state?

aspidites:
> Can you copy paste the URL in your browser when you are in that bad state?

http://10.10.12.188:8002/reports/view/dynamicoverview/#/set-up-report?filterJson={}&_k=s1qxhg

darrint:
> Would be easier/faster to fix these things in the future if they are catalogued.

Here's my first attempt at a catalog:

* The method `buildReportConfiguration` has a _lot_ of callbacks. Also `ReportFinder` maintains some callback subscribers. Break that state into redux reducers for the report list, current report setup, etc.
* Document the data model represented by the reducers.
* Convert all the filter manipulation methods into actions/reductions.
* Pull as much as possible of the state from `SetUpReport.jsx` into actions/reductions.

On the plus side, the code I have unit tests for in `reportEngineSpec` tends to stay working. I would keep/rewrite those tests to maintain that coverage. Moving state out of `SetUpReport` will make it possible to expand the unit tests.

Code does not actually work. I was thinking out loud at the keyboard:

https://github.com/DirectEmployers/MyJobs/compare/testredux?expand=1

darrint:
> Sometimes, going back in history leaves me with an invalid state.

I have fixed a number of bugs that were present on this branch. :-/

I'm trying not to be too aggressive about pushing the user out of bad states, like you saw in the data select box. It's easy to create an endless loop. i.e.:

1. Location state is bad.
2. Construct a better url + query.
3. Push history state to better url + query.
4. State is still bad. Goto 2.

That said, if you start by clearing the entire fragment part of the URL, (`#set-up...`) and pressing enter to navigate there, that should be a good state. From there it should not be possible to naturally navigate to a bad state. If you can, it's still broken. But hacking around directly with the URL can get you into a bad state and that's where I don't want to be too aggressive.

aspidites:
> New: I replaced the default index redirect one I wrote that figures out the menu entries before redirecting. This gives me a place to explicitly replace bad or missing report type selections. It also clears up some bad behavior that only showed up on Firefox.

Wrong. It showed up everywhere *except* Chrome. I could reproduce on IE, for instance.

PR #2284, opened 2016-05-04
[PD-2335, PD-2349] Human-Readable Export Values

aspidites:
# Testing
1. Run `manage.py reset_dynamic_data`
2. Generate a dynamic report and export it.
3. Compared to quality-control, you should see that communication records have human readable names. 
4. You should also see that column names before and after export are human-readable

~~*note: This does **not** handle column names in the exported file. This handles values in cells, and only for a particular column.~~ Went ahead and implemented that as well. Both column names on the export screen and in the resulting csv/excel should be correct.

### DIFF
--- a/gulp/src/reporting/ExportReport.jsx
+++ b/gulp/src/reporting/ExportReport.jsx
@@ -94,7 +94,7 @@ export default class ExportReport extends Component {
     const options = await reportFinder.getExportOptions(reportId);
 
     const fields = options.report_options.values;
-    const sortBy = get(fields, '[0].value');
+    const sortBy = get(fields, '[0].display');
     const fieldsSelected = map(fields, o =>
       ({...o, checked: true}));
 
--- a/myreports/report_configuration.py
+++ b/myreports/report_configuration.py
@@ -59,10 +59,11 @@ class ColumnConfiguration(object):
     filter_display: name to display alongside the filter_interface
     help: boolean, is help available for this column?
     """
-    def __init__(self, column, alias, format,
-                 filter_interface=None, filter_display=None, help=False):
+    def __init__(self, column, format,
+                 filter_interface=None, filter_display=None, help=False,
+                 alias=None):
         self.column = column
-        self.alias = alias
+        self.alias = alias or column
         self.format = COLUMN_FORMATS.get(format, '')
         self.filter_interface = filter_interface
         self.filter_display = filter_display
--- a/myreports/report_configuration.py
+++ b/myreports/report_configuration.py
@@ -34,7 +34,7 @@ class ReportConfiguration(object):
             ]
 
         return dict(
-            (c.column, c.extract_formatted(raw_data))
+            (c.alias, c.extract_formatted(raw_data))
             for c in columns)
 
     def find_column(self, column_name):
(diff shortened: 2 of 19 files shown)


gulp/src/reporting/ExportReport.jsx
```diff
@@ -94,7 +94,7 @@ export default class ExportReport extends Component {
     const options = await reportFinder.getExportOptions(reportId);
 
     const fields = options.report_options.values;
-    const sortBy = get(fields, '[0].value');
+    const sortBy = get(fields, '[0].display');
```
aspidites: By the time we're iterating over values, they are json, so we need to order by the human readable column name, not the database table column name.

myreports/report_configuration.py
```diff
@@ -34,7 +34,7 @@ def format_record(self, raw_data, values):
             ]
 
         return dict(
-            (c.column, c.extract_formatted(raw_data))
+            (c.alias, c.extract_formatted(raw_data))
```
aspidites: Use the column's alias for the serialized dictionary, rather than the database column name.

myreports/report_configuration.py
```diff
@@ -60,8 +60,10 @@ class ColumnConfiguration(object):
     help: boolean, is help available for this column?
     """
     def __init__(self, column, format,
-                 filter_interface=None, filter_display=None, help=False):
+                 filter_interface=None, filter_display=None, help=False,
+                 alias=None):
```
aspidites: Last so that it isn't a required argument. Defaults to what ever the column name is.

myreports/tests/setup.py
```diff
@@ -378,7 +378,7 @@ def create_full_fixture():
         filter_interface_type='search_multiselect',
         filter_interface_display='Communication Type',
         configuration=con_comm,
-        output_format="text",
+        output_format="comm_types_list",
```
aspidites: new output format to handle communication record types. In the future we'll probably do something more robust, but with DEAM around the corner, it was deemed (no pun intended) adequate to have a very specific formatter for communication types.

myreports/tests/test_views.py
```diff
@@ -692,7 +692,7 @@ def test_dynamic_contacts_report(self):
         response_data = json.loads(resp.content)
         self.assertEquals(10, len(response_data['records']))
 
-        first_found_name = response_data['records'][0]['name']
+        first_found_name = response_data['records'][0]['Name']
```
aspidites: Unit tests should be expecting the human readable versions, not the raw values.

myreports/views.py
```diff
@@ -707,7 +707,9 @@ def download_dynamic_report(request):
     response['Content-Disposition'] = disposition
 
     output = StringIO()
-    presentation.write_presentation(values, sorted_records, output)
+    columns = [
```
aspidites: Definitely a kludge. The alternative was to re-shape the serialized value to include the aliases, which would also have required me to update numerous other files (the export classes, configuration builder, test setup, model, view, and a few others).

fezick:
@aspidites, this branch has conflicts now.

koonsies:
Not sure if this is actually a bug, but just something I noticed. If all boxes are unchecked, the report is generated with all columns. Just a note.

koonsies:
Other than my small note about the checkboxes returning all when none are selected, I give a hearty 👍

aspidites:
About the check boxes: probably important but certainly not introduced by
my PR. I have a branch that's messing with those check boxes, so it may be
a good chance to address your concern.

aspidites:
@koonsies finished implementing your concern in another branch (no PR yet). The solution I came up with was to simply disable the export button when no columns are checked. I didn't disable the options selects because they aren't selects, which means I can't disable them without first adding a proptype to all selects that accounts for it, which is far outside of the scope of my tickets. I'll probably make a ticket for that and autofocus though.

aspidites:
I went ahead and [created the ticket for selects](https://directemployersfoundation.atlassian.net/browse/PD-2380).

PR #2266, opened 2016-04-26
[PD-1971] - Date Selection Component

lemniscate613:
# Note

This is a stripped down version of a previous, now closed, PR: (https://github.com/DirectEmployers/MyJobs/pull/2253)

# Overview

Selecting a `date` field brings up a new date widget. User can edit date through the widget or by editing text.
14784678-07b6-11e6-9b6f-a4f140d1b673

# How to Review

1. Visit https://secure.my.jobs/profile/view/
2. Find the module ID of a Education module (hover over its Edit button).
3. Replace `MODULEID` in this URL and visit: 
https://secure.my.jobs/profile/view/module/#/MODULEID?module=Education

# Notes

* Works in IE8
* Uses date format which works with backend (django-remote-forms)
* Introduces node package: https://www.npmjs.com/package/calendar-base
* Introduces node package: https://www.npmjs.com/package/react-onclickout I started with react-onclickoutside, but it's not compatible with our stack.
* Year list defaults to past 50 years

PR #2285, opened 2016-05-04
[PD-1971] - Date Selection Component 

lemniscate613:
# Note

This is an updated version of a previous PR: ( https://github.com/DirectEmployers/MyJobs/pull/2266 )

PD-1971 is this Date Selection Component. 

# Overview

The `DateField` component now more closely resembles original designs:

screen shot 2016-05-04 at 1 48 28 pm

Selecting a `date` field brings up a new date widget. User can edit date through the widget and by editing text:

screen shot 2016-05-04 at 1 48 36 pm

The `calendar-base` module renders the calendar with meaningful dates. If the user enters the date freehand, the `moment` package validates (client side) with leap years taken into consideration:

screen shot 2016-05-04 at 1 50 45 pm

# How to Review

1. Visit https://secure.my.jobs/profile/view/
2. Find the module ID of a Education module (hover over its Edit button).
3. Replace `MODULEID` in this URL and visit: 
https://secure.my.jobs/profile/view/module/#/MODULEID?module=Education

# Notes

Works in IE8 (closing calendar by click out does not, but this is a known degradation for IE8)

Uses date format which works with backend (django-remote-forms)

Year list defaults to past 50 years

Error messages follow new form standard (which is in progress: https://directemployersfoundation.atlassian.net/browse/PD-2076 )

Introduces new node packages:
* https://www.npmjs.com/package/calendar-base
* https://www.npmjs.com/package/moment Gives us excellent free input date validation
* https://www.npmjs.com/package/react-onclickout I started with react-onclickoutside, but it's not compatible with our stack.

MM/DD/YYYY is our new front-end date standard ( @darrint @misterjonscott @koonsies ). Keep in mind it needs to be translated to YYYY-MM-DD before saving to django-remote-forms (see: https://github.com/DirectEmployers/MyJobs/pull/2285/files#diff-f312072fbe26efed8b3fa22b29d3518fR115 )

Fixed some edge cases @koonsies imagined. In short, the user can switch between editing the date as free text and through the calendar widget.

# Further Work

Other client side validation (e.g. email address, etc.) can be handled with this: https://directemployersfoundation.atlassian.net/browse/PD-2373

Other UI improvements if desired (e.g. on tab out close calendar).


gulp/package.json
```diff
@@ -35,19 +35,17 @@
     "jquery-param": "~0.2.0",
     "less": "~1.6.1",
     "lodash-compat": "~3.10.1",
+    "moment": "^2.13.0",
```
lemniscate613: Enables date validation (ISO 8601) like:

`
moment('2011-01-03', 'YYYY-MM-DD', true).isValid() 
`

gulp/src/common/ui/CheckBox.jsx
```diff
@@ -44,14 +51,14 @@ CheckBox.propTypes = {
   /**
    * Should this bad boy focus, all auto like?
    */
-  autoFocus: React.PropTypes.bool,
+  autoFocus: React.PropTypes.string,
```
lemniscate613: My previous PRs played with this. Turns out, `django-remote-forms` sends us this element as a string. We'll let it set the standard for autoFocus.

aspidites:
@lemniscate613 PD-1971 seems to reference reporting, but this looks like it's only implemented for profile units. @fezick, didn't we say that the purpose of this ticket was to update the date components in reporting? If not, we'll need another ticket.

gulp/src/common/ui/DateField.jsx
```diff
@@ -0,0 +1,189 @@
...
+      month: month,
+      day: day,
+      displayCalendar: false,
+      dateInvalid: false,
+    };
+    this.closeCalendar = this.closeCalendar.bind(this);
+  }
+  onDaySelect(day) {
+    const {onChange} = this.props;
+    const fakeEvent = {};
```
lemniscate613: Normal elements have events of a certain structure (i.e. event.target.name). Even though this component (and others like Select) are fake elements, we're keeping that structure and naming convention for consistency.

aspidites:
> https://www.npmjs.com/package/react-onclickout I started with react-onclickoutside, but it's not compatible with our stack.

I think Darrin already introduced this a few PRs back. Can you make sure it isn't included twice?

gulp/src/common/ui/DateField.jsx
```diff
@@ -0,0 +1,189 @@
...
+      name,
+      onChange,
+      required,
+      maxLength,
+      isHidden,
+      value,
+      placeholder} = this.props;
+    const year = parseInt(value.substring(0, 5), 10);
+    const month = parseInt(value.substring(5, 7), 10);
+    const day = parseInt(value.substring(8, 10), 10);
```
lemniscate613: This is a hair more complicated than most props, because we allow the user to edit the value as a string or as an number (by means of the calendar popup).

gulp/src/common/ui/CheckBox.jsx
```diff
@@ -44,14 +51,14 @@ CheckBox.propTypes = {
...
 };
 
 CheckBox.defaultProps = {
   initial: false,
   isHidden: false,
   required: false,
-  autoFocus: false,
+  autoFocus: '',
```
aspidites: I thought we determined that autoFocus was to be a boolean?

gulp/src/common/ui/DateField.jsx
```diff
@@ -0,0 +1,189 @@
...
+    let error;
+    // Date value must match ISO 8601 format
+    if (moment(value, 'YYYY-MM-DD', true).isValid() === false) {
+      error = (
+        <div className="error-text">Date formatting error</div>
+      );
+    }
+
+    return (
+      <ClickOutHandler onClickOut={e => this.closeCalendar(e, this)}>
```
lemniscate613: Our new standard for triggering something when the user clicks __outside__ of a particular event (React makes this more complicated than normal).

gulp/src/common/ui/Datetime.jsx
```diff
@@ -1,72 +0,0 @@
-import React from 'react';
```
lemniscate613: Component named incorrectly. Should be Datefield.

gulp/src/common/ui/DateField.jsx
```diff
@@ -0,0 +1,189 @@
...
+      day = parseInt(value.substring(8, 10), 10);
+    }
+    this.state = {
+      year: year,
+      month: month,
+      day: day,
+      displayCalendar: false,
+      dateInvalid: false,
+    };
+    this.closeCalendar = this.closeCalendar.bind(this);
```
aspidites: I thought that we were using fat arrows to bind functions within the component itself instead of having to bind them in the constructor (eg. `<Component closeCalendar={() => this.closeCalendar()} />`?

gulp/src/common/ui/TextField.jsx
```diff
@@ -4,22 +4,25 @@ import React from 'react';
  * Simple input element with onChange handler
  */
 function TextField(props) {
-  const {name, onChange, required, maxLength, value, isHidden, placeholder, autoFocus} = props;
+  const {name, onChange, required, maxLength, value, isHidden, placeholder, autoFocus, onSelect, className} = props;
```
lemniscate613: I need to pass the className prop to all components in another PR...

gulp/src/common/ui/DateField.jsx
aspidites: Why do we need a fake event, again?

gulp/src/common/ui/CheckBox.jsx
lemniscate613: Good catch. See this: https://github.com/DirectEmployers/MyJobs/pull/2285#discussion_r62086135

gulp/src/myprofile/main.jsx
```diff
@@ -60,8 +60,29 @@ class Module extends React.Component {
     let value;
     if (event.target.type === 'checkbox') {
       value = event.target.checked;
-    } else if (event.target.type === 'select-one') {
-      value = event.target.value;
+    } else if (event.target.type === 'calendar-month') {
```
aspidites: Isn't moment taking care of the string munging for us? Seems like we should be able to use it to parse the string and convert it to a date object instead of having to do it ourselves here.

lemniscate613:
@aspidites I named this PR incorrectly. This is for the basic Datefield component ( https://directemployersfoundation.atlassian.net/browse/PD-2294 ). Folding it into Reporting will be another PR (after this one is merged).

gulp/src/common/ui/CheckBox.jsx
aspidites: I see.

lemniscate613:
@aspidites Good memory RE: react-onclickout. I discovered it awhile ago and it ended up being useful in reporting. They are using it, but it's not part of QC yet.

gulp/src/common/ui/DateField.jsx
lemniscate613: Whenever a user edits TextField, Textarea, DateField, Select, or CheckBox, an event is passed up through the onChange() callback. Calling it an "event" and giving it the same structure makes the payload consistent -- whether its a standard HTML form field or custom.

aspidites:
Functions properly as far as I can tell. A couple of nits:
- is it expected that the calendar pop up while I'm typing?
- the displayed format is yyyy-mm-dd, but if you look elsewhere (like [here](https://secure.my.jobs/prm/view/), we use mm/dd/yyyy

gulp/src/myprofile/main.jsx
lemniscate613: Excellent point. Forgot to go back and change this after adding moment.

gulp/src/common/ui/DateField.jsx
aspidites: Maybe you can walk me through this some time. If we aren't needing to raise an event, it seems we should be able to just not pass an "event". I'm sure I'm misunderstanding the use case though, which is why I requested a walk through. I could learn something new.

lemniscate613:
@aspidites Your first point was an annoying UX bug that I just fixed. As for the other point, yyyy-mm-dd is our new standard going forward with django-remote-forms.

JLMcLaugh:
@lemniscate613 we discussed that yyyy-mm-dd is the standard ISO 8601 format that we should use consistently in our code (I think we were originally passing mm/dd/yyyy to the backend). There was some back and forth about where the transformation should take place, but I thought you, @darrint, and I agreed that mm/dd/yyyy was more intuitive for displaying dates and free text input.

lemniscate613:
@JLMcLaugh @aspidites You're absolutely right. I totally forgot about that. Here you go: https://github.com/DirectEmployers/MyJobs/pull/2285/commits/1fedd6d8b9e8263675411d77a91656a914e7227a

gulp/src/myprofile/main.jsx
```diff
@@ -113,15 +113,17 @@ class Module extends React.Component {
...
   }
   processFormContents(formContents) {
-    for (const formItem in formContents) {
-      // Dates must be of form YYYY-MM-DD before POSTing
-      const valueIsDate = moment(formContents[formItem], 'YYYY/MM/DD', true).isValid();
-      if (valueIsDate === true) {
-        let dateValue = formContents[formItem];
-        dateValue = dateValue.replace(/\//g, '-');
-        formContents[formItem] = dateValue;
+    if (formContents.hasOwnProperty(formItem)) {
```
lemniscate613: We need to wrap for loops in hasOwnProperty thingies.

JLMcLaugh:
@lemniscate613 It was mostly the order we were talking about, MM/DD/YYYY.

lemniscate613:
D'oh! One second..

gulp/src/myprofile/main.jsx
tdphillips: formItem isn't defined yet, is it?

koonsies:
I found a bit of interesting behavior with the date picker. If you start typing a date, then select something from the popup, the input isn't formatted in the box.

For instance, I highlight the box and start typing "04".. After I type that, I notice the box opened, I go ahead and select the year (since month automatically populated from the "04" input) and select the day. The end result in 04182016 instead of 04/18/2016, thereby making the invalid input flag appear.

There's further weird behavior when I highlight the input and hit delete.. Then, the value from the bottom most box fills whatever date I was in. In this case, end date had a date of 05/26/2016. After messing around a bit with start date, I highlighted what I had entered and hit backspace. 05/26/2016 appeared in the text box.

lemniscate613:
@koonsies Good points. I've got some ideas...

koonsies:
In the date format error message, you should probably include what format you're looking for.

lemniscate613:
I addressed all of your points @koonsies except closing the calendar on tab. It can be addressed in a future "UI improvements" ticket.

koonsies:
I tested this and it appears to deal with a majority of the problems I found. 

That said, the timing on when the values in the text box are updated feels a bit off to me. I feel like the values in the text box should be updated on selecting a "day". As it stands, changing the month will cause the value in the text box to become 

<month selected>/<current day of the week>/<year selected>

This is a minor issue, but I see the user selecting the date in the following order.. Month.. Year.. and finally clicking a "Day" to complete the date. It feels strange to me that the value in the textbox be updated on every change in the widget.

Just my two cents. If this thought isn't shared, I give a :+1: to what I've seen thus far!

lemniscate613:
@koonsies There was an issue balancing 1) enabling user to edit text freely 2) parsing that string to be `int` input for the calendar and 3) real time validation. The solution I arrived at was reseting the date to the current date if, upon opening the calendar, the date was malformed.

I'd like future UX improvements to go into another ticket. That way, I can start folding this (even if it could be more advanced) into Reporting.

PR #2286, opened 2016-05-06
[PD-1839] Apply reusable table component

koonsies:
# Overview

This PR covers the application of a React table component that can be used in various applications.

The application is on the Non User Outreach project as a list of various Non User Outreach records. Application.

## Setup

To test this table, you first need to run migrations on mypartners

Afterward, you will need to execute the following script to create the outreach records needed. NOTE: You do not need to use Facebook as the company. Change that bit with any company you currently are a user/admin of.

```
from datetime import datetime
from factory import fuzzy
from seo.models import Company
from mypartners.models import (OutreachRecord, OutreachEmailAddress,
                               OutreachWorkflowState)

company = Company.objects.get(name="Facebook")
try:
    outreach_email = OutreachEmailAddress.objects.get(company=company,
                                                      email="testemail")
except:
    outreach_email = OutreachEmailAddress(company=company,
                                          email="testemail")
    outreach_email.save()

try:
    workflow = OutreachWorkflowState.objects.get(state="Reviewed")
except:
    workflow = OutreachWorkflowState(state="Reviewed")
    workflow.save()

i = 0
while i < 100:
    i += 1
    OutreachRecord(date_added=datetime.now(),
                   outreach_email=outreach_email,
                   from_email=str(fuzzy.FuzzyText(length=10).fuzz()) + "@example.com",
                   email_body=fuzzy.FuzzyText(length=200).fuzz(),
                   current_workflow_state=workflow
                   ).save()

```

## Testing

* Run npm install
* Run gulp/dev server build
* Run secure locally
* Navigate to https://secure.my.jobs/prm/view/nonuseroutreach (or whatever url you use to access your secure instance)
* Click Outreach Records
* Verify table works properly.. Pagination is turned off for IE8 as there is an issue with the module we are using.


mypartners/migrations/0031_add_api_activity_rename_outreach_record_table.py
```diff
@@ -0,0 +1,616 @@
...
+
+        ###
+        #Create activity for new API
+        ###
+
+        nuo = AppAccess.objects.get(name='NUO')
+        Activity.objects.create(
+            app_access=nuo,
+            name="read outreach records",
+            description="Read outreach records")
```
aspidites: Historically, I've been making description different from name:
2016-05-06-150016_670x529_scrot

gulp/src/nonuseroutreach/components/OutreachRecordButton.jsx
```diff
@@ -0,0 +1,23 @@
+import React, {Component, PropTypes} from 'react';
+import Button from 'react-bootstrap/lib/Button';
+
+
+// menu link to inbox management app screen
+export class OutreachRecordButton extends Component {
+  handleClick() {
```
aspidites: Not sure how much we care, but I think we've started in other apps using onFoo for events rather than handleFoo.

gulp/src/nonuseroutreach/nonuseroutreachEngine.js
```diff
@@ -54,3 +54,13 @@ export class InboxManagement {
     return result;
   }
 }
+
+export class OutreachRecordManagement {
```
aspidites: Should one of these exports be default? To be honest, I'm not even clear on when we decide we need to use a default export, especially as it has implications for import statements.

mypartners/migrations/0031_add_api_activity_rename_outreach_record_table.py
aspidites: Also, all other activities are singular, so 'read outreach record' rather than 'read outreach records' for the activity name.

templates/base-react.html
```diff
@@ -22,6 +22,7 @@
         {% compress css %}
             <link rel="stylesheet" href="{{ STATIC_URL }}bootstrap-3/css/bootstrap.min.css" rel="stylesheet" type="text/css">
             <link rel="stylesheet" href="{{ STATIC_URL }}custom.css" rel="stylesheet" type="text/css">
```
aspidites: Mind fixing the other two since you're in the file already(bootstrap 3 and custom)? By fix, I mean convert them to `{% static %}`.

aspidites:
FYI, the try/except could be replaced by a call to `get_or_create` which will accomplish the same thing. Future reference.

aspidites:
Seems functional. I tested IE8 a while back. I'll test again after my walk.

gulp/src/nonuseroutreach/components/OutreachRecordButton.jsx
koonsies: I'm a bit worried about making this onClick, as it would share a name with the event itself. Not sure if we have similar functions elsewhere.

mypartners/migrations/0031_add_api_activity_rename_outreach_record_table.py
koonsies: Fixed

gulp/src/nonuseroutreach/components/OutreachRecordButton.jsx
aspidites: Ah. I think we've just been using arbitrary names. Eg, I had a click handler I was messing with that was named onReorder and linked to onChange. Not a big deal. Especially if we end up using redux or something.

aspidites:
Unless someone else is looking at this, I'll be good with it once tests pass and i take a peek at the results of the updated migration.

aspidites:
Even without pagination, IE8 hates this. Open up a console to figure out what's up and my entire VM just freezes. Is the standard that "as long as the page loads without errors and is functional we're good, or simply "as long as the page loads and is functional we're good"? @fezick  @JLMcLaugh, @haliscooper ?

koonsies:
Weird. I don't get that type of behavior. A few errors in my console, but nothing outrageous.

koonsies:
Looks like my Jasmine test is malformed as well. Gonna fix that real quick

PR #2287, opened 2016-05-09
[PD-2366] Improved Drag & Click Behavior for Export Columns

aspidites:
# Description
If you go to reporting on qc.secure.my.jobs and visit the export screen, you'll see that you can only enable/disable a column by clicking on the small checkbox. This PR improves on that by allowing both the checkbox and the text next to the checkbox to be clicked. After a few days of research, I wasn't able to find anything with jQuery-UI's drag distance behavior, which would have allowed us to be able to click on the entire row as well as drag it. The only alternative I'm aware of for what I've done is to allow the entire row to be clickable, but only allow dragging to happen on the handle icon. From speaking with @fezick, that seemed to be undesirable, though his mind may have changed on that since then.

# Testing
You'll need to run `npm install` before testing this. I didn't mess with the backend, so it shouldn't be necessary to `reset_dynamic_data`.

I also addressed @koonsies  comments on another PR with respect to the fact that you could export a report without selecting any columns, which resulted in a report which had all columns selected. I've addressed this by simply disabling the export button when no columns are selected.

Finally, I made it so that manually selecting all columns also selects the "Select All" checkbox. In other words, the "Select All" checkbox properly reflects the state of the individual checkboxes.

~~# Caveat~~
~~This folds in some work misterjonscott did, such as including handles. I'm not sure how far along that work is, which is why I didn't wait on that to be merged. I didn't ignore it entirely because it had implications on my work (such as where at in the DOM the drag handle lived, and how style rules worked).~~ I went ahead and removed Jon's changes for now to clean up this PR.


aspidites:
Still need to write proptype documentation, rearrange a couple of them, and implement the select all behavior. Closing until the description accurately reflects the state of the branch.

gulp/src/common/ui/PopMenu.jsx
```diff
@@ -1,36 +1,25 @@
 import React, {Component, PropTypes} from 'react';
 import ClickOutCompat from 'common/ui/ClickOutCompat';
 
-
```
aspidites: Part of @misterjonscott 's changes.

gulp/src/common/ui/SvgIcon.jsx
```diff
@@ -0,0 +1,37 @@
+import React, {PropTypes} from 'react';
```
aspidites: Also part of @misterjonscott's changes.

gulp/src/reporting/SortableList.jsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  render() {
+    const {items, onReorder, onChange} = this.props;
+    return (
+      <Sortable
+        className="my-list"
+        options={{
+          ghostClass: 'placeholder',
+          delay: 'ontouchstart' in window ? 150 : 0,
```
aspidites: Generally reliable way to detect mobile. Here, we only want a delay before dragging starts on mobile.

static/custom.css
```diff
@@ -2194,12 +2194,16 @@ h3.bottom-bordered {
     border-radius: 3px;
 }
 
+.reporting.sidebar ul li .pop-menu ul li {
```
aspidites: More @misterjonscott things.

gulp/src/common/ui/SvgIcon.jsx
aspidites: Comments like these are no longer valid, so may be ignored.

misterjonscott:
Looks great, tested and working.

PR #2289, opened 2016-05-09
A few tweaks.

aspidites:
1. Disabled the "Comm records per month per partner" report, as it wasn't functioning properly and was meant to be disabled anyway.
2. Added "Dynamic Reports" to Employer menu, but only accessible by superusers. Got tired of URL munging, and it sounded like this was close to release anyway.
3. Removed `companyName` from react template, as we aren't using it.
4. Beta Menu - React profile units, new reporting, and nonuser outreach are now available in a beta menu which is restricted to super users with an @apps.directemployers.org email address. Should be easier to test react things now before official releases.
# Testing
You should see "Dynamic Reports" in the Employers menu. You should also notice that no aggregated reports are available to be run, after first doing a `./manage.py reset_dynamic_data'.


tdphillips:
Merge conflicts.

PR #2292, opened 2016-05-09
[PD-2333] Report Clone

darrint:
## Review

* `myreports/datasources/base.py` defines a new method for filling out a previously used filter with UI data needed to display it.
* other files in `myreports/datasources` fill out that method on existing datasources.
* UI code uses the state portion of browser history to retrieve/store the prefilled filter. (Redux will improve this in the future but we might still use the history mechanism this way.)
* The two pane select control in reporting needed some adjustments.
* I'm still staying clear of date controls.

## Testing

* Set some filters and run a report.
* Navigate somewhere else.
* Click the menu on the previously run report and choose clone.


aspidites:
Cloned reports should have a name of "Clone of <name of cloned report>". See old reporting, for example.

darrint:
Fixed. Not sure why I didn't get an Eslint warning on that though. :-/

tdphillips:
Maybe I'm blind but I'm not seeing where the report name is being modified.

darrint:
`ReportList.jsx`

PR #2295, opened 2016-05-10
Add regenerate [PD-2332]

tdphillips:


tdphillips:
There's no in

tdphillips:
There's no indicator that things happened outside of the popup dismissing.

aspidites:
Depending on which definition of refresh you use (old reporting has been told to do one or the other at different times during development), regenerate may or may not be functioning as intended. If You create a contact report, then retcon a new contact and refresh that report, the counts are the same as the old report, but creating a new contact report shows the new counts.

Example: I started with 162 rows. I added a new record for a new  contact and refreshed the report. Still showed 162 rows. Creating a new contact record with the same parameters showed 163 rows.

tdphillips:
Looking at DynamicReport.regenerate

haliscooper:
It should have the newer data.

tdphillips:
It's magical what saving a modified object will do.

PR #2297, opened 2016-05-10
PD-2382 - Added function to scroll report to top, when a report is run.

misterjonscott:
# It's nothing complex.
I added a loop to scroll over 250ms so the user sees they're being taken back to the top, and not to a new page.

### DIFF
--- a/gulp/src/reporting/SetUpReport.jsx
+++ b/gulp/src/reporting/SetUpReport.jsx
@@ -90,7 +90,7 @@ export default class SetUpReport extends Component {
 
   scrollToTop() {
     const scrollLoop = setInterval(() => {
-      if (document.body.scrollTop <= 100) {
+      if (document.body.scrollTop <= 0) {
         clearInterval(scrollLoop);
       }
       document.body.scrollTop = document.body.scrollTop + ((0 - document.body.scrollTop) / 250 * 10);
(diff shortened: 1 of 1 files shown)


aspidites:
Would it be handy to have this in common so we can use it elsewhere? I suspect User Management would be a good candidate for this next.

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -88,6 +88,15 @@ export default class SetUpReport extends Component {
     }
   }
 
+  scrollToTop() {
```
tdphillips: Should we call something "scrollToTop" if it doesn't scroll to the top?

aspidites: May be nice to parameterize this by the amount to scroll on each interval as well as the point at which to stop.

misterjonscott:
@aspidites This could certainly be expanded to take inputs and be re-used on other pages, but in the interest of time, and the context of the ticket, this seems appropriate for now.  If anyone feels we should make this into a full-featured function in the common tools, I'll close this request and focus on those changes.

gulp/src/reporting/SetUpReport.jsx
misterjonscott: @tdphillips I changed the "top" back to 0, assuming that's what you meant with your comment.  Thanks for catching that.

PR #2299, opened 2016-05-10
scrollTo

aspidites:
Didn't notice until after I merged @misterjonscott's work that it didn't work properly for Firefox. This corrects that, as well as moved it to a common function with documentation and parametrize input.


gulp/src/common/ui/SvgIcon.jsx
```diff
@@ -1,5 +1,5 @@
 import React, {PropTypes} from 'react';
-import {isIE8} from 'common/browserSpecific.js';
+import {isIE8} from 'common/dom.js';
```
misterjonscott: Do you care that you included the extension, here?

aspidites: Fixed.

PR #2301, opened 2016-05-11
[PD-1971] fix bug

lemniscate613:
Another simple fix.

If a date field is empty, it defaults to today. I was checking the wrong thing, so it was always "empty" and therefor always defaulting to today (and overwriting data).

### DIFF
--- a/gulp/src/myprofile/main.jsx
+++ b/gulp/src/myprofile/main.jsx
@@ -253,7 +253,7 @@ class Module extends React.Component {
           let month;
           let day;
           // If date value is empty use today's date
-          if ((!formContents[item]) || (formContents[item] === '')) {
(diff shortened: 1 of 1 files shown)


aspidites:
I'm not able to reproduce the erroneous behavior. Empty field defaults to today, but choosing something and tabbing to the date field opens the widget with the correct date in it. Unless you are referring to form submission?

lemniscate613:
@aspidites What *used* to happen was: you edit an existing module, and all dates are overwritten with the current date.

Now, when you edit an existing module, the date value is as it should be.

Essentially, this used to always be true: https://github.com/DirectEmployers/MyJobs/pull/2301/files Because formContents[item] didn't exist yet.

PR #2306, opened 2016-05-11
[PD-2351] Add "live" filtering to the partner and contact fields

darrint:
## Test

* Create a Contacts or Communication Record report
* Filter by some field.
* Notice how the partners/contacts update as you make changes.
* Try to break it with history navigation and cloning.

## Review

* If any filter changes, the fields "subscribed" to the filter change notification call help again.
* Two pane selects update the filter repeatedly when [de]selecting many items at once. That's why the debounce is there.

### DIFF
--- a/gulp/src/reporting/DynamicReportApp.jsx
+++ b/gulp/src/reporting/DynamicReportApp.jsx
@@ -14,7 +14,7 @@ export class DynamicReportApp extends Component {
 
   componentDidMount() {
     const {reportFinder} = this.props;
-    this.callbackRef =
+    this.unsubscribeToNewReports =
       reportFinder.subscribeToNewReports(
           (reportId, runningReport) =>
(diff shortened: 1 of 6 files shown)


aspidites:
That the contact are partners multi-selects are co-dependent leads to inconsistent behavior, as I showed you. That is, you can select a contact and it then filters the partner list to only those partners with that contact. If I then deselect a contact, the associated partners are removed. 

While you could certainly make each box filter when either is changed, you could probably reclaim some sanity by only having contacts depend on partners, but partners not depend on contacts. That is, partners would determine which contacts you can choose from, but selecting a contact wouldn't then refilter the partner list. Thus, partners would be picked explicitly by the user, and contacts to choose from would be governed by those chosen partners.

I know I repeated myself a few times there, but I'm hoping that in stating things a few different ways one of them might click with you more than the others.

PR #2310, opened 2016-05-12
Add max length to report name and enforce it on clone [PD-2370]

tdphillips:
I verified this in all window widths that I could, including down to my browser's minimum of 475px. I'm open to suggestions if that is not adequate.

### DIFF
--- a/gulp/src/reporting/SetUpReport.jsx
+++ b/gulp/src/reporting/SetUpReport.jsx
@@ -79,7 +79,8 @@ export default class SetUpReport extends Component {
   }
 
   onReportNameChanged(reportName) {
-    this.setState({reportName});
+    const {maxNameLength} = this.props;
+    this.setState({reportName: reportName.substring(0, maxNameLength)});
   }
 
   handleHistory(something, loc) {
@@ -178,6 +179,7 @@ export default class SetUpReport extends Component {
       category: reportType,
       dataSet: dataType,
     } = this.props.location.query;
+    const {maxNameLength} = this.props;
     const {
       loading,
       filter,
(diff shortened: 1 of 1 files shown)


gulp/src/reporting/SetUpReport.jsx
```diff
@@ -338,4 +341,9 @@ SetUpReport.propTypes = {
       reportDataId: PropTypes.string,
     }).isRequired,
   }).isRequired,
+  maxNameLength: PropTypes.number,
+};
+
+SetUpReport.defaultProps = {
+  maxNameLength: 24,
```
tdphillips: Because magic numbers are :hankey:

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -79,7 +79,8 @@ export default class SetUpReport extends Component {
   }
 
   onReportNameChanged(reportName) {
-    this.setState({reportName});
+    const {maxNameLength} = this.props;
+    this.setState({reportName: reportName.substring(0, maxNameLength)});
```
tdphillips: Ugly? Bad? Annoying? Thoughts?

misterjonscott: So we're just going to clip it if it's too long?

misterjonscott:
@tdphillips Looks good, but I'll wait to hear back about the question above.

gulp/src/reporting/SetUpReport.jsx
tdphillips: I think so.

[Here's Jason's comment on the matter.](https://directemployersfoundation.atlassian.net/browse/PD-2370?focusedCommentId=218838&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-218838) Let me know if you disagree.

PR #2311, opened 2016-05-12
[PD-2294] Add DateField component to Reporting

lemniscate613:
# This PR:

* adds DateField to reporting
* enables the DateField to display errors as props from above ( https://github.com/DirectEmployers/MyJobs/pull/2311/files#diff-c5b207334a4c11ac32df04a2b0ba657cR177 ) which enables validation like...
* adds a new validation to confirm end date is after begin date ( https://github.com/DirectEmployers/MyJobs/pull/2311/files#diff-d50ed71fb5078e252db0450db9459aedR54 ).

# Future

I'll create another ticket to update export so it takes this date range into account.

I *can* create another ticket to improve the implementation of this component, if that's important. For example, enable free-text editing (and not just selecting the date fromt he popup).

# Old
screen shot 2016-05-12 at 11 44 32 am

Which opens:

screen shot 2016-05-12 at 11 44 38 am

# New

screen shot 2016-05-12 at 2 20 42 pm

Which opens:

screen shot 2016-05-12 at 2 21 01 pm

### DIFF
--- a/gulp/src/common/ui/DateField.jsx
+++ b/gulp/src/common/ui/DateField.jsx
@@ -176,7 +176,7 @@ class DateField extends React.Component {
     }
 
     let errorComponent;
-    if (error) {
+    if (errorMessage) {
       errorComponent = (
         <div className="error-text">{errorMessage}</div>
       );
--- a/gulp/src/common/ui/DateField.jsx
+++ b/gulp/src/common/ui/DateField.jsx
@@ -169,7 +169,7 @@ class DateField extends React.Component {
       month = momentObject.month();
       year = momentObject.year();
     } else {
-      errorMessage = "Date error. Must be of format: MM/DD/YYYY";
+      errorMessage = 'Date error. Must be of format: MM/DD/YYYY';
       day = 0;
       month = 1;
       year = 2000;
@@ -178,10 +178,10 @@ class DateField extends React.Component {
     let errorComponent;
     if (error) {
       errorComponent = (
-        <div className="error-text">{error}</div>
+        <div className="error-text">{errorMessage}</div>
       );
     }
-    
+
     let calendar;
     if (this.state.displayCalendar) {
       calendar = (<div className="input-group datepicker-dropdown dropdown-menu"
--- a/gulp/src/common/ui/DateField.jsx
+++ b/gulp/src/common/ui/DateField.jsx
@@ -148,13 +148,20 @@ class DateField extends React.Component {
       maxLength,
       isHidden,
       value,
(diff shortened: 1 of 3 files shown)


gulp/src/common/ui/DateField.jsx
```diff
@@ -148,27 +148,40 @@ class DateField extends React.Component {
...
     if (moment(value, 'MM/DD/YYYY', true).isValid()) {
       momentObject = moment(value, 'MM/DD/YYYY');
       day = momentObject.date();
       month = momentObject.month();
       year = momentObject.year();
     } else {
-      error = (
-        <div className="error-text">Date error. Must be of format: MM/DD/YYYY</div>
-      );
+      errorMessage = 'Date error. Must be of format: MM/DD/YYYY';
```
tdphillips: Given the `if (error)` check on line 179, is it possible to reach this point if `error` is falsy? If so, we'll never display this error.

lemniscate613: @tdphillips thank you for pointing that out. I was checking the wrong value: https://github.com/DirectEmployers/MyJobs/pull/2311/commits/ce6555b6857e52e60ed81c28d6d9f193800e84bc

gulp/src/common/ui/DateField.jsx
```diff
@@ -176,7 +176,7 @@ class DateField extends React.Component {
     }
 
     let errorComponent;
-    if (error) {
+    if (errorMessage) {
```
lemniscate613: For example, within this component we are checking that the date is valid (MM/DD/YYYY). Reporting code which uses the `DateField` component has other validation to confirm one ending date is *after* another beginning date.

aspidites:
Validation isn't working:
2016-05-12-140510_608x96_scrot

Note To date is before From date.

aspidites:
We can probably fix it next sprint, but I do remember AJ mentioning a few weeks ago that he liked being able to type in dates. Just as a future reference, not expecting that for this PR.

aspidites:
Correction. Dates are fine. Didn't pay attention to the year.

lemniscate613:
@aspidites @tdphillips I made them side by side, which is a little improvement.

aspidites:
OK. Validation for the field works correctly, but the run report button isn't being disabled. I think there is a ticket somewhere tracking that though, so I'm fine ignoring it for the moment. "Date range" should probably be "Date Range" though. I do like the update though.

tdphillips:
I'm having issues with the path used for calendar.[svg|png] but no one else is. Let's see what happens.

PR #2312, opened 2016-05-12
City/State now dependent on each other.

aspidites:
There isn't an associated ticket for this. Just an improvement I decided to work on since I was out of tickets. Before, all states were shown in the state dropdown in reporting, though selecting something would update cities. Now, only states for which records can be located will show up in the state drop down. Additionally, changing a city impacts which states are shown in the select. ~~Finally, if a city filter results in a single choice, that choice will automatically be selected from the state drop down.~~ Resulted in some dodgy behavior where a users choice could be changed, so I removed that bit.


gulp/src/reporting/wizard/WizardFilterCityState.jsx
```diff
@@ -33,41 +34,41 @@ export class WizardFilterCityState extends Component {
...
       return (
         <span>
           <Select
             onChange={e =>
               this.updateField('state', e.target.value)}
             name=""
-            value={getDisplayForValue(states, currentLocation.state)}
-            choices={this.state.states}
+            value={getDisplayForValue(regions, currentLocation.state)}
+            choices={this.state.regions}
```
tdphillips: I admittedly don't know as much about this as others but is there a reason we can't use `regions` directly here?

aspidites: Partial refactor. I didn't originally destructure state when I wrote this.

misterjonscott:
This is awesome.

PR #2313, opened 2016-05-12
Fix export history

darrint:
* The UI was not updating when switching between export screens from the export screen.
* Fixed a missing key warning.

### DIFF
--- a/gulp/src/reporting/ExportReport.jsx
+++ b/gulp/src/reporting/ExportReport.jsx
@@ -23,7 +23,13 @@ export default class ExportReport extends Component {
   }
 
   componentDidMount() {
-    this.loadData();
+    const {history} = this.props;
+    this.historyUnlisten = (
+      history.listen((something, loc) => this.handleHistory(something, loc)));
+  }
+
+  componentWillUnmount() {
+    this.historyUnlisten();
   }
 
   onReorder(order) {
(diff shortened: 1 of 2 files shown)


gulp/src/reporting/ExportReport.jsx
```diff
@@ -237,6 +250,7 @@ export default class ExportReport extends Component {
 }
 
 ExportReport.propTypes = {
+  history: PropTypes.object.isRequired,
```
aspidites: Don't we usually use shapes for objects?

gulp/src/reporting/ExportReport.jsx
```diff
@@ -70,6 +76,13 @@ export default class ExportReport extends Component {
     });
   }
 
+  handleHistory(something, loc) {
+    const lastComponent = loc.components[loc.components.length - 1];
+    if (lastComponent === ExportReport) {
```
aspidites: Why does ExportReport require special logic but the other options don't?

gulp/src/reporting/ExportReport.jsx
darrint: Not for this one. It's an api being made available by the router.

gulp/src/reporting/ExportReport.jsx
darrint: SetUpReport needs it too.

I'm surprised they don't make a bigger deal about this in the `react-router` docs. It seems to be the only reliable way to respond to URL changes on already mounted components.

gulp/src/reporting/ExportReport.jsx
aspidites: Ah. Doesn't look like shape allows for extensible records, which would allow us to at least document which portions of the api we were using. :-( Can we switch to PureScript now?

gulp/src/reporting/ExportReport.jsx
aspidites: Yikes. Passing for now, I may be interested in revisiting this issue in the future though.

PR #2318, opened 2016-05-13
Fix date ranges

darrint:
## Test

* Start a new report. Defaults should match old system.
* Set date ranges in reports. See that partners/contacts live filter.
* Run reports. See that data is filtered.
* Clone reports. Date ranges should be preserved.
* History should work with date ranges.

### DIFF
--- a/gulp/src/reporting/SetUpReport.jsx
+++ b/gulp/src/reporting/SetUpReport.jsx
@@ -20,7 +20,7 @@ export default class SetUpReport extends Component {
     super();
     this.state = {
       loading: true,
-      filter: {},
+      filter: null,
       reportName: null,
       reportingTypes: [],
       reportTypes: [],
@@ -30,6 +30,7 @@ export default class SetUpReport extends Component {
 
   componentDidMount() {
     const {reportFinder, history} = this.props;
+    this.mounted = true;
     this.unsubscribeToMenuChoices = reportFinder.subscribeToMenuChoices(
         (...choices) => this.onMenuChanged(...choices));
     this.historyUnlisten = (
@@ -49,6 +50,7 @@ export default class SetUpReport extends Component {
   componentWillUnmount() {
     this.historyUnlisten();
     this.unsubscribeToMenuChoices();
+    this.mounted = false;
   }
 
   onIntentionChange(intention) {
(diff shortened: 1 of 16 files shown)


gulp/src/reporting/SetUpReport.jsx
```diff
@@ -227,23 +223,8 @@ export default class SetUpReport extends Component {
...
-            const day = now.getDate() < 10 ? '0' + now.getDate() : now.getDate();
-            const today = month + '/' + day + '/' + year;
-
-            begin = '01/01/2014';
-            end = today;
-          } else {
-            begin = reportConfig.currentFilter[col.filter][0];
-            end = reportConfig.currentFilter[col.filter][1];
-          }
+          const begin = (reportConfig.currentFilter[col.filter] || [])[0];
```
tdphillips: [][0] is an IndexError.

tdphillips:
Should I be seeing default dates in the two inputs like I do on old reporting? I'm not.

gulp/src/reporting/wizard/WizardFilterDateRange.jsx
```diff
@@ -10,24 +10,24 @@ export class WizardFilterDateRange extends Component {
 
     updateField(event, field) {
       let finalValue;
+      const today = moment().format('MM/DD/YYYY');
+      const rawExistingDate = (field === 'begin') ? this.props.begin : this.props.end;
```
aspidites:
```javascript
const rawExistingDate = this.props[field];
```

myreports/datasources/jsondriver.py
```diff
@@ -102,5 +118,11 @@ def encode_single_filter_interface(self, column_config):
         return {
             'interface_type': column_config.filter_interface,
             'filter': column_config.column,
-            'display': column_config.filter_display
+            'display': column_config.filter_display,
         }
+
+
+class CustomEncoder(json.JSONEncoder):
+    def default(self, obj):
```
aspidites: Probably need to call the default implementation of default in case the object isn't a datetime.

gulp/src/reporting/SetUpReport.jsx
darrint: It returns undefined.

tdphillips: :fearful: 

Thought I was reading Python for some reason.

myreports/datasources/jsondriver.py
darrint: I added it. I didn't need it though. I'm running all kinds of data through this. I double checked the docs and they have one so here it is.

darrint:
@tdphillips The only way I know to get a blank in the date fields is to carry in some previous state (from history?) or clone a previous report that never had dates. If you create a new report there should be default values. Can you elaborate on how you are doing that?

tdphillips:
No clue. I ran gulp build before I said that and it didn't work. I checked again and it does. ¯\\\_(ツ)_/¯

PR #2319, opened 2016-05-13
PD-2376 - Removed unnecessary class from date widget

misterjonscott:
This class had no associated styles, but was picking up styles from the new date range component.

### DIFF
--- a/templates/mypartners/includes/date_widget.html
+++ b/templates/mypartners/includes/date_widget.html
@@ -6,7 +6,7 @@
 <div class="partner-tag custom button disabled-tag" title="Activity within specific date range">Custom</div>
 <div class="date-picker-widget">
     <input class="datepicker picker-left" id="activity-start-date" type="text" placeholder="Activity Start Date" />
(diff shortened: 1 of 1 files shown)


tdphillips:
It may not have associated styles but there are several references in record_overview.167-10.js, prm.170-6.js, reporting.js, and custom.166-06.js.

misterjonscott:
@tdphillips 
record_overview.167-10.js and  prm.170-6.js both display the calendar when you click on a field with the class "datepicker", which is unaffected by this, as the fields still contain this class.

reporting.js uses the datepicker class the same way as above, and is unbroken by this change.

custom.166-06.js seems to be jQuery 1.7.1 and jQuery UI 1.8.16 jammed together and renamed.  I'm not sure why this exists, but it's not going to affect this case, either.

tdphillips:
Didn't notice that this was only being applied to that span. Never mind.

misterjonscott:
No worries.  ;)  That's why I researched it.

misterjonscott:
@tdphillips ^^

PR #2321, opened 2016-05-13
Started using native events for report.

aspidites:
`stopPropagation()` wasn't working because React doesn't use native events. We use this and `preventDefault()` a few other places, but I'd prefer to fix those on an as-need basis as I'm not entirely sure what other implications are to be had by using this method. 

This is using react-native-listener.

Reference fo this implementation: https://medium.com/@ericclemmons/react-event-preventdefault-78c28c950e46#.lq0ypqvz4

Running a report should now only happen if you click the run button or it's selected and you press enter.

### DIFF
--- a/gulp/npm-shrinkwrap.json
+++ b/gulp/npm-shrinkwrap.json
@@ -4,2412 +4,1477 @@
   "dependencies": {
     "accepts": {
       "version": "1.3.2",
-      "from": "accepts@1.3.2",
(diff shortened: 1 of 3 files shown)


darrint:
I'm suspicious that the real problem here is we don't actually need those form tags. At least the author of that article sometimes wanted to submit a form. We _never_ want to do that. (Still merging though.)

PR #2323, opened 2016-05-13
[PD-2387] Add refresh indicator for reports

koonsies:
## Overview

This PR simply adds a refresh spinner to the ReportList when a report is refreshed

## Testing

* Do the dev-server/gulp shuffle.
* Head to dynamic reporting
* Click an existing report in the sidebar
* Click "refresh"

_NOTE: Locally, refresh happens very quickly for me. Adding a sleep to myreports->views->refresh_report helps to see the spinner_

### DIFF
--- a/gulp/src/common/ui/PopMenu.jsx
+++ b/gulp/src/common/ui/PopMenu.jsx
@@ -31,7 +31,7 @@ export default class PopMenu extends Component {
 
   render() {
     const {toggleMenu, isMenuPending} = this.props;
-    if (isMenuPending){
+    if (isMenuPending) {
       return (
         <div style={{position: 'relative'}}>
           <span className="report-loader list-loader">
--- a/gulp/src/common/ui/PopMenu.jsx
+++ b/gulp/src/common/ui/PopMenu.jsx
@@ -31,11 +31,20 @@ export default class PopMenu extends Component {
 
   render() {
     const {toggleMenu, isMenuPending} = this.props;
+    if (isMenuPending){
+      return (
+        <div style={{position: 'relative'}}>
+          <span className="report-loader list-loader">
+          </span>
+          {this.menuContents()}
+        </div>
+      );
+    }
     return (
       <div style={{position: 'relative'}}>
         <span
           onClick={(e) => toggleMenu(e)}
-          className={isMenuPending ? "menuPending" : "menuEllipses"}>
+          className="menuEllipses">
           &hellip;
         </span>
         {this.menuContents()}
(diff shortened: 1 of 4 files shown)


gulp/src/common/ui/PopMenu.jsx
```diff
@@ -5,19 +5,18 @@ export default class PopMenu extends Component {
   handleItemClick(item, e) {
     e.preventDefault();
     item.onSelect();
-    this.setState({isMenuActive: false});
```
koonsies: I never saw that this was used.

gulp/src/common/ui/PopMenu.jsx
```diff
@@ -5,19 +5,18 @@ export default class PopMenu extends Component {
...
+    const {options, isMenuActive, closeAllPopups, isMenuPending} = this.props;
     const dropdownItems = options.map((item, index)=> {
       return (
         <li key={index} onClick={e => this.handleItemClick(item, e)}>
           {item.display}
         </li>
       );
     });
-    if (isMenuActive === true) {
+    if (isMenuActive === true & isMenuPending === false) {
```
koonsies: Is "===" the accepted method for existence checks/boolean comparisons? Is "===" preferable to "isMenuActive & !isMenuPending"?

gulp/src/common/ui/PopMenu.jsx
```diff
@@ -31,7 +30,16 @@ export default class PopMenu extends Component {
...
+    const {toggleMenu, isMenuPending} = this.props;
+    if (isMenuPending){
+      return (
+        <div style={{position: 'relative'}}>
+          <span className="report-loader list-loader">
+          </span>
+          {this.menuContents()}
+        </div>
+      );
+    }
```
koonsies: Is there a better way to handle this?

gulp/src/common/ui/PopMenu.jsx
aspidites: I'd prefer ! myself. That said, you should be using `&&` as  `&` is bitwise.

gulp/src/common/ui/PopMenu.jsx
darrint: You can call functions from inside the jsx:

      return (
        <div style={{position: 'relative'}}>
          {this.renderMenuStuff()}
          {this.menuContents()}
        </div>

    ...

    renderMenuStuff() {
      const {isMenuPending} = this.props;
      if (isMenuPending) {
        return <span pendingway .../>
      }
      return <span otherway .../>

PR #2328, opened 2016-05-13
[PD-2173] make react profile units live

lemniscate613:
I updated links from myprofile overview page to new react modules. I also removed old templates and the link in the new Beta toolbar dropdown.


myprofile/tests/test_views.py
```diff
@@ -51,38 +51,6 @@ def test_edit_summary(self):
         resp = self.client.get(link)
         self.assertEqual(resp.status_code, 200)
 
-    def test_handle_form_get_new(self):
```
lemniscate613: I just removed the tests associated with `profile_form.html`, a template I deleted. The other tests could be updated in the future if desired.

myprofile/views.py
```diff
@@ -176,10 +162,6 @@ def handle_form(request):
             remote_form = RemoteForm(form_instance)
             return HttpResponse(content_type='application/json',
                                 content=json.dumps(remote_form.as_dict()))
-        else:
```
tdphillips: With the removal of this (and other) else clause, there are branches in this view that do not return an HttpResponse. For this exact removal I'm commenting on, you probably need to remove the "if" as well and deindent the code that it was wrapping. I'm unsure what needs to happen for the others.

PR #2330, opened 2016-05-16
Just converted form to div instead.

aspidites:
> I'm suspicious that the real problem here is we don't actually need those form tags. At least the author of that article sometimes wanted to submit a form. We never want to do that. (Still merging though.)

I'd appreciate if you gave me a chance to respond to a concern next time, thanks.

`preventDefault` and `stopPropagation` will continue to be a problem, but this at least takes care of the run report button (except on IE 8, naturally).

### DIFF
--- a/gulp/npm-shrinkwrap.json
+++ b/gulp/npm-shrinkwrap.json
@@ -4,1990 +4,3239 @@
   "dependencies": {
     "accepts": {
       "version": "1.3.2",
+      "from": "accepts@1.3.2",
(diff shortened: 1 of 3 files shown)


darrint:
Need to run `npm run clean-shrinkwrap`

aspidites:
I did that already.

aspidites:
In fact, the only reason the shrinkwrap file is included in the diff is because I ran `npm run clean-shrinkwrap`.

PR #2331, opened 2016-05-16
[PD-2394] Data migration for Dynamic Reporting. No more reset command.

darrint:
Still need to delete the command.

I wrote a script to "parse" and rewrite the relevant portion of setup.py.

* What should I do with backwards migration? My gut says throw an exception/disallow, to protect existing data.
* Should I delete all the "dead" records and the compliance section that is unused?


tdphillips:
:+1: on the backwards migration idea.

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        "Write your forwards methods here."
+        # Note: Don't use "from appname.models import ModelName". 
+        # Use orm.ModelName to refer to models in this application,
+        # and orm['appname.ModelName'] for models in other applications.
+
+        orm["myreports.UserType"].objects.all().delete()
+        ut_emp_dead = orm["myreports.UserType"].objects.create(
```
aspidites: I don't think we want to convert all of these "dead" configurations, do we?

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+
+        orm["myreports.UserReportingTypes"].objects.all().delete()
+        orm["myreports.UserReportingTypes"].objects.create(
+            user_type=ut_emp, reporting_type=rit_prm, is_active=True)
+        orm["myreports.UserReportingTypes"].objects.create(
+            user_type=ut_emp, reporting_type=rit_comp, is_active=True)
+        orm["myreports.UserReportingTypes"].objects.create(
+            user_type=ut_emp, reporting_type=rit_dead, is_active=True)
+        orm["myreports.UserReportingTypes"].objects.create(
+            user_type=ut_emp, reporting_type=rit_maybe_dead, is_active=False)
```
aspidites: Same comment as above, but with the "maybe" types.

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            description="Contacts Report",
+            datasource="contacts", is_active=True)
+        rt_comm = orm["myreports.ReportType"].objects.create(
+            report_type="communication-records",
+            description="Communication Records Report",
+            datasource="comm_records", is_active=True)
+        rt_state = orm["myreports.ReportType"].objects.create(
+            report_type="state",
+            description="State Report", is_active=True)
+        rt_screen = orm["myreports.ReportType"].objects.create(
```
aspidites: More a question than anything else: do we want report types that aren't visibly on a road map, or is it better to create them as we make configurations for them?

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            data_type="count_comm_rec_per_month",
+            description="Number of Communication Records per Month per Partner")
+
+        orm["myreports.PresentationType"].objects.all().delete()
+        pre_dead = orm["myreports.PresentationType"].objects.create(
+            description="Inactive", is_active=False)
+        pre_maybe_dead = orm["myreports.PresentationType"].objects.create(
+            description="Maybe Inactive", is_active=True)
+        pre_csv = orm["myreports.PresentationType"].objects.create(
+            presentation_type="csv", description="Unformatted CSV", is_active=True)
```
aspidites: Are CSVs *ever* formatted?

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            name="Contact Basic Report", is_active=True)
+        con_part = orm["myreports.Configuration"].objects.create(
+            name="Partner Basic Report", is_active=True)
+        con_comm = orm["myreports.Configuration"].objects.create(
+            name="Communication Records Basic Report", is_active=True)
+        con_comm_count = orm["myreports.Configuration"].objects.create(
+            name="Partners Comm Record Count Per Month Report", is_active=True)
+
+        orm["myreports.ConfigurationColumn"].objects.all().delete()
+        orm["myreports.ConfigurationColumn"].objects.create(
```
aspidites: I don't think this is an actual column on ContactRecord.

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            multi_value_expansion=False, is_active=True)
+        orm["myreports.ConfigurationColumn"].objects.create(
+            column_name="notes",
+            alias="Notes",
+            order=107,
+            output_format="text",
+            configuration=con_con,
+            multi_value_expansion=False, is_active=True)
+
+        orm["myreports.ConfigurationColumn"].objects.create(
```
aspidites: Should there be a delete method call above this for con_part configuration columns?

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            multi_value_expansion=False, is_active=True)
+        orm["myreports.ConfigurationColumn"].objects.create(
+            column_name="primary_contact",
+            alias="Primary Contact",
+            order=106,
+            output_format="text",
+            configuration=con_part,
+            multi_value_expansion=False, is_active=True)
+
+        orm["myreports.ConfigurationColumn"].objects.create(
```
aspidites: Same question as above, but for delete on con_comm.

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            configuration=con_comm,
+            multi_value_expansion=False, is_active=True)
+        orm["myreports.ConfigurationColumn"].objects.create(
+            column_name="job_interviews",
+            alias="Job Interviews",
+            order=120,
+            output_format="text",
+            configuration=con_comm,
+            multi_value_expansion=False, is_active=True)
+        orm["myreports.ConfigurationColumn"].objects.create(
```
aspidites: This can actually be deleted. It'll come into play for the User Activity report, but that's a separate report type, and will thus never be active here, as far as I can tell. If so, it's easy enough to re-add it later.

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            is_active=False,
+            multi_value_expansion=False)
+        orm["myreports.ConfigurationColumn"].objects.create(
+            column_name="length",
+            alias="Length",
+            order=122,
+            output_format="text",
+            configuration=con_comm,
+            multi_value_expansion=False, is_active=True)
+        orm["myreports.ConfigurationColumn"].objects.create(
```
aspidites: We probably want a better alias here, as I don't know that most will know the difference between "Location" and "Contact Location".

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -0,0 +1,1035 @@
...
+            report_data=rtdt_con_unagg, is_active=True)
+        orm["myreports.ReportPresentation"].objects.create(
+            presentation_type=pre_xlsx,
+            display_name="Partner Excel Spreadsheet",
+            report_data=rtdt_part_unagg, is_active=True)
+        orm["myreports.ReportPresentation"].objects.create(
+            presentation_type=pre_xlsx,
+            display_name="Communication Record Excel Spreadsheet",
+            report_data=rtdt_comm_unagg, is_active=True)
+        orm["myreports.ReportPresentation"].objects.create(
```
aspidites: Do we use these anywhere? Is there a reason to have them? (json passthrough)

myreports/migrations/0012_initial_dynamic_setup.py
darrint: The delete for all of ConfigurationColumn covers this.

myreports/migrations/0012_initial_dynamic_setup.py
darrint: The delete for all of ConfigurationColumn covers this.

myreports/migrations/0012_initial_dynamic_setup.py
darrint: So I should change it to "Contact Location"?

myreports/migrations/0012_initial_dynamic_setup.py
```diff
@@ -396,6 +312,7 @@ def forwards(self, orm):
             output_format="text",
             configuration=con_comm,
             multi_value_expansion=False, is_active=True)
+        # Leaving this.
```
aspidites: Why? It was never meant to be a user readable field, especially not for a communication record report.

myreports/migrations/0012_initial_dynamic_setup.py
aspidites: No. Location here is the meeting location or the place the contact occurred. Eg, "At XYZ conference", or "Recruiting office". Compare that to contact location, which is generally the geographical location of the contact (or their recruiting home office's address). @fezick, @JLMcLaugh, better name?

myreports/migrations/0012_initial_dynamic_setup.py
darrint: Discussed in person. This comment was meant to apply to the "last_action_time" which I deleted.

myreports/migrations/0012_initial_dynamic_setup.py
fezick: "Meeting Location" and "Contact Location" work for me.

darrint:
Backwards migration is now disallowed, addressing @tdphillips comment above.

aspidites:
Tests probably need to be updated

aspidites:
To clarify. Testing dead stuff is probably fine, but if aliases are different in the data migration, they probably should be for the test set up as well.

myreports/migrations/0012_initial_dynamic_setup.py
darrint: Now it is "Meeting Location"

PR #2333, opened 2016-05-16
[PD-2387] Left Spinner

koonsies:
To the left, to the left.


tdphillips:
I'm not seeing a difference.

koonsies:
The spinner isn't on the left of the filename now?

tdphillips:
It was always on the left. If I go to QC, that's where it is.

koonsies:
If you head to reports on QC, click the ellipses to the right side of an existing report and click refresh. The spinner should show up on the right side. It's easily mistaken for the spinner that appears when a new report is generated, which does appear on the left. This PR brings those two in balance.

koonsies:
That being said, don't merge this just yet.. Checking something.

tdphillips:
Thanks for the explanation. Just verified.

koonsies:
After your comment, I looked back at how the "new report" spinner was populated and realized that I was creating an additional class where we didn't need one. Instead, I'm not using the same spinner case as the new report.

Less fluff code.

Works as before.

gulp/src/common/ui/PopMenu.jsx
```diff
@@ -8,15 +8,15 @@ export default class PopMenu extends Component {
   }
 
   menuContents() {
-    const {options, isMenuActive, closeAllPopups, isMenuPending} = this.props;
+    const {options, isMenuActive, closeAllPopups} = this.props;
```
koonsies: This and below is simply reverting back to the way the file was before this ticket.

static/custom.css
```diff
@@ -2225,12 +2225,6 @@ h3.bottom-bordered {
   vertical-align: top;
 }
 
-.reporting.sidebar .report-loader.list-loader {
-  display: block;
-  float: right;
-  margin-top:3px;
-}
-
```
koonsies: Remove styling that I added for the new class.

aspidites:
@koonsies this PD-2387 probably needs to be started and tagged here so that it transitions correctly.

koonsies:
Good call. Done and done!

aspidites:
@koonsies  is this waiting on anything? I see @tdphillips  passed it, and I just verified locally that it works.

koonsies:
Nope! Once tests pass, it should be good.

static/custom.css
```diff
@@ -2229,12 +2229,6 @@ h3.bottom-bordered {
   vertical-align: top;
 }
 
-.reporting.sidebar .report-loader.list-loader {
-  display: block;
-  float: right;
-  margin-top:3px;
```
aspidites: Were you going to re-add this, or just do it as part of the ticket dedicated to centering that icon?

PR #2337, opened 2016-05-18
Fix IE8 Reporting Bugs

darrint:
Fixes:

* Components importing React Sortable wouldn't load on ie8 at all, at least under dev server.
* ClickOutCompat was needed and not being used in DateField.
* ReportList context menu couldn't be closed on ie8.


aspidites:
I can confirm the first, but it is restricted to devserver. Runs fine if you just do a build.

PR #2345, opened 2016-05-26
PD-2406 PD-2404 PD-2403 - Tag Select, Avail. Record Count, and Mode Selection

misterjonscott:
# Wait, 3 tickets on one PR?
Yeah.  The record count uses the data from the tag select to display a count in the mode selector.  The tag selector needed a way to be presented, so I completed the mode selector and stubbed the tag query builder in so it can be added later, easily.

# What's new?
- Select can now be passed a "disabled" bool which deactivates the menu, complete with styles to change appearance.
- Report Builder now shows a filter option so users can disable filtering by Partner or Contact, select by name, and later select by Tag, matching the prototype. [url](https://invis.io/RQ75PQZ7M)
- The Tag Select is a Select component that offers multiselect in a tag-styled interface.  This behavior is styled after the very popular Chosen and Select2 plugins, and replaces the side-by-side multiselect with the same behavior but far greater scalability and in less space.
- The Tag Select has its own controller for singular use, which features a built in counter display.  In this PR, it's being controlled by its parent, though.  I just wanted to offer a complete option.
- The available record count is based on the conversation between @fezick and I, here: [url](https://directemployersfoundation.atlassian.net/browse/PD-2403) and is displayed on the Mode Selector as well as optionally in the individual Tag Select element.
- I've also updated the style for the loader, to get it properly centered, as well as other styles that I stumbled over while making this component.
- Controls that are loading will show a loader image instead of a counter, and will be disabled while loading.

# :shipit:


misterjonscott:
Not sure why this is failing...  Anyone have any ideas?

koonsies:
Do we want to handle the possibility that someone selects a contact before they select a partner? Currently, if you select a partner, it limits the contacts to that partner group.. but if you select a contact, it does not limit the partner list. So, if you select a contact first, you can select a partner that is unrelated to that contact.

misterjonscott:
@darrint Can you speak to Koonsies' comment on the filtering methods?

PR #2346, opened 2016-05-26
[PD-2436] Deploy redux in preparation for foreign tags

darrint:
## Status

* Filtering works. Including hairy behavior with field hints being dependent on each other.
* ~~Running reports and history don't work. At this point run and the associated report list stuff should be trivial to implement. History is still tricky so I'm going to try to touch it as little as possible for this ticket.~~ Everything works.
* Export and Preview haven't been touched.
* ~~I still need to update the JS style guide as there's a lot of conventions we follow that we haven't documented.~~ I updated it.

## Review Suggestions

* Understand how redux apps look. If you haven't done a basic tutorial yet, now is a good time.
* Understand how redux apps handle async actions. What I have here is very close to: http://redux.js.org/docs/advanced/AsyncActions.html
* Look at `report-state-reducer.js`, its actions, and its unit tests. This is the basic filtering logic.
* Look at `compound-actions.js` this is where all the hairy stuff is handled. A lot of hacks stuffed into arbitrary spots in the UI code are now here and unit tested.
  * The unit test for `doFilterUpdatesWithDependencies` is a lot hairier than I'd like.
* Look at `WizardFilterCityState.jsx` as an example of the kinds of simplifications I did as a result of this work.
  * `getHints` no longer returns a value. It just causes props to refresh.
  * Currently selected values are just passed in as props.
* Look at `WizardFilterMultiSelect.jsx` where the DANGER comment is gone. That is being handled (and unit tested) in `doFilterUpdatesWithDependencies`. The potential looping problem is gone.
* ~~`SetUpReport.jsx` won't get much shorter until I attack the history modifying parts of the app and eliminate `ReportFinder`. I don't want to do that in this ticket.~~ I ended up addressing history.


gulp/src/common/myjobs-api.js
```diff
@@ -61,3 +61,15 @@ export class MyJobsApi {
     return this.ajaxWithFormData('DELETE', url, data);
   }
 }
+
+export function is500Error(exc) {
```
aspidites: ```javascript```
Status = {
  NotFound: 404,
  InternalServerError: 500
};
```

I'd then inline the check or write a combinator if that bothers you
```javascript
const isError = (exc, status) => exc.response && exc.response.status === response;
console.log(isError(response, Status.NotFound));
```
Flipping the arguments would let you map over an array of status codes, but you probably don't need that.

aspidites: Actually, I'm not seeing where these functions are being used at all.

gulp/src/common/create-redux-store.js
```diff
@@ -0,0 +1,22 @@
+import {createStore, applyMiddleware, compose} from 'redux';
+import thunkMiddleware from 'redux-thunk';
+import promiseMiddleware from 'redux-promise';
+
+// Since we use this in unit tests occasionally.
+const devToolsMiddleware = typeof window === 'undefined' ?
```
aspidites: is this still run in production?

gulp/src/common/ui/SearchInput.js
```diff
@@ -46,9 +46,8 @@ export class SearchInput extends Component {
 
   onSelect(e, index) {
     e.preventDefault();
-    const {onSelect: selectCb} = this.props;
-    const {items} = this.state;
-    const selected = items[index];
+    const {hints, onSelect: selectCb} = this.props;
```
aspidites: I think it's confusing to have both a method and prop called onSelect. I've seen a few different conventions used between all of our projects, but the most prominent seems to be to name methods that are doing something with the callback handle* and the actual props on*. So in this case `handleSelect` for the method and `onSelect` for the prop. you'd also then have to alias the prop name on destructuring.

gulp/src/common/ui/SearchInput.js
```diff
@@ -101,14 +101,14 @@ export class SearchInput extends Component {
...
-      items: [],
+      dropped: false,
     };
   }
 
   async search(value) {
     const {getHints} = this.props;
-    const items = await getHints(value);
-    this.setState({items});
+    this.setState({dropped: true});
```
aspidites: If we're using redux, we shouldn't ever be mutating the state directly except in reducers.

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -73,6 +88,14 @@ export default class SetUpReport extends Component {
 
   onMenuChanged(reportingTypes, reportTypes, dataTypes,
       reportConfig) {
+    const {dispatch, reportName} = this.props;
+    dispatch(startNewReportAction({
```
aspidites: Bikeshed - IF all actions are in a single file and are invoked with dispatch, does the Action prefix buy us anything other than keystrokes?

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -171,6 +189,13 @@ export default class SetUpReport extends Component {
     history.pushState(null, href, query);
   }
 
+  dispatchFilterAction(action) {
```
aspidites: This raises a red flag for me. It seems bizarre for me that we would have an action creator that takes a payload of another action to then call another action creator which also takes an action as a payload. To illustrate, this feels like an anti-pattern:

```javascript
dispatch( (action, ...) => dispatch( action => dispatch( action => return { type: ... } ) ) )
```

Instead, I'd rather see:
```javascript
dispatch( (...) => { type: ... };
```
And
async (...) => {
    dispatch(actionCreator(...));
    const result = await asyncStuff(); 
    dispatch(actionCreator(result);
}
```

That is, an action creator is responsible for only creating an object with a key of "type" and some form of payload. If we need to do asyn things, that would not be an action creator, but would rather be a function which itself dispatches action creators that might depend on each other. I don't think it should ever the case that an action creator calls an action creator which calls another action creator. If actions need to be dispatched as a result of others, the framework should handle that, since dispatched actions are handled sequentially.

File: gulp/src/reporting/SetUpReport.jsx [hunk 7]
```diff
@@ -213,7 +242,7 @@ export default class SetUpReport extends Component {
     }
 
     const rows = [];
-    if (reportConfig) {
+    if (filterInterface.length > 0) {
```
aspidites: Maybe doesn't matter, but `if (filterInterface.length)` is sufficient here.

gulp/src/reporting/SetUpReport.jsx
```diff
@@ -225,21 +254,26 @@ export default class SetUpReport extends Component {
...
             name=""
             autoFocus="autofocus"
-            onChange={v => reportConfig.changeReportName(v.target.value)}
-            maxLength={maxNameLength}/>
+            onChange={v => dispatch(setReportNameAction(v.target.value))}/>
         </FieldWrapper>
       );
-      reportConfig.filters.forEach(col => {
+      filterInterface.forEach(col => {
+        function getHints(field, value) {
```
aspidites: I don't know that we want to redefine the same function for every value in the filterInterface. Should be sufficient to just declare this outside of your forEach.

gulp/src/reporting/compound-actions.js
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * User ran a report.
+ *
+ * idGen: a id generator, used for temporary ids of unfinished reports.
+ * api: api instance, used to run report and get the report list.
+ * reportDataId: thekind of report to run
+ * name: name of the report
+ * filter: filter to use for running this report.
+ */
+export function doRunReport(reportDataId, name, filter) {
```
aspidites: Do you not have to mark this as async since await is being used?

gulp/src/reporting/compound-actions.js
```diff
@@ -0,0 +1,151 @@
+import warning from 'warning';
```
aspidites: I don't know how I feel about these (the middleware mostly, that we are dispatching multiple actions based on the status of promises is fine). I am, however, happy that it's all centralized and reduced to about 4 such functions.

gulp/src/reporting/main.js
```diff
@@ -1,23 +1,43 @@
...
 installPolyfills();
 
 const myJobsApi = new MyJobsApi(getCsrf());
+const idGen = new IdGenerator();
 const reportingApi = new Api(myJobsApi);
 
 const configBuilder = new ReportConfigurationBuilder(reportingApi);
 const reportFinder = new ReportFinder(reportingApi, configBuilder);
 
+const reducer = combineReducers({
```
aspidites: At some point it'd probably make sense if we split reducers based on component. For instance, ReportList could have it's own.

gulp/src/reporting/report-list-actions.js
```diff
@@ -0,0 +1,30 @@
...
+ * payload is the numeric id for this running report.
+ */
+export const removeRunningReportAction = createAction('REMOVE_RUNNING_REPORT');
+
+/**
+ * A report completed running.
+ *
+ * payload is the read database id for this new running report.
+ */
+export const newReportAction = createAction('NEW_REPORT');
```
aspidites: Bikeshed: Your other reports have verbs. This is just an adjective and noun. I was expecting something like 'CREATE_NEW_REPORT'.

gulp/src/reporting/report-state-actions.js
```diff
@@ -0,0 +1,103 @@
+import {createAction} from 'redux-actions';
```
aspidites: BikeShed -- I really miss Algebraic Data Types after reading this. 

On a serious note -- I appreciate the documentation. I wonder if using purescript style signatures might cut down on the typing (no pun intented). For instance:

```javascript
// createItemAction :: forall a b. String -> a -> b -> { field :: a, item :: b }
function createItemAction(name) {
    return createAction(name, (field, item) => ({field, item}));
}
```
Which reads as "For any types a and b, given a String, you'll receive a function that takes an a and a b and returns a dictionary with keys field of type a and item of type b". The Universal quantification is unnecessary and the currying could be made explicit:

```javascript
// createItemAction :: String -> (a -> b -> { field :: a, item :: b })
```

gulp/src/reporting/report-state-reducer.js
```diff
@@ -0,0 +1,221 @@
+import {handleActions} from 'redux-actions';
+import {map, filter as lodashFilter} from 'lodash-compat/collection';
+import {omit} from 'lodash-compat/object';
+
+const maxNameLength = 24;
+
+function addOrReplaceByValue(items, item) {
```
aspidites: As a generic function, are we worried about duplicates? If I run this on something with values [1, 2, 2, 3], I'll end up with values [1, 2, 3, 2].

gulp/src/reporting/report-state-reducer.js
```diff
@@ -0,0 +1,221 @@
...
+ *  errors[key] = [messages]
+ *    errors in this report configuration
+ *    key: see filter in filterInterface or "$ALL" for errors associated with
+ *      no particular field. 'reportName' is served for the reportName control.
+ *    messages: a list of strings, intended to be shown as a bulleted list.
+ *
+ * }
+ */
+export const reportStateReducer = handleActions({
+  'START_NEW_REPORT': (state, action) => {
```
aspidites: Why not just extend the object using the spread operator instead of destructuring just to restructure?

```javascript
export const reportStateREducer = handleActions({
    START_NEW_REPORT: (state, action) => ({...payload, errors: {}, hints: {}}),
});
```

gulp/src/reporting/report-state-reducer.js
```diff
@@ -0,0 +1,221 @@
...
+    const {currentFilter} = state;
+
+    if (!currentFilter.hasOwnProperty(field)) {
+      return state;
+    }
+    const orFilter = currentFilter[field];
+    const newOrFilter = lodashFilter(orFilter, i => i.value !== item.value);
+
+    // If the group is empty, return the filter without this key.
+    const newFilter = newOrFilter.length > 0 ? {
```
aspidites: `> 0` is redundant.

gulp/src/reporting/spec/compound-actions-spec.js
```diff
@@ -0,0 +1,269 @@
+import {
```
aspidites: Would be nice if we used a freezing library to ensure that we aren't mutating state when dealing with our reducers. The redux videos show examples of this. We don't litter our code with line noise, but still get immutability guarantees that way.

gulp/src/reporting/wizard/WizardFilterCityState.jsx
```diff
@@ -5,90 +5,73 @@ import {getDisplayForValue} from 'common/array';
...
+  componentDidMount() {
+    const {getHints} = this.props;
+    // Prepopulate state select.
+    getHints('state');
+  }
 
-    updateField(field, value) {
-      const {updateFilter} = this.props;
-      const {currentLocation} = this.state;
+  updateField(field, value) {
```
aspidites: At a brief glance through this file and the file that uses the component, it doesn't look like this is using redux. No actions being dispatched and it looks like we're calling the API directly.

aspidites:
> Running reports and history don't work. At this point run and the associated report list stuff should be trivial to implement. History is still tricky so I'm going to try to touch it as little as possible for this ticket.

Given that, we probably don't want this pointing to QC as dynamic reporting is in production. I'd hate for us to have to change gears before you get around to making these things work again.

aspidites:
Now that the critical stuff is out of the way:
- I like where this is heading. In most cases, it was easier for me to either track state changes, or suspend enough of my logic to continue reading and come back to reducers later. A key reason here is that for the pure reducers, I knew that there wouldn't be any strange state mutation going on.
- There's a lot of test coverage here - I didn't read them in a lot of detail, but the fact that they even exist is fantastic
- For elaborate SPA's the time traveling debugger will be nice (though the last time you showed me, it looked like there were still a few kinks to work out)
- consistency of formatting aside, you're doing a much better job at documenting things than what I've seen from you before. Still some work to do, but I appreciate the strides you've made in this area.

I *think* that's all I've got. Granted, 3/4 of the way down I started losing focus, but I think I touched on the biggest stuff.

gulp/src/common/ui/SearchInput.js
darrint: So far I'm only fixing SearchInput since you called it out. There may be some other components that need addressed in the future. I think the `handle*` naming departs from convention used elsewhere in the community but it reads more clearly, IMHO.

aspidites: I'm not sure we follow the community on most things. :-) Really, I don't mind what it's called. My biggest issue was with us shadowing a name, then renaming during restructuring when unambiguous names would resolve the issue ahead of time.

gulp/src/common/ui/SearchInput.js
darrint: Are you concerned that I'm undermining redux? Redux *does not* currently manage the dropped state of the control. Redux only manages the more generic state like potential items to show and selected value, etc.

http://redux.js.org/docs/FAQ.html#organizing-state-only-redux-state

The case you've called out is (coincidentally) the case mentioned in the FAQ.

In the future I'm happy to participate in pushing more of that kind of state into Redux.

FWIW search input is one of the more complex controls I've made and currently has 4 internal state values:

* partialValue, currently typed input which is independent of the last value selected
* mouseInMenu, is the mouse currently over the menu? Need to know this during onBlur of some dropdown controls.
* keySelectedIndex, if the user is navigating the control with the keyboard, where should the highlight be?
* dropped, should the dropdown menu be shown?

Redux controls:

* the current value which has actually been selected
* the current hint list to be shown if the control is dropped.

gulp/src/reporting/SetUpReport.jsx
darrint: I want to be clear that calling startNewReportAction returns an object, but does not actually do anything else. I can be talked out of it.

gulp/src/reporting/compound-actions.js
darrint: What I'm doing is nearly exactly like what is being done here:

http://redux.js.org/docs/advanced/ExampleRedditAPI.html Skip to "fetchPosts" and "fetchPostsIfNeeded". Both return a function that returns a promise. The async is just a very nice syntax sugar for handling the promises.

I wouldn't mind writing a custom middleware just for us that exposes the same stuff with a less weird interface. The thunk/promise middlewares are less than 50 LOC combined.

gulp/src/reporting/compound-actions.js
darrint: Covered all functionality except export (1 more maybe?) and I'm up to 7. A few of these go a long way.

The scariest of them is `doReportDataSelect` which is run after landing on `/set-up-report` or cloning a report. I'm quite happy with the fact that I can lay out end to end everything that could happen or go wrong when loading setup.

gulp/src/reporting/main.js
darrint: See `report-list-reducer.js` and friends.

gulp/src/reporting/report-state-reducer.js
darrint: In its current form the function will replace all instances of the given items. (Might have been different when you read it.)

gulp/src/reporting/report-state-reducer.js
darrint: I don't want to accidentally leave behind any state from the previous report.

gulp/src/reporting/spec/compound-actions-spec.js
darrint: Do you remember which video it was?

gulp/src/reporting/report-state-actions.js
darrint: I really want to encourage us to explain what field and item mean. Although I guess since it's a generic function I don't say much there. When I wrote that function I used it 4 times. Since I "pluralized" the interface for or filters I only use it twice. I might just take it out if it's confusing.

gulp/src/common/create-redux-store.js
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * reducer: reducer for the store
+ * defaultState: optional default state for the reducer. otherwise undefined
+ * thunkExtra: extra arg for thunk middleware.
+ */
+export default function createReduxStore(reducer, defaultState, thunkExtra) {
+  return createStore(reducer, defaultState, compose(
+    applyMiddleware(thunkMiddleware.withExtraArgument(thunkExtra)),
+    applyMiddleware(promiseMiddleware),
+    devToolsMiddleware));
```
aspidites: Why doesn't devToolsMiddleware need to be wrapped in applyMiddleware?

gulp/src/common/create-redux-store.js
aspidites: In case it wasn't obvious, my point here was that I'm not sure that we should be enabling devtools during production deployments.

gulp/src/common/ui/SearchInput.js
aspidites: I read the FAQ entry you linked to. What I read from your comment is that managing component-level state isn't possible. What I'm reading from that FAQ is that it isn't exactly expected that you be able to. I'll probably fork this and play with it, but I'd suspect that you either needn't have component-level state, or that app state is the composition of all child states (as it happens in Elm (from which Redux is derived) and Pux (which is arguably what Redux wanted to be).

gulp/src/reporting/SetUpReport.jsx
aspidites: Action creators by definition return an object:

> Action creators are exactly that—functions that create actions. It's easy to conflate the terms “action” and “action creator,” so do your best to use the proper term.

I'd also argue that "start*" sounds like it _does_ something, so I'm not sure that the name clarifies anything. I know how it works because I know Redux, in which case, *Action is redundant to me. I'd prefer naming that was evocative of the action being performed over a quasi-Hungarian notation. We can continue this outside of this PR though. It's non-essential, which is why I marked it as bike shedding.

gulp/src/reporting/SetUpReport.jsx
aspidites: Bump. @darrint, I see you haven't commented on this remark.

gulp/src/reporting/compound-actions.js
aspidites: Ok, you're using a closure to "trap" the environment. Difference between the example and your code is that you're exposing more to the environment, where as the example is just passing along dispatch. Still pondering this, so I'll reserve judgment until a future date.

Much clearer after having had real sleep (cumulative, I only got 2 hours last night).

gulp/src/reporting/main.js
aspidites: Ah. Must have missed this on my first read through.

gulp/src/reporting/report-state-actions.js
aspidites: I wouldn't say confusing, but I suspect it being inlined (as it is so short) would be clearer, at least for me. I don't need to go looking up a function definition.

aspidites:
This isn't running locally for me. All I get is a company name. I know the tables are populated correctly because dynamic reporting works on quality-control. I did run `npm install` before running. I do have a lot of lint warnings.

gulp/src/common/ui/SearchInput.js
```diff
@@ -12,44 +12,43 @@ export class SearchInput extends Component {
     this.state = this.getDefaultState(value);
```
aspidites: I know this wasn't changed in this PR. On line 10, should we be passing props to super?

darrint: Done

gulp/src/reporting/SetUpReport.jsx
darrint: I fixed the code but forgot to comment. This action creator no longer takes an action as a parameter. I changed the method in `SetUpReport` to first dispatch the state change action, then do the dependent field updates.

gulp/src/reporting/report-state-actions.js
darrint: I inlined it.

gulp/src/reporting/report-state-reducer.js
aspidites: I don't understand your comment. How is this:
```javascript
const {foo, bar, baz} = object;
return {foo, bar, baz, errors: {}, hints: {}};
```
Any different than:
```javascript
return { ...object, errors: {}, hints: {} };
```

If the concern is that there is something in payload you don't want, I'd argue that the payload should only include information intended to be used by the reducer. if there is left over state you wish to be reinitialized, that's what the arguments after the spread are for.

aspidites:
I think there's one outstanding comment from me, an it'd be a trivial refactor. All I really have left to do is run this locally and I'm done looking at it. Didn't before because of the devtools issue, which seems to be addressed now.

gulp/src/reporting/report-state-reducer.js
darrint: I misread. I thought you meant `...state`, not `...payload`.

If I include `...payload` I'm going to end up with `defaultFilter` and `filters` in the reducer state. So if I do this, I should probably rename them in `startNewReportAction`. Arguably I should make that names match in the action as that's probably clearer. Will that work?

aspidites:
Few regressions with respect to flickering components we spoke about in person that aren't apparent on QC. Also, it'd be great if these were organized in subdirectories. actions, reducers and components, for instance (I think having a components subdirectory was talked ab out a long time ago, which is why I even introduced them in the NUO app).

gulp/src/reporting/report-state-reducer.js
aspidites: Sounds like it should work. Though, I still don't know why we are passing things as payload if we don't intend to use it. Seems like if it's an ignored argument, it shouldn't be part of the payload to begin with.

PR #2348, opened 2016-05-31
[PD-2405] Add Empty List Handling to Reporting

koonsies:
## Overview

I have modified the filter_query_set function of both the communication records handler and the contacts handler. The following logic applies to both.

If one of the following fields receives an empty list, return no results:
contact, partner

If the tags object receives an empty list, return all untagged items of that category

If either of these receives a None, or the parameter is not passed in, do not filter on that field.

## Testing

I have also modified/added tests to properly "poke" these scenarios. Without a UI, this is kind of hard to hand test.

### DIFF
--- a/myreports/datasources/comm_records.py
+++ b/myreports/datasources/comm_records.py
@@ -314,15 +314,21 @@ class CommRecordsFilter(DataSourceFilter):
         if self.communication_type:
             qs = qs.filter(contact_type__in=self.communication_type)
 
+        list_empty = True
         if self.tags:
             or_qs = []
             for tag_ors in self.tags:
-                or_qs.append(
-                    reduce(
-                        __or__,
-                        map(lambda t: Q(tags__name__iexact=t), tag_ors)))
+                if tag_ors:
+                    list_empty = False
+                    or_qs.append(
+                        reduce(
+                            __or__,
+                            map(lambda t: Q(tags__name__iexact=t), tag_ors)))
             for or_q in or_qs:
                 qs = qs.filter(or_q)
+            if list_empty:
+                # if an empty tags list was received, return only untagged items
+                qs = qs.filter(tags=None)
 
         if self.locations is not None:
             contact_qs = Contact.objects.all()
@@ -338,10 +344,10 @@ class CommRecordsFilter(DataSourceFilter):
             contact_qs.distinct()
             qs = qs.filter(contact=contact_qs)
 
-        if self.partner:
+        if self.partner is not None:
             qs = qs.filter(partner__pk__in=self.partner)
 
-        if self.contact:
+        if self.contact is not None:
             qs = qs.filter(contact__pk__in=self.contact)
 
         return qs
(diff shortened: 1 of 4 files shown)


myreports/datasources/comm_records.py
```diff
@@ -314,15 +314,21 @@ def filter_query_set(self, qs):
         if self.communication_type:
             qs = qs.filter(contact_type__in=self.communication_type)
 
+        list_empty = True
```
koonsies: self.tags here is a list of lists. This flag will easily show whether or not it was a list of empty lists or otherwise.

myreports/datasources/comm_records.py
```diff
@@ -338,10 +344,10 @@ def filter_query_set(self, qs):
             contact_qs.distinct()
             qs = qs.filter(contact=contact_qs)
 
-        if self.partner:
+        if self.partner is not None:
```
koonsies: Is self.partner is an empty list, we still want to filter on it, but not if it is None

myreports/tests/test_comm_records_datasource.py
```diff
@@ -197,7 +197,10 @@ def test_filter_by_city(self):
         self.assertEqual(expected, subjects)
 
     def test_filter_by_tags(self):
-        """Show only commrec with correct tags."""
+        """
+        Show only commrec with correct tags.
+
+        """
```
koonsies: Formatting change here.

tdphillips:
The test failure won't be fixed by rerunning until today is over. See PD-2421 if you want more info.

myreports/datasources/comm_records.py
```diff
@@ -314,15 +314,21 @@ def filter_query_set(self, qs):
...
 
+        list_empty = True
         if self.tags:
             or_qs = []
             for tag_ors in self.tags:
-                or_qs.append(
-                    reduce(
-                        __or__,
-                        map(lambda t: Q(tags__name__iexact=t), tag_ors)))
+                if tag_ors:
```
darrint: What is `if tag_ors:` for? I don't see it in `contacts.py`.

koonsies: Unlike contacts.py, self.tags in this context is a list of lists, instead of a simple list. Basically, if `self.tags == [[]]` then line 324-326 would choke on the empty list. I believe it was the reduce function that struggled with it.

koonsies: Actually, now that I look at it, I'm less certain. It should be a list of lists in both cases, but the tests that were written for contacts.py only provide a simple list. I'm looking back over this.

aspidites: While it's currently the case that it's a list of lists, I was under the impression that the work @misterjonscott was doing would operate on a simple list (eg, there's no oring). Is that not the case? Are we using a completely different set of functionality for that case?

tdphillips: I thought I heard an offline conversation about this. Did that happen or am I misremembering?

koonsies: I do not have an update on this. @misterjonscott is the UI sending a list of lists or a simple list for the tags selections?

darrint: List of lists. You can see that in the `TagAnd.jsx` proptypes. `selected` is array of array of shape. `available` is just array of shape.

koonsies: Alright, good. So what I have here is correct.

PR #2350, opened 2016-06-02
[PD-2427] Logging for requests to prm@my.jobs

aspidites:
When you send an email to prm@my.jobs with known contacts in PRM in the bcc line, we parse that to create communication records. All this PR does is add a bunch of logging to the mypartners view involved in that process. No new unit tests as I assume we don't want to be calling out to newrelic's API or doing even more IO (by attempting to read a log file). I'm unaware of a way to test this functionality locally, and running tests doesn't seem to trigger logging as the level is set as ERROR, even when I force it to INFO using override_settings or by manually touching django.conf.settings.

I also raised the logging level on QC so that more is logged. Production will still only log errors.


mklauber:
:+1:

mypartners/views.py
```diff
@@ -1179,6 +1187,8 @@ def process_email(request):
 
     to = request.REQUEST.get('to', '')
     cc = request.REQUEST.get('cc', '')
+    newrelic.agent.add_custom_parameter("to", to)
```
tdphillips: If an exception isn't generated and then sent to NewRelic, these won't show up anywhere.

default_settings.py
```diff
@@ -408,9 +408,9 @@
         'pysolr': {
             'level': 'ERROR'
         },
-        'views': {
-            'level': 'INFO',
-            'handlers': ['file']
+        'mypartners.views': {
```
JLMcLaugh: We use the views logger in a few places.

aspidites: Well it doesn't work.

mypartners/views.py
```diff
@@ -1202,18 +1220,23 @@ def process_email(request):
...
-        if not element.lower() == 'prm@my.jobs' and validate_email(element):
+        if not element.lower() == PRM_EMAIL and validate_email(element):
             validated_contacts.append(element)
 
     contact_emails = validated_contacts
 
     try:
         contact_emails.remove(admin_email)
     except ValueError:
+        newrelic.agent.record_exception(*sys.exc_info())
```
JLMcLaugh: This probably isn't worth recording, especially as an exception that shows as an error in new relic. The most common use-case is that the sender's email isn't going to be in the to or cc fields.

aspidites: Really, I just added labels and called new relic on exceptions so that they were eventually called. Given that newrelic is only called on exceptions, is it even worth using?

JLMcLaugh: Newrelic records a lot of data outside of exceptions. We pass on this exception because it's not an error. It wouldn't normally show up in NewRelic. If you manually call record_exception, it will get recorded as an error. We don't want that in this case.

aspidites: I just removed them. Not convined new relic is buying us anything here. Especially if it's a silent failure not caused by an exception. In any case, the file logging still remains.

mypartners/views.py
```diff
@@ -1194,13 +1193,6 @@ def process_email(request):
     contact_emails = filter(None,
                             [email[1] for email in recipient_emails_and_names])
 
-    # This info will only be sent to newrelic if an exception is raised.
-    newrelic.agent.add_custom_parameter("to", to)
```
JLMcLaugh: We still want these parameters. They'll be available in insights, even if an error isn't recorded.

aspidites: As far as I understand, if the error isn't recorded, these won't be either. @tdphillips?

tdphillips: I was probably mistaken.

PR #2351, opened 2016-06-02
[PD-2430] Classic is now default.

aspidites:
If the version cookie isn't set, you'll be directed to the classic reporting experience when you click on reporting in the navigation.

### DIFF
--- a/myjobs/templatetags/common_tags.py
+++ b/myjobs/templatetags/common_tags.py
@@ -328,10 +328,10 @@ def get_menus(context):
             'dynamic': url("reports/view/dynamicoverview"),
             'classic': url("reports/view/overview"),
         }
-        reporting_version = request.COOKIES.get('reporting_version', 'dynamic')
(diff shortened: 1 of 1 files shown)


myjobs/templatetags/common_tags.py
```diff
@@ -328,10 +328,10 @@ def get_menus(context):
             'dynamic': url("reports/view/dynamicoverview"),
             'classic': url("reports/view/overview"),
         }
-        reporting_version = request.COOKIES.get('reporting_version', 'dynamic')
+        reporting_version = request.COOKIES.get('reporting_version', 'classic')
         employer_menu["submenus"].append({
             "id": "reports-tab",
-            "href": version_urls.get(reporting_version, "beta"),
```
aspidites: This should have been classic to begin with. While writing the feature originally, I renamed beta to dynamic, but apparently missed a spot.

PR #2354, opened 2016-06-02
[PD-2420] manage users - add check when editing self as user

lemniscate613:
You've always had to have at least one role to save the User form

We were not checking that that role had the 'read roles' activity -- which is required to view Manage Users at all.


myjobs/views.py
```diff
@@ -1060,6 +1060,16 @@ def api_get_users(request):
                         content_type="application/json")
 
 
+def api_get_current_user(request):
```
lemniscate613: @darrint suggested something this. We're trying to get away from passing information from templates into React directly. Rather use APIs like this.

aspidites: You're already making one call. No need to make another. I'd prefer we simply added the current user to the context of whatever request we originally made.

myjobs/views.py
```diff
@@ -1060,6 +1060,16 @@ def api_get_users(request):
...
+
+    ctx = {}
+    ctx['email'] = request.user.email
```
aspidites: email address are unique, so no need to pass the id as well.

aspidites: In fact, if you aren't doing anything with it other than performing checks, the id would be better, as the response size would be much smaller.

myjobs/views.py
lemniscate613: Good idea @aspidites https://github.com/DirectEmployers/MyJobs/pull/2354/commits/ad2587c7c982f25de6365132b5cd05af81b238d8

tdphillips: Given the linked commit, is api_get_current_user still necessary?

gulp/src/manageusers/User.jsx
```diff
@@ -132,11 +132,48 @@ class User extends React.Component {
...
       });
       return;
     }
 
+    // If a user is editing their own account, they must have at least one role
+    // with the 'read role' activity, otherwise they'll be kicked out of
+    // manage users.
+
+    // Is user editing their own account?
+    if (parseInt(userId, 10) === parseInt(currentUserID, 10)) {
```
aspidites: The parseInt here is redundant as you're doing nothing with the parsed result: `if (userId === currentUserID`. Also, It'd be nice if the naming were consistent ("Id" vs "ID").

gulp/src/manageusers/User.jsx
```diff
@@ -132,11 +132,48 @@ class User extends React.Component {
...
+      // What roles are currently assigned?
+      const assignedRolesAsStrings = _.map(assignedRoles, function getName(role) {
```
aspidites: We've been using arrow functions elsewhere for anonymous closures.

gulp/src/manageusers/User.jsx
tdphillips: Is it possible for userId and currentUserID to be different types at this point? If one is a string and the other's an int, parsing makes sense.

myjobs/views.py
aspidites: @tdphillips  No. If you look at the commit, what he's referring to is sending only the id instead of both id and email. There is still a separate API call to get said ID (which I am still opposed to).

PR #2356, opened 2016-06-03
Allow the upload of exported csv files [PD-2379]

tdphillips:
Testing:

- Head to the DestinationManipulation admin
- Search for something
- Select some rows If you select duplicate (view_source, action_type) pairs, undefined things will happen on reimport.
- In the dropdown, select "Save selected as CSV" and hit "Go"
- Modify if you wish.
- Head to /ajax/source-codes locally and fill out everything.
    - "Buids" is comma-separated. These do not have to match the buids in the spreadsheet - if they don't match, the values in the spreadsheet will be copied to the provided buids. If not provided, we update rows based on the buids from the csv.
- Submit and verify that stuff happened in the DestinationManipulation admin

### DIFF
--- a/automation/forms.py
+++ b/automation/forms.py
@@ -2,7 +2,8 @@ from django import forms
 from django.core.files.storage import default_storage
 from django.core.files.base import ContentFile
 
-from automation.source_codes import process_spreadsheet, add_source_codes
+from automation.source_codes import (
+    add_source_codes, add_destination_manipulations, process_file)
 from redirect.models import Redirect
 
 ATS_PARAMETERS = {
@@ -67,7 +68,7 @@ class SourceCodeFileUpload(forms.Form):
         if all(field in cleaned_data for field in ['source_code_parameter',
                                                    'buids']):
             # TODO: Detect if the file already contains parameters
-            cleaned_data['source_codes'] = process_spreadsheet(
+            cleaned_data['source_codes'] = process_file(
                 cleaned_data['source_code_file'],
                 cleaned_data['buids'],
                 cleaned_data['source_code_parameter'],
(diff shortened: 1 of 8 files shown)


tdphillips:
I'm not super pleased with this and we may not actually need it, hence the "GUIDANCE REQUESTED".

tdphillips:
Discussed with @aspidites.

- Fix the issue with source code being optional
- Do file type detection, which allows us to...
- ... make buid optional if we're looking at a csv

aspidites:
Comment for posterity: Gave feedback in person..

For anyone else curious about the use case: When adding entries in the destination manipulation tool, sometimes Seth would like to batch update some when there are errors, instead of having to open x number of tabs with an edit screen open.

automation/forms.py
```diff
@@ -22,56 +23,57 @@
...
 class SourceCodeFileUpload(forms.Form):
-    source_code_file = forms.FileField(required=True)
-    buids = forms.CharField(required=True)
+    source_code_file = forms.FileField(
+        required=True,
+        widget=forms.ClearableFileInput(attrs={'accept': '.xls,.xlsx,.csv'}))
+    buids = forms.CharField(required=False)
     source_code_parameter = forms.CharField(required=False)
 
-    def clean_buids(self):
```
tdphillips: All three fields are related in some way, which makes the various clean_field methods unpleasant to work with.

automation/source_codes.py
```diff
@@ -21,6 +25,14 @@ def get_book(location):
     return book
 
 
+def get_csv(location):
+    if not isinstance(location, (unicode, str)):
```
aspidites: basstring would work here. Not Python3 compatible, but I doubt we ever move to Python 3.

automation/forms.py
```diff
@@ -22,56 +23,57 @@
...
+        if is_excel:
+            if not buids:
+                self._errors['buids'] = [
+                    'Buids must be provided on Excel upload']
+            # Checks for jobs belonging to the buids provided and
+            # auto-determines an acceptable parameter based on the job. If
+            # this fails, a parameter must be provided in the form
+            test_job = Redirect.objects.filter(buid__in=buids,
+                                               expired_date__isnull=True)
+            if test_job.count():
```
aspidites: .exists()?

PR #2363, opened 2016-06-07
Correctly parse new email field format in PRM [PD 2428]

JLMcLaugh:
We weren't correctly matching users in PRM when the from field was formatted "Lastname, Firstname <name@example.com>".


mypartners/views.py
```diff
@@ -1192,10 +1192,25 @@ def process_email(request):
...
     recipient_emails_and_names = getaddresses(["%s, %s" % (to, cc)])
-    admin_email = getaddresses([admin_email])[0][1]
     contact_emails = filter(None,
                             [email[1] for email in recipient_emails_and_names])
 
+    admin_emails = [name_email[1] for name_email in getaddresses([admin_email])]
+
+    # Get the first valid user based on admin_emails. getaddresses can 
+    # return an extra tuple with an invalid email address.
+    admin_email = None
```
aspidites: Don't know if you'll like it better or not, but there's this:

```
from itertools import ifilter
```python
admin_user = next(ifilter(lambda email: User.get_email_owner(email, only_verified=True), admin_emails), None)
```

aspidites: NVM. Would need another traversal to get the user instance back from this.

aspidites: Just for completeness:
```python
admin_user = User.objects.get_email_owner(next(ifilter(lambda email: User.get_email_owner(email, only_verified=True), admin_emails, '')
```

Which returns the user where as the above returned an email address.

mypartners/views.py
```diff
@@ -1341,11 +1347,10 @@ def process_email(request):
 
         created_records.append(record)
 
-    logger.info("created_contacts: %(contacts)",
-                extra={"contacts": ", ".join([
-                    contact.email for contact in created_contacts])})
-    logger.info("unmatched_contacts: %(contacts)",
-                extra={"contacts": ", ".join(unmatched_contacts)})
+    logger.info("created_contacts: {contacts}".format(contacts=", ".join([
```
aspidites: The reason I didn't do this is that my linter complained, suggesting we should be using the loggers formatting strings.

PR #2371, opened 2016-06-09
[PD-2447] Re-enable multi-word tags

aspidites:
On production ,the only way to add a tag with multiple words is to surround each tag with quotes, as pressing space will cause a new tag to be created immediately. With this PR, new tags are created by pressing Tab, not space.

Other things I did:
- added an eslintrc file to static so that we could lint es5 scripts independently of es6 ones. 
- converted from STATIC_URL to {% static %} and {% compress %}. The only versioned file I updated was tags.js, as doing the others would have meant touching even more files. 
- made tags.js meet linting guidelines.
- fixed a typo that's been around since 2014 and stopped shadowing variables

All frontend template work so no new tests.


static/tags.js
```diff
@@ -0,0 +1,57 @@
+/* global $ */
+/* global createTags */
+
+$(document).ready(function() {
+  var validTags = [];
+  $('#p-tags').hide();
+  $('#p-tags').tagit({
+    allowSpaces: true,
```
aspidites: This was the actual fix. The rest of this PR was basically house cleaning.

static/tags.170-6.js
```diff
@@ -1,51 +0,0 @@
-$(document).ready(function() {
-  var validTags = [];
-  $("#p-tags").hide();
-  $("#p-tags").tagit({
-      showAtuocompleteOnFocus: !create_tags,
```
aspidites: Typo that I created and subsequently fixed.

static/tags.170-6.js
```diff
@@ -1,51 +0,0 @@
...
-      tagSource: function(search, showChoices) {
-          var value = $(".tagit-new > input").val(),
-              search = {value: value},
```
aspidites: Good old shadowed variables...

PR #2372, opened 2016-06-10
PD-2445 Minor style change to fix UL padding in menus and text.

misterjonscott:
Whoever merges this gets their choice of Chicken Tortilla or Steak Burger variety soup from Progresso.  Progresso® - Because it's Still Edible.

# Notes
Per [this ticket](https://directemployersfoundation.atlassian.net/browse/PD-2445), there was a weird margin on the left side of Select menus.  I've fixed that without causing weird list action elsewhere.

### DIFF
--- a/static/custom.css
+++ b/static/custom.css
@@ -100,6 +100,10 @@ textarea{
 }
 
 /*Lists*/
+ul {
+    margin: 0 0 0 20px;
+}
+
 li {
     color: #000;
 }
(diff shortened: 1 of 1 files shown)


tdphillips:
Chunky or nothing.

misterjonscott:
@tdphillips I mean, they're all pretty chunky.  I'm not sure what the measuring unit or test is for chunkiness, but I bet they're both satisfactorily chunky.

PR #2375, opened 2016-06-16
[PD-2439] Implement redux (bug fixes)

darrint:
Most of this change reworks a large complex action that was interacting with history into a couple of smaller ones.

I also found some flickers at load time and reworked the actions and dispatches to mitigate them.

The history behavior is greatly improved with this PR.


gulp/src/reporting/actions/compound-actions.js
```diff
@@ -131,87 +131,122 @@ export function doSetUpForClone(history, reportId) {
...
+ * We may need to fill in the data set menus.
  *
- * history: history object, in case we have to navigate
  * intention, category, dataSet: strings that together describe which report
  *   the user wants to run.
- * reportDataId: optional, the database id for the reportData previously in use
- * reportFilter: optional, prepopulate the filter
- * reportName: optional, use this name instead of a default name
  */
-export function doReportDataSelect(history, intention, category, dataSet,
```
darrint: This is the compound action creator that was split.

gulp/src/reporting/actions/compound-actions.js
```diff
@@ -131,87 +131,122 @@ export function doSetUpForClone(history, reportId) {
...
+export function doReportDataRedirect(history, intention, category, dataSet) {
+  return async (dispatch, getState, {api}) => {
+    try {
+      dispatch(markPageLoadingAction(true));
+      dispatch(markOtherLoadingAction('dataSetMenu', true));
+      const menu = await api.getSetUpMenuChoices(
+        intention || '',
+        category || '',
+        dataSet || '');
+      dispatch(replaceDataSetMenu({
```
aspidites: Not worried for this PR, but it may be worth having the API returning exactly what you want so you don't have to create all these aliases. Then payload just becomes "reportConfiguration" or whatever.

aspidites: We should probably rename dataSet in the future. Very confusing as it's not an array or an object, despite having 'set' in the name.

gulp/src/reporting/components/ReportRedirect.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+    dispatch(doReportDataRedirect(history, intention, category, dataSet));
+  }
+
+  render() {
+    return <div/>;
+  }
+}
+
+ReportRedirect.propTypes = {
```
aspidites: Document prop types or note that they are documented elsewhere.

aspidites:
Left 3 actionable comments. 1 which should be addressed before merge (documentation).

aspidites:
Confirmed history bug and highlight bugs fixed. Will thumbs up once comments are updated. Will merge if someone reviews by the time I'm ready to thumbs up.

gulp/src/reporting/components/ReportRedirect.jsx
darrint: Added docs.

gulp/src/reporting/actions/compound-actions.js
darrint: * https://directemployersfoundation.atlassian.net/browse/PD-2459
* https://directemployersfoundation.atlassian.net/browse/PD-2460

PR #2376, opened 2016-06-16
[PD-2446] Non-User Outreach in Redux

aspidites:
# Description
__*note:* You'll need to run `npm install` as some dependencies have been updated__

This started out as just porting NUO to Redux, but turned into a few more things:
- I redesigned the inbox component, as the original components made state management tricky, even with Redux (especially when trying to eschew component-level state)
- I addressed some styling issues that cropped up. I suspect that as we weren't actively working on this app tha when we styled components elsewhere it had unintended side consequences on this one, such as double/triple borders
- I implemented navigation and history, except for the outreach table, since that's an external component. I may investigate adding history to the pagination there in a separate ticket
- I consolidated a lot of modules (both components and utility modules. 
- I added tests for *all* state manipulation, while at the same time removed the existing API tests, as they were really just testing mocks. I think they may also have done some state comparisons, but those should be covered in the aforementioned tests
- I made liberal use of react-bootstrap, which also meant I needed to update this dependency. An extra benefit of doing so is that components styled this way aren't interfered with by whatever overrides we may have done (in 99% of cases anyway). 

# Testing
When comparing with the original implementation, you should notice the following:
- overview, inbox management, and outreach records have their own url, and can be navigated to via history
- focus is always changed to the first item in the list , which is where new inboxes are added
- when an inbox is added, you should see it immediately (before, you had to click on overview, then back to inbox management to see your changes, possibly having to refresh as well)
- you will be warned if attempting to add an inbox which already exists

The table in the outreach records page is largely unchanged other than using redux. As such, there is no history for pagination. To test it out, you'll need to populate records (not part of the intereface yet, separate ticket). I did so using @koonsies  script [here](https://github.com/DirectEmployers/MyJobs/pull/2286).

# Before
2016-06-16-151305_1916x1056_scrot
2016-06-16-151312_1916x1056_scrot
2016-06-16-151330_1916x1056_scrot

# After
2016-06-16-145721_1916x1056_scrot
2016-06-16-145732_1916x1056_scrot
2016-06-16-145756_1916x1056_scrot
2016-06-16-145846_1916x1056_scrot


aspidites:
Forgot to re-add tips:
2016-06-16-162839_1148x1056_scrot

aspidites:
I should also note that I manually serialized api responses. Wasn't a fan of Django's serialization format, so cut it down to something simpler.

aspidites:
@darrint  I confirmed this works on IE8, Chrome, and Firefox.

aspidites:
Just realized I need to write documentation. I'll get on that tomorrow morning.

aspidites:
The loading indicator will be fixed with [this](https://directemployersfoundation.atlassian.net/browse/PD-2457) ticket, thus I'm intentionally ignoring it here.

aspidites:
@darrint correction: IE8 is choking on the outreach record page now.

aspidites:
Finished writing documentation.

PR #2379, opened 2016-06-17
[PD-2458] 10% project: Fix modal

darrint:
This PR eliminates the constant warnings about the use of `confirm`.

For example usage see `main.jsx` and `User.jsx` in manage users.

As a bonus, `myprofile` and `mypartners` each have a minimal redux store now.


aspidites:
>  As a bonus, myprofile and mypartners each have a minimal redux store now.

Nice!

gulp/src/common/actions/confirm-actions.js
```diff
@@ -0,0 +1,53 @@
...
+ * dialog box.
+ *
+ * dispatch: store dispatch function
+ * message: optional, confirmation message to display.
+ *
+ * For this to work there should be exactly 1 <Confirm> component included
+ * in the app somewhere. A `react-router` root component tends to be a
+ * good spot.
+ */
+export async function runConfirmInPlace(dispatch, message) {
```
aspidites: Are we no longer using 'do' prefix for impure action creators?

gulp/src/common/actions/confirm-actions.js
```diff
@@ -0,0 +1,53 @@
...
+  let innerReject;
+  let innerResolve;
+  const waiter = new Promise(function confirmPromise(resolve, reject) {
+    innerResolve = resolve;
+    innerReject = reject;
+  });
+
+  while (!innerResolve) {
```
aspidites: When do we ever clear this timer? Indeed, why do we need a timer at all?

gulp/src/common/reducers/confirm-reducer.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * current state of a singleton confirmation dialog box: {
+ *
+ *  data: {
+ *    show: is the dialog showing right now?
+ *    message: string to show in the box. i.e. "Are you sure?"
+ *  },
+ *
+ *  These are the resolve/reject functions of a promise. If present, it is
+ *  assumed that these are associated with an active promise somewhere.
```
aspidites: "an active promise somewhere" -- makes me think this will be notoriously difficult to debug if things ever go wrong, since the location of that promise isn't known precisely.

gulp/src/common/ui/Confirm.jsx
```diff
@@ -0,0 +1,55 @@
...
+  const {show, message, onResolve} = props;
+
+  return (
+    <Modal show={show} onHide={() => onResolve(false)}>
+      <Modal.Body>
+      {message}
+      </Modal.Body>
+      <Modal.Footer>
+          <button
+            className="button primary wide"
```
aspidites: There is a component for this:
```jsx
<Button
    bsStyle="primary"
    block
    onClick={e => {e.preventDefault(); onResolve(false);}>
    Ok
</Button>
```

aspidites: Also, would it make more sense for onResolve to also take an event as a parameter and have it call preventDefault instead?

gulp/src/common/ui/Confirm.jsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+Confirm.defaultProps = {
+  message: 'Are you sure?',
+};
+
+export default connect(s => ({
+  show: s.confirm.data.show,
+  message: s.confirm.data.message,
+  onResolve: s.confirm.resolve,
```
aspidites: Bikeshed: I don't know about 'onResolve' - why not continue calling it 'onClick'. If it's delayed, maybe 'onSubmit'/'onConfirm'?

gulp/src/manageusers/Role.jsx
```diff
@@ -295,11 +300,12 @@ class Role extends React.Component {
...
+  dispatch: React.PropTypes.func.isRequired,
   location: React.PropTypes.object.isRequired,
   params: React.PropTypes.object.isRequired,
   callRolesAPI: React.PropTypes.func,
   history: React.PropTypes.object.isRequired,
   api: React.PropTypes.object,
 };
 
-export default Role;
+export default connect()(Role);
```
aspidites: For anyone else who is curious -- connect is being used here so that the component has access to `dispatch`.

gulp/src/common/actions/confirm-actions.js
darrint: This one is different. You don't dispatch it. It wants to dispatch several actions _and_ return a promise with a boolean.

I can write it in do form, dispatch it, and return the result with an await, but it has the side effect of dispatching the result (a boolean) through redux before returning it. That's theoretically harmless but a change in middleware might change that.

What do you think?

gulp/src/common/actions/confirm-actions.js
darrint: I added more comments to indicate the reason for the timer and promise.

I think i'm ok with not clearing the timer (it's a timeout, not an interval):

http://stackoverflow.com/questions/7391567/when-using-settimeout-do-you-have-to-cleartimeout

aspidites:
Modal isn't centered on the page vertically.

aspidites:
See PRM for example:
2016-06-20-145232_1916x1056_scrot

aspidites:
Also note the button placement in comparison to  my screenshot above:
2016-06-20-145431_956x1056_scrot

PR #2380, opened 2016-06-17
Fix counts of referral and communication records in reporting  pd 2388

JLMcLaugh:
We were assuming that contact name was enough to uniquely identify a row in contact reports, which led to double-counting when we were showing counts for each individual contact. This uses a combination of Partner, Contact, and contact email to uniquely identify rows and get the correct counts.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -589,20 +589,20 @@ class ContactRecordQuerySet(SearchParameterQuerySet):
         """
 
         all_contacts = self.values(
-            'partner__name', 'partner', 'contact__name',
+            'partner__name', 'partner', 'contact__name', 'contact__id',
             'contact_email').distinct().order_by('partner__name')
 
         records = dict(self.exclude(contact_type='job').values_list(
-            'contact__name').annotate(
-                records=models.Count('contact__name')).distinct())
+            'contact__id').annotate(
+                records=models.Count('contact__id')).distinct())
 
         referrals = dict(self.filter(contact_type='job').values_list(
-            'contact__name').annotate(
-                referrals=models.Count('contact__name')).distinct())
+            'contact__id').annotate(
+                referrals=models.Count('contact__id')).distinct())
 
         for contact in all_contacts:
-            contact['referrals'] = referrals.get(contact['contact__name'], 0)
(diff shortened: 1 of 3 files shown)


mypartners/models.py
```diff
@@ -588,22 +610,28 @@ def contacts(self):
...
-            'partner__name', 'partner', 'contact__name',
-            'contact_email').distinct().order_by('partner__name')
+        all_contacts = self.values('partner__name', 'partner',  'contact__name',
+                'contact', 'contact_email').distinct().order_by('partner__name')
 
-        records = dict(self.exclude(contact_type='job').values_list(
-            'contact__name').annotate(
-                records=models.Count('contact__name')).distinct())
+        # Fields required to identify distinct instances. Names are redundant.
+        distinct_fields = ('partner', 'contact', 'contact_email')
```
JLMcLaugh: Using these 3 fields instead of just contact__name is the main change.

mypartners/models.py
```diff
@@ -578,6 +578,28 @@ def hires(self):
     def referrals(self):
         return self.filter(contact_type='job').count()
 
+    @staticmethod
+    def _dict_from_values(values, key_fields, value_field):
```
aspidites: Seems pretty generic to live in models. Would universal.helpers be a better place?

mypartners/models.py
```diff
@@ -578,6 +578,28 @@ def hires(self):
...
+        """
+        Sorts a single value from values into a dictionary keyed by key_fields.
+        Used to match distinct items between querysets based multiple field 
+        values.
```
aspidites: The wording of this is a confusing. I only know what this method does by reading the implementation.

JLMcLaugh: Yeah, I tried a few variations on list of dictionaries to dictionary of dictionaries that didn't sound confusing, but didn't have much success. I figured values and values_list are common enough that it's ok to check the documentation.. Honestly, this is just meant to be a helper function preventing the copy/paste that was going on between referrals and records. I don't think it's universal enough to factor out of the app or model without causing more confusion. I'll put an example input/output to help clear it up Monday. I'm open to other suggestions though.

JLMcLaugh:
I see a failing test anyway. I'll close this until I look into it.

PR #2383, opened 2016-06-20
[PD-2420] manage users - add check when editing self as user 

lemniscate613:
You've always had to have at least one role to save the User form

We were not checking that that role had the 'read roles' activity -- which is required to view Manage Users at all.


gulp/src/manageusers/User.jsx
```diff
@@ -132,11 +132,45 @@ class User extends React.Component {
...
       });
       return;
     }
 
+    // If a user is editing their own account, they must have at least one role
+    // with the 'read role' activity, otherwise they'll be kicked out of
+    // manage users.
+
+    // Is user editing their own account?
+    if (parseInt(userID, 10) === parseInt(currentUserID, 10)) {
```
lemniscate613: Not necessary, but our linter doesn't like ==.

gulp/src/manageusers/User.jsx
```diff
@@ -56,8 +56,8 @@ class User extends React.Component {
     const {action} = this.props.location.query;
 
     if (action === 'Edit') {
-      const results = await api.get('/manage-users/api/users/' + this.props.params.userId + '/');
```
lemniscate613: Making variables standard for, thanks @aspidites

myjobs/views.py
```diff
@@ -1060,6 +1060,15 @@ def api_get_users(request):
                         content_type="application/json")
 
 
+def api_get_current_user(request):
```
JLMcLaugh: I thought this was going on the existing api_get_users api. Is there anywhere we call this without also calling api_get_users?

https://github.com/DirectEmployers/MyJobs/pull/2354#discussion_r65903024

gulp/src/manageusers/User.jsx
```diff
@@ -56,8 +56,8 @@ class User extends React.Component {
...
-      const userObject = results[this.props.params.userId];
+      const results = await api.get('/manage-users/api/users/' + this.props.params.userID + '/');
```
aspidites: [This](https://github.com/DirectEmployers/MyJobs/pull/2354#discussion_r65903024) has still not been addressed.

lemniscate613: Most of the disagreement about how to implement this has been offline, like at stand ups. I back tracked and folded this user_id bit into a preexisting API.

myjobs/views.py
lemniscate613: Addressed. The team was split about how to do this, but most of the discussion was offline. I switched implementation so this bit of data into a preexisting API (`api_get_users`).

PR #2384, opened 2016-06-20
Nuo intake process [PD-1749]

tdphillips:
I've got a bit of work ahead but thought I should solicit feedback on what I currently have.

Remaining:
- [x] Multiple companies
- [x] Query OutreachEmailDomain
- [x] Test what happens if a user sends to one of these addresses but is not attached to the company
- [x] ~~Move all of the non-views from views.py (to helpers.py? :sob:)~~ Unnecessary
- [x] ~~Create entries in SendGrid when an OutreachEmailAddress is created/updated~~ Not necessary
- [x] Modify redirect to forward emails that contain outreach email addresses
- [x] Ensure secure actually tests BCC
- [x] :cow::bell: 
- [x] Insert your own here


mypartners/views.py
```diff
@@ -1194,32 +1197,114 @@ def process_email(request):
     recipient_emails_and_names = getaddresses(["%s, %s" % (to, cc)])
     contact_emails = filter(None,
                             [email[1] for email in recipient_emails_and_names])
+    # TODO: confirm mysql is case-insensitive
+    NUO_HOSTS = OutreachEmailAddress.objects.filter(email__in=[
+        contact.split('@')[0] for contact in contact_emails])
+    NUO_LOCAL = [host.email for host in NUO_HOSTS]
+    admin_user, admin_email, is_nuo, error = get_admin(admin_email, NUO_HOSTS,
```
tdphillips: Almost every chunk that I split out of process_email contains the possibility of returning an HttpResponse. Is this a bad pattern to follow?

mypartners/views.py
```diff
@@ -1194,32 +1197,114 @@ def process_email(request):
...
+    if error is not None:
+        return error
+
+    newrelic.agent.add_custom_parameter("to", to)
+    newrelic.agent.add_custom_parameter("cc", cc)
+    newrelic.agent.add_custom_parameter("admin_email", admin_email)
+    newrelic.agent.add_custom_parameter("contact_emails",
+                                        ", ".join(contact_emails))
+
+    contact_emails, date_time = determine_contacts(
```
tdphillips: I don't like how long some of these argument lists are.

mypartners/views.py
```diff
@@ -1194,32 +1197,114 @@ def process_email(request):
...
-        admin_user = User.objects.get_email_owner(admin_email, only_verified=True)
-        if admin_user is not None:
+        try:
+            email_validator(admin_email)
+        except ValidationError:
+            pass
+        else:
             break
 
+    def is_proper_domain(admin_email, companies):
```
tdphillips: It was at this point that I became more sad than normal. In order to adequately ensure things happen as they're supposed to, different things need to happen in the four separate branches below based on some combination of is_nuo (defined a few lines below) and the return value of this function call.

mypartners/views.py
```diff
@@ -1315,46 +1424,49 @@ def process_email(request):
...
-
-    logger.info("created_contacts: {contacts}".format(contacts=", ".join([
-                    contact.email for contact in created_contacts])))
-    logger.info("unmatched_contacts: {contacts}".format(contacts=", ".join(
-                    unmatched_contacts)))
-    send_contact_record_email_response(created_records, created_contacts,
-                                       attachment_failures, unmatched_contacts,
-                                       None, admin_email)
-    return HttpResponse(status=200)
+        if admin_user:
```
tdphillips: All instances of admin_user will probably be modified at some point to include an "is_nuo" check.

lemniscate613:
Your list needs more cowbell.

tdphillips:
Added just for you.

darrint:
It might be helpful to expound a bit on why it's important in `determine_contacts` if we are dealing with a "forward".

mypartners/helpers.py
```diff
@@ -264,16 +264,26 @@ def get_records_from_request(request):
 
 def send_contact_record_email_response(created_records, created_contacts,
                                        attachment_failures, unmatched_contacts,
-                                       error, to_email):
+                                       error, to_email, companies=None,
+                                       is_nuo=False, buckets=None):
+    bucket = ''
```
tdphillips: I'm going to modify how we determine the list of outreach email addresses to limit it to a single company. That will simplify the changes that will need to be made to this function.

tdphillips:
Feel free to ask as many questions as you want. I'm sure there are sticking points given how long I've been working on this.

tdphillips:
I'm guessing we probably don't want the "for additional assistance" bit.

2016-06-24-160604_666x843_scrot

PR #2387, opened 2016-06-22
[PD-2440] Added detrack to various page templates

aspidites:
Copied from Jira ticket:
> Search pages (no results functionality included on page)
`<script id="detrack" defer src="https://d2e48ltfsb5exy.cloudfront.net/t/t.js?i=0,2"></script>`
Results pages (can include a search)
`<script id="detrack" defer src="https://d2e48ltfsb5exy.cloudfront.net/t/t.js?i=0,3"></script>`
Job view pages
`<script id="detrack" defer src="https://d2e48ltfsb5exy.cloudfront.net/t/t.js?i=0,1"></script>`
static pages
`<script id="detrack" defer src="https://d2e48ltfsb5exy.cloudfront.net/t/t.js?i=0"></script> `

If a template had a what/where box, but didn't iterate through jobs, I used the first. Otherwise, if it had a what/where box and did iterate through jobs, I used the second. I added the third to the search results template, and added the fourth to the two templates with static in their name.

### DIFF
--- a/templates/home_page/home_page_billboard.html
+++ b/templates/home_page/home_page_billboard.html
@@ -1,3 +1,6 @@
 {% extends "seo_billboard_homepage_base.html" %}
 {% load seo_extras %}
 {% block after_billboard %}{% logo_carousel %}{% endblock after_billboard %}
+{% block extra-js %}
+    <script id="detrack" defer src="https://d2e48ltfsb5exy.cloudfront.net/t/t.js?i=0"></script> 
+{% endblock %}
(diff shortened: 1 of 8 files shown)


tdphillips:
Do job_detail.html and job_listing.html need modification?

aspidites:
>  job_detail.html
Yeah. I misspoke earlier, too. This is what the third case was for. Search results by themselves are on the second case. 

> job_listing.html
I assume that instead of search_results.html being modified, I should be modifying this one.

koonsies:
:+1:

PR #2394, opened 2016-06-23
[PD-2311] Empty partner/contact filter now disables run button.

aspidites:
Enabled:
2016-06-23-160010_1920x1080_scrot

Disabled:
2016-06-23-160003_1920x1080_scrot

### DIFF
--- a/gulp/src/common/ui/HelpText.jsx
+++ b/gulp/src/common/ui/HelpText.jsx
@@ -3,7 +3,7 @@ import React, {PropTypes} from 'react';
 /* HelpText
  * Component for displaying a validation error.
  */
-export function HelpText(props) {
+export default function HelpText(props) {
   const {message} = props;
   return (
     <div className="input-error">
--- a/gulp/src/nonuseroutreach/components/Inbox.jsx
+++ b/gulp/src/nonuseroutreach/components/Inbox.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {HelpText} from './HelpText';
+import HelpText from '../../common/ui/HelpText';
 import {
   Button,
   ButtonGroup,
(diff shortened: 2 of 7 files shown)


koonsies:
In reviewing the ticket and checking this functionality, I give adherence to the ticket a big ol 👍 

That said.. In witnessing this functionality, I think without some messaging that explains why the "run report" button is turned off, this will create more questions than it solves. If the reason "run report" is grayed out is because the report would be empty, could we perhaps add a small bit of text that explains that? Or at very least, some alt text on the button itself?

Furthermore, @darrint @misterjonscott I don't see that the UI is sending empty lists when filter is set to "Filter by name" but no contact/partner is selected. Am I incorrect in thinking that setting the filter but not selecting any partners/contacts was supposed to result in no results??

aspidites:
> That said.. In witnessing this functionality, I think without some messaging that explains why the "run report" button is turned off, this will create more questions than it solves. If the reason "run report" is grayed out is because the report would be empty, could we perhaps add a small bit of text that explains that? Or at very least, some alt text on the button itself?

Addressed:
2016-06-24-162745_1916x1056_scrot

> Furthermore, @darrint @misterjonscott I don't see that the UI is sending empty lists when filter is set to "Filter by name" but no contact/partner is selected. Am I incorrect in thinking that setting the filter but not selecting any partners/contacts was supposed to result in no results??

I think the meeting we had the other day kind of threw that logic out the window. As I understand it, @darrint  is currently reworking that logic as we speak. I don't quite remember the result of the meeting, but I think the new behavior will be:
-  user selects "No Filter", in which case we don't filter on that field
- The user selects "filter by name" but doesn't actually add a name - we treat it as an incomplete form and just filter as though "No Filter" was selected
- An explicit "only untagged items" entry will be added for tag filtering

So the only time an empty list should be sent is when asking for untagged items, and even then, Darrin might just opt to send a null or something.

PR #2399, opened 2016-06-27
[PD-1694] Changed date range to allow for future dates.

aspidites:
- reduced date range to 10 years.
- rather than range being from 2016-2006, it is now 2011-2020

### DIFF
--- a/gulp/src/common/ui/CalendarPanel.jsx
+++ b/gulp/src/common/ui/CalendarPanel.jsx
@@ -17,14 +17,18 @@ export default class CalendarPanel extends React.Component {
   renderDay(currentDay) {
     const {onSelect, day} = this.props;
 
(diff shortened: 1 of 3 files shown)


lemniscate613:
Some contexts that use this component need a range larger than 10. We should probably pass it in instead of having it built in.

For example, I'm using this in my profile when you select your educational background. Some people need to say they went to a university from 1964-1968. I replicated current funcitonality in allowing this large of a range.

aspidites:
@lemniscate613, addressed:

New prop types:
- numberOfYears - determines how many years will appear in the dropdown
- pastOnly - determines whether or not the current year should be the highest available year, or sit in the middle of the available years

At some point we may find it better to have a date range prop or a pair of startYear/endYear props, but this should be good enough for now, as it accomplishes what I set out to do without breaking existing code.

PR #2400, opened 2016-06-28
Prefer local data files if they exist

tdphillips:
Handy in the event of a total network outage. Runserver doesn't work otherwise.

### DIFF
--- a/myprofile/templatetags/selection_boxes.py
+++ b/myprofile/templatetags/selection_boxes.py
@@ -123,7 +123,7 @@ def _load_json_data(json_file):
     Retrieves an internal json file, opens it, and reads it in as a dict.
 
     Inputs:
-    :jason_file:     The location of the json data file
+    :json_file:     The location of the json data file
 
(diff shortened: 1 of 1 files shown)


myprofile/templatetags/selection_boxes.py
```diff
@@ -26,9 +28,12 @@ def country_select(selected="usa", html_id="", input_name="country",
...
     """
     selected = selected.lower()
-    data_url = 'https://d2e48ltfsb5exy.cloudfront.net/myjobs/data/countries_167-24.json'
-    data_list = _load_json_data(data_url)
-    country_list = data_list["countries"]
+    data_file = os.path.join(settings.PROJ_ROOT,
+                             'jsondata/countries_167-24.json')
+    data_list = _load_json_data(data_file)
+
+    country_list = data_list['countries']
```
tdphillips: I know this could result in a KeyError but that possibility existed before this modification. This erroring is not a regression.

PR #2402, opened 2016-06-28
[PD-2485] Stopped using grid as it's adding container which we don't want.

aspidites:
Apparently `<Grid />` means `<div class="container"`, which is of course a very bad idea in the middle of a layout that's already started. I didn't simply remove it since there can only be one root node in a react component (eg. two rows side by side is a no go, but two rows wrapped in a div is fine).


lemniscate613:
TIL

PR #2403, opened 2016-06-29
[PD-1683] Make "nofollow" count facet slugs properly

koonsies:
## Overview

This is the ACTUAL PR for 1683. The previous PR was actually PD-2478 in disguise!

Basically, this PR causes the filter counts to properly reflect multiple filters on the facet_slugs category. Now, after more than one facet_slug filter is applied, the robots "nofollow" meta tag will be properly applied

## Testing

Navigate to the following links and check the source for the following line..
`<meta name="robots" content="nofollow">`

http://ANYSTATE.jobs/operator-jobs/new-jobs/   (should NOT have the above tag)
http://ANYSTATE.jobs/manager-jobs/operator-jobs/new-jobs/  (should have nofollow)
http://ANYSTATE.jobs/business-jobs/manager-jobs/operator-jobs/new-jobs/  (should have nofollow)

### DIFF
--- a/myblocks/models.py
+++ b/myblocks/models.py
@@ -822,13 +822,20 @@ class Page(models.Model):
     def get_template(self, request):
         filters = context_tools.get_filters(request)
 
+        facet_slugs = []
+        if filters.get('facet_slugs', None):
+            facet_slugs = filters['facet_slug'].split('/')
+
         context = self.context(request)
         context.update({
             'body': mark_safe(self.get_body()),
             'google_analytics': context_tools.get_google_analytics(request),
             'head': mark_safe(self.get_head()),
             'max_filter_settings': settings.ROBOT_FILTER_LEVEL,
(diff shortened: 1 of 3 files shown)


myblocks/models.py
```diff
@@ -822,13 +822,20 @@ def get_head(self):
     def get_template(self, request):
         filters = context_tools.get_filters(request)
 
+        facet_slugs = []
+        if filters.get('facet_slugs', None):
```
JLMcLaugh: extra_facet_slugs = filters.get('facet_slugs', '').count('/') would get rid of a couple lines of code and remove the extra k != 'facet_slug' logic at 838.

For future reference, None is the default for the second argument in dict.get, so you wouldn't need to include it here.

JLMcLaugh: I may be missing something in the above comment. In search_views we need the facet_slugs themselves. Do we not use them here?

PR #2406, opened 2016-06-30
[PD-2464] Add job data to Mongo

mklauber:
Tasks for adding job data to mongo when we run imports on a company.  

@fezick, @JLMcLaugh, @haliscooper How should we plan to QC this?


tdphillips:
I take it everyone who wanted to review has by now.

PR #2408, opened 2016-07-05
Update default_settings.py

tdphillips:
This runs every minute. If we define a value for minute, it should run every hour.

Outstanding question: Do we need to continue updating solr? The solr collection in this instance is being used for candidate dashboard. @fezick, @JLMcLaugh 

Similarly, do we really need to run update_solr_from_model every five minutes?


mklauber:
:+1:

PR #2412, opened 2016-07-06
[PD-2457] Loading Indicators in NUO

darrint:
This implements loading indicators for NUO.

## Test

This testing procedure is a little flaky but it does allow you to see the indicator after a few tries.

1. In chrome, open the debugger and go to the networking tab.
2. Go to throttling and create a new profile. "High Latency".
3. Set upload and download to 30000 and the latency to 2000. Save it.
4. Restart the debugger and refresh the page.
5. Go to networking again and select the high latency profile.
6. Refresh the page again. (just in case) It should be slower.
7. Flip frequently between the pages using the buttons in the sidebar.

Note that loading the spinner image is affected by the network latency as well as the XHR requests so it may not show up right away.

### DIFF
--- a/gulp/src/nonuseroutreach/components/NonUserOutreachApp.jsx
+++ b/gulp/src/nonuseroutreach/components/NonUserOutreachApp.jsx
@@ -3,6 +3,10 @@ import {Col, Row} from 'react-bootstrap';
 import {connect} from 'react-redux';
 import {Loading} from 'common/ui/Loading';
 import {Menu} from './Menu';
+import InboxManagementPage from './InboxManagementPage';
+import {markPageLoadingAction} from '../../common/actions/loading-actions';
+import {doGetInboxes} from '../actions/inbox-actions';
+import {setPageAction} from '../actions/navigation-actions';
 
 
 /* NonUserOutreachApp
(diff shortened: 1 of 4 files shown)


aspidites:
Haven't looked in detail, but I do like that only 1 component (the top-level component at that) needs the componentWillMount nastiness, now.

gulp/src/nonuseroutreach/components/NonUserOutreachApp.jsx
```diff
@@ -11,6 +17,43 @@ import {Menu} from './Menu';
...
+
+  componentWillUnmount() {
+    this.unsubscribeToHistory();
+  }
+
+  async handleNewLocation(_, loc) {
+    const {dispatch} = this.props;
+
+    const lastComponent = loc.components[loc.components.length - 1];
+    if (lastComponent === InboxManagementPage) {
```
aspidites: I'd prefer a switch here instead of if/thens with equality checks on the same variable.

koonsies:
:+1: I see spinners and they are good.

lemniscate613:
"Let there be spinners." -Darrin

PR #2414, opened 2016-07-07
[PD-2487, PD-2493] User Management - Activities Refactor

aspidites:
Original plan was to just port some things to Redux. I ended up getting distracted. Rather than submit a single PR when I'm done, I decided to submit PRs for each section of User Management that I work on. This particular ticket reorganizes the app so that it more closely resembles the other work we did. It was original written before we all started converging on strategies, so I wanted to get it up to date before things started to shift again. I also refactored the activities API so that we didn't have to do so much work to munge data on the front end. Finally, `confirm` is a built-in function (at least in DOM environments), so I renamed it `confirmation` to prevent confusion (and get my editor to stop yelling at me). 

Next, I'll be working on the User(s) components.


koonsies:
👍  functional, looks good.

PR #2417, opened 2016-07-07
Hide the country if it doesn't exist on the job detail page.

mklauber:


templates/job_detail.html
```diff
@@ -43,14 +43,16 @@
...
+                        <span itemprop="addressLocality">{{ the_job.city }}</span>
                     {% endif %}
                   {% endif %}
                   {% is_special_case the_job as special_case %}
                   {% if special_case %}
                       <meta itemprop="addressCountry" content="{{ the_job.country }}" />
                   {% else %}
-                      <span itemprop="addressCountry">{{ the_job.country }}</span>
+                      {% if the_job.country %}
+                          , <span itemprop="addressCountry">{{ the_job.country }}</span>
```
tdphillips: This will probably result in a space between city and country, won't it?

mklauber: I'm trying to hide the comma when we don't have a country to add, but I'm not sure, it might end up being `<city>[space],[space]<country>`

JLMcLaugh: There's a space inside the {{ }} and after the comma. You'll have 2 spaces.

koonsies:
Test failed

PR #2418, opened 2016-07-07
[PD-2463] add support for new templates

koonsies:
## Overview

This ticket involves creating the logic for switching between versions of templates. The basic idea is that versions of templates can be added to the choices tuple on the configuration model. The first value of that tuple will be the path of the template subdirectory containing the new version of templates.

For instance, 
```
    TEMPLATE_VERSION_CHOICES = (
        ('v1', 'Version 1'),
        ('v2', 'Version 2'),
    )
```

If a configuration has a template set of v2, the path to put the templates would be

` templates/v2/<path_to_template> `

The exception are v1 templates, which refer back to the base templates directory.

## Testing

You may notice the tiny, worthless template added to /v2/job_listing.html. That template exists solely to test that the switching does work when a configuration is set to use version 2 templates. To test, simply go to the admin for the configuration of a microsite, change the template version to v2 and go navigate to <thatmicrosite>.jobs/jobs/

## Moving forward

As we develop version 2 templates and put them in the templates/v2 directory, we will simply need to wrap any calls to that template in the `site_config.get_template` method. An example is found in the code:

```
    return render_to_response(site_config.get_template('job_listing.html'),
                              data_dict, context_instance=RequestContext(request))
```

### DIFF
--- a/seo/views/search_views.py
+++ b/seo/views/search_views.py
@@ -1802,8 +1802,8 @@ def search_by_results_and_slugs(request, *args, **kwargs):
         'widgets': widgets,
         'analytics_info': get_analytics_info()
     }
-    return render_to_response('job_listing.html', data_dict,
-                              context_instance=RequestContext(request))
+    return render_to_response(site_config.get_template('job_listing.html'),
+                              data_dict, context_instance=RequestContext(request))
 
 
 def get_analytics_info():
--- a/templates/v2/job_listing.html
+++ b/templates/v2/job_listing.html
@@ -1,0 +1,1 @@
+<h1>I'm a version 2 template!! Behold!!</h1>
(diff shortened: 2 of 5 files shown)


seo/views/search_views.py
```diff
@@ -1802,8 +1802,8 @@ def search_by_results_and_slugs(request, *args, **kwargs):
         'widgets': widgets,
         'analytics_info': get_analytics_info()
     }
-    return render_to_response('job_listing.html', data_dict,
-                              context_instance=RequestContext(request))
+    return render_to_response(site_config.get_template('job_listing.html'),
+                              data_dict, context_instance=RequestContext(request))
```
koonsies: This technically doesn't have to be wrapped yet, as we don't have a v2 template for job_listing, but this allows testing of the below template

templates/v2/job_listing.html
```diff
@@ -0,0 +1 @@
+<h1>I'm a version 2 template!! Behold!!</h1>
```
koonsies: You'll know if you're seeing the "right" template..

mklauber:
:+1:

PR #2420, opened 2016-07-08
Add redirect job views to Mongo [PD-2489]

tdphillips:
Requires a deployment change.

Includes just as many tests as the Analytics version. I can come up with contrived examples if anyone wants coverage.


haliscooper:
Why not use .utcnow().isoformat()?  Just curious, I don't know how that works in python...

tdphillips:
I assume you're talking about the bit where I'm doing `now.isoformat()[:-6] + 'Z'`.

I should probably just do that. What exists now was an attempt to keep the times exact between DirectTraffic and the new analytics platform. A millisecond difference probably doesn't matter.

To do that, I was reusing the datetime that we were using to calculate expired dates. That datetime has to be timezone-aware (`.utcnow().isoformat()` isn't timezone-aware despite sounding like it would be).

tdphillips:
Adding some documentation around the weird conditionals I added.

haliscooper:
My "to" datetime comes from the computer that is viewing the page it's called (what it thinks UTC time is), so it's not guaranteed either if the time is off on the computer, you'd hope most are...  
So, as long as the correct UTC time gets stored so it fits most of the time in the chain of events for that jobseeker, its good, I was just worried about having to go change the -6 when the time changes, is that what has to happen, or am I off in thnking that?

tdphillips:
`.isoformat()` returns a string. `[:-6]` strips off the last six characters which, given the datetime that we're using, are always "+00:00"

haliscooper:
Ha, thought it might be subtracting 6 hours, clear now, thanks.  Python string manipulation is not something I know at all ;-)

PR #2424, opened 2016-07-11
[PD-2455] Password Expiration

darrint:
## Test

Use the admin to turn on password expiration for a company.
Exercise the following limits:

* 90 day password expiration. You can mess with the `password_last_modified` attribute of the user to test this. (sorry no admin, use the shell)
* A history of 4 known past passwords. You can't reuse these. They rotate.
* Lockout after 6 attempts. You can reset this by resetting your password via email.


backends.py
```diff
@@ -16,7 +17,23 @@ class CaseInsensitiveAuthBackend(ModelBackend):
...
             user = User.objects.get(email__iexact=username)
+            if user.is_locked_out():
+                raise PermissionDenied("User is locked out")
             if user.check_password(password):
                 return user
         except User.DoesNotExist:
             return None
+
+
+class CaseInsensitiveAuthFailCatcher(ModelBackend):
```
tdphillips: Is it intended that this is being applied to all users?

darrint: The way it's written now it will note the failed login, but no lockout will be enforced unless expiration is being enforced.

lemniscate613:
Probably unrelated to this PR specifically, but I don't know what this error message means (it appears when trying to use a word like "dog" as a password):

"Based on a common sequence of characters"

lemniscate613:
I tried to use this password: mGO0oB95G@mz#2x

And got this error: "Must be more complex (Must contain 1 or more uppercase characters)"

So I added a capital letter: Pinky2fpdinky2f

And got this error: "Must be more complex (Must contain 1 or more punctuation character)"

I'm surprised the note about a punctuation character didn't appear at first.

lemniscate613:
This worked for me: a rolling history of the 4 last used passwords that cannot be reused

lemniscate613:
After one failed login, I got two errors, but only one had a message:

screen shot 2016-07-12 at 2 54 28 pm

lemniscate613:
I was successfully locked out after multiple attempts, but there is not a message indicating that I'm locked out (or what I need to do -- ie use reset password form -- to fix it).

lemniscate613:
Other than the above concerns, it looks good.

tdphillips:
The "two errors, one message" thing exists on production as well.

PR #2425, opened 2016-07-12
Forbid deletion of all roles from last company admin user [PD-2491]

JLMcLaugh:
We were ensuring that we didn't remove the admin role from the last admin user when editing roles, but not when deleting all company roles from a user. This adds a check to myjobs.api_delete_user.

### DIFF
--- a/myjobs/tests/test_api.py
+++ b/myjobs/tests/test_api.py
@@ -15,7 +15,7 @@ from setup import MyJobsBase
 
 class UserResourceTestCase(MyJobsBase):
     def setUp(self):
-        super(UserResourceTests, self).setUp()
+        super(UserResourceTestCase, self).setUp()
         create_api_key(User, instance=self.user, created=True)
         self.client = TestClient(
             path='/api/v1/user/',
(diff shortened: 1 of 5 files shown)


myjobs/tests/test_api.py
```diff
@@ -54,8 +55,42 @@ def test_existing_user(self):
             self.assertFalse(content['user_created'])
             self.assertEqual(content['email'].lower(), 'alice@example.com')
 
+class UserManagementTestCase(MyJobsBase):
+    def setUp(self):
+        super(UserManagementTestCase, self).setUp()
+        self.role.activities = self.activities
```
JLMcLaugh: This seems like something we'd want to do in the base setUp, but it was causing errors in test_activities. I didn't want to dive into those tests for the hotfix.

myjobs/tests/test_api.py
```diff
@@ -12,7 +13,7 @@
 from setup import MyJobsBase
 
 
-class UserResourceTests(MyJobsBase):
+class UserResourceTestCase(MyJobsBase):
     def setUp(self):
         super(UserResourceTests, self).setUp()
```
tdphillips: super call needs to be updated with the new class name.

myjobs/tests/test_api.py
aspidites: At one point I attempted to do so. It got very messy, which is why it's stayed this way. If it's tackled, it's probably better to tackle as a separate ticket.

tdphillips:
I can no longer merge this. It's on you, rest of the team.

PR #2439, opened 2016-07-20
[PD-2494, PD-2504] User Page for User Management -> Redux

aspidites:
The title is misleading, as I did way more than that. I've annotated the diff to give some highlights of the changes.

Before:
2016-07-20-132328_654x540_scrot

After:
2016-07-20-132131_663x296_scrot


gulp/package.json
```diff
@@ -37,7 +37,6 @@
     "jquery-param": "~0.2.0",
     "less": "~1.6.1",
     "lodash-compat": "~3.10.1",
-    "lodash.unionby": "^4.6.0",
```
aspidites: I was able to accomplish what I needed with a regular union after changing some logic to compare strings instead of objects.

gulp/src/common/myjobs-api.js
```diff
@@ -60,6 +60,37 @@ export class MyJobsApi {
   async delete(url, data) {
     return this.ajaxWithFormData('DELETE', url, data);
   }
+
+  /** Upload data to a server by submitting the contents of :data: as
```
aspidites: The pre-existing `post` method wasn't parsing data correctly, so I back-ported this from the 10% project I had lying around for Excel imports (hence the name). At some point, it may make sense to rename to `post` and get rid of the pre-existing version.

gulp/src/common/reducers/confirm-reducer.js
```diff
@@ -1,5 +1,11 @@
 import {handleActions} from 'redux-actions';
 
+export const initialConfirmation = {
```
aspidites: Initial state. I'm allergic to `undefined` as an argument value.

gulp/src/manageusers/actions/company-actions.js
```diff
@@ -0,0 +1,102 @@
+import {createAction} from 'redux-actions';
+import {errorAction} from '../../common/actions/error-actions';
+
+export const setErrorsAction = createAction('SET_ERRORS');
```
aspidites: Pure Actions (eg, no side effects).

gulp/src/manageusers/actions/company-actions.js
```diff
@@ -0,0 +1,102 @@
...
+// Note: Each of the asynchronous calls will dispatch an `errorAction` if an
+// exception was thrown.
+
+/**
+ * doRefreshUsers
+ *
+ * Asynchronously fetches an updated users object where keys are user ids and
+ * values represent pertinent user information.
+ */
+export function doRefreshUsers() {
```
aspidites: Asynchronous impure (eg side effects) actions.

gulp/src/manageusers/components/ManageUsersApp.jsx
```diff
@@ -1,47 +1,107 @@
...
+  }
+
+  async handleNewLocation(_, loc) {
+    const {dispatch} = this.props;
+    const lastComponent = loc.components[loc.components.length - 1];
+    const userId = loc.params.userId;
+
+    dispatch(markPageLoadingAction(true));
+
+    // refresh company data, including roles, users, and available activities
```
aspidites: This is very catch-all-y and probably not efficient. WE could do an optimization pass to see which parts of the UI actually need the results of each API call, which would reduce the number of calls needed in general. Given how many calls reporting does, for instance, I didn't feel to bad leaving it this aggressive for the time being.

gulp/src/manageusers/components/ManageUsersApp.jsx
```diff
@@ -66,41 +126,19 @@ export default class ManageUsersApp extends React.Component {
       rolesAPIResults: results,
     });
   }
-  async callUsersAPI() {
```
aspidites: This is now an action.

gulp/src/manageusers/components/ManageUsersApp.jsx
```diff
@@ -126,13 +164,10 @@ export default class ManageUsersApp extends React.Component {
 
           <div className="col-sm-8 col-xs-12">
             <div className="card-wrapper">
-              {this.props.children && React.cloneElement(
+              {loading ? <Loading /> : this.props.children && React.cloneElement(
```
aspidites: Once the roles page is converted, the `React,cloneElement` can be eliminated entirely.

gulp/src/manageusers/components/UsersList.jsx
```diff
@@ -1,29 +0,0 @@
-import React from 'react';
```
aspidites: Wasn't being used by anything else, so I folded the layouting bits into the parent component.

gulp/src/manageusers/formatForMultiselect.js
```diff
@@ -1,22 +0,0 @@
-// Input:
```
aspidites: As the API now returns the data in the right shape, there is no need to data munge on the frontend.

gulp/src/manageusers/reducers/company-reducer.js
```diff
@@ -0,0 +1,178 @@
+import {difference, forOwn, union} from 'lodash-compat';
```
aspidites: Given more time (and higher priority), I think I could fold some of this state into itself such that there are less bits to keep track of. These were the bits that I came up with first and seemed to do what I want. Plenty of room for improvement if someone feels lucky.

myjobs/tests/test_manage_users.py
```diff
@@ -342,56 +367,6 @@ def test_get_specific_role_contains_users(self):
         first_available_user_name = first_available_user['name']
         self.assertIsInstance(first_available_user_name, unicode)
 
-    def test_get_specific_user(self):
```
aspidites: This is no longer an end point, so no reason to test it.

myjobs/tests/test_manage_users.py
```diff
@@ -400,84 +375,56 @@ def test_get_users(self):
...
         Require POST
         """
-        response = self.client.get(reverse('api_create_user'))
-        output = json.loads(response.content)
-        self.assertEqual(output["success"], "false")
-        self.assertEqual(output["message"], "POST method required.")
+        response = self.client.get(reverse('api_add_user'))
+        self.assertEqual(response.status_code, 405)
 
     def test_create_user_user_must_have_role(self):
```
aspidites: This is covered by the test cases that Jason wrote (and I updated). That those tests had to be written suggests that this wasn't testing what we thought it was anyway.

myjobs/urls.py
```diff
@@ -61,6 +61,10 @@
     url(r'^manage-users/api/roles/$',
         'api_get_roles',
         name='api_get_roles'),
+    # TODO: Remove once roles page converted
```
aspidites: The reason I made a new end point instead of updating the existing one is that I suspect that the Roles page (which I haven't looked at yet) uses different information. I intended to see what the roles page needed and add those bits to `api_get_all_roles` and rename it to just `api_get_roles`.

aspidites:
Note that the proptypes are pretty generic. Spoke with @darrint  and found out what I need is `React.PropTypes.ObjectOf`. I plan to tackle that in a separate 1-2 point ticket if I have time.

PR #2447, opened 2016-07-26
Only attempt to add jobs to mongo if jobs exist.

mklauber:


import_jobs/mongo.py
```diff
@@ -32,13 +32,15 @@ def jobsfs_to_mongo(guid, buid, name, mongo=settings.MONGO_HOST):
...
-    
-    client = MongoClient(mongo, w="majority")
-    collection = client.analytics.jobs
-    bulk = collection.initialize_unordered_bulk_op()
-    for job in jobs:
-        bulk.find({'guid': job['guid']}).upsert().replace_one(job)
-    bulk.execute()
+    jobs = list(jobs)
+
+    if len(jobs) > 0:
```
tdphillips: This should work fine on a tuple.

PR #2449, opened 2016-07-27
[PD-2507] Start processing page with new search component

darrint:
This creates a new page for processing emails, with a control for searching. The control only works for partners so far, although there is a placeholder for contacts.

Once you select or partner or choose to add one, the control goes into a permanent state with a loading indicator and you have to refresh the page to interact with it again. This is **by design**. The control will disappear at this point when more of the page is implemented.

## Review

* I had to tweak the navigation to expose the recordId from history. Should I have exposed `recordId` directly from redux instead of the location query object?
* A `Button` became a `Link`.

## Test

1. Go to the outreach records page.
2. Pick a record to review.
3. Interact with the new control. See notes above.


koonsies:
Small note. When someone clicks "Cancel" after entering a new contact and receiving the "Cancel/Confirm" box, the focus should return to the input box if possible.

tdphillips:
Rebound seems like a lot of code just for smooth scrolling.

Anyway, works. :+1: pending @koonsies' feedback.

darrint:
I fixed the focus after cancel.

koonsies:
I'll merge when the tests pass, but the cancel logic looks good 👍

PR #2450, opened 2016-07-28
Fix script path

tdphillips:


redirect/log_to_mongo.py
```diff
@@ -9,6 +9,8 @@
 
 from django.conf import settings
 
+sys.path.insert(0, '..')
```
tdphillips: That's annoying. Suggestions?

PR #2458, opened 2016-08-01
[PD-2511] modify outreach table

koonsies:
## Overview

Turns out, there IS a ticket for this.

This is basically the overhaul of the react-bootstrap-table implementation to bring it in line with the design.

## Testing

Head to non user outreach, outreach records. Test the filter functionality and verify that the design is in line with the mockup.


gulp/src/nonuseroutreach/actions/navigation-actions.js
```diff
@@ -3,3 +3,57 @@ import {createAction} from 'redux-actions';
...
   'SET_PAGE',
   (page, query, args) => ({page, query, args}));
+
+export const setWorkflowStateChoicesAction = createAction('SET_WORKFLOW_CHOICES');
+export const updateTermFilterAction = createAction('SET_TERM_FILTER');
+export const updateWorkflowFilterAction = createAction('SET_WORKFLOW_FILTER');
+export const setFiltersActiveAction = createAction('SET_FILTERS_ACTIVE');
+export const filterRecordsAction = createAction('FILTER_RECORDS');
+
+export function doGetWorkflowStateChoices() {
```
darrint: Needs docstring.

dk2/dockerdev.sh
```diff
@@ -143,10 +143,13 @@ rebuilddev() {
 }
 
 doruncd() {
+    if [ -z ${port+x} ]; then
+     nethost="--net=host"
+    fi
```
koonsies: if no explicit port declared, use net=host.. change for using Docker for Mac.

PR #2461, opened 2016-08-03
PD-2531 Add right aligned contact counts to the drop down.

mklauber:
image

### DIFF
--- a/gulp/src/nonuseroutreach/components/SearchDrop.jsx
+++ b/gulp/src/nonuseroutreach/components/SearchDrop.jsx
@@ -157,6 +157,9 @@ export default class SearchDrop extends Component {
               className={classnames(
                   {'active': i === activeIndex})}>
               {result.display}
+              <span className="partner-count">
+                {result.count}
+              </span>
             </li>
           ))}
         </ul>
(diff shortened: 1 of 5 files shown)


gulp/webpack.config.js
```diff
@@ -39,12 +39,12 @@ module.exports = {
       },
     ],
     // ie8 catchall. Some imported react components need this.
-    postLoaders: [
-      {
-        test: /\.js$/,
-        loaders: ['es3ify'],
-      },
-    ],
+//    postLoaders: [
```
tdphillips: Accidental?

darrint: Shouldn't commit this.

gulp/src/nonuseroutreach/components/SearchDrop.jsx
```diff
@@ -157,6 +157,9 @@ export default class SearchDrop extends Component {
               className={classnames(
                   {'active': i === activeIndex})}>
               {result.display}
+              <span className="partner-count">
+                {result.count}
```
darrint: Should be something like `({result.count} contacts)`. Refer to the design.

mklauber: Fixed.

PR #2463, opened 2016-08-04
[PD-2522] Add Migration to Rename Workflow State Names

koonsies:
## Overview

It's all in the subject. Simple data migration to capitalize/slightly reword state names.

## Testing

Run the migration


tdphillips:
We reference the "new" state by name in mypartners/views.py

koonsies:
Good catch. Updated.

PR #2468, opened 2016-08-08
[PD-2536] wizard submission logic

darrint:
This leaves NUO still buggy but able to, under limited circumstances, convert an outreach record to a communication record.

Unfortunately GitHub isn't picking up that I renamed process-email-* to process-outreach-*.

## Test

1. Make sure you have an outreach record to work with.
1. Pick an outreach record.
1. Pick an existing partner.
1. Pick an existing contact.
1. Fill in some data for the communication record.
1. Submit.
1. Using the shell and the outreach record ID you can see that the outreach record is in reviewed state and there is a new communication record.


tdphillips:
Any idea what some of those limited circumstances are?

darrint:
I added description of how to test.

koonsies:
:+1: Good as a checkpoint. I was able to create a comm record based on the testing criteria. Created a spin off ticket for some CSS styling problems I saw.

koonsies:
I know that most of these are simply not yet implemented, but for posterity, the most notable issues I saw:

1) History/Back does not work properly.
2) If you use browser "backs" to get back to the table, the state of the partner/contact selects remains.
3) There is no link to return to the table view from the creation wizard
4) Selecting an existing contact doesn't give the opportunity to add notes/view contact info
5) The asterisks used to indicate whether a field is required on contact record need to be revisited, some are marked "required" that are not.

PR #2473, opened 2016-08-09
Consolidate State

darrint:
Consolidate form and item selection state into the `record` property. See the docstring in `process-outreach-reducer.js` for details.

Behavior shouldn't change. Forms should display. The happy path of select partner, select contact, record form, submit should still work.


mklauber:
:+1:

PR #2474, opened 2016-08-10
[PD-2547] Forms Work

darrint:
## Review

This PR isn't as big as it looks. `forms.js` is the bulk of it and doesn't need read in detail.

Trying to use the `mypartners` Django forms turned out not to be a good line of development. I took out the api around that and the client code. In its place is a largish file: `forms.js` which contains the information we might have gotten from the forms api.

I ran the docs example for the `convert` api through a pretty printer.

## Test

Run through the UI creating partners and contacts. Use the network debugger (the `convert` call) to see if you were successful.


gulp/src/nonuseroutreach/forms.js
```diff
@@ -0,0 +1,668 @@
+const communicationTypes = [
+  {display: 'Email', value: 'email'},
+  {display: 'Phone', value: 'phone'},
+  {display: 'Meeting or Event', value: 'meetingorevent'},
+  {display: 'Job Followup', value: 'job'},
+];
+
+const states = [
```
tdphillips: I see some territories here. Does this also need to contain American Samoa, Northern Mariana Islands, or Virgin Islands? They're all listed in jsondata/usa_regions/json[p], as well as Guam and Puerto Rico (which are already in this list).

darrint: I created [PD-2549](https://directemployersfoundation.atlassian.net/browse/PD-2549)

tdphillips:
Regarding the testing instructions, I never get far enough to see a call to `convert`.

`TypeError: s is undefined`

gulp/src/nonuseroutreach/forms.js
```diff
@@ -0,0 +1,668 @@
...
+        attrs: {
+          maxlength: '255',
+        },
+        input_type: 'text',
+      },
+      min_length: null,
+      max_length: 255,
+    },
+    date_time: {
+      label: 'Date & Time *',
```
tdphillips: I'm seeing a double asterisk (possibly because this is marked "required"?)

darrint:
Is there a stack trace?

tdphillips:
I'm not set up to run off of unminified JS so the stack trace wasn't very helpful.

tdphillips:
How in-depth of a review do you want?

For instance, I can't create a contact without adding notes because `IntegrityError: (1048, "Column 'notes' cannot be null")`

darrint:
If you can create Partners and Contacts at all we're good for this PR. I'll make a ticket for that error.

PR #2475, opened 2016-08-11
Analytics improvements [PD-2521]

tdphillips:
- Parse things into their proper types
- Add ip to analytics header
- Change where analytics header is constructed

See https://github.com/DirectEmployers/MyJobs-Analytics/pull/6

### DIFF
--- a/redirect/log_to_mongo.py
+++ b/redirect/log_to_mongo.py
@@ -4,6 +4,7 @@ import json
 import os
 import sys
 
+from dateutil import parser
 import boto
 from pymongo import MongoClient
 
@@ -70,6 +71,19 @@ def to_mongo(file_name):
             # dictionaries. If "json_line" is a string, it's not a
             # document we wish to keep.
             if not isinstance(json_line, basestring):
+                for key, value in json_line.items():
+                    if key in ['to', 'sv', 'nv', 'fv'] and value:
+                        # parser.parse('') results in today's date; we probably
+                        # don't want that. Ensure the parameter has a value.
+                        try:
+                            json_line[key] = parser.parse(value)
+                        except (ValueError, TypeError):
+                            pass
+                    elif isinstance(value, basestring) and value.isdigit():
+                        json_line[key] = int(value)
+                        if key == 'time':
+                            json_line[key] = datetime.fromtimestamp(
+                                json_line[key])
                 json_line['file_id'] = file_id
                 json_lines.append(json_line)
 
--- a/redirect/views.py
+++ b/redirect/views.py
@@ -170,7 +170,7 @@ def home(request, guid, vsid=None, debug=None):
                 se = '?' + se
         else:
             hn = request.get_host()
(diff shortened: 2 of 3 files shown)


redirect/views.py
```diff
@@ -156,33 +157,40 @@ def home(request, guid, vsid=None, debug=None):
...
+            parsed = urlparse(redirect_url)
+            pn = parsed.path
+            pr = parsed.scheme + ':'
+            hn = parsed.netloc
+            se = parsed.query
+            if se:
+                se = '?' + se
+        else:
+            hn = request.get_host()
+            pr = request.is_secure() and 'https:' or 'http:'
```
koonsies: I wouldn't have thought to do this. (truthy statement) and (return value) or (default value). Cool! That's new to me.

koonsies:
:+1: I believe you said in SCRUM that this isn't easily tested? Code looks good.

redirect/views.py
mklauber: Actually, that's not the best way to do that.  I'd prefer 

```python
'https:' if request.is_secure() else 'http:'
```

tdphillips: > when the return value is a variable

I don't know what this means.

mklauber: In this case, nothing could happen.  But if instead of `"https:"` you had 

```python
https = ""
pr = request.is_secure() and https or 'http:'
```

It would always return `'http:'` because the and condition would consider `""` to be false.

mklauber:
Thanks.  :+1:

redirect/views.py
```diff
@@ -156,33 +157,40 @@ def home(request, guid, vsid=None, debug=None):
                                             request.get_host(),
                                             aguid)
 
-    if redirect_url and not expired:
```
tdphillips: Constructing the header here was catching views we didn't want to catch.

See the ticket for a more in-depth explanation if desired.

redirect/log_to_mongo.py
```diff
@@ -70,6 +71,19 @@ def to_mongo(file_name):
...
             if not isinstance(json_line, basestring):
+                for key, value in json_line.items():
+                    if key in ['to', 'sv', 'nv', 'fv'] and value:
+                        # parser.parse('') results in today's date; we probably
+                        # don't want that. Ensure the parameter has a value.
+                        try:
+                            json_line[key] = parser.parse(value)
+                        except (ValueError, TypeError):
+                            pass
+                    elif isinstance(value, basestring) and value.isdigit():
```
tdphillips: There are integers being passed in query strings. We want to store them as ints for ease of searching.

redirect/log_to_mongo.py
```diff
@@ -70,6 +71,19 @@ def to_mongo(file_name):
...
+                        json_line[key] = int(value)
+                        if key == 'time':
```
tdphillips: Unless that int's name is "time" - this is a timestamp and should be stored as a datetime.

mklauber:
:+1:

PR #2476, opened 2016-08-11
[PD 2537] Mini Cards!

koonsies:
## Overview

This ticket covers the introduction of the mini cards for tracking progress through NUO. This is not a completed product, but instead represents how the cards will look and populates based on state of the app.

## Testing

To test whether the cards work, select a partner from the dropdown. Then, select a contact. There should be cards displayed for both.

## Further work

This is the first of two tickets I will push for this ticket, and there is a follow up ticket for some additional logic that needs to be added. This is merely a checkpoint so that Darrin can begin some work on the cards as well.

**Work remaining on this ticket**
1) Modify SVG based on card type (partner, contact, etc)
2) Only display delete "X" when hovering over the card
3) Display contact name instead of PK on contact card
4) Write tests

**Work on follow up ticket**
1) Implement ability to click a card to open and edit a from (partner, contact)
2) Implement ability to delete a card and change state of app based on that.


darrint:
I created tickets for all the items on your list and linked them to [PD-2526](https://directemployersfoundation.atlassian.net/browse/PD-2526).

PR #2481, opened 2016-08-15
Forgot to connect the new field in the API to the state.

mklauber:


tdphillips:
I'm still getting a subject of undefined.

tdphillips:
process-outreach-actions.js names the subject "outreachSubject" while OutreachDetails.jsx is bringing it in with "outreachSubject: state.process.outreach.subject"

mklauber:
Should be fixed.

PR #2483, opened 2016-08-16
[PD-2537] mini card additional features

koonsies:
This ticket covers a couple of spin off tickets.

1) Make icon type change based on the type of mini card
2) Make the delete icon only appear when hovering over the card
3) Adjust width on SVG icons to bring them in line with design
4) Modify the way we handle keys on the mini cards

See annotations for notes.


gulp/src/nonuseroutreach/components/OutreachCard.jsx
```diff
@@ -4,25 +4,41 @@ import {expandStaticUrl} from 'common/assets';
...
-          alt="[partner]"
-          src={expandStaticUrl('svg/partner-card.svg')} />
-        </div>
-        <div className="tray-items">
-            <img
-             alt="[delete]"
-             src={expandStaticUrl('svg/delete.svg')} />
+          style={{width: '20px',
+                  height: '20px'}}
+          alt={'[' + type + ']'}
```
koonsies: Keep alt text enclosed in brackets, as per other alt text in our React code.

gulp/src/nonuseroutreach/components/OutreachCardContainer.jsx
```diff
@@ -38,8 +38,9 @@ class OutreachCardContainer extends Component {
 
     return (
       <OutreachCard
-        key={contact}
+        key={index}
```
koonsies: Use the contact's list index instead of any field given by the contact itself. I also considered using contact name, since new contacts don't have a PK (PK being the more obvious choice)

gulp/src/nonuseroutreach/components/OutreachCardContainer.jsx
```diff
@@ -49,15 +50,17 @@ class OutreachCardContainer extends Component {
 
     return (
       <OutreachCard
-        key={partner.pk}
+        key="partner"
```
koonsies: So, there is only ever one partner. Is it sufficient to hardcore the key?

gulp/src/nonuseroutreach/components/OutreachCardContainer.jsx
```diff
@@ -49,15 +50,17 @@ class OutreachCardContainer extends Component {
...
         displayText={partner.partnername}
         onNav={() => dispatch(editPartnerAction())}/>
     );
   }
 
   handleCommunicationRecord(communicationRecord) {
     return (<OutreachCard displayText={communicationRecord}
-                         key={communicationRecord} />);
+                          type="communicationrecord"
+                          key="commrec" />);
```
koonsies: Same question about keys here. @darrint any opinions?

gulp/src/nonuseroutreach/components/OutreachCardContainer.jsx
darrint: Yes.

gulp/src/nonuseroutreach/components/OutreachCardContainer.jsx
darrint: Yes, here too.

tdphillips:
Looks good. Mini cards stick around after deletion but the ticket for that is already in progress.

PR #2484, opened 2016-08-17
Remove contact count from NUO contact dropdown [PD-2564]

tdphillips:
Would we do something like this in-line or define a variable prior to the return?


darrint:
I would have gone with a variable to keep the lines short. I'm still merging this though.

PR #2488, opened 2016-08-18
[PD-2547] forms submit

darrint:
@koonsies, this one.


mypartners/tests/test_api.py
```diff
@@ -299,7 +336,8 @@ def test_outreach_api_contact_location_by_pk(self):
 
         """
         existing_location = LocationFactory()
-        self.request_data['contacts'][0]['location'] = {"pk":existing_location.pk}
+        record = self.request_data['forms']['contacts']
+        record[0]['location'] = {"pk": {'value': existing_location.pk}}
```
koonsies: I wonder if we'll ever actually be doing locations by PK. So far, we don't have a control for that type of thing.

PR #2491, opened 2016-08-19
[PD-2555] mini card delete

koonsies:
## Overview

This ticket covers two areas

1) Allow cards to be deleted (contact, partner, etc)
2) Redefine "flow" in the app. Determine what view to display based on what info is present.

## Testing

To test this PR, simply navigate around the NUO process email view. Delete cards, add cards, explore state in the app. It should work very similarly to how it operated previously, but the flow of "state" is much cleaner.


tdphillips:
If I choose a partner and then delete said partner, I go back to the section of the wizard where you pick/choose a partner as I expected.

The parter input is still filled out, however. The spinner is present and I can't edit the text.

koonsies:
Good call! I haven't attempted to tackle that issue yet. Same thing happens if you use backspace to navigate to the main table and click "review" on another record. I'll capture investigation into that in a separate ticket.

gulp/src/nonuseroutreach/actions/process-outreach-actions.js
```diff
@@ -88,6 +78,22 @@ export const editCommunicationRecordAction =
...
+
+/**
+ * User wants to remove contact from the record
+ *
+ * contactIndex: which one
+ */
+export const deleteContactAction = createAction('NUO_DELETE_CONTACT',
+  contactIndex => ({contactIndex}));
+
+export const deleteCommunicationRecordAction =
```
darrint: docstring

gulp/src/nonuseroutreach/reducers/process-outreach-reducer.js
```diff
@@ -149,13 +164,51 @@ export default handleActions({
...
+
   'NUO_EDIT_COMMUNICATIONRECORD': (state) => {
     return {
       ...state,
       state: 'NEW_COMMUNICATIONRECORD',
     };
   },
 
+  'NUO_DELETE_PARTNER': (state) => {
+    const newContacts = state.record.contacts.filter(contact => !contact.pk);
```
darrint: A comment here explaining why this works would be good.

PR #2492, opened 2016-08-22
[PD-2547] Forms Submit

darrint:
This change improves the ability to submit code and handle the resulting errors.

## Test

Create contacts and contact records with problems. You should be able to navigate back to contacts and contact records and see the errors.

## Review

I've reworked the `convert` api call a bit. The overall structure is the same but the class for holding and sending errors back to the client is completely replaced.

In the client the code is much more intentional about converting `process.record` into the shape required by the API and back into the shape required by the forms.


gulp/src/nonuseroutreach/actions/process-outreach-actions.js
```diff
@@ -160,53 +159,109 @@ export function extractErrorObject(fieldArray) {
...
-        const errors = {
-          partner: extractErrorObject(get(formErrors, 'partner', [])),
-          contacts: map(get(formErrors, 'contacts', []), ce =>
-            extractErrorObject(ce)),
-          communicationrecord: extractErrorObject(
-            get(formErrors, 'contactrecord', [])),
-        };
-        dispatch(noteErrorsAction(errors));
+        const apiErrors = e.data.forms;
+        if (apiErrors) {
```
tdphillips: What is the difference between these two ifs?

universal/api_validation.py
```diff
@@ -92,96 +93,198 @@ def build_error_response(self):
...
+        'name': {'value': 'Alice',
+    }
+
+    to:
+
+    {'name': 'Alice'}
+    """
+    result = {}
+    for (k, v) in data.iteritems():
+        if isinstance(v, dict) and 'value' in data[k]:
```
tdphillips: If v is a dict, couldn't this be `'value' in v`

universal/api_validation.py
```diff
@@ -92,96 +93,198 @@ def build_error_response(self):
...
+            result[k] = v['value']
+        elif isinstance(v, list):
+            result[k] = [collapse_values(inner) for inner in v]
+        else:
+            result[k] = collapse_values(v)
```
tdphillips: If a thing is not a dict or list, isn't this an infinite loop?

tdphillips: What type could v be if not a dict or list?

gulp/src/nonuseroutreach/actions/process-outreach-actions.js
darrint: One is checking truthiness of `e.data`, the other of `e.data.forms`. It won't happen today but in the future `e.data` could contain something else entirely.

universal/api_validation.py
darrint: String or int.

gulp/src/nonuseroutreach/actions/process-outreach-actions.js
tdphillips: Both are checking e.data.forms.

universal/api_validation.py
tdphillips: Won't that result in a 500? Neither string nor int have .iteritems

gulp/src/nonuseroutreach/actions/process-outreach-actions.js
darrint: Good Catch...

darrint: Fixed.

koonsies:
Communication type triggers a validation error when submitting the original partner form, which causes the communication record card to appear. Beyond that, the error for not having a communication record selected is displaying as "Value u'{}' is not a valid choice."

This also causes issues with the workflow PR I put up before. Since that CommRecord object is present after validation on partner, it tries to skip the comm record step.

koonsies:
It's not directly related to this PR, but I noticed that ContactRecord's datetime field has a blank=True set, but not null=True. This is causing the full_clean method to say that having a blank date time is fine, but when it goes to save the model, it is triggering a 500 error. We need to look into why this field is set up this way (without having a default value) in the future.

tdphillips:
There is no meaningful default. Any default value we could pick for "date this event happened" would almost certainly be wrong.

PR #2493, opened 2016-08-22
Add return to table button

mklauber:
Added return to records list button as part of ProcessRecordPage.render.
Added styles for nuo-buttons to match the mockup.
Updated classes on the "Form Ready for Submission" page to use new styles.


tdphillips:
Falls afoul of the bad form states that I noted in #2491 but I've been assured that those are fine for now.

PR #2494, opened 2016-08-22
Fixed overflow styling on partner/contact drop downs

mklauber:
Fix styling on overflow drop downs.  Should display as much of the name as visible, ending with an ellipsis if it extends off screen.

### DIFF
--- a/gulp/src/nonuseroutreach/components/SearchDrop.jsx
+++ b/gulp/src/nonuseroutreach/components/SearchDrop.jsx
@@ -156,11 +156,11 @@ export default class SearchDrop extends Component {
               onMouseEnter={() => this.handleMouseOver(i)}
               onClick={() => this.handleSelect()}
               className={classnames(
-                  {'active': i === activeIndex})}>
-              {result.display}
+                  {'active': i === activeIndex, 'list-result': true})}>
               {typeof result.count !== 'undefined' ?
                 <span className="partner-count">({result.count} contact{result.count === 1 ? '' : 's'})</span>
                 : ''}
+              {result.display}
             </li>
           ))}
         </ul>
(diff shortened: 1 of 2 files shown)


gulp/src/nonuseroutreach/components/SearchDrop.jsx
```diff
@@ -156,11 +156,11 @@ export default class SearchDrop extends Component {
               onMouseEnter={() => this.handleMouseOver(i)}
               onClick={() => this.handleSelect()}
               className={classnames(
-                  {'active': i === activeIndex})}>
-              {result.display}
+                  {'active': i === activeIndex, 'list-result': true})}>
               {typeof result.count !== 'undefined' ?
                 <span className="partner-count">({result.count} contact{result.count === 1 ? '' : 's'})</span>
                 : ''}
+              {result.display}
```
mklauber: Moved this to after the count, so that this is the "last" text, and thus the one cutoff. Since partner-count has `float:right` the count text still shows on the right.

gulp/src/nonuseroutreach/components/SearchDrop.jsx
```diff
@@ -156,11 +156,11 @@ export default class SearchDrop extends Component {
               onMouseEnter={() => this.handleMouseOver(i)}
               onClick={() => this.handleSelect()}
               className={classnames(
-                  {'active': i === activeIndex})}>
-              {result.display}
+                  {'active': i === activeIndex, 'list-result': true})}>
```
mklauber: Not sure if this is how the `classnames(` function is supposed to work, I just know this resulted in the `'list-result'` class being added to all the elements, which is what I wanted.  @koonsies @darrint Is this right?

darrint: https://github.com/JedWatson/classnames

This is fine. You could also have said `classnames({'active': i === activeIndex}, 'list-result')

PR #2496, opened 2016-08-23
Remove blank=True from ContactRecord model.

darrint:
As discussed, remove `blank=True` from `ContactRecord.date_time`.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -674,7 +674,7 @@ class ContactRecord(ArchivedModel):
                               null=True)
     subject = models.CharField(verbose_name="Subject or Topic", max_length=255,
                                blank=True, default="")
-    date_time = models.DateTimeField(verbose_name="Date & Time", blank=True)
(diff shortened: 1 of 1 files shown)


tdphillips:
Creating a communication record still works. :+1:

darrint:
Also I get "This field cannot be null" on the ContactRecord datetime field.

koonsies:
Hmm.. Field cannot be null isn't a good error message to display to an end user. I can't think of a way to display replace the error message, though.

tdphillips:
Presumably we use a form somewhere.

http://stackoverflow.com/questions/1481771/django-override-default-form-error-messages

koonsies:
Yeah. That is what I was thinking. We were trying to avoid having to swap the convert API from being model-centric to form-centric, but it's looking like we may have to. If we switch to a form-centric API, this change isn't mandatory.

darrint:
I've made several new tickets including one to address the form error messages. See PD-2572. (And somewhat related, PD-2577)

PR #2497, opened 2016-08-24
[PD-2573] append only contact form

darrint:
Show a special append notes only form when clicking an existing contact mini card.

## Test

1. Select an existing partner and contact. You should be at the communication record form.
2. Click on the contact's mini-card.
3. See the new append-notes-only form.

### DIFF
--- a/gulp/src/common/ui/CheckBox.jsx
+++ b/gulp/src/common/ui/CheckBox.jsx
@@ -8,7 +8,16 @@ import React from 'react';
  */
 
 function CheckBox(props) {
-  const {name, onChange, required, initial, isHidden, autoFocus} = props;
+  const {
+    name,
+    onChange,
+    required,
+    initial,
+    isHidden,
+    autoFocus,
+    disable,
+  } = props;
+
   return (
     <input
       type="checkbox"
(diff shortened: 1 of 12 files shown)


koonsies:
A couple of visual notes:

The contact name still appears editable from this view (it is not grayed out)

The Notes field placeholder text is grayed like the disabled fields and doesn't appear to be modifiable. Not sure how to differentiate this field at the moment, but we should fix that eventually.

I'd say the first one should be fixed, but the second is less important.

tdphillips:
Are you supposed to be able to submit a functional record after this?

Process:

- Choose partner
- Choose contact
- Now you're on the communication record form; click on the contact that you added
- Modify as appropriate and click "Add Contact"
- Now you're on the  "form ready for submission" form with a communication record named "unknown", bypassing the commrec form entirely.

darrint:
@tdphillips I hate to say it but I'm going to go with yes, that is "supposed" to happen, at least for now. Once you show the communication record form there's a communication record, so clicking add contact is going to take you to the next empty form.

darrint:
Also, @koonsies suggestion about graying out text didn't turn out to be necessary. There is other grayed out text present in other controls but those are placeholders. The background of disabled controls is grayed out so there is a visual indication that the control is disabled.

PR #2504, opened 2016-08-26
Make contact required [PD-2568]

tdphillips:
This fixes the immediate problem. I just want to get some feedback on if this is how we want to solve it.

### DIFF
--- a/mypartners/forms.py
+++ b/mypartners/forms.py
@@ -359,6 +359,8 @@ class ContactRecordForm(NormalizedModelForm):
             partner = kwargs.pop('partner')
         super(ContactRecordForm, self).__init__(*args, **kwargs)
         self.fields['contact'].required = True
+        if not hasattr(self.instance, 'contact'):
+            self.fields['contact'].required = False
 
         instance = kwargs.get('instance')
         if partner:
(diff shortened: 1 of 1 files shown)


mypartners/forms.py
```diff
@@ -358,6 +358,9 @@ def __init__(self, *args, **kwargs):
         if 'partner' in kwargs:
             partner = kwargs.pop('partner')
         super(ContactRecordForm, self).__init__(*args, **kwargs)
+        self.fields['contact'].required = True
```
tdphillips: Contact should generally be required.

mypartners/forms.py
```diff
@@ -358,6 +358,9 @@ def __init__(self, *args, **kwargs):
...
+        if not hasattr(self.instance, 'contact'):
```
tdphillips: Unless self.instance.contact doesn't exist (the contact was deleted).

mypartners/forms.py
```diff
@@ -444,7 +447,12 @@ def save(self, request, partner, commit=True):
         PRMAttachment.objects.filter(
             pk__in=self.cleaned_data.get('attach_delete', [])).delete()
 
-        identifier = instance.contact.name
+        if instance.contact:
```
tdphillips: Do some checking to determine what the contact identifier should be.

I'm open to suggestions on the default.

JLMcLaugh: This looks more complete than what we're logging elsewhere. We record the contact id and the __unicode__ value also, so it should be fine.

PR #2507, opened 2016-08-29
Bypass expired page for VetCentral (vs 99) [ST-2804]

tdphillips:


tdphillips:
Checked the job in the ticket. It fails exactly as expected - with a "this job is expired" page on Taleo instead of my.jobs.

koonsies:
:+1: code looks good. Is there an easy way to hand test this?

tdphillips:
Not really. You could manually add an expired job to your database then check it out with view source 99.

PR #2508, opened 2016-08-30
[PD-2579] - moved contact name above contact's organization for the Cont…

huy-nguyen-colorado:
Moved contact name above contact's organization for the Contact Search ajax auto-complete box.


huy-nguyen-colorado:
To test in browser: 
Navigate to NUO > Outreach Records > Contact Search > type "m", look at "Monique Melchor ...", notice the contact's name appears above the contact's organization name (which happens to be the longest org name in the history of org names).

PR #2509, opened 2016-08-30
Display the `common/ui/Confirm.jsx` dialog when the delete button is pressed

mklauber:
* Fixed a typo in a comment
* Added confirmation function and tied it to delete button on Inbox component.
* Added confirmation box to InboxManagementPage component.
* Added the necessary reducers to `nonuseroutreach/main.jsx`

### DIFF
--- a/gulp/src/nonuseroutreach/components/InboxManagementPage.jsx
+++ b/gulp/src/nonuseroutreach/components/InboxManagementPage.jsx
@@ -2,7 +2,6 @@ import React from 'react';
 import {connect} from 'react-redux';
 import Inbox from './Inbox';
 import {Col, Row} from 'react-bootstrap';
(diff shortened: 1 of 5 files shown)


gulp/src/nonuseroutreach/components/InboxManagementPage.jsx
```diff
@@ -13,6 +14,7 @@ class InboxManagementPage extends React.Component {
 
     return (
       <div className="cardWrapper">
+        <Confirm/>
```
darrint: Does it still work if you move this to `NonUserOutreachApp`?

tdphillips:
:+1: 

Our React confirmations have reversed buttons from our older stuff but that's not an issue with this PR.

PR #2514, opened 2016-09-01
[PD-2576] update tags control

koonsies:
## Overview

This ticket covers introducing the tag control to the partner form in NUO. In it's current state, it can only add new tags and those new tags aren't submitted when the form is submitted. This is due to the overhaul @darrint is doing to the way we manage forms in NUO.

## Testing

As this ticket does not cover existing tags or submitting tags, simply test the controls by adding and removing tags on the various forms.

Some functionality that should exist:
* newly added tags should be available for selection on all forms once added
* you should be able to select and remove tags, tags do not delete when unselected
* select all and filter functionality is still present


gulp/src/nonuseroutreach/components/OutreachCardContainer.jsx
```diff
@@ -62,6 +63,7 @@ class OutreachCardContainer extends Component {
         onNav={() => dispatch(editContactAction(index))}
         onDel={() => {
           dispatch(deleteContactAction(index));
+          dispatch(removeNewTagsFromForm('contact' + index));
```
koonsies: @darrint you may want to revisit this logic depending on how you handle storage/retrieval of contacts.. currently, they're stored as a string with contact appended with it's index.

PR #2515, opened 2016-09-01
Create outreach records regardless of contact/partner presence [PD-2596]

tdphillips:
Also adds this new info to the ui.

### DIFF
--- a/gulp/src/nonuseroutreach/actions/process-outreach-actions.js
+++ b/gulp/src/nonuseroutreach/actions/process-outreach-actions.js
@@ -132,6 +132,8 @@ export function convertOutreach(record) {
     outreachInbox: record.outreach_email,
     workflowState: record.current_workflow_state,
     outreachSubject: record.subject,
+    outreachTo: record.to_emails,
+    outreachCC: record.cc_emails,
   };
 }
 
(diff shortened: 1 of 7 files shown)


gulp/src/nonuseroutreach/actions/process-outreach-actions.js
```diff
@@ -132,6 +132,8 @@ export function convertOutreach(record) {
     outreachInbox: record.outreach_email,
     workflowState: record.current_workflow_state,
     outreachSubject: record.subject,
+    outreachTo: record.to_emails,
+    outreachCC: record.cc_emails,
```
tdphillips: Might look weird but I think this is accurate casing.

mypartners/views.py
```diff
@@ -1246,21 +1246,27 @@ def process_email(request):
     logger.info("unmatched_contacts: {contacts}".format(contacts=", ".join(
                 unmatched_contacts)))
```
tdphillips: The changes below separate NUO and PRM record creation into two discrete things. Before, we were trying to do both at the same time.

mypartners/views.py
```diff
@@ -1641,45 +1665,33 @@ def make_records(request, date_time, possible_contacts, created_contacts,
         return None, None, HttpResponse(200)
 
     for contact in all_contacts:
```
tdphillips: This function is largely unchanged, with the outreach record functionality being split into `make_outreach_records`

koonsies:
:+1: Code looks good. I'll wait for a bit to let anyone else review before I merge.

PR #2517, opened 2016-09-06
Webpacking custom.css

mklauber:
This is a PR to automatically compile `static/custom.css` from `gulp/src/sass/custom.scss`.  It appears to work, at least on my machine.


gulp/src/sass/custom.scss
```diff
@@ -4327,7 +4327,7 @@ table td, table th {
   margin-top: 1px;
   line-height: 14px;
   vertical-align: text-top;
-  background-image: url("../img/glyphicons-halflings.png");
```
mklauber: Webpack dislikes relative urls to files it can't find.  (at least, with how it's setup now.) These files exist on our cloudfront server, lets use them.

gulp/src/sass/custom.scss
```diff
@@ -4327,7 +4327,7 @@ table td, table th {
...
+  background-image: url("//d2e48ltfsb5exy.cloudfront.net/content_mj/files/images/glyphicons-halflings.png");
```
mklauber: Webpack dislikes relative urls to files it can't find. (at least, with how it's setup now.) These files exist on our cloudfront server, lets use them.

darrint:
We should update the `compile-sass` target in `package.json`. That will be a nice escape hatch in case someone is working on pure Django and doesn't want to mess with devserver.

tdphillips:
compile-sass appears to work just fine without devserver.

mklauber:
@darrint @tdphillips, is `compile-sass` working then, or not?  

@tdphillips, with this change, the intent is that deployment will be able to handle creating the custom.css file, and thus we won't need to checkin custom.css anymore.  That said, thoughts on whether this PR should include removing and ignoring static/custom.css from the repo?

tdphillips:
@darrint, any input?

huy-nguyen-colorado:
Yes, static/custom.css should be ignored.  Developers should only be modifying or adding .scss files or if there's really really really really really really really reallllllyyyyyyyyyyyyyy a need for it..... .sass files.

Unless of course Murphy's Law creeps in and we have to manually do something with the custom.css file.

PR #2519, opened 2016-09-07
pd-2605 Fixed Word Wrapping issue within Firefox and IE8 in the body …

mwarnsleyDE:
pd-2605 Fixing the Word Wrapping issue in the Body of the non user outreach section


huy-nguyen-colorado:
This fix works, but going forward, we should place styles in the style sheet for easier maintenance and sanity.

PR #2522, opened 2016-09-08
[PD-2590] Better forms validation

darrint:
Some regressions/additional annoyances I am aware of:

* The state field isn't a select. The state list in `forms.js` should probably be moved to the new NUO forms.
* Communication record objects are created before we are ready and show up as having errors at the wrong time.
* Tags aren't colored.
* Existing contacts allow for an append notes form. This form isn't actually loading the rest of the contact data so the fields in the form are blank.

Improvements:

* The tags control is fully functional.
* Django forms are doing the validation. Labels and error messages are friendlier.

## Test

* Use NUO to convert some outreach records to communication records.
* You should be able to see your data in PRM. (Best strategy I know of this is to put sane dates on the communication records.


mypartners/views.py
```diff
@@ -1941,244 +2012,250 @@ def api_convert_outreach_record(request):
...
-
-    outreach_record.current_workflow_state = workflow_status
-    outreach_record.save()
+            if create_communication_record:
+                communication_record_form = NuoCommunicationRecordForm(
+                    communication_record_data, company=user_company)
+            else:
+                api_errors.append(
+                    "User does not have permission to create a " +
+                    "communication record.")
```
koonsies: We should probably raise this error to an NUO access requirement at some point.. because without being able to create a comm record, there's no point in using NUO. Unlike Partner/Contact, which you could simply select existing.

koonsies:
Few notes:

Comm Recs and Contacts seem to share the notes field. On submission, the notes are only added to the comm rec, not contact.

Multiple contacts are not working properly.. If you add more than one contact to a comm record, only one comm record is created and it is tied to a single contact.

gulp/src/common/ui/RemoteFormField.jsx
```diff
@@ -14,7 +14,7 @@ export default class RemoteFormField extends Component {
     const {fieldName, form, value, onChange,
       tagActions, selectedTags, availableTags} = this.props;
     const field = form.fields[fieldName] || {};
-    const errors = form.errors[fieldName];
+    const errors = (form.errors || {})[fieldName];
```
JLMcLaugh: This will always throw a keyerror if form.errors is falsy.

PR #2525, opened 2016-09-08
[PD-2611] Multiple Communication Records

darrint:
This fixes a number of bugs regarding writing records to the database.

## Test

1. Pick a partner and contact.
2. Add an additional contact.
3. Add tags to the communication record. Set a date that is recent.
3. Go to PRM and verify that 2 communication records have been created and all tags are present on all records.

### DIFF
--- a/mypartners/models.py
+++ b/mypartners/models.py
@@ -703,6 +703,25 @@ class ContactRecord(ArchivedModel):
 
         super(ContactRecord, self).save(*args, **kwargs)
 
+    def recreate(self):
+        """
+        Becomes a new copy of the original record.
+
+        returns a reference to the original record.
+        """
+        original_record = ContactRecord.objects.get(pk=self.pk)
+        self.pk = None
+
+        new_status = self.approval_status
+        new_status.pk = None
+        new_status.save()
+        self.approval_status = new_status
+
+        self.save()
+        self.tags.add(*original_record.tags.all())
+
+        return original_record
+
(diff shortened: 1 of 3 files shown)


mklauber:
:+1:

mypartners/models.py
```diff
@@ -703,6 +703,25 @@ def save(self, *args, **kwargs):
 
         super(ContactRecord, self).save(*args, **kwargs)
 
+    def recreate(self):
```
tdphillips: I know the documentation explains this, but the premise of this function feels weird to me.

mypartners/views.py
```diff
@@ -2242,6 +2246,14 @@ def all_forms_iter():
             contact = contact_result['contact'].instance
         else:
             contact = contact_result['instance']
+
+        # on subsequent passes, make new communication records for each contact
+        if i > 0:
+            communication_record.recreate()
+            attach_new_tags(
```
tdphillips: Given tags are added before recreate() is called, is this necessary?

mypartners/models.py
darrint: I don't think there's another way to write it. The technique is supposed to be set `pk` to `None` and `save()`. So you're stuck with mutating the original object in place.

mypartners/views.py
darrint: You were correct here. I'll put the fix for this in my next PR.

PR #2527, opened 2016-09-08
[PD-2616] See New Partners in PRM

darrint:
New partners weren't linked to the user's company and weren't showing up in PRM. This fixes that. Also, @tdphillips found that some new tag attachment code was redundant. That is fixed in this PR.

## Test

1. Create a new partner in NUO. Look for it in PRM.
2. In NUO create a new communication record with an existing tag and a new tag and attached to multiple contacts. (got all that?) Verify in PRM that the new and existing tags are on both communication records.


tdphillips:
Verified using @darrint's instructions. :+1:

PR #2529, opened 2016-09-09
Fix for only text in minicard being clickable.

mklauber:


koonsies:
I am getting an error when I add more than one mini card for contact. 

My workflow:
Add a partner and contact via the select contact box
Add another contact.
Error.
`Form.jsx:20 Uncaught TypeError: Cannot read property 'ordered_fields' of undefined`

mklauber:
Based on my testing, that issue is on the QC branch.  It's not caused by my changes.

koonsies:
Ahh. I was just about to check that. I had compared to QC, but I hadn't pulled the latest from the repository when I did.

gulp/src/nonuseroutreach/components/OutreachCard.jsx
```diff
@@ -25,7 +25,10 @@ class OutreachCard extends Component {
           <img
           alt="[delete]"
           src={expandStaticUrl('svg/delete.svg')}
-          onClick={() => this.props.onDel()} /> : null}
+          onClick={(event) => {
+            this.props.onDel();
+            event.stopPropagation();
```
darrint: Interesting. Without this did it try to delete, then navigate?

PR #2531, opened 2016-09-09
[pd-2619] Fixed error for opening up records along with fixing the over…

mwarnsleyDE:
Fixed the issue with the records not opening along with the overflow on the records table.


templates/nonuseroutreach/nuo_main.html
```diff
@@ -57,6 +57,36 @@
     <script src="{{ STATIC_URL }}pickadate/legacy.js"></script>
     SteelToe: https://github.com/jclem/steeltoe -->
     <!-- <script src="{{ STATIC_URL }}steeltoe.js"></script> -->
+    <script type="text/javascript">
```
darrint: This needs to be in `polyfills.js` if possible.

PR #2532, opened 2016-09-09
Fix for PD-2622

mklauber:
This populates the `outreach_record` form from the beginning to avoid error messages or missing form erros.

### DIFF
--- a/mypartners/views.py
+++ b/mypartners/views.py
@@ -1898,10 +1898,13 @@ def merge_contact_forms(contact_forms, company):
 @requires("convert outreach record")
 def api_get_nuo_forms(request):
     company = get_company_or_404(request)
+    outreach_pk = request.GET.get('outreachId', None)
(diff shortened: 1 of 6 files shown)


mypartners/views.py
```diff
@@ -1898,10 +1898,13 @@ def merge_contact_forms(contact_forms, company):
 @requires("convert outreach record")
 def api_get_nuo_forms(request):
     company = get_company_or_404(request)
+    outreach_pk = request.GET.get('outreachId', None)
```
tdphillips: mypartners.tests.test_api.NonUserOutreachTestCase.test_tag_names (what a mouthful) doesn't pass any parameters to this view, causing this key to be None. If None is valid, we may need to guard the following lines with checks.

mklauber: `None` should not be valid.  Fixing.

PR #2533, opened 2016-09-09
[PD-2621] Prune communication record form.

darrint:
This matches the dynamic behavior of PRM's communication record form in NUO.

## Test

1. Look at the communication record form.
2. Change the contact type.
3. See how the form changes.

### DIFF
--- a/gulp/src/nonuseroutreach/forms.js
+++ b/gulp/src/nonuseroutreach/forms.js
@@ -326,8 +326,15 @@ export function pruneCommunicationRecordForm(form, formContents) {
   const fieldListsByType = {
     '': [...before, 'subject', 'date_time', ...after],
     email: [...before, 'contact_email', 'subject', 'date_time', ...after],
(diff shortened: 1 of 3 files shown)


gulp/src/nonuseroutreach/forms.js
```diff
@@ -320,6 +320,34 @@ function readonlyField(field) {
   };
 }
 
+export function pruneCommunicationRecordForm(form, formContents) {
+  const before = ['contact_type'];
+  const after = ['tags', 'notes'];
+  const fieldListsByType = {
+    '': [...before, 'subject', 'date_time', ...after],
+    email: [...before, 'contact_email', 'subject', 'date_time', ...after],
```
tdphillips: Is it intentional for email to have subject then date_time while phone and meetingorevent have them transposed?

darrint: It was intentional but I see that I had them wrong. I went through the PRM ui and matched that directly this time.

PR #2534, opened 2016-09-09
[pd-2617] Datetime-Picker-Widget-Styling adding the leading zeros to th…

mwarnsleyDE:
Adding in the leading zeros to the date picker fields

### DIFF
--- a/gulp/src/common/ui/DateTimePicker.jsx
+++ b/gulp/src/common/ui/DateTimePicker.jsx
@@ -51,9 +51,8 @@ export default class DateTimePicker extends Component {
   }
 
   rangeToDropDownArray(rangeStart, rangeEnd) {
-    const lz = '0';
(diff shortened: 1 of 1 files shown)


gulp/src/common/ui/DateTimePicker.jsx
```diff
@@ -51,8 +51,9 @@ export default class DateTimePicker extends Component {
   }
 
   rangeToDropDownArray(rangeStart, rangeEnd) {
+    const lz = '0';
     return map(range(rangeStart, rangeEnd),
-      (n) => ({value: n, display: String(n), render: () => ''}));
+      (n) => ({value: n, display: parseInt(String(n), 10) < 10 ? String(lz + n) : String(n), render: () => ''}));
```
koonsies: This can be accomplished via the lodash.padLeft(string, count, character) function as seen on line 81-84

PR #2535, opened 2016-09-09
[PD-2611] Set created_by on new records.

darrint:
We weren't setting `created_by` on new records. Now we are.

## Test

1. Create new records in NUO.
2. Find records in PRM.
3. See that your user created those records.


tdphillips:
Conflicts.

PR #2537, opened 2016-09-09
[PD-2620], scroll to the top of form if there is an error in the form

huy-nguyen-colorado:
To test:

In NUO > Communication Record, in the "Communication Type *" drop down, if you leave that drop down option to the default option of "------" and click on the "Add Record" button, the browser should return to the top of the page to show the end user where the error is.


gulp/src/nonuseroutreach/actions/process-outreach-actions.js
```diff
@@ -247,6 +247,7 @@ export function doSubmit(validateOnly, onSuccess) {
         }
         if (e.data.forms) {
           dispatch(noteFormsAction(formsFromApi(e.data)));
+          document.body.scrollTop = document.documentElement.scrollTop = 0;
```
darrint: The scroll code should not live here. Use a callback like `onFormErrors`.

PR #2546, opened 2016-09-15
[PD-2610] Replace HTTP with HTTPS for Redirect Links

koonsies:
## Overview

To assist with tracking the refer(r)er values within analytics, we need to cause links to redirect to be sent via HTTPS.

This PR covers logic to force HTTPS on apply links on microsites, as well as overwriting some hardcoded values

## Testing

To test the apply links, simply load the branch and view the target of the links

To test the hardcoded values, navigate to a view containing them.

## Further work

This represents my current best stab at tracking down links to http://my.jobs. Any input toward other places to look would be appreciated.

### DIFF
--- a/templates/job_detail.html
+++ b/templates/job_detail.html
@@ -1,6 +1,7 @@
 {% extends "seo_base.html" %}
 {% load i18n %}
 {% load seo_extras %}
+{% load redirect_urls %}
 {% block direct_extraHeaderContent %}
 <link rel="stylesheet" href="/style/def.ui.dotjobs.results.css" type="text/css">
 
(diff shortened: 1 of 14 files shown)


templates/job_detail.html
```diff
@@ -79,7 +80,7 @@
         </div>
         <div class='direct-action-btn' id="direct_applyButtonBottom">
             {% if the_job.link %}
-            <a href="{{the_job.link}}" onclick="goalClick('/G/apply-click', this.href); return false;">{% blocktrans %}Apply Now{% endblocktrans %}</a>
+            <a href="{{the_job.link|swap_http_with_https:"my.jobs"}}" onclick="goalClick('/G/apply-click', this.href); return false;">{% blocktrans %}Apply Now{% endblocktrans %}</a>
```
tdphillips: Not saying this needs work, just posing some questions.

Given job links in Solr will solve themselves eventually as jobs expire, does this replacement need to happen?

Another option would be updating the jobs in Solr directly, though @mklauber knows better than I.

koonsies: That's a good question. I almost removed this bit of logic after speaking with Jason, who said that content acquisition would be updating their process to change the job.link to https. I ended up leaving it in as a catch all. But I'm more than willing to reconsider that.

PR #2550, opened 2016-09-20
[ST-2975] Mark some columns as filter only and use that.

darrint:
**This may need to be a hotfix and it contains a migration.**

Some columns needed to be marked `filter_only` and their blank output suppressed from the report. 

**Edit:** I see that some locations columns already had `filter_only` set. Make sure we aren't creating some kind of regression here. The affected reports are Partner and Communication Record.


tdphillips:
Is this a hotfix?

darrint:
It will be. I'm still working on qc though.

PR #2553, opened 2016-09-21
Docker improvements.

mklauber:


darrint:
Test: run `dk rebuilddev` then restart all the app containers (not background ones) and make sure things still work.

PR #2560, opened 2016-09-29
[PD 2658] network template widgets

mwarnsleyDE:
Edited all of the facets on the left hand side making them into individual include widgets that way we can reorganize them or switch them in and out depending on what page we are on. Also, not apart of this task but since I had it done I included the new logo that I created that wasn't pixelated like the original logo.


koonsies:
Is there a step to compile the CSS for this? Styling isn't loading for me.

PR #2561, opened 2016-09-29
[PD-2551] correct erroneous titles

koonsies:
## Overview

In the slug field for titles, certain characters are automatically trimmed. A few examples are '?', '.' and ','

If a title is erroneously created to contain ONLY those special characters, the job title slug will be saved as blank. This PR covers modifying the code slightly to handle blank titles

This also fixes a small bug that happens when you try to load job search results when there are no jobs in solr.

## Testing

The easiest way to load a job into the system with erroneous characters is to create one via postajob. If you have postajob set up on your system, simply create a job with a title such as "..." and view that job. It should work fine. On QC branch, it will raise a 500.

PostAJob docs
http://directemployers.github.io/MyJobs/developers/postajob/index.html

### DIFF
--- a/seo/search_backend.py
+++ b/seo/search_backend.py
@@ -37,7 +37,6 @@ class DESearchQuerySet(SearchQuerySet):
 
         facet_counts1 = self.facet_counts()
         facet_counts2 = sqs2.facet_counts()
-        import ipdb; ipdb.set_trace()
         # field_counts = {facet_field: [(facet_field_value, count), ...], ...}
         field_counts1 = facet_counts1.get('fields', {})
         field_counts2 = facet_counts2.get('fields', {})
@@ -54,7 +53,8 @@ class DESearchQuerySet(SearchQuerySet):
             field_counts1[facet_field] = field_dict.items()
             # Sort the list of field count tuples by count
             # (2nd value in each tuple)
-            field_counts1[facet_field].sort(key=lambda tup: tup[1], reverse=True)
+            field_counts1[facet_field].sort(key=lambda tup: tup[1],
+                                            reverse=True)
 
         facet_counts1['fields'] = field_counts1
         return facet_counts1
--- a/seo/search_backend.py
+++ b/seo/search_backend.py
@@ -37,10 +37,10 @@ class DESearchQuerySet(SearchQuerySet):
 
         facet_counts1 = self.facet_counts()
         facet_counts2 = sqs2.facet_counts()
-
+        import ipdb; ipdb.set_trace()
         # field_counts = {facet_field: [(facet_field_value, count), ...], ...}
-        field_counts1 = facet_counts1.get('fields', [])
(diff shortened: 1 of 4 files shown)


seo/search_backend.py
```diff
@@ -37,10 +37,9 @@ def add_facet_count(self, sqs2):
 
         facet_counts1 = self.facet_counts()
         facet_counts2 = sqs2.facet_counts()
-
         # field_counts = {facet_field: [(facet_field_value, count), ...], ...}
-        field_counts1 = facet_counts1.get('fields', [])
-        field_counts2 = facet_counts2.get('fields', [])
+        field_counts1 = facet_counts1.get('fields', {})
+        field_counts2 = facet_counts2.get('fields', {})
```
koonsies: facet_counts1.fields is a dict, this was defaulting to list for some reason and causing downstream exceptions when there were no jobs.

seo/views/search_views.py
```diff
@@ -442,12 +442,15 @@ def job_detail_by_title_slug_job_id(request, job_id, title_slug=None,
...
     # We only need the job id to be in the url, but we also put the title.
     # The offshoot of that is that if someone mistypes or mispells the title
     # in the url, then we want whoever clicks the link to be directed to the
     # canonical (and correctly spelled/no typo) version.
-    if (title_slug == the_job.title_slug and
+
+    def nvl(i, s):
+        return i if i else s
+
+    if (nvl(title_slug, 'na') == nvl(the_job.title_slug, 'na') and
```
koonsies: Not sure 'na' is a perfect default here. Basically, it keeps the URL from having a double slash (my.jobs//usa/jajdj1kam8371902as/job/)

templates/job_detail.html
```diff
@@ -123,11 +123,13 @@ <h3 id="direct_searchCriteriaHeader" style="float:left;">{% blocktrans %}Current
             </a>
           </li>
           {% endif %}
+          {% if crumbs.title %}
```
koonsies: Prevent a blank breadcrumb from appearing.

seo/views/search_views.py
```diff
@@ -442,12 +442,15 @@ def job_detail_by_title_slug_job_id(request, job_id, title_slug=None,
...
 
     # This check is to make sure we're at the canonical job detail url.
-
     # We only need the job id to be in the url, but we also put the title.
     # The offshoot of that is that if someone mistypes or mispells the title
     # in the url, then we want whoever clicks the link to be directed to the
     # canonical (and correctly spelled/no typo) version.
-    if (title_slug == the_job.title_slug and
+
+    def nvl(i, s):
```
tdphillips: Does nvl mean anything outside of SQL-land?

koonsies: Good call. I'll convert the naming into something more obvious haha.

PR #2562, opened 2016-09-30
Preserve newlines in NormalizedModelForm [PD-2602]

tdphillips:
# Testing
- Go to the "Add Partner" page
- Enter stuff containing multiple spaces
- In the notes textarea, include newlines in your input
- Save
- Verify that newlines were preserved

### DIFF
--- a/mypartners/forms.py
+++ b/mypartners/forms.py
@@ -205,21 +205,16 @@ class NewPartnerForm(NormalizedModelForm):
     def save(self, commit=True):
         # self.instance is a Contact instance
         company_id = self.data['company_id']
-        partner_url = self.data.get('partnerurl', '')
-        partner_source = self.data.get('partnersource', '')
+        partner_url = self.cleaned_data.get('partnerurl', '')
+        partner_source = self.cleaned_data.get('partnersource', '')
 
         status = Status.objects.create(approved_by=self.user)
-        partner = Partner.objects.create(name=self.data['partnername'],
+        partner = Partner.objects.create(name=self.cleaned_data['partnername'],
                                          uri=partner_url, owner_id=company_id,
                                          data_source=partner_source,
                                          approval_status=status)
         log_change(partner, self, self.user, partner, partner.name,
                    action_type=ADDITION)
-
-        self.data = remove_partner_data(self.data,
-                                        ['partnername', 'partnerurl',
-                                         'csrfmiddlewaretoken', 'company',
-                                         'company_id', 'ct'])
 
         create_contact = any(self.cleaned_data.get(field)
                              for field in self.__CONTACT_FIELDS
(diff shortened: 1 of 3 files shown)


universal/forms.py
```diff
@@ -19,9 +18,12 @@ class NormalizedModelForm(ModelForm):
...
     def clean(self):
-        self.cleaned_data = {key: ' '.join(value.split())
-                             # I don't see us porting to Python 3 any time soon
-                             if isinstance(value, basestring) else value
-                             for key, value in self.cleaned_data.items()}
-
-        return super(NormalizedModelForm, self).clean()
+        super(NormalizedModelForm, self).clean()
+        self.cleaned_data = {
+            key: '\n'.join(' '.join(line.split())
```
tdphillips: This appears to condense non-line-break whitespace. Do we wish to do the same for line breaks (replace multiple consecutive line breaks with a single one)?

mypartners/forms.py
```diff
@@ -205,22 +205,17 @@ class Meta:
     def save(self, commit=True):
         # self.instance is a Contact instance
         company_id = self.data['company_id']
-        partner_url = self.data.get('partnerurl', '')
```
tdphillips: `partnerurl`, `partnersource`, and `partnername` were being pulled from self.data, which is pre-clean. This resulted in those potentially containing a bunch of spaces/newlines despite clean having been run already.

mypartners/forms.py
```diff
@@ -205,22 +205,17 @@ class Meta:
...
         status = Status.objects.create(approved_by=self.user)
-        partner = Partner.objects.create(name=self.data['partnername'],
+        partner = Partner.objects.create(name=self.cleaned_data['partnername'],
                                          uri=partner_url, owner_id=company_id,
                                          data_source=partner_source,
                                          approval_status=status)
         log_change(partner, self, self.user, partner, partner.name,
                    action_type=ADDITION)
 
-        self.data = remove_partner_data(self.data,
```
tdphillips: This appears to be unnecessary. We explicitly define the fields we're getting (and pulling their values from self.cleaned_data, not self.data) so we should never run into problems having redundant params.

mypartners/tests/test_api.py
```diff
@@ -394,14 +394,11 @@ def test_outreach_api_contact_append_notes(self):
         self.contact.save()
         response = self.client.post(
             reverse('api_convert_outreach_record'),
-            data = {'request': json.dumps(self.request_data)}
+            data={'request': json.dumps(self.request_data)}
         )
         self.check_status_code_and_objects(response, 200, 8)
         self.contact = Contact.objects.get(id=self.contact.id)
-        # This is how we wish this could be written. Can be fixed along with
```
tdphillips: :boom: Your wish is my command. :boom:

universal/forms.py
koonsies: @fezick @JLMcLaugh Do we have a preference to this question? Should notes allow consecutive newlines, or trim them to one?

koonsies:
:+1: This looks good. I'll give Jason a bit longer for input on the newlines thing and then I'll merge it.

universal/forms.py
tdphillips: :ping_pong:

PR #2563, opened 2016-09-30
[PD-2631] Created v2 top bar

huy-nguyen-colorado:
"New" v2 topbar to match mock up.  v2 topbar uses  seo_base_styles.scss and bootstrap.min.css for styles, it still calls topbar.js for for mobile, however this will change once I'm done with PD-2630.  I changed myjobs > templatetags > common_tags.py file because the mock-up does not use the envelope icon anymore, instead it uses the text "Inbox".  I added conditional statement to myjobs > views.py to render the v2 topbar when v2 is requested, also added v2=1 to jsonp ajax call in seo-base-script.js

To test:
1. Change "admin" page to "My.jobs - Flexible flat Network (Texas Copy)...", and select: "home_page_static_header_footer_bootstrap3.html".

2. In the browser, navigate to www.my.jobs, please ensure that the "Employers / Post Jobs"
and "Create Account / Log In" is there and works when you click on them.

3. Please log in to www.my.jobs and ensure all on-hover drop down links works and go where they are suppose to go.

4. The mobile view will be in PD-2630.

### DIFF
--- a/myjobs/templatetags/common_tags.py
+++ b/myjobs/templatetags/common_tags.py
@@ -278,8 +278,7 @@ def get_menus(context):
     message_menu = {
         "label": "Messages",
         "id": "menu-inbox",
-        "icon": "icon-envelope icon-white",
-        "iconLabel": str(new_messages.count() if new_messages else 0),
+        "iconLabel": str(new_messages.count() if new_messages else "Inbox"),
         "submenus": [
             {
                 "id": "menu-inbox-all",
--- a/myjobs/templatetags/common_tags.py
+++ b/myjobs/templatetags/common_tags.py
@@ -278,7 +278,8 @@ def get_menus(context):
     message_menu = {
         "label": "Messages",
         "id": "menu-inbox",
-        "iconLabel": str(new_messages.count() if new_messages else "Inbox"),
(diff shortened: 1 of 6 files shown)


myjobs/templatetags/common_tags.py
```diff
@@ -278,8 +278,7 @@ def get_menus(context):
     message_menu = {
         "label": "Messages",
         "id": "menu-inbox",
```
tdphillips: Here's where the id should be coming from.

myjobs/templatetags/common_tags.py
```diff
@@ -278,8 +278,7 @@ def get_menus(context):
...
-        "icon": "icon-envelope icon-white",
-        "iconLabel": str(new_messages.count() if new_messages else 0),
+        "iconLabel": str(new_messages.count() if new_messages else "Inbox"),
```
tdphillips: And here's where I think the bug might be. If there are no messages, we don't show a number.

huy-nguyen-colorado: Agreed, I fixed it and re-commited about 20 minutes ago, Jenkins' still building.

PR #2565, opened 2016-10-03
[PD 2658] network template widgets

mwarnsleyDE:
This is more than just editing the widgets and making them into individual pieces. There is also the design concept merged in along with the functionality of the left hand facets having functionality in them along with the design to them. Also, the job details, homepage, job results, etc pages are also in order with the design and formatting with widgets as well.

To Test:
In the admin page, make sure that you are set on home_page_listing_bootstrap3.html and then the url to go to should be my.jobs. This will take you to the homepage for testing around on.


huy-nguyen-colorado:
Perhaps we should provide some instructions for the testers on how to test this, along with how to configure the settings in the Admin page.

huy-nguyen-colorado:
In the Chrome developer tool console, whenever I type something in the jobs search box or the location search box, I get this error in the console: "TypeError: a.curCSS is not a function" .  It seems to be happening on the second character of a string.

mwarnsleyDE:
That error seems to be tied in with the jQuery UI that we are using. I'm guessing it has to do with the auto complete function

huy-nguyen-colorado:
In the left column, in the "Filter By ..." section, the show more and show less links doesn't quite show more nor show less.  Also, the Filter by Company and by Country isn't there, I'm not sure if that is needed.

seo/tests/test_views.py
```diff
@@ -220,108 +220,108 @@ def check_for_string_on_results_pages(self, string):
         resp = self.client.get(reverse('all_jobs'))
         self.assertIn(string, resp.content)
 
-    def test_custom_search_box_labels(self):
+    # def test_custom_search_box_labels(self):
```
darrint: We need these tests either working or removed.

koonsies:
I see wide_header_bootstrap3.html in the root templates folder still. Is this supposed to be in the v2 folder as well?

mwarnsleyDE:
I'm not sure of those can go in there or not. Since Huy was working on the header and made the seo_base files, I left them where they were just because I didn't know if it would effect the topbar or not. @huy-nguyen-colorado might be able to chime in better and let me know but if they can go in there I will put them in there. I just didn't want to edit the name of his files or change the location because I don't know the effect they might have on topbar if any at all.

mwarnsleyDE:
There's also a topbar_v2 that I don't know if it can go in the V2 folder as well. I'm not too sure on the functionality or routing or anything that has to do with the header and the base template

koonsies:
It's not a huge deal to leave them there for now. For conciseness, I'd say it would be better to get them there eventually.

templates/includes/job_list.html
```diff
@@ -42,15 +42,14 @@
...
     </div>
 {% endif %}
-{% else %} 
+{% else %}
     {% arrange_jobs %}
     {% with fj_len=featured_jobs|length dj_len=default_jobs|length %}
         {% if fj_len|add:dj_len > site_config.num_job_items_to_show %}
         <span id="direct_moreLessLinks_listingDiv" data-type={% if data_type %}"{{ data_type }}"{% else %}{% trans "listing" %}{% endif %} class="more_less_links_container" offset="{{site_config.num_job_items_to_show|multiply_value:2}}" data-num-items="{{site_config.num_job_items_to_show}}">
-            <a id="button_moreJobs" href="#" rel="nofollow">More</a>
+            <a id="button_moreJobs" class="btn show-more-jobs" href="#" rel="nofollow">Show More Jobs</a>
```
JLMcLaugh: This template shouldn't be changing.

templates/seo_base_bootstrap3.html
```diff
@@ -89,260 +76,65 @@
                 <div class="row">
                   <div class="col-xs-12 col-md-12">
                     <div class="location-logo">
-                      <a class="location-logo-text" href="/">
-                        <img class="logo" src="{{ STATIC_URL }}images/dotJobsLogo.png" id="page_logo" alt="" width="150" height="90">
-                        <h1 class="logo-text">Virginia</h1>
-                      </a>
+                        <!-- <h1 class="logo-text">Virginia</h1> -->
+                        {% if site_config.home_page_template == "home_page/home_page_my_jobs.html" %}
```
JLMcLaugh: This should be an empty block that gets overridden on home_page_my_jobs.html.

JLMcLaugh:
:+1:

mwarnsleyDE:
Can I merge this or does someone else merge it instead?

PR #2568, opened 2016-10-04
Add space to profile page title [PD-2663]

tdphillips:
Update Docker docs

### DIFF
--- a/doc/source/developers/docker/use.rst
+++ b/doc/source/developers/docker/use.rst
@@ -214,7 +214,7 @@ Add to dev_settings.py::
 
 Run::
 
-    dkg npm run devserver
+    dk rundevserver
 
 Visit the webpack base url above in a browser. Accept the certificate.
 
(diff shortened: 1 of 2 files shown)


tdphillips:
Before:
2016-10-04-084231_312x99_scrot

After:
2016-10-04-093140_238x72_scrot

doc/source/developers/docker/use.rst
```diff
@@ -214,7 +214,7 @@ Add to dev_settings.py::
 
 Run::
 
-    dkg npm run devserver
+    dk rundevserver
```
tdphillips: This is what @mklauber said he was using and it works for me. If anyone is using the former, we might want to take a deeper look at why it was failing.

koonsies: Yup. Good catch. This command was changed but the docs weren't updated.

PR #2574, opened 2016-10-07
[PD 2666] styling of templates

mwarnsleyDE:
To Test:
In the admin switch over to the home_page_listing_bootstrap3.html template for the home template with Version 2 selected.

Styles have been added for the rebranding along with some functionality for desktop and mobile. Need to check and make sure all of the pages have the styling included along with the left hand facets, footer, navigation, etc.


fezick:
This looks good to me. @huy-nguyen-colorado or @JLMcLaugh, thoughts?

huy-nguyen-colorado:
I think we should merge this so we can start testing on an actual mobile device.  In the future, perhaps we should change the hyperlinks color in the submenu to look like hyperlinks everywhere else on the app, because in mobile there is no on-hover state when end users are using their fingers.

PR #2575, opened 2016-10-07
[PD-2630] started new mobile layout

huy-nguyen-colorado:
New mobile layout with improved responsiveness and mobile navigation.  The bottom nav submenu isn't complete yet, but the functionality should be there, knock on proverbial wood.

To test:
1. Either on a mobile web browser simulator of choice or, in a Jedi way, configure your phone to work with your dev environment, switch to v2 and navigate to www.my.jobs in mobile view (iPhone 5, 6, 6 plus, iPad.

2. Please start with "log out" state and ensure that there is an "Employers" and a "Log In" icon and text. Please click on both icons.

3. Please log in as admin user and test all links in bottom nav, it's fugly, but it should go where it's supposed to.

4. Please log in as "James" persona and test links, and ensure that there are only "Envelope" and "Profile" icons, and ensure that the submenu links works.  Again, fugly but functional.

### DIFF
--- a/gulp/src/sass/seo_base_styles.scss
+++ b/gulp/src/sass/seo_base_styles.scss
@@ -223,18 +223,11 @@ a.direct_optionsLess {
       display: none;
   }
 
-  #menu-inbox::before {
-    content: "Inbox ";
-  }
-
   #menu-inbox ~ ul li > p {
       color: white;
       padding-left: 4px;
   }
 
-  #menu-inbox > div.icon-envelope {
-      display: none;
-  }
  /* End legacy Topbar styles*/
 
 #last-microsite-name {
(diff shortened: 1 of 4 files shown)


koonsies:
I'm wondering if 991 pixels is a nominal size to switch to mobile view. It seems to be a bit big. Any particular reason we chose this size?

koonsies:
Other than the question about width, this looks good 👍

huy-nguyen-colorado:
J. Sole, Hal, nor the product owners, did not specify specific mobile breakpoints, so for now, I just went with 991 for the newer iOS and Samsung devices.  We can add more breakpoints if they request it.

arikamiles:
In the meeting in the war room, it was suggested some point in the 700s was a good spot, can’t remember exactly, but Jason S said that was where we break now.

huy-nguyen-colorado:
It would be better to create additional break points once Marcus and I merge our respective branches into quality-control

PR #2587, opened 2016-10-18
[PD-2667] Add time field for redirects

darrint:
@tdphillips Will this work?


tdphillips:
This should work.

darrint:
Okay.

PR #2591, opened 2016-10-20
Updated templates to manually version files. PD-2665.

mklauber:


tdphillips:
We used to version files using what's returned by `./version.py -b`.

mklauber:
Fixed.

PR #2592, opened 2016-10-20
[PD 2683] d3/react front end for analytics ui

mwarnsleyDE:
TEST:

You will need to make sure when you pull down the repository to do an NPM install inside of the gulp folder to install the dependencies that are added for the Recharts library. If you have the correct privileges inside of secure under the Beta tab there should be a tab that says Analytics. When you click on that, it should show the charts put on the page with the data from the API inside of them which you can also see is being retrieved inside of the code from the Action, Reducer, and API.js all inside of the Analytics folder inside of gulp. This is all being consumed inside of the respected Components inside of the components folder. Task was a basic demo of what the charts can look like with dumbie data, however, real data from the API has been supplied inside of the component charts instead. Will still need to refractor and do some more naming conventions for the Actions, Reducers, Components, etc within React but for now this gives an example of the charts with data.


analytics/migrations/0001_add_activities_roles.py
```diff
@@ -0,0 +1,38 @@
+# -*- coding: utf-8 -*-
```
tdphillips: Should this migration be located in the analytics app? I thought activity migrations were mostly located in myjobs.

tdphillips:
Anyone opposed to me merging this? I have one commit on this branch but that's resolving a merge conflict.

tdphillips:
I guess that answers that question.

PR #2600, opened 2016-10-25
[PD-2690] Editing the mobile solution for the logo alignment

mwarnsleyDE:
Editing the logo alignment in mobile and desktop view.

To Test:
In the admin panel, switch to any .jobs domain. Make sure you are on V2 templates and also that you are using the home_page_listing_bootstrap3.html template. Also it should be set to HTML5. Specifically domains with longer names should be used and check the logo and text in mobile and desktop view. Should be somewhat centered for most of the time. Essentially it is a quick fix for the problem as we take time to explore possible other options to handle longer and shorter names together.

### DIFF
--- a/static/seo-base-scripts.js
+++ b/static/seo-base-scripts.js
@@ -244,14 +244,3 @@ JAVASCRIPT FOR THE NEW REBRANDING HOMEPAGE FUNCTIONALITY
   }
 
   showHideHeaderScroll();
-
-// (function(){
-//   var siteName = $('.site-name').text();
-//
-//   if(siteName.length > 10){
-//     var currentPosLeft = 0;
-//     var newLength = siteName.length - 10;
-//     var posLeft = newLength * 4;
-//     $('.site-name').css('left', currentPosLeft - posLeft + '%');
-//   }
-// })();
--- a/static/seo-base-scripts.js
+++ b/static/seo-base-scripts.js
@@ -245,13 +245,13 @@ JAVASCRIPT FOR THE NEW REBRANDING HOMEPAGE FUNCTIONALITY
 
   showHideHeaderScroll();
 
-(function(){
-  var siteName = $('.site-name').text();
-
-  if(siteName.length > 10){
-    var currentPosLeft = 0;
-    var newLength = siteName.length - 10;
-    var posLeft = newLength * 4;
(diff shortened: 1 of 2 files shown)


static/seo-base-scripts.js
```diff
@@ -245,13 +245,13 @@ JAVASCRIPT FOR THE NEW REBRANDING HOMEPAGE FUNCTIONALITY
...
-  var siteName = $('.site-name').text();
-
-  if(siteName.length > 10){
-    var currentPosLeft = 0;
-    var newLength = siteName.length - 10;
-    var posLeft = newLength * 4;
-    $('.site-name').css('left', currentPosLeft - posLeft + '%');
-  }
-})();
+// (function(){
```
tdphillips: Should this be commented out?

mwarnsleyDE: Yes, I commented it out to make sure that my other solution was okay first and then I plan on going back in and deleting it.

koonsies: We typically don't commit commented code, as it has a way to getting forgotten about over time. What is a bit nicer is to simply use git to store the history.. Make a commit that has the code you may want in the future, then follow it with a commit that removes that code. Therefore, if you need it, it's in source control.. but not something you have to clean up later.

static/seo-base-scripts.js
```diff
@@ -244,14 +244,3 @@ JAVASCRIPT FOR THE NEW REBRANDING HOMEPAGE FUNCTIONALITY
...
-(function(){
-  var siteName = $('.site-name').text();
-
-  if(siteName.length > 10){
-    var currentPosLeft = 0;
-    var newLength = siteName.length - 10;
-    var posLeft = newLength * 4;
-    $('.site-name').css('left', currentPosLeft - posLeft + '%');
-  }
-})();
```
koonsies: 👍 much better.

mwarnsleyDE:
@tdphillips @koonsies I've taken out the commented code from the Javascript file.

mwarnsleyDE:
lol

mwarnsleyDE:
However, I'm still not too sure as to why my test are failing because of the Python test?

mwarnsleyDE:
@tdphillips @koonsies Any ideas why the python test keep failing on this one?

PR #2604, opened 2016-10-26
[ST-3219] Another huge PR

koonsies:
## Overview

This is an update to the existing URL redirect logic we use for DSEO search results. I added an additional conditional that checks if the company_slug filter has any capitals, and if it does, redirect to the proper, lower-case company name.

## Testing

Go here (or another company, but make a few letters capital)
http://www.my.jobs/USA/jobs/bridgestone-AMericas-inc/careers/

Notice how it corrects the URL error.

Wooooooooooo.

### DIFF
--- a/seo/helpers.py
+++ b/seo/helpers.py
@@ -1142,7 +1142,7 @@ def determine_redirect(request, filters):
     query_loc = request.GET.get('location', None)
     filter_moc = filters.get('moc_slug', None)
     query_moc = request.GET.get('moc', None)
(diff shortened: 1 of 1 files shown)


seo/helpers.py
```diff
@@ -1187,6 +1188,10 @@ def determine_redirect(request, filters):
         filters['location_slug'] = None
         needs_redirect = True
 
+    if any(x.isupper() for x in filter_company):
```
koonsies: I thought about that as well. Not sure which is actually faster.

PR #2605, opened 2016-10-27
[St-3226] autocompletes for search fields

huy-nguyen-colorado:
Updated jQuery to 2.2.4 (latest release in the series) and custom jQuery-UI to the latest version but only included methods for autocomplete and accordion (to keep things light), added manual versioning to deferred JavaScript files, added auto complete.

To test:

1. Switch to v2 templates.

2. Auto-complete will fire on second character of string.

3. If the "Job Name" and the "Job Location" is valid, then the form will automatically submit.  In MOC, if the "Job Name",  the "Job Location", and the Military Occupational Specialty box is valid, the form will automatically submit, I made it like this to match what we already have in v1 Production app.


tdphillips:
When we talked about versioning, we were thinking of just our code - the jQuery files are already versioned.

huy-nguyen-colorado:
I removed the manual versioning of the the jQuery files just now.

templates/seo_base_bootstrap3.html
```diff
@@ -166,15 +166,12 @@ <h1 class="logo-text">
...
-<script defer src="{{ STATIC_URL }}seo-base-scripts.js"></script>
-
+<script defer src="{{ STATIC_URL }}jquery-2.2.4.min.js"></script>
+<script defer src="{{ STATIC_URL }}jquery-ui-1.12.1.custom.min.js"></script>
+<script defer src="{{ STATIC_URL }}seo-base-scripts-181-27.js"></script>
 <script defer src="{% static "utils.181-20.js" %}"></script>
 <script defer src="{% static "secure-blocks/secure-block.181-20.js" %}"></script>
-<script defer src="{% static 'search.181-20.js' %}"></script>
-
+<script defer src="{{ STATIC_URL }}pager.164-21.js"></script>
```
mklauber: I'd prefer if we used the `{% static 'pager.164-21.js' %}` format for lines like this.  `{{STATIC_URL}}` bypasses some of the handling that the `static` tag does.

templates/includes/topbar_v2.html
```diff
@@ -117,5 +117,5 @@
 </script>
 {% endif %}
 {% compress js %}
-<script defer src="{% static "topbar_v2.js" %}"></script>
+<script defer src="{% static "topbar_v2-181-27.js" %}"></script>
```
mklauber: Oops, need to remove the compress tags around this, or you'll lose the defer attribute on it.

huy-nguyen-colorado:
I fixed all items on the code review feedback and checked them in.  Also updated the UI testing instructions for J. Sole.

PR #2621, opened 2016-11-10
PD-2735, moved global saved search style properties to the root scope

huy-nguyen-colorado:
Purpose:  Fix broken styles for the Saved Search element.

To test: 

1. Please switch to v2 template.

2. Please search for a job without location nor MOC, to trigger the rendering of the Save Search "box".

3. If things goes correctly, the Saved Search box should no longer look like a naked poodle.  Please compare with before and after pics.

Before:
before

After:
after

### DIFF
--- a/gulp/src/sass/seo_base_styles.scss
+++ b/gulp/src/sass/seo_base_styles.scss
@@ -360,24 +360,60 @@ a.direct_optionsLess {
 }
 
 #mobile_direct-saved-search-box, #direct-saved-search-box {
+  padding: 0 15px;
+  margin-bottom: -3.1rem;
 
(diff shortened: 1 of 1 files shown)


koonsies:
Could we switch the placeholder text for the email slot to something more fitting than "clothilde_crooks@gmail.com"? Does have to be on this PR, but probably worth doing on a following PR.

koonsies:
It also may be nice to add a little padding below "Cancel Anytime". It feels a bit crowded to me. Not much, I mean 2-5 pixels.

haliscooper:
Ticket in to remove cancel anytime already

PR #2625, opened 2016-11-14
PD-2747, changed save search email placeholder text

huy-nguyen-colorado:
Purpose: change Save Search email placeholder verbiage to something more intuitive.  Also removed commented out html codes that probably shouldn't be in production.

To test:

1. Please switch to v2 template.

2. Please save a job title search, please ensure that the email placeholder text says "Please type your email address here".

Before:
before

After:
after


koonsies:
"Please type your email address here" is a bit wordy and winds up being cut off on many browsers. Perhaps just "Your email" or something similar would work better.

huy-nguyen-colorado:
I modified the placeholder text.

PR #2627, opened 2016-11-15
PD-2733, refactored v2 seo base templates to prevent double filters

huy-nguyen-colorado:
Purpose:  Render 1 filter list for both desktop and mobile through responsive design, rather than 2 filter lists for the exact same code like it is currently.

1.  Please switch to v2 template.

2.  In the desktop and mobile views, please use web developer tool to inspect the DOM and ensure that the direct_disambiguationDiv element only appears once.  Everything else should look exactly the same.

Before fix, notice 2 filter DOM elements performing the exact same work:
before

After, 1 filter element for desktop and mobile:
after

### DIFF
--- a/templates/home_page/home_page_listing_bootstrap3.html
+++ b/templates/home_page/home_page_listing_bootstrap3.html
@@ -39,6 +39,10 @@
 {% endblock directseo_main_content %}
 
 {% block directseo_right_hand_column %}
+<section class="mobile-search-criteria">
+    <button id="mobile_search" class="mobile-search-btn">Filter Search Criteria</button>
+</section>
+<section class="mobile-search-facets">
 {% if default_jobs or featured_jobs %}
 <div id="direct_disambiguationDiv" class="facets direct_rightColBox">
   <div class="filter-box">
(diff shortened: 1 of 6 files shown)


templates/home_page/home_page_listing_bootstrap3.html
```diff
@@ -5,7 +5,6 @@
 {% cache 600 joblisting %}
 
 {% block direct_extraHeaderContent %}
```
huy-nguyen-colorado: Calling unnecessary v1 stylesheet, v2 should be develop in isolation from v2 anyway to avoid UI bugs.

templates/home_page/home_page_listing_bootstrap3.html
```diff
@@ -41,7 +40,6 @@
 
 {% block directseo_right_hand_column %}
 {% if default_jobs or featured_jobs %}
-<a name="filters"></a>
```
huy-nguyen-colorado: Anchor tag name attribute no longer part of html5 standard.

templates/home_page/home_page_listing_bootstrap3.html
```diff
@@ -54,25 +52,6 @@
 {% endif %}
 {% endblock directseo_right_hand_column %}
```
huy-nguyen-colorado: Duplication of work of the "direct_disambiguationDiv" element.

koonsies:
:+1: awaiting test completion

PR #2629, opened 2016-11-15
ST-3339, hot fix for filter by verbiage

huy-nguyen-colorado:
To test:

In v1 template the text "Filter by" should precede filter name.


tdphillips:
This appears to revert the v2 fix. Is that intended?

tdphillips:
@JLMcLaugh answered. Looks good to me. I'll merge once tests pass.

PR #2633, opened 2016-11-15
PD-2748, remove the 'Filter by' text within the v2 conditional statement

huy-nguyen-colorado:
Purpose:  Remove the "Filter by" verbiage only within the v2 conditional statement.  Leave v1 alone man!

To test:

1. In v1 template, please ensure that the verbiage "Filter by" precede filter criterias.

2. In v2 template, please ensure that the verbiage "Filter by" is not  preceding the filter criterias.

### DIFF
--- a/seo/filters.py
+++ b/seo/filters.py
@@ -253,10 +253,14 @@ class FacetListWidget(object):
         accordion_class = '<h3>'
         if self.version == 'v2':
             accordion_class = '<h3 class="filter-accordion">'
+            column_header = (accordion_class +
+                             '<span class="direct_highlightedText">%s</span></h3>')
+            column_header = column_header % self.get_title()
+            return column_header
         column_header = (accordion_class +
+                         '<span class="direct_filterLabel">%s</span> '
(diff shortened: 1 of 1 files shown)


seo/filters.py
```diff
@@ -253,10 +253,14 @@ def _render_header(self):
         accordion_class = '<h3>'
         if self.version == 'v2':
             accordion_class = '<h3 class="filter-accordion">'
+            column_header = (accordion_class +
+                             '<span class="direct_highlightedText">%s</span></h3>')
+            column_header = column_header % self.get_title()
+            return column_header
         column_header = (accordion_class +
```
tdphillips: This will overwrite what you just set on line 256. You'll probably need an `else` somewhere.

huy-nguyen-colorado:
I added an else statement for the v1 and return the column_header for the function.

tdphillips:
Sorry, didn't see the return in the if block. I still think the latest commit is an improvement, though. Thanks.

PR #2634, opened 2016-11-16
PD-2739, added Google Search Engine Result Pages (SERP) json-ld

huy-nguyen-colorado:
Purpose: Add Google SERP, JIRA task PD-2739

To test:

1.  Switch to v2.

2. Go to www.my.jobs and/or veterans.jobs, open browser web developer tool, please ensure this code is in the head section: 

Should look like this:

my_dot_jobs

veterans_dot_jobs

### DIFF
--- a/templates/home_page/home_page_listing_bootstrap3.html
+++ b/templates/home_page/home_page_listing_bootstrap3.html
@@ -14,10 +14,10 @@
         {
             "@context": "http://schema.org",
             "@type": "WebSite",
-            "url": "http://www.indianapolis.jobs/",
+            "url": "{{request.get_host}}",
             "potentialAction": {
                 "@type":"SearchAction",
-                "target":"http://www.indianapolis.jobs/jobs/?q={q}&loc=en_us&source=sitelinks_searchbox",
+                "target":"{{request.get_host}}/jobs/?q={q}&loc=en_us&source=sitelinks_searchbox",
                 "query-input":"required name=q"
             }
         }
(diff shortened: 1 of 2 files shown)


templates/home_page/home_page_listing_bootstrap3.html
```diff
@@ -9,6 +9,21 @@
 
 {% endblock direct_extraHeaderContent %}
 
+{% block meta_serp_schema %}
+    <script type="application/ld+json">
+        {
+            "@context": "http://schema.org",
+            "@type": "WebSite",
+            "url": "http://www.indianapolis.jobs/",
```
koonsies: What is the significance of this url being hardcoded to indianapolis.jobs?

haliscooper: That would be bad.  The ticket says
"Where the URL and Target must be driven by which site the user is on."

templates/home_page/home_page_listing_bootstrap3.html
```diff
@@ -9,6 +9,21 @@
...
 
+{% block meta_serp_schema %}
+    <script type="application/ld+json">
+        {
+            "@context": "http://schema.org",
+            "@type": "WebSite",
+            "url": "{{request.get_host}}",
+            "potentialAction": {
+                "@type":"SearchAction",
+                "target":"{{request.get_host}}/jobs/?q={q}&de_n=google&de_m=web&de_c=sitelinks",
```
tdphillips: Is q a template variable or will it be interpreted by whatever makes use of this script?

If it's a Django template variable, the brackets need to be doubled.

huy-nguyen-colorado: It's for the script, it acts as sor of a placeholder.

PR #2643, opened 2016-12-01
PD-2757, moved seo-base-scripts.js to gulp directory

huy-nguyen-colorado:
Purpose:  Relocate the seo-base-scripts.js file from the static directory to the gulp directory in order to utilize the goodness of WebPack and es6.  Seems like a simple thing but this PR actually lays the foundation for using es6 project wide in the v2, in the future.  I also removed the "defer" attribute on the script tags which I added a couple of months ago.  I removed the defer because it was conflicting with the webpack-dev-server version 1.14.1 (known webpack issue).  Unexpectedly, page load times are faster with es6 and WebPack without the defer attribute than the old version with the defer attribute.  I emptied cache and hard reload 10 times in each version, the median load times for the new version is faster, also, in production the webpack-dev-server and websocket will not be used so load times will even be faster than in my local dev environment.
Before:
before

After (more file request because of local webpack dev server):
es6-webpack

Suggested Testing Plan: 

1. Switch to v2 template.

2. In desktop view, please test job and location input boxes for the autocomplete dropdown suggestion box, which should be triggered on the second character input.  Please ensure that when both input boxes are completed, the search function should fire (all 3 input boxes in the case of MOC).  Please ensure the "Save & Email this search" facet appears after the api has return jobs. Please ensure that the "Show More Jobs" button works.

3. Please click the Save button in the "Save & Email this search" facet and ensure an email is generated and sent to your designated inbox.

4. In the facets, with the exception of the "Save & Email this search" facet, please ensure that the facets are collapsible and expandable.  Inside of the facets, please ensure the "Show More" and "Show Less" (where applicable) works.

5. In mobile view, please click on the "Filter Search Criteria" button and repeat steps 2 through 4 above.

6.  In mobile view, please refresh the page. Please scroll down, the top bar should disappears as you scroll down and reappears as you scroll up.

### DIFF
--- a/gulp/src/sass/seo_base_styles.scss
+++ b/gulp/src/sass/seo_base_styles.scss
@@ -587,9 +587,6 @@ margin-bottom: 50px;
     @extend .global-btn;
     text-align: center;
   }
-  .share-social.search-submit-moc {
-    min-width: 7rem;
-  }
   .search-show-all {
     font-size: 12px;
     position: relative;
@@ -1221,12 +1218,6 @@ margin-bottom: 50px;
     .search-submit-moc {
       width: 94%;
     }
-    .share-social.search-submit {
-      display: none;
-    }
-    .share-social.search-submit-moc {
-      display: none;
-    }
     .breadcrumb-row {
       display: none;
     }
@@ -1548,12 +1539,6 @@ margin-bottom: 50px;
     }
     .submit-section {
       padding: 0;
-    }
-    .share-social.search-submit {
-      display: inline-block;
-    }
-    .share-social.search-submit-moc {
-      display: inline-block;
     }
   }
   #direct_listingDiv {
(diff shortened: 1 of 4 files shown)


gulp/src/sass/seo_base_styles.scss
```diff
@@ -587,9 +587,6 @@ margin-bottom: 50px;
     @extend .global-btn;
     text-align: center;
   }
-  .share-social.search-submit-moc {
```
tdphillips: This seems unrelated to moving seo-base-scripts.js

huy-nguyen-colorado: In the old seo-base-scripts-181-27.js file, lines ~ 294-305, there was code to manipulate this bit of styles. It's to control the "Share" social media button.  That button is now gone and social media icons are placed below the "Show More Jobs" button. So since this functionality is no longer in the new seo-base-scripts file, I decided to whack this css selector in this PR.

PR #2651, opened 2016-12-06
PD-2776, remove vendor.js file and references to it

huy-nguyen-colorado:
Purpose:  Remove the vendor.js file and all references to it because it was no longer being used, furthermore, it was adding 830K of extra weight to the load time.

To Test:

1.  Please switch to the v2 template.

2. In the dev tool network tab, please ensure that the seo_base_scripts.js file isn't 800k or more in size.  In the actual QC env, this file might be call 4bb123434566.js.

3. Please also do a functionality check of NUO, the profile page, dynamic reports, and  analytics.

### DIFF
--- a/gulp/gulpfile.js
+++ b/gulp/gulpfile.js
@@ -14,9 +14,6 @@ process.on('SIGINT', function() {
 
 // This build produces several javascript bundles.
 //
-//     In production this bundle also contains whatever code webpack deemed
-//     as common to all app bundles.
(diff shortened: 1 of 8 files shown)


darrint:
I checked all the affected apps under secure and they are working.

huy-nguyen-colorado:
I removed the old comment about vendor.js bundle in the gulp.js file.

PR #2654, opened 2016-12-07
[PD-2766] Switch to using the same interface to MongoDB that MyJobs-Analytics uses.

almathew:
This allows for improved consistency between MyJobs and MyJobs-Analytics and also allows for MyJobs to connect to Mongo Atlas. 

It requires a few settings:
MONGO_SSL (new setting, a boolean for whether or not a secure connection is required)
MONGO_DBNAME (new setting)
MONGO_HOST or MONGO_CN (MONGO_CN is a new setting, and is the same as MONGO_HOST except pointing to a cluster instead)

### DIFF
--- a/analytics/api.py
+++ b/analytics/api.py
@@ -13,6 +13,7 @@ from dateutil import parser as dateparser
 from myjobs.decorators import requires
 from universal.helpers import get_company_or_404
 
+
 @requires("view analytics")
 def views_last_7_days(request):
     """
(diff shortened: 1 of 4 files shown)


tdphillips:
@mklauber, any thoughts? Those failures don't make any sense.

almathew:
I'm thinking it's because settings are now being imported somewhere they weren't previously or something similar to that.

tdphillips:
Any reason we can't just import secrets instead of settings?

almathew:
No real reason other than I always preferred settings as an interface to secrets rather than accessing secrets directly. 

I was considering switching to that anyway, but I'm writing tests on pymongo-env right now and hadn't had a chance to try it out. I would feel a lot more comfortable doing it that way if there were some definition around the expected contents of settings vs. secrets.

PR #2655, opened 2016-12-07
[PD-2762] analytics start api

darrint:
This creates a starting point api for analytics.

## Review

* There are unit tests for the happy path
* I use `ApiValidator` for handling errors in a consistent way. It's also consistent with reporting.
* There are unit tests for the error cases.
* I use `value` to represent keys and ids that we want returned to us. I use `display` to represent things shown to the user. (We should convert the other analytics api call to follow that.)
* I'm still _not_ dealing with the fact that this api assumes that we are interested in analytics. We can add a parameter to `api/available-reports` when we are ready.

## Test

* If you have a working postman interface you can hit the api:
  * `https://secure.my.jobs/analytics/api/available-reports`
  * `https://secure.my.jobs/analytics/api/report-info?analytics_report_id=job-locations`

### DIFF
--- a/analytics/api.py
+++ b/analytics/api.py
@@ -9,10 +9,15 @@ from django.shortcuts import HttpResponse
 from django.http import HttpResponseNotAllowed, Http404
 from django.views.decorators.csrf import csrf_exempt
 
+from universal.api_validation import ApiValidator
+
 from dateutil import parser as dateparser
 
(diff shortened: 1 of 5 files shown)


koonsies:
Hahaha. It wasn't until I started reviewing this PR that I noticed my API's lack of tests. I guess I have a little side PR to put up...

PR #2659, opened 2016-12-08
[PD 2761] setup ui to make initial api call for primary dimension

mwarnsleyDE:
This task was for consuming the API given to me by Darrin for setting the initial dimensions for the starting points. The dimensions are loaded in the sidebar coming from the API. Also inside of here is the task for changing and updating the initial starting point data date-range which should be the last 30 days. That can be seen inside of the analytics/components/analyticsapp.jsx component. 
Ignore console logs and commented out code that will be displayed inside of ESLint.

To Test:

You may need to possibly run a dkm migrate.
Will also need to make sure that you run an NPM install as some of the package.json has changed.
When pulling up the analytics app in the BETA tab, you should see that the initial data is still coming from the API provided by Jonathan with a date range of the last 30 days BUT the sidebar is being populated by the starting point API provided by Darrin.


gulp/src/analytics/components/Charts/ChartContainer.js
```diff
@@ -1,7 +1,11 @@
 import React from 'react';
 import {Component} from 'react';
 import {Row, Col} from 'react-bootstrap';
-import SimpleBarChart from './Bar/SimpleBarChart';
+// import SimpleBarChart from './Bar/SimpleBarChart';
```
darrint: We should clean up the commented out code.

gulp/src/analytics/components/Tabs/TabsContainer.js
```diff
@@ -1,17 +1,29 @@
 import React from 'react';
 import {Component} from 'react';
+// import $ from "jquery";
```
darrint: Does this mean we can remove the jquery requirement from `package.json`?

mwarnsleyDE: I want to say YES for right now, but I may need it in the future. I was experimenting with a little bit of jQuery and some responsive stuff so that's why the dependency is there. I don't think I use it in anything else but I just had it there experimenting with some things.

darrint: Let's take it out. We can put it back if we really need it. For now you can just tell eslint that `$` is global in certain files and it will work without adding the dependency.

PR #2668, opened 2016-12-12
Remove encoding from opengraph [ST-3417]

tdphillips:


tdphillips:
As far as I can tell, we only need to quote company name in the MSCCN section.

PR #2669, opened 2016-12-12
[PD-2740] Send an email if fewer than 1000 solr-related tasks have been processed.

almathew:


tasks.py
```diff
@@ -808,6 +814,37 @@ def requeue_failures():
         print "Requeuing task with args %s" % task.args
 
 
+@task(name="tasks.check_total_throughput", ignore_result=True)
+def check_total_throughput():
```
almathew: Good call. I missed adding default_settings.py when adding my changes. It's there now.

PR #2670, opened 2016-12-12
[PD 2785] loading initial data into chart for starting point

mwarnsleyDE:
This is a combination of 2 PR. 

To Test:
You might have to run npm install when you pull down the branch if you aren't up to date.
The initial data should be loading into the table from the default job-locations that we have selected. As well, the locations World Chart should be rendering with the countries available in the table highlighted. This is coming from the initial data that is displayed on first render of the page as well. If need be, you can check the Network request tab and see the request for the page data coming from the available-request and dynamic fetch request.


koonsies:
So, as of this point, nothing is clickable, correct? Drilldown, main nav, the "x" on tab? Wanted to make sure that was intentional and not an issue with my local.

koonsies:
As of now, I can't interact with any of the controls on this PR. If anyone is able to get the drilldown links etc working, let me know.

PR #2681, opened 2016-12-15
[PD 2802] store active filters for api

mwarnsleyDE:
/*
NEEDS TO BE MERGED AFTER PR-2785 AS IT'S A BRANCH OFF OF IT
*/

Storing the active filters inside of Redux so the correct data is being posted to the API for data return.

To Test:
Pull Down the repository (If not up to date might have to run an NPM install) and use the filtering on the table and the left hand sidebar that gives you the primary dimensions. The return should be the correct data coming back from the API from what's being posted. You can also switch the tabs and filter again which will send the correct data off and send a tigger to remove ALL of the tabs in front of it as requested.


tdphillips:
Now that the other PR has been merged, this has conflicts.

mwarnsleyDE:
PR should be fixed now with the pull request

PR #2682, opened 2016-12-16
[PD 2808] loading in all charts for display

mwarnsleyDE:
This task was to load in the charts for display. The charts are switching depending on the data coming in, however the data is current dumbie data that is being supplied. The actual data will go back in once the API is complete for chart switching in which I will have to make some modifications to the logic behind this. 

To Test:
Pull down the repository and open analytics. You WILL need to run an NPM install as the packages have been changed and updated. You can filter down inside of the table and see the charts switch depending on which data is present from the Maps to the Bar Chart.


mwarnsleyDE:
I will do a pull once the other PR is merged which should fix this testing error

gulp/src/analytics/components/Charts/Bar/BarChart.js
```diff
@@ -1,100 +1,43 @@
-import d3 from 'd3';
 import React from 'react';
 import {Component} from 'react';
-import Axis from '../Common/Axis';
-import Grid from '../Common/Grid';
+import {BarChart, Bar, XAxis, YAxis, CartesianGrid, Tooltip, Legend, ResponsiveContainer} from 'Recharts';
```
tdphillips: If I lowercase "Recharts", my bundle builds fine.

mwarnsleyDE: Ok, I don't know why I uppercased that in the first place but I just changed it and pushed

PR #2683, opened 2016-12-16
[ST-3428] Refresh Role data on new page load.

almathew:
This ensures that changes made on the Users page are reflected on the Roles page.


koonsies:
Now -THIS- is a PR I can handle.

PR #2686, opened 2016-12-16
Adds (some) initial data.

almathew:
### Goal
Given a completely empty database, created via `syncdb --all`, create data that provides a minimal starting point for most of the functionality we work with.

This is capable of creating or "fixing" the initial data (if, for some reason, it already exists).

### What has been done
- Sets up an initial superuser and a couple generic users, mainly for User Management purposes.
- Create all apps and activities. **I've done this in a way that some might not agree with**
- Creates a company, assigns a business unit to that company and the "Admin" role for that company (with all apps/activities properly assigned).
- Sets up a few microsites - a copy of www.my.jobs (a few actually, but all using the same config), a company site, and a regional site.

### Known remaining items
- Adding fake jobs to Solr.
- (Possibly?) Adding fake data to MongoDB.
- Set up filtering for indianapolis.jobs so only fake jobs for Indianapolis show up on that site.

Now that I've gotten this far, I would be interested in any feedback about what 100% absolutely is needed next, or how I should approach things differently.


almathew:
Found out what we were using the initial data for....

tdphillips:
What would that be?

myjobs/management/commands/loaddevdata.py
```diff
@@ -0,0 +1,235 @@
...
+        defaults = {
+            'email': 'dev@apps.directemployers.org',
+            'is_staff': True,
+            'is_superuser': True,
+            'is_verified': True,
+            'first_name': 'DirectEmployers',
+            'last_name': 'Developer',
+            'user_guid': 'dev',
+        }
+        user, _ = update_or_create(User, pk=1, defaults=defaults)
```
tdphillips: I would argue that this should possibly only be done on an empty database.

almathew:
A bunch of the tests rely on the initial data to be there. I'm not sure they're relying on that specific data, but they definitely expect something to be in the database.

myjobs/management/commands/loaddevdata.py
almathew: I did it this way because there were points in time I just wanted to fix the data without rebuilding the database (MySQL is slow!).

I have no strong feelings one way or the other, and can see your argument for safety, so consider it done.

almathew: Addressed.

myjobs/management/commands/loaddevdata.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # Clone of www.my.jobs.
+        defaults = {
+            'domain': 'www.my.jobs',
+            'group': group,
+            'site_title': 'Dev My.jobs',
+            'site_heading': 'Dev My.jobs',
+            'site_description': 'The Right Place for Developers.',
+        }
+        site = SeoSite.objects.create(pk=1, **defaults)
```
tdphillips: With a fresh syncdb, I've already got a site with pk=1.

darrint: Also, why did it turn out to be important to have a site with pk=1?

PR #2689, opened 2016-12-19
[PD 2809] replacing calendar

mwarnsleyDE:
This task was to replace the bootstrap daterangpicker with one that we currently have.

To Test:
Pull down the repository. You can click on the button with the Date in the header and see the calendar come down. The first part is the quick range selections which will trigger an API call and you can see the correct data going into the table in terms of filtering. You can also see the response coming back in the network Tab. If you click custom range inside of the dropdown it will pop out the calendar which has NOT been linked as of yet. This is be a different PR to set it with two calendars as a date range selection.

### DIFF
--- a/gulp/src/analytics/actions/calendar-actions.js
+++ b/gulp/src/analytics/actions/calendar-actions.js
@@ -1,11 +1,11 @@
 import {createAction} from 'redux-actions';
 
-// import {markNavLoadingAction} from '../../common/actions/loading-actions';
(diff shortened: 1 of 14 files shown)


mwarnsleyDE:
@darrint Hey Darrin, I'm not sure what test these are but they look like something from another PR that might be going right now. I pulled from quality control before I submitted and I'm not sure why it would be waiting on other PR's.

darrint:
@mwarnsleyDE Do another merge and push.

PR #2691, opened 2016-12-19
[PD 2786] user friendly no data display

mwarnsleyDE:
This task was applying a user friendly message displaying when there is no data to go into the chart or the table.
SHOULD BE MERGED AFTER 2809 as it is a branch off of it.

To Test:
Pull down the repository. You can change the date range in the calendar to today or yesterday which should provide no data at all. This should show the no results found component for the table and the chart.


mwarnsleyDE:
I believe checks are expected to fail until PR 2809 is merged.

PR #2694, opened 2016-12-19
[PD 2813] Editing all of the files documentation to meet standards. Took out un…

mwarnsleyDE:
This task is simply documenting the code written to meet with the standards and how we are doing it in other apps. Trying to stay updated and keep everything uniformed.

To Test:

Pull down the repository or just view the files changed and you can see the documentation inside of the files. You can also just make sure everything is still working correctly which it should. Code was removed but mostly all of it was commented out and not used or empty data I used for testing before.


mwarnsleyDE:
Love the enthusiasm @darrint

PR #2696, opened 2016-12-20
MyJobs now with more Vagrant.

almathew:
### What is it?
Run MyJobs in a VM that builds itself. An alternative option, similar to what we use Docker for.

Realistically I don't expect anyone other than @JLMcLaugh and maybe @mklauber to try this, but if you're interested the README has some good basic info, and I'd be thrilled to help too. 

### To Do
* I'd like to do a custom Vagrant box that already has all of our default packages installed. This will *really* speed up the setup process.
* It needs nginx before it's ready for completely easy usage, but I don't want to deal with switching between branches and stuff until that's ready, because that will probably take me a little while to configure.
* Tons more custom aliases and commands to make things we do all the time easier.
* I'm still experiencing flakiness with nodejs, so I'd like to figure out why that is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,10 @@
 dk2/dev/requirements.txt
 dk2/revproxy/
 
+# Vagrant
+.vagrant
+vagrant/id_rsa*
+
 # Misc
 files/
 reports/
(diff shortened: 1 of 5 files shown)


almathew:
I'm going to self-maintain this until I work out some wonky-ness with nginx (since it's unlikely to otherwise be used anyway).

PR #2697, opened 2016-12-21
[PD 2816] creating custom range selection for calendar

mwarnsleyDE:
Task was to create a date range selection with two calendars from the header.

To Test:

Pull down the repository. You can click on the date in the header and bring up the dropdown for quick selections. When you click custom range you will see two calendars appear. One being the start date and one being the end date. You can select that range and click apply and it will apply the filters to the data with those dates being sent to the API. You can also check the Network tab and see the request sent to Dynamic and make sure the headers are sending off the custom range that you selected with start_date and end_date in the request.

### DIFF
--- a/gulp/src/analytics/actions/calendar-actions.js
+++ b/gulp/src/analytics/actions/calendar-actions.js
@@ -1,5 +1,6 @@
 import {createAction} from 'redux-actions';
 import {markNavLoadingAction} from '../../common/actions/loading-actions';
+import {errorAction} from '../../common/actions/error-actions';
 
 export const setSelectedEndMonth = createAction('SET_SELECTED_END_MONTH');
 export const setSelectedEndYear = createAction('SET_SELECTED_END_YEAR');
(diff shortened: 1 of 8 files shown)


koonsies:
When you click the drop down for year, you have to scroll down to get to any relevant dates. This is mostly due to the fact that you can select up to the year 2040. Rather than implement some fancy scroll logic, we shouldn't include any years beyond the present, as we can't possibly have analytics/job view data for the future.

tdphillips:
That depends. If members are able to re-run analytics reports once analytics and reports are integrated more, selecting future dates may be desired.

koonsies:
Small nitpick, but would it be easy to make the calendar display be the same height as the initial dropdown? Right now, there's some whitespace that appears below the Cancel and Apply buttons.

mwarnsleyDE:
I noticed that as well. I would have to tinker with some of the css for it. They are individual components wrapped inside of the same Div container so even if the Calendar drops down farther it won't stretch the dropdown so I would have to take a look at that and see what I could do

koonsies:
@mwarnsleyDE it is fine to leave it as-is. It's just something we could polish up in the future.

@tdphillips I see where you're coming from.. As it stands, we don't save analytics like we save reports. For now, I think it's safe to shorten the date range. A compromise could be to include the next year, or two.. but perhaps not all the way to 2040 haha.

Of course, if it's easy to have the dropdown scroll to the current year, that also works.

PR #2702, opened 2016-12-22
[PD 2823] closing current tab leaves blank screen

mwarnsleyDE:
Task was to fix the closing of the tabs and active tabs

To test:

Pull down repository and filter the tabs to create new tabs. You can exit out of tabs and close tabs according to the X inside of the tab itself. They should close and if you close the active tab it should switch to another one being active.

### DIFF
--- a/gulp/src/analytics/components/Table/Table.js
+++ b/gulp/src/analytics/components/Table/Table.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import {Component} from 'react';
-import TableSearch from './TableSearch';
 
 class Table extends Component {
   render() {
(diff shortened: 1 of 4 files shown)


gulp/src/analytics/components/Table/Table.js
```diff
@@ -1,14 +1,12 @@
...
 import {Component} from 'react';
-import TableSearch from './TableSearch';
 
 class Table extends Component {
   render() {
     const {tableData} = this.props;
     return (
       <div id={'table_data_tab_' + tableData.navId} className="table-data">
         <div id={'table_search_tab_' + tableData.navId} className="table-search">
-              <TableSearch searchData={tableData}/>
```
koonsies: What is this change? I don't remember it being in the description.

mwarnsleyDE: Oh sorry Jonathan, Jason wanted to remove the search from the table since it wasn't active so I just deleted the component

PR #2704, opened 2016-12-27
[PD-2828] Analytics bar graph label and tooltip display improvements.

almathew:
In this pull request:
* Currently the tooltip text is truncated. This prevents the job title in the tool tip from being truncated, even when the label on the graph is truncated.
* Currently, when the screen is resized, some of the labels containing the job titles are dropped. This prevents the job titles from being dropped, and angles the job titles so they display nicely on any screen type. 

## Tooltip is no longer truncated:
screen shot 2016-12-27 at 4 27 33 pm

## Label is show for each bar, regardless of screen size:
screen shot 2016-12-27 at 4 11 40 pm
screen shot 2016-12-27 at 4 11 55 pm
screen shot 2016-12-27 at 4 12 06 pm


mwarnsleyDE:
Looks good to me as well.......

PR #2705, opened 2016-12-28
[PD-2772] Got rid of base-react-analytics.html

almathew:
Most of the custom stuff has been moved to `analytics_main.html`, with `base-react.html` being updated to handle some additional custom stuff.

The only difference is that `react-bootstrap-table.min.css` will be included on the analytics pages. I thought this would/did make a difference, but switching back and forth between branches I'm not seeing the difference.

I've messed around with analytics, reporting, and user management, and I haven't noticed any issues, but this might benefit from someone else who knows more doing the same.


koonsies:
New
image

Old
image

almathew:
Fixed. The order of the css files is important, unfortunately. 

Now:
screen shot 2016-12-28 at 11 32 23 am

PR #2707, opened 2016-12-29
Add redirect qc settings file [PD-2636]

tdphillips:
Basically a copy of the prod settings but using QC databases, etc.


koonsies:
@mklauber? I'm not familiar with deployment file layouts, so I wouldn't feel comfortable merging this myself.

PR #2706, opened 2016-12-28
[PD 2822] primary menu wont display

mwarnsleyDE:
Task was to edit the main menu display on mobile.

To Test:

Pull Down the repository. You can resize your screen and use the arrow button to show the main menu. Also did more css fixes for responsiveness which might have to get changed later but for now it all works and it's all responsive.

PR #2708, opened 2016-12-29
[PD 2815] adding legends to map

mwarnsleyDE:
Task was to add legends to the maps.

To Test:
Pull down the repository. You can view the maps and see that the legends have been added at the top of the maps. There is also data associated with the tooltips now inside of the tooltip display if you hover over the maps. The cities view has been changed to a bar chart since we currently don't have a cities type of map to display so I figured a bar chart would be better suited.


koonsies:
@fezick do we want to have the variable shading on our bar charts? Since the bar charts are designed to show the data visually via the sorting/size of the bars, it's a bit redundant to have both coloring and size indicating count.

koonsies:
The map's sizing (especially World) becomes pretty shoddy on smaller screens. Is that handling going to be part of the mobile ticket?

mwarnsleyDE:
That would probably have to be a separate ticket for that. The d3 and maps aren't responsive based off CSS so it would probably have to be a lot of Javascript and re-drawing the maps somehow for responsive.

fezick:
@koonsies, I agree on the shading of bar charts and line charts, although we should keep it for pie and area charts 9should we ever implement them.

koonsies:
Correct me if I'm wrong, but this PR depends on https://github.com/DirectEmployers/MyJobs/pull/2706, correct? I noticed that changes related to mobile are also present in this PR.

gulp/src/analytics/components/Charts/Bar/BarChart.js
```diff
@@ -9,9 +10,10 @@ class SimpleBarChart extends Component {
     const {chartData, width, height} = this.props;
     const barData = chartData.PageLoadData.rows;
     const xAxis = chartData.PageLoadData.column_names[0].key;
-
+    const ranges = ['rgb(103,0,13)', 'rgb(165,15,21)', 'rgb(203,24,29)', 'rgb(239,59,44)', 'rgb(251,106,74)', 'rgb(252,146,114)', 'rgb(252,187,161)', 'rgb(254,224,210)', 'rgb(255,245,240)', 'rgb(255,245,235)'];
```
koonsies: As per Jason's response, we want to remove color ranges for bar charts

mwarnsleyDE:
Merge conflicts should be good to go and the other requested changes are in my current PD I'm working on

PR #2712, opened 2017-01-04
Only do logout timer shenanigans if a user is logged in [PD-2841]

tdphillips:
What do we think? Any edge cases I'm not thinking of?

### DIFF
--- a/static/utils.184-4.js
+++ b/static/utils.184-4.js
@@ -34,7 +34,8 @@ utils.readCookie = function readCookie(cookie) {
  **/
 utils.logoutTimer = function readCookie(url) {
   if (!timer) {
-    if (utils.readCookie('sessionid')) {
+    if (loggedin) {
+      // We were logged in when this was called. Set up our logout timer.
       timer = window.setInterval(function redirect() {
         // if we are logged out and not already on the home page
         // If the loggedout cookie is set, a log out was initiated by the
--- a/static/utils.184-4.js
+++ b/static/utils.184-4.js
@@ -34,18 +34,20 @@ utils.readCookie = function readCookie(cookie) {
  **/
 utils.logoutTimer = function readCookie(url) {
   if (!timer) {
-    timer = window.setInterval(function redirect() {
-      // if we are logged out and not already on the home page
-      // If the loggedout cookie is set, a log out was initiated by the
-      // user via.
-      // If the exp cookie is missing, the user's session has expired.
-      if (utils.readCookie('loggedout') && window.location.pathname !== url) {
-        window.location.assign(url);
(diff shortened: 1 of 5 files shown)


tdphillips:
I didn't touch topbar_v2.<...>.js. If anyone thinks I should have, I will.

static/utils.184-4.js
```diff
@@ -34,18 +34,20 @@ utils.readCookie = function readCookie(cookie) {
...
-      // if we are logged out and not already on the home page
-      // If the loggedout cookie is set, a log out was initiated by the
-      // user via.
-      // If the exp cookie is missing, the user's session has expired.
-      if (utils.readCookie('loggedout') && window.location.pathname !== url) {
-        window.location.assign(url);
-      } else if ((!utils.readCookie('exp')) && (window.location.pathname !== url)) {
-        window.location.reload(true);
-      }
+    if (utils.readCookie('sessionid')) {
```
koonsies: I may be a bit off on this, but is the only change is having it check for the presence of a sessionid cookie before continuing? If that's the case, due to SESSION_SAVE_EVERY_REQUEST being set to True in default settings, I believe everyone will have a session, whether logged in or not.

almathew: @koonsies you're correct about sessionid being HTTP only, that's why I had to create the exp cookie (which caused this problem :-1:). 

The exp cookie is just a cookie that expires at the same time as sessionid, since sessionid is unavailable. So if all you're concerned about is the presences of the cookie, checking for one should be just like checking for the other. Which leaves me a little confused about something here - wouldn't sessionid and exp stop being accessible at the same time, causing the same exact issue we're already seeing? If I'm wrong about that, swapping sessionid for exp should work.

tdphillips:
Guess I get to go around creating random js vars in our five base templates.

tdphillips:
How's this?

koonsies:
Still not perfect. I left Analytics open to trigger the expiration logic, and when I checked the window again, it had begun to do that endless refresh loop. Strangely enough, it stopped doing the loop after I witnessed about 10-20 refreshes. Can't tell what made it resolve.

tdphillips:
Jason wants this hotfixed so rather than spending a few hours fixing the reload mechanic, I reverted that line. If anyone else wants to take it upon themselves to make this do "the right thing", I don't mind.

koonsies:
I agree with that. It would be good to get this issue fixed (the auto-log out for react apps) but we'll have to try to tackle it again without it getting this far into the pipeline.

huy-nguyen-colorado:
The equivalent code in utils.184-4.js for v2 is in: static > topbar_v2-183-9.js, lines ~1- 57

PR #2713, opened 2017-01-05
Add persistence to mongo

koonsies:
When I was working from home, I discovered (to my dismay) that the Mongo docker container was not set up to persist. 

This PR adds a new volume container, "mongodata", which the container "mongo" uses to store data from the local mongo instance.


koonsies:
Anyone have an idea what that test failure is?

PR #2718, opened 2017-01-06
[PD-2844] updated topbar_v2-184-6.js utils code

huy-nguyen-colorado:
**Purpose:** 
Update the "utils" javascript codes in topbar_v2-184-6.js because the codes changed in PD-2841 > utils.184-4.js.  v1 and v2 utilize the same code, except in v2, the "utils" codes lives in  topbar_v2-184-6.js.   There should be no visual changes for this PR.

To Test:
1. Ensure password reset page isn't stuck in endless refresh loop.


tdphillips:
184-6 appears to be a zero-byte file.

Your changes will also have no impact on the password reset page.

huy-nguyen-colorado:
Let's hope so, because this is a direct copy and paste kind of operation.

huy-nguyen-colorado:
I don't know what the hell happened there, I re-committed the v2 topbar js file.  Only changes are updating the "utils" portion of the code to match v1.

PR #2726, opened 2017-01-17
[PD 2854] enable paging

mwarnsleyDE:
Task was to add pagination to the table with all of the data inside

To Test:

Pull down the repository. The best way to test this with the most data would be switching the main dimension on the left to job titles and looking at the table. You will see the pagination for the table at the bottom.

Per Jason this morning. The breadcrumbs have been hidden for right now until the functionality is done for them.

### DIFF
--- a/gulp/src/analytics/components/Table/PageSize.js
+++ b/gulp/src/analytics/components/Table/PageSize.js
@@ -3,7 +3,7 @@ import {Component} from 'react';
 
 class PageSize extends Component {
   render() {
-    const {pageSize, handlePageSizeChange, showCount} = this.props;
+    const {handlePageSizeChange, showCount} = this.props;
     const options = showCount.map((show, i) => {
       return (
         <option key={i} value={show}>{show}</option>
@@ -12,7 +12,7 @@ class PageSize extends Component {
     return (
       <div>
         <label htmlFor="pageSize">Per Page: </label>
-        <select className="page-size" value={pageSize} onChange={handlePageSizeChange}>
+        <select className="page-size" onChange={handlePageSizeChange}>
           {options}
         </select>
       </div>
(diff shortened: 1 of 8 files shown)


gulp/src/analytics/components/Table/PageSize.js
```diff
@@ -0,0 +1,51 @@
...
+      <div className="page-size-container">
+        <span className="show-page-size">Show:</span>
+        <select className="page-size" onChange={this.switchToStartPage.bind(this)}>
+          {options}
+        </select>
+      </div>
+    );
+  }
+}
+// handlePageChange={this.handlePageChange.bind(this, startPage)} currentPage={currentPage}
```
koonsies: commented code

gulp/src/analytics/components/Table/TableContainer.js
```diff
@@ -6,13 +6,49 @@ import TableRows from './TableRows';
 
 
 class TableContainer extends Component {
+  constructor() {
+    super();
+
+    this.state = {
+      pageSize: 10,
+      currentPage: 1,
+    };
```
koonsies: I'm not sure about starting to store component level state, as it breaks our redux-ie approach. I'm not sure this is definitely a no-go, but I want to make sure it's done sparingly and for good reason. Since a pagination control is unlikely to need to be modified by other components, this is probably alright.

mwarnsleyDE: I think the main reason that I did that was based on using the page size state and passing it to a couple of the components that were nested inside of table container. I'm not sure of a different way to handle this as far as putting it inside of the Page Size Component. I was kinda trying a lot of things out and it worked this way so  I rolled with it LOL. I was having a few troubles trying to build this component

PR #2729, opened 2017-01-18
[PD-2831] Allow for relationships between sites.

almathew:
Relationships between sites are stored in `relationships.SiteRelationship`.

On save/delete, they get "flattened" into `relationships.DenormalizedSiteRelationship`. This info can then be added to Solr using `manage.py update_index --remove`.

The flattening is a rather aggressive approach that can be sped up considerably by being more sensible about what it's doing. We might consider being more sensible if we start to see performance issues, but right now it gets the job done a lot more safely than try to do it quickly or in a clever way. 

I'd like to still add rudimentary searching capability, but this can be reviewed before I do that.


relationships/__init__.py
```diff
@@ -0,0 +1 @@
+from relationships.signals import update_relationships
```
tdphillips: Stuff like this has caused issues in the past. Could we do something else instead? Possibly this:
http://stackoverflow.com/a/9811066

relationships/signals.py
```diff
@@ -0,0 +1,34 @@
...
+from relationships.utils import Graph
+
+
+@receiver(post_save, sender=SiteRelationship)
+@receiver(post_delete, sender=SiteRelationship)
+def update_relationships(sender, instance, *args, **kwargs):
+    site = instance.parent
+
+    is_parent_of = DenormalizedSiteRelationship.objects.filter(parent=site)
+    list(is_parent_of)
```
tdphillips: What is the purpose of this? Calling delete on the cursor prior to executing it should still work as intended.

almathew: Left over from when I was trying to use that information after the delete has already been run (I'm doing it on line 26 now). Cleaning it up.

relationships/__init__.py
almathew: Fixed, although I had to do some horrible importing to get it there. Although horrible importing is less likely to cause problems in the future, so I'll consider it a win.

relationships/signals.py
```diff
@@ -0,0 +1,37 @@
...
+    from relationships.utils import Graph
+    from relationships.models import DenormalizedSiteRelationship
+
+    site = instance.parent
+
+    is_parent_of = DenormalizedSiteRelationship.objects.filter(parent=site)
+    is_parent_of.delete()
+
+    has_as_child = DenormalizedSiteRelationship.objects.filter(child=site)
+    has_as_child = DenormalizedSiteRelationship.objects.filter(
```
koonsies: If this step will find all parents of a given node (all the way to the highest hierarchy) and delete all relationships of those parents, wouldn't that also delete the children of that given node, rendering line 19-20 obsolete? Since during flattening, all children of a node are children of that node's parent node as well.

Or am I confused? I have been looking at this bit of logic for a little longer than I should have haha.

almathew: Lets say I have the relationships a->b->c

That would give me denormalized relationships a->b, a->c, and b->c. 

If I want to remove b the steps would be:
Lines 19-20: Find and delete denormalized relationship b->c.

Line 22: Find denormalized relationships a->b (where b is child). 
Line 23: Delete anything where a is a parent at all (since we don't know whether or not that relationship is a product of the a->b relationship).

Line 29: Make a note that you deleted a and b.

Line 37: Re-run a and b to re-create new, more correct denormalized relationships. 

If we didn't do Lines 19-20, Line 22 would still only be getting a->b. I might be able to switch Line 22 to be `filter(child=site | parent=site)` but at that point the logic is relatively the same. So I don't think that there's duplicate work being done? I might be misunderstanding the question though.

koonsies: Ah, you're right. My question came from working to understanding the flow, but I get what is happening now.

koonsies:
I am trying to hand test this feature, and when doing so I received an error on adding a relationship through admin. 

To reproduce, simply create a relationship type, select two sites, and assign that relationship type. 

screen shot 2017-01-25 at 9 00 18 am

tdphillips:
That comes from `relationships/signals.py`, line 32. The query used by that queryset is as follows:

```
SELECT `relationships_denormalizedsiterelationship`.`id`, `relationships_denormalizedsiterelationship`.`weight`, `relationships_denormalizedsiterelationship`.`created_on`, `relationships_denormalizedsiterelationship`.`parent_id`, `relationships_denormalizedsiterelationship`.`child_id`, `relationships_denormalizedsiterelationship`.`by_id`, `relationships_denormalizedsiterelationship`.`depth` FROM `relationships_denormalizedsiterelationship` WHERE `relationships_denormalizedsiterelationship`.`parent_id` IN (SELECT U0.`parent_id` FROM `relationships_denormalizedsiterelationship` U0 WHERE U0.`child_id` = 3 )
```

You can't modify the table you're selecting from. If I change line 24 to cast the subquery as a list (thus no longer doing a subquery), I get a different error.

almathew:
@tdphillips I cast it to a list, and it seems to be working fine for me now. @koonsies you might pull down my latest changes and try again.

By now you think I would've learned my lesson about testing on sqlite.

PR #2730, opened 2017-01-18
[PD 2852] breadcrumb functionality

mwarnsleyDE:
Task was to add the functionality to the breadcrumbs. The breadcrumbs when clicked will go back to the tab that created the crumb. Also, if a tab is deleted and a breadcrumb clicked on that is related to a tab that's deleted, it will undo and bring back the original tab related to the breadcrumb.

To Test:

Pull Down the repository. You won't see breadcrumbs on the first tab so you will have to filter or add tabs from the table or chart. You will see the breadcrumbs appear depending on what has been clicked inside of the table or chart. If you click the tabs you will see that they go to the tabs that created the crumbs.


gulp/src/analytics/reducers/analytics-reducer.js
```diff
@@ -149,6 +152,45 @@ export default handleActions({
...
+      navigation: newNavigation,
+    };
+  },
+  'DELETE_STORED_DELETED_TAB': (state, action) => {
+    const deleteStored = action.payload;
+    return {
+      ...state,
+      deletedNavigation: state.deletedNavigation.filter((deleted) => {
+        if (deleted.navId === deleteStored.navId) {
+          return deleted.navId !== deleteStored.navId;
```
koonsies: This is confusing me a bit. Can this be simplified to...

```
  'DELETE_STORED_DELETED_TAB': (state, action) => {
    const deleteStored = action.payload;
    return {
      ...state,
      deletedNavigation: state.deletedNavigation.filter((deleted) => {
        return deleted.navId !== deleteStored.navId;
      }),
    };
  },
```

gulp/src/analytics/reducers/analytics-reducer.js
```diff
@@ -149,6 +152,45 @@ export default handleActions({
...
+        return {
+          ...nav,
+          active: false,
+        };
+      }),
+    };
+  },
+  'REPLACE_DELETED_TAB': (state, action) => {
+    const replacedTab = action.payload;
+    const newNavigation = state.navigation.slice(0, state.navigation.length);
```
koonsies: Not important, but slice(0) does the same thing as this. The second argument isn't necessary.

gulp/src/analytics/actions/tab-actions.js
```diff
@@ -1,6 +1,10 @@
 import {createAction} from 'redux-actions';
 
 export const switchActiveTab = createAction('SWITCH_ACTIVE_TAB');
+export const breadCrumbSwitchTab = createAction('BREADCRUMB_SWITCH_ACTIVE_TAB');
+export const restoreDeletedTab = createAction('REPLACE_DELETED_TAB');
```
koonsies: It may be more readable to keep the function name in line with the action name. "replace" and "restore" aren't really synonyms and that can hamper code readability.

mwarnsleyDE:
For the duplicate tabs I can put it in this PR. I will take a look at it now. Sometimes they fall in the right order I guess depending on which one you click on. It worked for me sometimes yesterday and sometimes not and I fixed some stuff and then it worked and I tested and it works but good thing I have you looking at it because I was trying to test all the scenarios I could think of for replacing them back in the right order and I knew I would miss something somewhere along the lines but I'll work on it for this PR

mwarnsleyDE:
Okay @koonsies when these test finish you can view all of these changes again that I made. It's working for me now but we know how that goes when I'm look at it and trying the scenarios lol. The order of the tabs coming back and being replaced should be good to go now along with the other changes you requested.

PR #2732, opened 2017-01-20
[PD 2889] Fixing the deleting and refiltering of tabs

mwarnsleyDE:
To test:

Pull down the repository. You can add tabs by filtering. Then you can delete a couple of tabs or one if you like. Once you go back to the second or first tab in line and re-filter, it should remove all the tabs before it and start filtering again. Prior, it would just duplicate the very last tab instead of deleting it.


koonsies:
If you click a country to filter, go back to country tab and click another country, you'll get a second "State" tab.

mwarnsleyDE:
@koonsies Ok, I found the issue, it should be fixed now.

mwarnsleyDE:
Yea, I have a couple of other things to fix on this one real quick

PR #2734, opened 2017-01-23
[PD-2867] Ensure parameters are passed along to the URL on the job view page.

almathew:
Ensure that parameters used by analytics are passed along to the job view page from the job listing pages. 

The modifications themselves are pretty simple, although it took some searching to confirm that the request was being passed along as context to everything that ended up using `joblist_url()`. In case anyone else is wondering how things end up getting to `joblist_url()`:
joblist_url
The templates highlighted in red are unused as far as I can tell. I made the changes there too, but they should be non-impacting. I'd be interested if someone knew otherwise, as I'd like to test manually as well. 

The unused templates are:
* search_results.html
* v2/search_results.html
* v2/job_results.html
* listing_search_item.html

### DIFF
--- a/seo/templatetags/job_setup.py
+++ b/seo/templatetags/job_setup.py
@@ -5,6 +5,7 @@ from django import template
 
 register = template.Library()
 
+
 @register.inclusion_tag('job_result.html', takes_context=True)
 def arrange_jobs(context):
     featured_jobs = context.get('featured_jobs')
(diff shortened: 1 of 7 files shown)


huy-nguyen-colorado:
I don't know where: v2/search_results.html nor v2/job_results.html are being used in the v2 template.  My own manual testing doesn't show any dependency upon them.  Marcus added these two files to the v2 templates directory last year, probably just to be on the safe side since they were in the v1.

PR #2735, opened 2017-01-23
Add campaign tags to partner saved search emails [PD-2873]

tdphillips:
Adds three tags used in DE analytics to partner saved search emails:
`de_n=PRM+Saved+Search`
`de_m=email`
`de_c={{partner_name}}`

### DIFF
--- a/mysearches/models.py
+++ b/mysearches/models.py
@@ -164,6 +164,12 @@ class SavedSearch(models.Model):
                                                                     extras)
                         self.url = mypartners.helpers.add_extra_params(self.url,
                                                                        extras)
+                    campaigns = ("de_n=PRM Saved Search"
+                                 "&de_m=email&de_c={partner}").format(
+                        partner=self.partnersavedsearch.partner.name)
+                    mypartners.helpers.add_extra_params_to_jobs(items, campaigns)
+                    self.url = mypartners.helpers.add_extra_params(self.url, campaigns)
+
                 if self.custom_message and not custom_msg:
                     custom_msg = self.custom_message
 
@@ -504,6 +510,12 @@ class SavedSearchDigest(models.Model):
                     mypartners.helpers.add_extra_params_to_jobs(items, extras)
                     search.url = mypartners.helpers.add_extra_params(search.url,
                                                                      extras)
+                campaigns = ("de_n=PRM Saved Search"
+                             "&de_m=email&de_c={partner}").format(
+                    partner=pss.partner.name)
+                mypartners.helpers.add_extra_params_to_jobs(items, campaigns)
+                search.url = mypartners.helpers.add_extra_params(search.url, campaigns)
+
             search_list.append((search, items, count))
 
         saved_searches = [(search, items, count)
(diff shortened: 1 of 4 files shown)


mysearches/templatetags/email_tags.py
```diff
@@ -1,8 +1,8 @@
 from django import template
 from django.core.urlresolvers import reverse
 
+from mypartners.helpers import add_extra_params
 from registration.models import ActivationProfile
-from universal.helpers import update_url_param
```
tdphillips: This could probably be replaced in its entirety at this point.

mysearches/tests/test_models.py
```diff
@@ -477,6 +479,22 @@ def test_send_partner_saved_search_as_saved_search(self):
...
+        self.partner_search = PartnerSavedSearch.objects.get(
+            pk=self.partner_search.pk)
+        self.partner_search.send_email()
+        email = mail.outbox[0]
+
+        for campaign_tag in ["de_n=PRM+Saved+Search",
+                             "de_m=email",
+                             u"de_c={partner}".format(
+                                partner=self.partner_search.partner.name)]:
+            query = QueryDict(campaign_tag).urlencode()
```
tdphillips: QueryDict automagically does quoting and such.

mysearches/models.py
```diff
@@ -580,6 +595,7 @@ class PartnerSavedSearch(SavedSearch):
     unsubscriber = models.EmailField(max_length=255, blank=True, editable=False,
                                      verbose_name='Unsubscriber')
     last_action_time = models.DateTimeField(default=datetime.now, blank=True)
+    _campaigns = None
```
koonsies: Save me some research.. I don't understand what is going on here. What gets assigned to _campaigns? Why is it named with the private attribute convention? Not saying this is wrong in any way, just trying to understand how this is all working.

tdphillips: Private would be two underscores. Ideally users will use the read-only `campaigns` property which will construct `_campaigns` if it doesn't exist then return it.

JLMcLaugh: A single underscore is a cue for suggesting devs treat a method or function as private. Functions like __init__ are special attributes, but not private. Neither of these actually mean anything to the interpreter.

Also, I just learned that double leading underscores have a special meaning to the interpreter for name mangling:
https://shahriar.svbtle.com/underscores-in-python

koonsies: Excellent. Thanks, guys.

So, the underscore here is to prevent (hopefully) from people doing object._campaigns, and instead use the property which essentially has a default value to return in the event that _campaigns is empty. So, other than by the read_only property, when is _campaigns altered?

tdphillips: Never. It's probably a pointless optimization. _campaigns isn't being built unless it's necessary and it only gets built once rather than once per call to a method that makes use of it.

koonsies: That's fine. I was just making sure I understood the big picture. Thanks for taking the time to explain!

PR #2743, opened 2017-01-25
Upgrade pymongoenv version.

almathew:


almathew:
It seems likely that the errors are actually real errors unless we're seeing them somewhere else. Hopefully we can talk about things and reach a better conclusion to this issue this morning anyway.

PR #2744, opened 2017-01-26
[PD 2824] more compact layout

mwarnsleyDE:
Task had these requirements per Jason

List of required modifications:
Eliminate margin around page_content. Report panel should be flush against the primary nav menu. Top should be flush against tabs. Report panels should span full width of window with no grey on either side.
Reduce space above tabs by 50%
reduce space above table to 10px
remove gap between the chart and the table
There may be more as we go through it. Right now, we just have a lot of distracting negative space.

To Test:

Pull down repository and check that the extra spacing and margins are gone from the table and charts area.


huy-nguyen-colorado:
Code changes looks good to me.  I'm not able to actually see the UI due to an error I get in analytics.js.  Somebody else should look at the UI changes in the actual browser before merging.

PR #2747, opened 2017-01-26
[PD-2864] Switch guid to reqid in analytics.

almathew:
If you just want to switch guid to reqid, the only thing actually required is the migration. That's pretty awesome.

Since reqid might have a false-y/none-y value (which wasn't the case for guid), the way we do searches and display information to the user had to be updated. 
The value we display to the user is stored in `MISSING_VALUE`. Currently I have is as `"None"`. This could cause some problems if someone did something like had a job title of "None".

The grouping portion of this solution isn't really compatible with server-side pagination. After talking with @darrint and @koonsies it sounds like the best solution to that specific issue is to solve that problem when we get there.

I'm a little concerned about some of my logic in `group_rows()` not being generalized enough or inadvertently dropping required information, so I'd appreciate someone who knows more than me about how analytics works paying a little bit more attention there.

### DIFF
--- a/analytics/api.py
+++ b/analytics/api.py
@@ -19,6 +19,11 @@ from myreports.models import (
     ReportTypeDataTypes)
 
 
+# Value displayed to the user when the actual value is not stored in the
+# database or does not exist.
+MISSING_VALUE = 'None'
+
+
 @requires("view analytics")
 def views_last_7_days(request):
     """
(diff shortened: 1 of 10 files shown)


almathew:
And I accidentally based my branch off of the changes I made a little bit ago for the settings files.... oops.

analytics/api.py
```diff
@@ -162,6 +167,39 @@ def format_return_dict(record, group_label):
...
+
+        if row_id in grouped_rows:
+            grouped_rows[row_id] += row[aggregation_key]
+        else:
+            grouped_rows[row_id] = row[aggregation_key]
+
+    grouped_rows = [{'_id': key, aggregation_key: value} for key, value in
+                    grouped_rows.iteritems()]
+
+    return sorted(grouped_rows, key=lambda x: x[aggregation_key], reverse=True)
```
koonsies: One potential issue I see with this is that we'll eventually want to add additional columns to the charts. Right now, it's just job_views, but we may eventually want additional information per row.

I'm not sure this is a PR-breaking caveat.. As with the pagination issue, tackling that may be an issue to cross when we actually encounter it.

koonsies: That said, all information on a given row outside of the ID column will be numeric, so simply summing up the values of all other keys on the row would be sufficient. Of course, that changes the shape of the return object.

almathew: I struggled with that issue quite a bit before just figuring we'd solve it when we get there. It might be as easy as using `original.update(duplicate)`, but if we want to do something like aggregation on another field or prefering one type of value over another it gets messy. I held off because of just how messy trying to predict what that might look like was getting.

PR #2748, opened 2017-01-27
Use activity display text or name in activity selection widgets [PD-2886]

tdphillips:
See if you can spot which activity was given its own value for display_text.

2017-01-27-090155_414x277_scrot


gulp/src/manageusers/components/Role.jsx
```diff
@@ -48,6 +48,9 @@ class Role extends React.Component {
       roleName: event.target.value,
     });
   }
+  displayText(activity) {
```
tdphillips: I need to use this exact function in components/Activities.jsx. Can a React guru tell me where such code should live? It's not shared everywhere, just in this app.

tdphillips: I need to use this exact function in components/Activities.jsx. Can a React guru tell me where such code should live? It's not shared everywhere, just in this app.

darrint: A `.js` file in `gulp/src/manageusers` is a good place.

PR #2751, opened 2017-01-30
[PD 2882] date range overhaul

mwarnsleyDE:
Task was to make the Date Global across all tabs and also carry over to switching over to a new dimension as well.

To Test:

Pull down the repository. You can change the date range and see that it changes for all tabs when switched. The tabs will also have to hit the API again once the date has been changed so you will see that as well.


mklauber:
@mwarnsleyDE Tests passed for this, jenkins failed because it couldn't post the success to IRC.  I'm rerunning tests so it shows up, but if you need this merged immediately, let me know and I can override the failure status.

mwarnsleyDE:
@mklauber I don't need it merged immediately. I was actually waiting for one of my other PR's to be merged so I could merge them together and go through some of the UI stuff to make sure it matched which I just did but it is ready to be checked whenever someone is available which usually Jonathan is the one to verify for these but there's no rush for it or anything

PR #2752, opened 2017-01-30
[PD 2855] metric sums

mwarnsleyDE:
Task was to add the metric sums at the bottom of the table

To Test:

Pull down the repository. You can see at the bottom that the Showing is displayed for how many are showed on a specific page along with the grand total. You can change pages and also change the amount per page and it will update accordingly.


koonsies:
It looks like the code you wrote works perfectly. That said, I double checked on what Jason expects from this ticket, and while the "Showing 1-10 of X" works great, the item he's looking for would aggregate values from the columns (or column, job_views, in our current case). He provided a link to an example from Direct Traffic.

pasted image at 2017_01_31 08_44 am

mwarnsleyDE:
@koonsies those changes should be implemented. The way I did it, I was guessing this was what Jason was referring to.

koonsies:
The totals columns look good. That said, even though the example has both "Total for all 38 rows" and "Total for Time Period", these rows are redundant on our system. Is there an instance where those values would be different? If not, I would remove the "Total for time period" row. Thoughts, @fezick ?

pasted image at 2017_01_31 08_43 am

fezick:
Yeah, just the two. Total for time period is indeed redundant. The screenshot was just as an illustration as to why I was asking for 2 values.

fezick:
See the updated screenshot from analytics, I would recommend these changes to the text:
1. Subtotal for rows 1-10 --> Subtotal (current page)
2. Total for all 38 rows --> Total (all pages)

Drop the total for time period. It is redundant, and we already tell the user how many rows in the paging control.

gulp/src/analytics/components/Table/PageSize.js
```diff
@@ -12,18 +12,28 @@ class PageSize extends Component {
     }
   }
   render() {
-    const {showCount} = this.props;
+    const {showCount, dataSize, pageSize, currentPage, pageLength} = this.props;
```
koonsies: I noticed dataLength is renamed to dataSize when passed in here. What is the distinction? If there is none, we should do our best to maintain naming of variables that represent the same info.

koonsies: That said, it may be worth figuring out if we can name the variables a bit closer to what they are. "pageLength" and "pageSize" mean the same thing to me, but in this instance, pageSize is the selected size, and pageLength is the actual rows. Perhaps "pageSize" and "actualLength"? Something to make it clearer, as I had to console.log the values to grasp what we were looking at.

dataSize -> totalRows
pageSize -> rowsPerPage
pageLength -> rowsThisPage

Something like that? I could be accused of being over-verbose, but I highly value readability.

gulp/src/analytics/components/Table/Table.js
```diff
@@ -4,8 +4,27 @@ import {Component} from 'react';
...
+    const paginationData = tableData.PageLoadData.rows;
+    let numPages = Math.floor(paginationData.length / pageSize);
+    if (paginationData.length % pageSize > 0) {
+      numPages++;
+    }
+    return numPages;
+  }
   render() {
-    const {tableData, handlePageSizeChange, pageSize, currentPage, pageChange, count, dataLength} = this.props;
+    const {tableData, handlePageSizeChange, pageSize, currentPage, pageChange, count, dataLength, pageDataLength, metricRowsTotal, metricTotal} = this.props;
```
koonsies: Same as above in terms of variable names. 

count -> paginationOptions
dataLength -> totalRows, 
pageDataLength -> rowsThisPage
metricRowsTotal -> metricTotalThisPage
metricTotal -> metricOverallTotal

By no means am I saying you have to follow this exact naming convention, but rather I'm giving examples of conveying meaning in variable naming.

gulp/src/analytics/components/Table/Table.js
```diff
@@ -14,8 +33,20 @@ class Table extends Component {
...
+            </tr>
+            <tr className="metric-current-rows-totals">
+              <td>Total for all {totalRows} rows</td>
+              <td className="metric-total">{metricOverallTotal + ' (' + numPages + ')'}</td>
+            </tr>
+          </tbody>
+        </table>
         <div className="pagination-container">
-          <Pagination dataLength={dataLength} showCount={count} handlePageSizeChange={handlePageSizeChange} tableData={tableData} pageSize={pageSize} currentPage={currentPage} pageChange={pageChange} />
+          <Pagination pageLength={rowsThisPage} totalRows={totalRows} paginationOptions={paginationOptions} handlePageSizeChange={handlePageSizeChange} tableData={tableData} pageSize={pageSize} currentPage={currentPage} pageChange={pageChange} />
```
koonsies: Thanks for making those naming changes! A little goes a long way with variable naming. 

Is it possible to have pageLength change to rowsThisPage  in Pagination as well? Above, there is still pageSize and pageLength, which are interchangeable in meaning, not to mention the afore-mentioned swapping variables names on the same data as it is passed through the chain.

mwarnsleyDE:
Suggested changes for naming have been made. Page Size and Rows This Page are not interchangeable. Page Size will always be the Page Size from the Selection Box and Rows this Page displays the Rows of data for this page so it could be 1 and Page Size would still be 10 for the selection box.

PR #2756, opened 2017-02-01
[PD-2871] Toggle whether or not deleted partners show up in the all partners list.

almathew:
~~I'm still fighting with a checkbox, so I'm not technically "done", but the only remaining work is going to be a small amount of CSS (and tests, but that's another story). There's a fair amount here to review and I'm expecting a pretty wide range of feedback, so I wanted to get this up before I was completely committed to what I've done.~~

Archived/deleted partners can now optionally show up in the partner listing. They have subtle coloration differences and an additional tag saying they were deleted.

Existing partner:
screen shot 2017-02-01 at 8 43 48 am

Versus archived/deleted partner:
screen shot 2017-02-01 at 8 42 52 am

When viewing details links, I went ahead and hid all the edit/create links I could find for archived partners. I also added the deleted tag thing to the header to make it clear that the current partner was deleted. 
screen shot 2017-02-01 at 8 44 17 am

Styling isn't my thing, and I've only talked to @mwarnsleyDE about a few things, so I'm sure there are some real improvements that could be made there if anyone has any feedback.


mypartners/forms.py
```diff
@@ -279,16 +279,21 @@ class PartnerForm(NormalizedModelForm):
     """
     def __init__(self, *args, **kwargs):
         super(PartnerForm, self).__init__(*args, **kwargs)
-        contacts = Contact.objects.filter(partner=kwargs['instance'],
-                                          archived_on__isnull=True)
+        partner = kwargs['instance']
+        contacts = partner.get_contacts()
```
almathew: This is the only form that I think there's any reason at all to use an archived object (since it's how details about the partner are viewed). I didn't go as far as making the form completely read-only yet, but I'm leaning toward doing that.

universal/models.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    def __init__(self, field_with_rel, related_manager):
+        self.related_manager = related_manager
+        super(ManagedReverseDescriptor, self).__init__(field_with_rel)
+
+    def get_queryset(self, **db_hints):
+        db = router.db_for_read(self.field.rel.to, **db_hints)
+        manager = getattr(self.field.rel.to, self.related_manager)
+        return manager.using(db)
```
almathew: The real magic for both `ManagedForeignKey` and `ManagedReverseDescriptor` happens in `get_queryset`. The rest is just filler to set up `get_queryset` so it's usable.

almathew:
This should be good to go now.

PR #2760, opened 2017-02-02
[PD 2922] export ui

mwarnsleyDE:
Task was to create the UI for the export options. Currently we only have the excel and current tab in to make sure they work before going forward with everything else.

To Test:

Pull down the repository. At the top of the header next to the Calendar you should see the export button. When you click on it you will see a panel dropdown showing the export options. THIS IS JUST UI so exporting won't currently work but you can check what options you want as well as close out, etc.


gulp/src/analytics/components/Header/ExportOptions.js
```diff
@@ -0,0 +1,90 @@
...
+  pageClick() {
+    if (this.mouseIsDownOnCalendar) {
+      return;
+    }
+    this.setState({
+      showExportMenu: false,
+    });
+  }
+  mouseDownHandler() {
+    this.mouseIsDownOnCalendar = true;
```
koonsies: Why is this mentioning a calendar?

gulp/src/analytics/components/Header/ExportOptions.js
```diff
@@ -0,0 +1,90 @@
...
+          </li>
+        </ul>
+      </div>
+    );
+  }
+}
+
+ExportOptions.propTypes = {
+  dispatch: React.PropTypes.func,
+  analytics: React.PropTypes.object,
```
koonsies: Analytics isn't used here.

gulp/src/analytics/components/Header/Header.js
```diff
@@ -42,6 +45,16 @@ class Header extends Component {
...
+      showExportMenu: true,
+    });
+  }
+  hideExportMenu() {
+    this.setState({
+      showExportMenu: false,
+    });
+  }
   render() {
     const {analytics, toggleTabs, tabsActive} = this.props;
```
koonsies: I know this is used elsewhere, but "analytics" doesn't explain itself well in terms of what the variable is/contains.

If you need the entire state object, or in this case, the entire pageLoadData, it would be helpful to include an indicator that it is the state object.. Perhaps "analyticsState"

That said, in this case and in many others, it's not necessary to pull the entire state object. In this, we are only using analytics.stateCustomRange, so the prop handling would be cleaner as 

```
export default connect(state => ({
  stateCustomRange: state.pageLoadData.stateCustomRange
}))(Header);
```

mwarnsleyDE:
@koonsies I'm not stuck waiting on you for this one right now. I haven't gotten to the UI changes yet, I just managed to do the naming conventions. I'm working on the UI fixes right now with the checks and stuff

mwarnsleyDE:
NOT FINISHED YET. STILL WORKING ON UI

koonsies:
Since this PR isn't completed/ready for merge, we can probably close it and reopen when it is ready to be reviewed.

mwarnsleyDE:
@koonsies I fixed the UI for this PR so this one is good to go compared to before. I just have to use some of the stuff from it for the functionality stuff I'm working on in the export functionality PR.

mwarnsleyDE:
Also for this PR, the validation and some of the other stuff is in the separate PR for the functionality. This one was strictly just for the look of the export box.

PR #2763, opened 2017-02-02
[PD-2924] Properly re-set wildcard redirect at the beginning of every request.

almathew:


almathew:
I accidentally pulled in Troy's changes as well, but since they're not really going to be impacting production I don't care.

PR #2770, opened 2017-02-07
Merge django 1.9

mklauber:
This is the django-1.9 code that went into production. We should get it onto QC before friday (before we cut the next staging), but I want to give everyone a chance to look at things.

Of particular note:

The process for creating migrations will change.
All the old migration data has been reset. The initial migrations in this code base should match the state of the database in production.


apache/django.wsgi
```diff
@@ -13,6 +13,6 @@ if PROJECT_DIR_PARENT not in sys.path:
 os.environ['CELERY_LOADER'] = 'django'
 os.environ['DJANGO_SETTINGS_MODULE'] = 'settings'
 
-import django.core.handlers.wsgi
-application = django.core.handlers.wsgi.WSGIHandler()
+from django.core.wsgi import get_wsgi_application
```
tdphillips: Might want to update wsgi.py as well, unless we're saving Redirect for later.

PR #2771, opened 2017-02-07
[PD-2938] Account for args and kwargs that Django wants to send with the 404 page.

almathew:
Specifically `exception` is the arg/kwarg we're missing that Django is upset about.


almathew:
I think this might not be passing since it's branched against the django-1.9 branch, which doesn't pass tests on Jenkins yet based on what I last heard from @mklauber

mklauber:
I'd recommend you put this against quality-control rather than django-1.9.

almathew:
It's a 1.9-specific change.

PR #2774, opened 2017-02-08
Track campaigns on redirect views [PD-2866]

tdphillips:
There's no UI you can use to test this.

Have redirect running and my.jobs in your hosts file.
Open up a management shell, `from redirect.tests.factories import RedirectFactory; RedirectFactory()`
Grab the guid for the redirect that you just created and strip out any non-alphanumeric characters.
Navigate to my.jobs/\<guid\>+
Examine your headers and ensure that this info shows up in X-JSON-Headers.


almathew:
Unrelated to the actual code change - + isn't working for me (I just get a 301) so I had to do a head request. Is this a Docker thing or am I just doing it wrong?

tdphillips:
That's what I get for not previewing. GitHub saw that I wrote \<guid\> and helpfully treated it as a random HTML element.

PR #2775, opened 2017-02-08
[PD-2924] Apply permissions to PRM UI

almathew:
This mostly only accounts for permissions that are applied on the view. There are separate tickets to review/fix the places where the view is wrong or needs expanded. 

There are a few places in the templates where I duplicated some HTML in order to make the conditional logic less complex. This is because template conditional evaluation in Django doesn't support () and as a result I've been burned in the past by trying to get it to play nicely with exactly what I wanted.

~~Still writing tests, but I'm interested in feedback/if I missed something while I do.~~


almathew:
Tests are wrapped up. I still have to go in and remove tags from forms for users that don't have `read tag` permissions, but I think the rest of the templates are fine.

mypartners/tests/test_templates.py
```diff
@@ -0,0 +1,1400 @@
...
+        Renders the class's template to a string.
+
+        :param context: Any context required for rendering the template.
+        :return: The rendered template as a string.
+
+        """
+        return render_to_string(self.template, context)
+
+    @staticmethod
+    def remove_activities(activities):
```
tdphillips: This and `add_activities` work fine but it seems like they add needless complexity. We're only ever removing one activity. Whenever we add multiple activities, we do so one at a time.

templates/mypartners/includes/contact_card.html
```diff
@@ -19,12 +23,14 @@
...
     {% endif %}
+    {% if can_read_tag %}
     <div class="partner-tags">
         {% for tag in contact.tags.all %}
         <span class="tag-name small" style="background-color: #{{tag.hex_color}}">{{ tag.name }}</span>
         {% endfor %}
     </div>
-    {% if not partner.archived_on %}
+    {% endif %}
+    {% if not partner.archived_on and can_create_partner %}
```
tdphillips: The following link is for editing contacts. Was this intended to check if you can create partners?

templates/mypartners/includes/partner_sidebar.html
```diff
@@ -43,13 +49,10 @@ <h2 class="top">Filter Partners</h2>
...
         {% endif %}
-        <a class="btn" href="{% url 'create_partner' %}?ct={{ partner_ct }}">Create New Partner</a>
-        <a class="btn" href="{% url 'partner_library' %}">Search OFCCP Partner Library</a>
-        <a class="btn" href="{% url 'partner_tagging' %}">Tag Management</a>
-        {% can user company 'read outreach email address' as has_nuo %}
-        {% if has_nuo %}
-        <a class="btn" href="{% url 'nuo_main' %}">Non-User Outreach</a>
-        {% endif %}
+        {% if can_create_partner %}<a class="btn" href="{% url 'create_partner' %}?ct={{ partner_ct }}">Create New Partner</a>{% endif %}
+        {% if can_read_partner %}<a class="btn" href="{% url 'partner_library' %}">Search OFCCP Partner Library</a>{% endif %}
```
tdphillips: Should searching the partner library require "read partner" permissions? You're not reading your own partners.

templates/mypartners/main_records.html
```diff
@@ -46,11 +50,13 @@ <h2 class="top">Filters</h2>
...
                 <label>Tags:</label>
                 <input id="record_tag" type="text" placeholder="Tags" {% if request.GET.tag %}value="{{request.GET.tag}}"{% endif %} />
+                {% endif %}
                 <label>Date Range</label>
                 {% include "mypartners/includes/date_widget.html" %}
-                {% if not partner.archived_on %}
+                {% if not partner.archived_on and can_create_communication_record and can_update_communication_record %}
                     <div class="navigation">
                         <h2>Navigation</h2>
                         <a class="btn" href="{% url 'partner_edit_record' %}?partner={{ partner.id }}">New Record</a>
```
tdphillips: Does this really require update permissions?

templates/mypartners/overview.html
```diff
@@ -16,17 +17,23 @@
...
     <div class="row desktop_hide" id="prm-mobile-row-buttons">
         <div class="span12">
-            <a class="btn" href="{% url 'partner_records' %}?partner={{ partner.id }}">Communication Records</a>
-            <a class="btn" href="{% url 'partner_searches' %}?partner={{ partner.id }}">Saved Searches</a>
-            <a class="btn" href="{% url 'partner_details' %}?partner={{ partner.id }}">{% if not partner.archived_on %}Edit{% else %}View{% endif %} Partner Details</a>
-            <a class="btn" href="{% url 'report_view' %}?partner={{ partner.id }}">Reports</a>
+            {% if can_read_communication_record %}<a class="btn" href="{% url 'partner_records' %}?partner={{ partner.id }}">Communication Records</a>{% endif %}
+            {% if can_read_partner_saved_search %}<a class="btn" href="{% url 'partner_searches' %}?partner={{ partner.id }}">Saved Searches</a>{% endif %}
+            <a class="btn" href="{% url 'partner_details' %}?partner={{ partner.id }}">{% if not partner.archived_on and can_update_partner %}Edit{% else %}View{% endif %} Partner Details</a>
```
tdphillips: What if the user lacks "read partner"?

almathew:
@tdphillips I've added PD-2952 which addresses the question about why certain permissions were used.

templates/mypartners/overview.html
almathew: That's a good point. I left it off because they can't access this page without `read_partner`, but in most places I tried not to assume what permissions were already on the view (in case we decided to use the template in a new way). 

This will be a good thing to update on a second pass (which is definitely coming given PD-2949)

PR #2780, opened 2017-02-10
Bad import

almathew:


mklauber:
Think we can do this for all the `git+git...` requirements as part of this PR?

almathew:
@mklauber I am fine with this and so it is done.

PR #2782, opened 2017-02-13
Fix 404 page

mklauber:


mklauber:
Opening this against QC since we've merged django-1.9 now.

PR #2783, opened 2017-02-14
[PD-2924] Users without read tag permission should not see tags on forms.

almathew:
Users without create tag permissions can't create tags already, so that should cover all the tag-related permissions requirements.


almathew:
Ok, the process now cleans up after itself. Also, I switched to using `local()` over `current_thread()` since that's what everyone else/Django uses.

For fun, some examples of other people doing what I've done with middleware:
https://djangosnippets.org/snippets/2179/
https://github.com/jedie/django-tools/blob/master/django_tools/middlewares/ThreadLocal.py
https://github.com/nebstrebor/django-threadlocals/blob/master/threadlocals/threadlocals.py

And some examples where Django is using the thread similarly:
https://github.com/django/django/blob/d2e7d15b4c594f64ee9d37bf40e61920cea41487/django/urls/base.py#L15
https://github.com/django/django/blob/cecc079168e8669138728d31611ff3a1e7eb3a9f/django/core/cache/__init__.py#L65
https://github.com/django/django/blob/9718fa2e8abe430c3526a9278dd976443d4ae3c6/django/utils/translation/trans_real.py#L24
https://github.com/django/django/blob/a87d6b69a75d0f9a66652e3f0bd1c2c85624c7d5/django/db/utils.py#L143

mypartners/forms.py
```diff
@@ -61,6 +65,12 @@ def __init__(self, *args, **kwargs):
             self.fields['notes'] = notes
 
         init_tags(self)
+        # For some reason requests time out if init_tags() isn't done first.
+        # So rather than skipping init_tags if a user can't read tags,
+        # instead pop the tag field off after the fact.
```
almathew: If anyone has any idea why this might be I'd love to fix it. Even using runserver it times out rather than producing an actual error. `pdb` didn't pick out any loops that looked meaningful or made sense to me after a few minutes of looking, and this worked, so I just left it and moved on. It was bothering my on my drive in this morning though.

PR #2789, opened 2017-02-15
Introducing MemSQL

darrint:
## Introducing MemSQL

This is an additional database. In this project we will have read only access through the given models.

## Review Questions

**Is merging this going to have any impact on deployment?** Based on my testing migrations should be unaffected until Jenkins has commands added to explicitly migrate the analytics database. I could be wrong though.

This works and it allows us to manage the schema of the analytics database like any other SQL database. **Are we sure this is what we want?**

Django is pretty weak at expressing analytical queries. I can imagine us choosing to introduce an SqlAlchemy bridge in the future or writing a significant amount of raw SQL. **Are we okay with this?**

Migration `0002` is really complex. A useful test is to run `dkm makemigrations analytics` to make sure I didn't miss anything.

I will have an upcoming PR for analytics that uses this code, and introducing MyJobs as a dependency. **If you want to see that before merging this, let me know.**

~~I haven't done any work yet to make memsql available via Docker. In the near future I'll need to add it to the background containers. **Does that need added to this PR or can it wait until the future.**~~ I added the docker support. You can test migrations with it now.

I introduced `setup.py` so analytics could import and use Django models. I manually specified just the package I watned in `setup.py`, not the whole project with `find_packages()`. **Is that what we want?**

## Code

* Migration `0002` is really complex. Does it look sane?
* Note that the router interface for migrations has changed. Someday we might need to update the router for `RedirectArchive`.

## Settings Needed

You need access to a memsql database. Add the following to `DATABASES` in `dev_settings.py`:

```
    'analytics': {
        'NAME': 'analytics',
        'ENGINE': 'django.db.backends.mysql',
        'USER': 'root',
        'HOST': 'memsql.myjobs.',
    },
```

Also run

```
$ dk initmemsqldata
$ dk backgroundstop
$ dk background
```

Now you should be able to run migrations.

```
$ dkm migrate --database analytics --list analytics # the closest thing to --dry-run available
$ dkm migrate --database analytics analytics
$ dkm makemigrations analytics # It should have nothing to do, unless I messed up.
```


almathew:
Extra info for anyone else who is oblivious like me: you will need to get onto the memsql docker instance and add the analytics database. Syntax for adding a database is the same as mysql (so `create database analytics`). Command to get to the memsql cli is `memsql`.

darrint:
The mysql cli works too.

almathew:
As someone who was involved in some of the decision making process, I feel like my opinion is unnecessary, but...

> This works and it allows us to manage the schema of the analytics database like any other SQL database. Are we sure this is what we want?

100% yes. If we're not doing schemaless, we should be managing our schema. If it's an available option, we should definitely be managing it the same way we've always managed it. 

> Django is pretty weak at expressing analytical queries. I can imagine us choosing to introduce an SqlAlchemy bridge in the future or writing a significant amount of raw SQL. Are we okay with this?

It's preferred over what we're doing with MongoDB right now.

> I manually specified just the package I watned in setup.py, not the whole project with find_packages(). Is that what we want?

Indifferent


At this point I can understand and run migrations, and do basic CRUD operations on the new table. Looks good to me :+1:

tdphillips:
Just a reminder, don't forget about dbshell.

Once your database has been created,
`dkm dbshell --database=analytics`

PR #2791, opened 2017-02-15
[PD-2959] add search functionality

mwarnsleyDE:
## What's this PR do?
Allows for searching the current analytics tab.

This should work on any of the analytics charts.

Go to any (or all!) of the analytics charts. Search for something that's on the current page. Search for something that's on another page. Search for something that isn't present at all.

All of them should provide exactly what you would expect from search results.

Search terms should only be applied to the current column displayed to the user (so if the user is looking at a list of job titles, search should only be done on the job titles).

Search results should update the history.

After reviewing how search was implemented, I decided to hold off on implementing search on the other datasources for now. I don't think the way search is implemented is necessarily applicable to other datasources, so implementing it now will just introduce a useless maintenance nightmare.


almathew:
I'm taking this over, and will re-open when I'm ready for review.

almathew:
This is ready for review.

analytics/management/commands/make_data.py
```diff
@@ -43,8 +43,8 @@ def handle(self, *args, **options):
                       for x in range(0, days_to_generate)]
 
         job_views = []
-        num_jobs_to_create = random.randint(num_in_use_buids * 200,
-                                            num_in_use_buids * 2000)
```
darrint: This got reverted. It should not.

almathew: I added two zeros the other pull request. This bumps it down to 1. Turns out two zeros is a lot.

PR #2792, opened 2017-02-15
Add default reporting objects

tdphillips:


myreports/migrations/0002_auto_20161021_0752.py
```diff
@@ -43,4 +647,5 @@ class Migration(migrations.Migration):
             name='configuration',
             field=models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='myreports.Configuration'),
         ),
+        migrations.RunPython(set_up_reporting, lambda apps, schema_editor: None),
```
tdphillips: Do we need the ability to reverse this migration?

I included this one for testing purposes but I can leave it if desired. We should never back out of this migration unless we're also going to back out of 0001 (which removes the relevant tables entirely) so it should be fine.

mklauber: I'd want @almathew's opinion, but I don't see any reason to be able to reverse that part of this migration, at least.

almathew: I'm fine with not being able to reverse it. 
This is already a dev-only thing already, and I can't foresee a dev not being able to just delete them all manually in the extremely unlikely event that the dev wants them gone.

tdphillips: Works for me.

PR #2793, opened 2017-02-15
[PD-2942] Changed bar chart colors so it's easier to differentiate between bars

almathew:
screen shot 2017-02-15 at 2 25 00 pm


gulp/src/analytics/components/Charts/Bar/BarChart.js
```diff
@@ -11,6 +12,54 @@ class SimpleBarChart extends Component {
...
+    };
+    let colorIndex = 0;
+    let prevPrimaryColor = '';
+
+    /**
+    * Determines the color/shade of the color that a Bar on the Bar Chart
+    * should be, based on the number of views.
+    *
+    * Assigns colors so that overall count is sectioned off (i.e. The top 25%
+    * of groupings by job view are red, the next 25% are purple, etc.) but
```
tdphillips: Is this accurate? The bars with the lowest counts (in that screenshot, at least) are red and those with the most are purple.

almathew: I guess e.g. would've been more appropriate, not i.e.

It was correct when I wrote it, but either way d3 is doing the color selection so really it will depend on the job view distribution. For one person the comment might be correct, for another it might not. The spirit of the comment was more to describe what grouping by color meant more than to describe exactly what was going to happen.

PR #2796, opened 2017-02-16
[PD-2892] initial commit of cbo ui for knowledge transfer to Congress…

ghost:
…man Carlos Danger

Will video chat with Representative Danger for knowledge transfer.


koonsies:
I will be taking over this project. I will leave this up for visibility, but will not attempt to get it merged until I get a better handle on the code.

gulp/package.json
```diff
@@ -66,8 +67,12 @@
...
     "sortablejs": "^1.4.2",
     "tinycolor2": "~1.3.0",
+    "toastr": "2.1.2",
     "warning": "^2.1.0",
     "webpack": "^1.12.9",
     "webpack-dev-server": "^1.14.1"
-  }
+  },
+   "devDependencies": {
+     "redux-immutable-state-invariant": "1.2.3"
```
darrint: This needs to move to dependencies.

gulp/src/cbo/store/configureStore.dev.js
```diff
@@ -0,0 +1,12 @@
+import {createStore, applyMiddleware} from 'redux';
+import rootReducer from '../reducers';
+import reduxImmutableStateInvariant from 'redux-immutable-state-invariant';
+import thunk from 'redux-thunk';
+
+export default function configureStore(initialState) {
+  return createStore(
+    rootReducer,
+    initialState,
+    applyMiddleware(thunk, reduxImmutableStateInvariant())
```
darrint: This needs to be moved to the createStore function and only engaged in development, not production.

darrint:
The npm-shrinkwrap file appears to have been modified without running npm clean-shrinkwrap. The diff shouldn't be that big.

gulp/package.json
koonsies: I haven't researched this myself, but does is devDependencies only for packages required when the node settings are set to "development"?

koonsies: I know I've done --save and --save-dev (or something like that before). When would we use --save-dev?

gulp/src/cbo/store/configureStore.dev.js
koonsies: They way he has it set up, configureStore is..

```
if (process.env.NODE_ENV === 'production') {
  module.exports = require('./configureStore.prod');
} else {
  module.exports = require('./configureStore.dev');
}
```

Is that not sufficient?

koonsies:
I'm not seeing any references to the command npm clean-shrinkwrap.. Is that the right command? If so, what does it do?

gulp/package.json
darrint: For now never. It's too hard to keep the shrinkwrap file straight with `devDependencies`.

gulp/src/cbo/store/configureStore.dev.js
darrint: Where is that?

darrint:
https://github.com/DirectEmployers/MyJobs/blob/quality-control/gulp/package.json#L7

You just run it.

koonsies:
Closing pending further review

PR #2800, opened 2017-02-17
Add a network timeout to SEOXML imports.

mklauber:
By default a call to `urllib.urlretrieve` does not have a timeout associated with it.  The call itself does not accept a timeout parameter, so we set the global timeout.  To prevent the global timeout from affecting anything other hung SEOXML requests, we set it to an unusually large timeout 30 minutes.


mklauber:
The urllib.urlretrieve request in question is on line 488 in the new code.

PR #2803, opened 2017-02-20
Deployment Settings For MemSQL.

darrint:
From our experience with MongoDB Enterprise we learned that it is important that our QA server config closely match the production one in case additional security features, etc. change the behavior or code. That should be less of a problem with MemSQL than with MongoDB. But stil...

To test with SSL I asked @mklauber to set up a CA key and included below is the _public_ cert for the CA key. MemSQL is doing the same thing. Since these _should_ rarely change it is probably best to just distribute them with the projects that need them as that keeps the code and settings simple. Feel free to comment if you disagree.

I've added the needed accounts to `secrets.py` in Jenkins; so in theory this should all just work. :crossed_fingers:


deploy/settings.myjobs_qc.py
```diff
@@ -3,7 +3,8 @@
 import datetime
 import os
 
-from secrets import REDIRECT_QC, ARCHIVE_STAGING, REDIRECT_STAGING
+from secrets import (
+    REDIRECT_QC, ARCHIVE_STAGING, REDIRECT_STAGING, ANALYTICS_STAGING)
```
tdphillips: What issues will arise if qc and staging point to the same database?

almathew:
@darrint didn't you say something about a 1 month key yesterday?

@mklauber this looks good to me but I'll let you have the final say.

PR #2805, opened 2017-02-22
Maybe make Webpack great again.

darrint:
Can someone on OSX test this PR? If it drastically reduces your cpu usage go ahead and merge it.

We tested this change waaaay back when we switched from Docker Toolbox to Docker for Mac. There was some reason why we couldn't do at the time. Whatever that reason was doesn't appear to be a problem anymore. (It might have been before we updated npm.)


koonsies:
This has made a big difference in how much CPU is used by docker on my computer. @mwarnsleyDE, could you check this out and see if you notice any problems?

mwarnsleyDE:
@darrint I will check in on my system here in a second

PR #2806, opened 2017-02-22
[PD 2960] gui for report intention

mwarnsleyDE:
Task was to build a UI for reporting intentions so that the user has a good starting point of where they want to go.

To Test:

Pull down the repository and go to reports in the topbar. You should see the starting point includes PRM Reports, Job Analytics, and Site Analytics. Clicking on them will take you to the correct page. The switching to dynamic reporting facet on the right should be gone as well. However, site analytics currently doesn't go anywhere and Jason Sole is aware of that so clicking on it currently won't do anything. PRM and Job Analytics work however.


koonsies:
To clarify, Jason is fine with the button being visible, but non functional? In the past he's wanted non-functional controls to be hidden.

gulp/src/reporting/components/ReportingIntention.jsx
```diff
@@ -0,0 +1,62 @@
...
...w+fslEfV6h1LPAOgLk8FLTZwMxc6dZ8/OEsAmAKelAnZ1cvlQ85Q0Bg+/l5fW1ERFgv3OOJsAoQbU7XzmYovmZPUlzniHqQwAmdDB/j/l4GZjF28qSlcIPbTVwEezlk5REkt71RmCauI39XqkQDtTZL+5aaSr4CkKUqWKoiMRZhTv2aZpA651hxBNMtAoIgRlFleaD/4hxa2oicm+MlUglULNN56wiCWCCyVVVQQsQt6UuoiuBuwEkaHbaKA/wOCTUVActwKjmFRG7P9zpvKKaEHr5t20qVHgKSIDHSpaUBj8E/Hbd8ogF33epJwMnf3oSRF0AnMht4slwmI0zxSE+HtbwmbpVPJgYauGvdbVdy1OVbwrrVSPjeYNalh14U0is7YfsawhKgITa6ogE/0JehqmPVnxjS4ZyyRdMoyoAdvqfTX4pVA2eCpJ9dPK2pL7gGI+XinLHpy2uBAJ19qt15oeBES0rIS2Kn67bJ5aHd23aOcFte4ro8xj9VrAaDLT6qVVX+fiWhanfGE08XVM1ZNThYzbXGQZ/SFsDpCysjMyGlCNVE+QdrocTvnycGEokUL8FoioAvsJuhniyp4ONlZ4Slp01PMPjNG3xKyO+aeTjLN6wwzHNO6GpRjchNGr7CkgpqYzY3KTVlesX3BbVbH++pxg67a93/Ry/d6z+tkHU5wCFzKMkc86wOHOIYG7jSUqwZZ2Cx8UOsgldbfGT8g6dJE8ZdyjfC8bNJNnpY91JYVmlkhaO37AcXhJvhfFizT+5ukqZK0RycoobQBtceXDMORuWuuWfuXOI+U+LEOWgM0f5gAWtAggSgCABCBKAIAEIEoAEIAFIABKABIiIAIkZdpia7SEsBMYkYgIQMQMJQAKQACQACUAC6ArQfuIMxCKF5TWXrdLReC+x2o8kAAlAApAAJAAJEGvs2X/ks43bd39kBfBeSAAiZgX4gTqi1/IDCnCTOqLXcrPP4Y7T9dQRvROMfZ+2tss/Ff+4QR3S67iBsVfSgUePHv1Z+/EzTeI/bwt+pM7psfyoxFjEGmMu3QvYcu3aT86fP/8k0fPA2Aa9GUT0bKgTSADqBBKAIAEIEoAgAQgSgCABCBKAIAEIEoAgAQgSgCABCBKAIAEIEoAgAQgSgOgx/A/bEeD5EQKTCAAAAABJRU5ErkJggg=='} alt="PRM Reports"/>
```
darrint: Can we put these images in static?

gulp/src/reporting/components/ReportingIntention.jsx
```diff
@@ -0,0 +1,62 @@
...
+import {Link} from 'react-router';
+
+class ReportingIntention extends Component {
+  render() {
+    return (
+      <div id="reporting_intention_container">
+       <div className="row reporting-intention-row">
+          <div className="col-xs-12 col-md-12">
+            <ul id="report_site_container">
+              <li className="report-site">
```
darrint: These need to be dynamic based on the same variables used in the original dropdown on the Reports page.

gulp/src/reporting/components/DataTypeSelectBar.jsx
```diff
@@ -6,9 +6,6 @@ import {connect} from 'react-redux';
 class DataTypeSelectBar extends Component {
   render() {
     const {
-      intentionChoices,
```
darrint: If these aren't needed anymore take them out of `propTypes`.

gulp/src/reporting/components/ReportingIntention.jsx
mwarnsleyDE: @darrint I was going to try and put them inside of the static file. I just don't know how to read a static file from django into a React Component. I was trying some url structure for the img src but couldn't figure it out for some reason

mwarnsleyDE:
@koonsies Jason Sole said that only specific people would have access to Analytics anyways so he was fine just leaving it there for right now.

gulp/src/reporting/components/DataTypeSelectBar.jsx
mwarnsleyDE: @darrint These have been removed

gulp/src/reporting/components/ReportingIntention.jsx
mwarnsleyDE: @darrint Do we have that setup on the backend as far as the names being the same as in the Dropdown list for intentions? Was that something that Troy was working on? Because I can use the same code as the dropdowns if that's the case depending on how that data is coming back.

mwarnsleyDE: @darrint I'm guessing if the intentions are coming back dynamically from the server that I would need some kind of json with the name, the description, and the icon image unless I'm thinking of it in the wrong way

darrint: The code you had to remove from reporting is already hooked up to the dynamic intentions. You can follow it from there.

gulp/src/reporting/components/ReportingIntention.jsx
darrint: Search the code for function `expandStaticUrl`. This will let you reference files in `static/`.

mwarnsleyDE:
@koonsies and @darrint after the meeting this morning we ARE going to hide Site Analytics for now. I've setup some code to do it dynamically so I would just need something from the API that we are currently hitting. I've created a task for it PD-2988 as far as a way for the data to come back but I'm sure you guys have a better grasp of handling that than I do but the task is there. Test are failing because of ESLint errors which I suspected since I have some code in that I'm going to use once the API is changed.

PR #2809, opened 2017-02-23
[PD 2977] remove access

mwarnsleyDE:
## What's this PR do?
This PR removes all references to Classic Reporting in Reports and strictly goes to dynamic reporting 100% of the time.

N/A

Start at the reports home page found in the navigation bar of secure.my.jobs

Pull Down the Repository and go to the navigation bar clicking on Reports. Reports should not ask for or allow for you to go to Classic Reporting and should go straight to dynamic reporting view.

All Laughs from the Python guys are welcomed as far as the views.py file that was edited by the FRONT END Developer lol.


gulp/src/reporting/components/ReportList.jsx
```diff
@@ -138,12 +138,6 @@ class ReportList extends Component {
               {reportLinks}
             </ul>
           </div>
-          <h2>Reporting Version</h2>
-          <button
-            className="button primary wide"
-            onClick={() => this.handleSwitchVersions()}>
```
tdphillips: Should handleSwitchVersions be removed as well?

mwarnsleyDE: @tdphillips I just took a look at this and removed it and I believe removing it doesn't give you the chance to create a new report unless I'm not doing it correctly

mwarnsleyDE: nvm troy

mwarnsleyDE: I was looking in the wrong spot that's my fault

tdphillips:
I suspect there's going to be lots of test fixes. We've probably got a bunch still looking at old reporting.

Discussion point that we should probably talk about:

Remove the "overview" view in its entirety and modify myreports/urls.py to point view/overview to the dynamicoverview view.
After that, is a RedirectView from view/dynamicoverview to view/overview necessary?

PR #2810, opened 2017-02-24
Some graphical issues remain

tdphillips:
## What's this PR do?
This PR is going to reduce duplication of ancient templates by making use of existing blocks rather than reimplementing them.
moc_coding/customcareer did some weird stuff. The existing JS was bad and had syntax errors so I don't feel bad about this looking different.
Left: new, right: old
2017-02-24-093623_1953x520_scrot

In the admin.
Check admin pages, I guess?
We had lots of code in templates. Now we have less.

Autofocus is a thing that the Django admin likes to do, though it's not good at it.
Several of our admin sections have `save_on_top=True` and Django is dumb, constantly focusing the first visible input. The result of that is the "Save" button gaining focus. A list of such pages (updated as I go along):
seo/billboardimage
seo/configuration
seo/seosite
social_links/sociallink


tdphillips:
Should be good to go.

PR #2811, opened 2017-02-27
Generate fake JobViews and put them into MemSQL.

almathew:
## What's this PR do?
Generates fake job views and adds them to MemSQL. Useful for testing analytics locally, since all QC MemSQL requires VPN access.

I tried to create a useful distribution of cities/regions/countries/job titles/etc, and my results thus far suggest that I was successful. 

Data is random and will not represent true job views. The foundations for the data were pulled from actual job views, but should have any company-identifying information stripped.
 
screen shot 2017-02-27 at 8 09 50 am

With a solid understanding of what a job view is, and opinions on what fake data should look like.

Try running the command (`python manage.py make_data` or `dkm make_data` if you're using Docker) against your local MemSQL db. Confirm that you are satisfied with what you see in analytics.

No.


analytics/management/commands/make_data.py
```diff
@@ -0,0 +1,78 @@
...
+        view_sources = view_sources.distinct()
+
+        # Date range we're generating fake data for
+        days_to_generate = 14
+        date_range = [datetime.datetime.now() - datetime.timedelta(x)
+                      for x in range(0, days_to_generate)]
+
+        job_views = []
+        num_jobs_to_create = random.randint(num_in_use_buids * 2,
+                                            num_in_use_buids * 20)
```
darrint: I think you need at least one more zero here.

darrint:
Having `make_data_datasource` in that directory causes it to show up as a command when you run `python manage.py` or `dkm` with no arguments.

analytics/management/commands/make_data.py
almathew: It's the number of entries, not the number of job views per entry.
I'm happy to increase the number of zeros, but I don't think the data is quite diverse enough to support that while still being worthwhile.

darrint: A few thousand means you'll have plenty of job titles in every location. I was getting only a few tens of job views running this. IMHO the data is more interesting when there are a few thousand records.

almathew:
The datasource file has been moved, so it no longer shows up as a manage command.

analytics/management/commands/make_data.py
almathew: Fair request, and it doesn't make a huge difference in run time. Two zeros have been added.

PR #2814, opened 2017-02-27
Update to VPC cache.

mklauber:
## What's this PR do?
We've moved the QC server into the VPC.  Unfortunately, aws doesn't permit you to allow VPC instances to access non VPC caches.  So we need to start using the VPC cache.  

It can't it's an infrastructure change, not a code change.


almathew:
👍

PR #2821, opened 2017-02-28
Tests fail under keepdb without this.

darrint:
## What's this PR do?

Makes `myreports` tests run with `--keepdb`.

It's short.

`dkm test myreports --keepdb`

There are probably other places in the tests that need this. I didn't go looking for more trouble.

I'm mystified as to why this worked before.


almathew:
I'm not sure how I missed it, but next time we might not want to provide a default option a unique field. That's really only useful once.

PR #2825, opened 2017-02-28
Reporting - dynamic report intentions [PD-2962]

tdphillips:
## What's this PR do?
This PR makes the report intention select box dynamic.

As of right now, I'd like to ensure this meets the expectations of the team. I think it's going in the right direction but additional input is always nice.
2017-02-28-132455_392x344_scrot
Take a look at the initial reporting screen. There should be more than two options.
Click "PRM Reports". Note that you've got a bunch of fields to fill out.
Click one of the other options. Note that you get redirected to analytics.


myreports/models.py
```diff
@@ -123,6 +124,13 @@ class ReportingType(models.Model):
     def __unicode__(self):
         return self.reporting_type
 
+    @property
+    def link(self):
```
tdphillips: Just testing out some stuff. If we like how the other bits look, this can be moved to the database or something.

tdphillips:
Questions, concerns, thoughts? I'll clean it up and move everything to the database if people like where this is going.

gulp/src/reporting/components/SetUpReport.jsx
```diff
@@ -38,10 +38,10 @@ import TagAnd from 'common/ui/tags/TagAnd';
 class SetUpReport extends Component {
   onIntentionChange(intention, selectedObject) {
     const {history, category, dataSet} = this.props;
-    if (selectedObject && selectedObject.link) {
-      const link = selectedObject.link;
-      if (link === 'analytics') {
-        window.location = '/analytics/view/main';
+    if (selectedObject) {
+      console.log(selectedObject); // eslint-disable-line
```
darrint: We want to keep this?

gulp/src/reporting/components/SetUpReport.jsx
```diff
@@ -38,10 +38,10 @@ import TagAnd from 'common/ui/tags/TagAnd';
...
+      if (selectedObject.link) {
+        window.location = selectedObject.link;
```
darrint: I like this a lot.

gulp/src/reporting/components/SetUpReport.jsx
tdphillips: Removed.

myreports/models.py
darrint: I like it. Somebody has to know what apps are at what urls. I'd rather it be Django's route system than hard coded strings in JS.

analytics/urls.py
```diff
@@ -16,6 +16,7 @@
 view_patterns = patterns(
     'analytics.views',
     url(r'^main', 'analytics_main', name='analytics_main'),
+    url(r'^site', 'analytics_main', name='analytics_site'),
```
darrint: This is the only part I'm unsure about. Instead of reversing on various views we might be able to handle this by just passing a (ReportType?) id parameter to analytics_main through the url itself. (If we go with separate urls for everything than my other "I like it" comments apply. If we can get away with less of these urls then that's better, IMHO.)

tdphillips: Sounds good to me. I'll get to work.

tdphillips: Is having split urls like this meaningfully different from tacking a random parameter onto it somewhere?

darrint: The random parameter is a reporting intention. I expect those to grow at least a few times per year.

mwarnsleyDE:
@tdphillips are we still doing this even though the dropdown is going away?

tdphillips:
We may not be doing the dropdown but we're still dynamically generating whatever the selection page will be so I'd say the work is still relevant.

tdphillips:
Remaining work: Add "link" field to ReportingType and add data supporting it to the database.

tdphillips:
I mistagged all commits. :cry: 

This should be PD-2962

darrint:
On my system going to reports lands you on an blank form for "Job Analytics". In my PR which builds on this the problem goes away. I think we should merge this as is now but make a note of the possible known problem for whoever QC's this ticket.

PR #2827, opened 2017-03-01
[PD-2982] Make the last few holdouts for tag lookup/creation case insensitive.

almathew:
## What's this PR do?
Most of the places in the codebase already look up tags using `__iexact` before doing the create. This corrects the last few that didn't.

I would really be happy if there were a way to centralize all this logic so tag-case-insensitivity didn't need to be some super-secret knowledge passed from developer to developer.

It's a small change. The biggest question is if there anything I missed?

Create a tag.
Try to re-create that same tag, just different case sensitivity, everywhere you can. In all instances it should use the existing tag. It shouldn't create a new tag.

The ticket calls for correcting existing tags, but I wasn't able to find any in production. There may be a following PR with a migration if someone proves me wrong about those.


mypartners/helpers.py
```diff
@@ -593,16 +593,21 @@ def new_partner_from_library(request):
                             ('veteran', '5EB94E')]:
 
         if getattr(library, 'is_%s' % interest):
+            interest = interest.replace('_', ' ').title()
             tag, _ = Tag.objects.get_or_create(
-                company=company, name=interest.replace('_', ' ').title(),
-                defaults={'hex_color': color})
+                company=company,
+                name__iexact=interest,
```
tdphillips: Do we need both `.title()` and `name__iexact`?

I know the speed saving would be basically nonexistent for something like this.

mypartners/helpers.py
```diff
@@ -593,16 +593,21 @@ def new_partner_from_library(request):
                             ('veteran', '5EB94E')]:
 
         if getattr(library, 'is_%s' % interest):
+            interest = interest.replace('_', ' ').title()
             tag, _ = Tag.objects.get_or_create(
-                company=company, name=interest.replace('_', ' ').title(),
```
darrint: Probably to make things nice for the JavaScript?

almathew: I don't think anything other than this look ever sees the snake cased stuff in the dictionary, but JavaScript possibly using it would be a good example of why I just didn't touch it.

PR #2829, opened 2017-03-02
[PD-2954] Make all database transactions for views sendgrid uses atomic.

almathew:
## What's this PR do?
Ensures that we don't commit anything to the database until after we successfully send the 200 response back to sendgrid. 

Specifically, this is to prevent duplicate contact record creation in the event that the view times out or fails after a few contact records have been created.


redirect/views.py
```diff
@@ -222,6 +223,7 @@ def myjobs_redirect(request):
         'http://www.my.jobs' + request.get_full_path())
 
 
+@transaction.atomic
```
tdphillips: What does this get us on the redirect side?

almathew: We don't create a log entry.

Outside of middleware it's the last thing we do, and one could argue that we might want the log entry either way, so it might be pointless.

tdphillips: We don't create log entries for PRM emails. The bit at the end of the view only occurs for a specific set of emails.

almathew: Removed it.

redirect/views.py
```diff
@@ -5,6 +5,7 @@
 from urlparse import urlparse
 import uuid
 
+from django.db import transaction
```
mklauber: Why are we importing this here?

almathew: I was using it, and then @tdphillips pointed out I didn't need to. I removed the decorator but not the import. Removed.

PR #2833, opened 2017-03-03
Limit task count per worker.

mklauber:
## What's this PR do?
Adds a limit of the number of tasks executed before the worker task is respawned.  

N/A

N/A

N/A

Currently, our workers have a memory leak that appears to result in them hanging for a extended period.  Hopefully, limiting the number of tasks per process should prevent the memory leak from climbing to the point where it causes issues.


default_settings.py
```diff
@@ -135,6 +135,7 @@
 CELERY_RESULT_BACKEND = 'amqp'
 CELERY_IMPORTS = ('tasks',)
 CELERY_PREFETCH_MULTIPLIER = 0
+CELERYD_MAX_TASKS_PER_CHILD = 10
```
almathew: Is restarting going to impact email sending on myjobs negatively? Or any other workers I might not know about.

mklauber: I don't believe so.  It may slow down emails slightly, due to the overhead of spinning up processes.  But without testing it we can't know to what degree.  I expect it would probably be small however.

PR #2847, opened 2017-03-08
[PD 3011] replace loading spinners

mwarnsleyDE:
## What's this PR do?
This PR replaces all of the loading spinners with the one we use for analytics

N/A

Start inside PRM, Reports, or Manage Users. Anywhere that may have the loading spinners for data loading

Pull Down the Repository. There are a couple of the apps that utilize the loading spinners which should be Analytics, PRM, Reports, and Manage Users. Check and Make sure the loading spinner is the same blue and gray one that we use for analytics

Styling doesn't have to be the same as far as matching analytics goes. Only the spinner has to be the same. For all of the apps except for Analytics, we won't have an overlay for it and just the spinner in general


mwarnsleyDE:
BEFORE MERGING I need to pull in the repository for the UI Reporting Intentions and edit the Loading Spinner inside of there first

tdphillips:
The reporting intention work has been merged.

mwarnsleyDE:
This PR is good as far as checking and merging now

PR #2848, opened 2017-03-08
[PD-2988] [PD-2960] [PD-2961] GUI and API for new report intention screen

darrint:
## What's this PR do?

Replaces the report intention dropdown with a new screen.

menu
prm-select

* The UI code: the `prm-reports` route is now responsible for handling PRM reports.
  * There's still a special case for PRM.
  * We haven't handed new reporting intentions that use the PRM ui. There can still be only one.
* Migrations: I've added a number of new fields to `ReportingIntention` and populated them.
  * Note the new 'ui' field, which switches explicitly between the old PRM ui and the new analytics ui.
* `views.py`/`models.py`: adds the new information to the api as needed.

1. `dkm migrate`
2. Go to reporting.
3. Try different options.

NA.


mwarnsleyDE:
DO NOT MERGE AS OF YET

Darrin still needs to disabled Site Analytics inside of the database since we currently don't have it and it's not being utilized so once that's done I will go through and make sure everything else looks good and then we can go ahead merging it from there.

mwarnsleyDE:
PR should be good to merge now once all of the test come back Green

PR #2853, opened 2017-03-10
[PD-3010] Fix query redirect admin

mklauber:
## What's this PR do?
Makes the QueryRedirect page in the admin render and behave correctly.  Prior to this PR, the fields redirect from, redirect to, and site fields did not display.  Now they do, and the site field will auto search as you type in it.  

New: image

The QueryRedirect page in the Admin (under Seo).  

Make sure the section of the form above displays, and that typing in the site field displays a drop down.

This issue arose from the django-1.9 upgrade, and the included version bump for django-extensions.  As part of the version bump, django-extensions stopped bundling it's own version of jquery when it dropped support for django-1.5, so I've removed that from our forms.


tdphillips:
When I try to use the autocomplete widget to choose a site, I get a bunch of JS errors regarding a lack of jQuery.

koonsies:
I get an error on page load, but everything seems to work properly. Thoughts?
screen shot 2017-03-12 at 2 30 47 pm

tdphillips:
Looks like something in our stack expects a version of jQuery that has $.browser. That has apparently been removed in the version that we use.

I suppose that means this might fail on IE. If that's super important, I suppose we can either try to fix it or include [MORE JAVASCRIPT](https://github.com/jquery/jquery-migrate/#readme).

mklauber:
I get that error too, but I don't think it's something to do with this.

PR #2854, opened 2017-03-10
[PD 2960] gui for report intention

mwarnsleyDE:
## What's this PR do?
This PR will hide the right hand saved search facet on the Reporting Intentions homepage

N/A

Starting in reports from the dropdown menu

Pull down the repository. Go into reports from the navigation. You should see that the reporting intentions are still there but the saved search facets on the right hand side is gone. If you click and go into PRM the facet will come back with the saved search reports. It is only hidden on the starting point main homepage

Minor fix right now until we get the canned reports in and then this code will go away. This is a spin off of the backend work that Darrin did for the reporting intentions UI page


mwarnsleyDE:
This is just using the old branch that I had with merging the stuff from Quality Control and hiding the saved search on the right hand side of it

PR #2858, opened 2017-03-10
[PD-3018] Simplify NUO Notes

darrint:
## What's this PR do?

The notes field was not working as users expected. This makes it simpler.

1. Look at the fix in `Textarea.js`. 
2. Note the fix in MyProfile.
3. Note the similar fix in NUO.
4. Note the removal of append logic in NuoContactAppendForm. (That may be due for a rename soon.)
5. A few test updates.

1. In NUO, view an existing contact. Notes should be present.

All uses of `<Textarea>` which included MyProfile and NUO were broken, not showing existing values. The underlying `textarea` control was uncontrolled.

We probably would have noticed earlier but this control is rarely used.


gulp/src/common/ui/RemoteFormField.jsx
```diff
@@ -54,6 +54,8 @@ export default class RemoteFormField extends Component {
           name={fieldName}
           onChange={e => onChange(e, fieldName)}
           required={field.required}
+          value={value}
+          maxLength={field.widget.maxlength}
```
koonsies: I don't see maxLength being used in the Textarea component

PR #2862, opened 2017-03-15
[PD-3031] user locations from ip

almathew:
WARNING: A tiny bit of the logic surrounding the database might change depending on some on exactly what version of the database we decide to go with. This also does not add location to reporting. Neither of these will have a huge impact on the logic in this pull request. Both of those will be handled, as needed, in a different pull request.

## I highly recommend that everyone adds the following line to dev_settings.py
`MMDB_PATH = 'db.mmdb'`
This prevents constantly reloading the database every time you create a new docker container (so use runserver, open a shell, etc).

## What's this PR do?
Given the IP address of a user, maps the IP to a location.

Most of the new logic starts in `models.py`. More specifically, `UserLocationBaseModel` will be the entry point, while the code in `user_locations.py` does the heavy lifting.

`UserLocationBaseModel` has custom logic for `save` and `__init__` in order to ensure locations are added. Is there any other case I missed besides those two?

The logic in `user_locations.init_db` and `user_locations.reload_db` is meant to ensure the database can be loaded/updated dynamically and, more importantly, without managing the database in source control.

Run the necessary migrations and create some new `JobView` objects with IP addresses. Confirm that the user location is getting added.


analytics/country_code_map.py
```diff
@@ -0,0 +1,249 @@
+COUNTRY_CODE_MAPPING = {
```
almathew: I am so sorry for the very existence of this file.

The only comparable resource was the `seo/fixtures/countries.json` file, and that isn't a perfect match and also ins't loaded by the Analytics app. So instead this happened.

analytics/models.py
```diff
@@ -1,7 +1,72 @@
...
 
+from analytics.user_locations import UserLocation
 
-class JobView(models.Model):
+
+class UserLocationBaseModel(models.Model):
+    ip = models.CharField(max_length=255, null=True)
+
+    user_location_city = models.CharField(max_length=255, null=True)
+    user_location_country_code = models.CharField(max_length=255, null=True)
```
tdphillips: This is either two or three characters, isn't it?

analytics/models.py
```diff
@@ -1,7 +1,72 @@
...
+    user_location_zip_code = models.CharField(max_length=255, null=True)
```
tdphillips: I know significantly less about zip codes than country codes but I suspect this could be shortened as well.

analytics/models.py
```diff
@@ -1,7 +1,72 @@
...
+    user_location_state = models.CharField(max_length=255, null=True)
+    user_location_lat = models.FloatField(null=True)
+    user_location_long = models.FloatField(null=True)
+
+    class Meta:
+        abstract = True
+
+    def update_location(self):
+        location = UserLocation(self.ip)
```
tdphillips: Should this be `UserLocation.objects.get` or something similar?

analytics/models.py
```diff
@@ -1,7 +1,72 @@
 from django.db import models
 
+from analytics.user_locations import UserLocation
 
-class JobView(models.Model):
+
+class UserLocationBaseModel(models.Model):
+    ip = models.CharField(max_length=255, null=True)
```
tdphillips: `null=True` on a boatload of CharFields.

default_settings.py
```diff
@@ -728,3 +728,6 @@
 
 # The email host used to parse communication records
 PRM_EMAIL_HOST = 'my.jobs'
+
+MMDB_PATH = '/tmp/db.mmdb'
+MMDB_URL = 'https://s3.amazonaws.com/de-mmdb/db.mmdb'
```
tdphillips: This appears to be world-accessible. Is that intended/allowed?

analytics/models.py
almathew: That is not Django recommended, but we really are storing them as null in the database in places.

default_settings.py
almathew: It's already available world readable on their website.
I just don't want to rely on their update cycle or them keeping the URL the same.
Worst that happens from this is someone decides to download it a bunch of times, but they could already do that with any of our static files.

analytics/models.py
almathew: For map UI purposes it should always be 3, but some countries don't have 3 and I can't guarantee it actually will be.

analytics/models.py
almathew: UserLocation isn't a model. It's just an object I wrote because @darrint wanted a named tuple, and an object offered all the named tuple advantages plus a more convenient place to group the logic.

analytics/models.py
almathew: No clue here. We have other zip code fields that are that size, so at this point on some level I would just call it inertia. 
I could ask Ron what he expects for international zip codes, but since we're using a new library that his team has 0 involvement with, I'm not even sure he's dealing with them in a way similar to what we're doing.

PR #2863, opened 2017-03-15
pd-3037 New tables and db for batch jobs.

darrint:
## What's this PR do?

Sets up new MySQL tables in a separate DB for handling batch work in analytics. In staging and QC I've added settings for deployment.

I did _not_ handle production settings. I created this sub task:

https://directemployersfoundation.atlassian.net/browse/PD-3028

* Tables
* Router
* Settings

You will need this in your `dev_settings.py`:

```
DATABASES = {
    ...
    'analytics_batch_control': {
        'NAME': 'db',
        'ENGINE': 'django.db.backends.mysql',
        'USER': 'root',
        'PASSWORD': secrets.MY_PASSWORD,
        'HOST': mysql_host,
        'PORT': mysql_port,
    },
    ...
```

Run unit tests on analytics to make sure the migrations work.

Currently we have a mechanism for this which works for job views in MongoDB.

It has two problems:

1. It's in MongoDB so it needs to move out of there.
2. A single record represents both the work needing to be done and the work done. There's no nice way to represent that each batch needs multiple jobs run against it.

Additional concerns:

1. MemSQL's transactional tables don't play well with Django and we don't need in ram only for this.
2. There will be several reads and writes per second to these tables so we want a separate db instance in production.

This PR creates two tables in MySQL, one for batches, and one for work done, and assumes a separate DB instance. In development the separate db can just be configured to be the same as the usual development db.


deploy/settings.myjobs_qc.py
```diff
@@ -19,6 +19,12 @@
         'HOST': 'db-redirectqc.c9shuxvtcmer.us-east-1.rds.amazonaws.com',
         'PORT': '3306',
     }, **REDIRECT_QC),
+    'analytics_batch_control': dict({
+        'NAME': 'redirect',
```
mklauber: If we're expecting this to be a different DB, can we name it something different? Or are we expecting this to be the same DB in QC/Staging?

darrint: Same in QC/Staging.

analytics/migrations/0009_merge.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.10 on 2017-03-15 19:54
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
```
almathew: Did you try this after you merged? The dependencies are weird.

almathew: Nevermind, it's right, there are just two...

almathew:
We need to make sure that the migration command gets for analytics_batch_control gets added to the deployment script still.

PR #2864, opened 2017-03-16
Added new production db for analytics batching [PD-3038]

almathew:
## What's this PR do?
Adds a new production database for analytics batch jobs (see https://github.com/DirectEmployers/MyJobs/pull/2863)

Database is in the VPC and locked down to only connections coming from inside the VPC, so any testing of the database connection requires VPN access.

The database is Aurora (a scalable MySQL replacement) and not MySQL.
secrets.py has already been updated.


darrint:
We tested these settings against the migration for this database in my PR and verified that the migration worked as expected.

PR #2865, opened 2017-03-16
[PD-3014] ui improvements

mwarnsleyDE:
## What's this PR do?
PR updates a lot on the CBO application in terms of coding and adding in the ability to add a CBO as well as adding additional contacts, distribution locations, etc.

n/a

Start in the CBO app under the Beta Tab

Pull Down Repository and go through the CBO app under the Beta Tab. You should be able to add a new cbo using the cbo wizard that can be found on the right hand facet. As well, you will be able to add additional contacts and distribution locations to the current cbo contacts that are already created if need be.

Big PR due to all of the refactoring that was needed for it. May need to stub out a bunch of other task and branch them off of it so it doesn't grow too much out of control


mwarnsleyDE:
Still a lot of work to do on this PR. It is pretty big as of yet but there are bunch of linting errors from some validation that I still have to finish up and do along with some changes and updates that need to be made to the Add CBO wizard. Also still need to make the ability to add additional contacts and distribution locations with the CBO contacts that are already created.

gulp/src/cbo/components/contact/ContactListRow.js
```diff
@@ -1,37 +1,57 @@
...
+            onClick={() => {onDelete(zipperedCbo);}}>
+            <i className="glyphicon glyphicon-trash delete-icon"/>
+            {deleting ? ' Deleting...' : ' Delete'}
+          </button>);
+    }
+    return (
+      <tr>
+        <td>
+          <Select
+            name="state"
```
koonsies: This has the dropdown as States, but the title is category. Additionally, I don't think we need any edit-field functionality on the list row itself. Better to keep that in the ContactForm.

gulp/src/cbo/components/cbo/AddContact.js
```diff
@@ -0,0 +1,235 @@
...
+    for (let i = 0; i < categoryClass.length; i++) {
+      if (categoryClass[i].checked) {
+        categoryList.push(categoryClass[i].nextSibling.innerHTML);
+      }
+    }
+    this.setState({
+      categoryList: categoryList,
+    });
+  }
+  formatPhoneNumber(e) {
```
koonsies: This prevent the user from using the arrow keys to navigate to a section of the number to edit, or clicking in and area and deleting more than one character (as it pushes the cursor to the end of the text at every keyUp)

koonsies:
The add new contact link looks good to me. Having that same option on the cbo list, the distribution record list and (eventually) the rules list will be good.

gulp/src/cbo/components/cbo/AddContact.js
mwarnsleyDE: Good Call, I will edit that. I didn't even notice it until I tried it here

gulp/src/cbo/components/cbo/AddCbo.js
```diff
@@ -0,0 +1,76 @@
...
+    const classes = this.state.errors ? ['form-control', 'add-cbo-input', 'add-cbo-error'] : ['form-control', 'add-cbo-input'];
+    const btnText = cboPreviousState === 'NEW_CBO_REVIEW' ? 'Update' : 'Next';
+    return (
+      <div id="add_new_cbo">
+        <h2 className="top">Add CBO</h2>
+        <div className="row">
+          <div className="col-md-12">
+            <TextInput
+              id="add_cbo_input"
+              name="name"
```
koonsies: We should be careful with naming this name, as it picks up Google's autocomplete for "name". "Jonathan" isn't a valid input for my CBO haha.

cbo/api.py
```diff
@@ -11,7 +11,7 @@
                                add_pagination, get_object_or_none)
 from django.views.defaults import bad_request
 from django.http import JsonResponse
-
+import ipdb;
```
mklauber: Probably need to remove all instances of this.

mwarnsleyDE: @mklauber I removed them from that file. Not too sure if they are located in any of the other files or not but I can check

mwarnsleyDE:
@koonsies you can check those changes now. They should be good from when I was testing them I think the only issues I'm having is on the edit contact details as far as populating the state and categories again for a contact.

mwarnsleyDE:
I think storing the state Id and the categories and their id's and stuff were what you were referring too so they are all stored inside of redux for when they are needed

koonsies:
A few notes from Tejaswini reviewing the UI:

* Pressing NewCBO -> Enter CBO NAME -> Hit NEXT ->(oops!! I’ve given CBO rather a bad name)-> Hit back button….Comes to cbo/view. Then on pressing New CBO, the prompt the info will be lost comes up
* I went back from the creating CBO to view CBOs page. The clicked on New CBO. It goes directly to create contact information and keeps the CBO name created in previous instance
* CBO name doesn’t take any special characters. A validation and appropriate message might be better, coz when I put in a spl character, it just says “Name field must contain at least 2 or more characters”
This strangely doesn’t happen when we just hit Edit on existing CBO and update CBO name
* On edit CBO -> On pressing EDIT Distribution Location, Dropdown for state doesn’t appear

gulp/src/cbo/components/cbo/AddContact.js
```diff
@@ -0,0 +1,391 @@
...
+      stateValue: 'selectState',
+      stateId: '',
+      categories: [],
+    };
+
+    this.mouseUpHandler = this.mouseUpHandler.bind(this);
+    this.mouseDownHandler = this.mouseDownHandler.bind(this);
+  }
+  componentDidMount() {
+    window.addEventListener('mousedown', this.pageClick.bind(this), false);
```
koonsies: I think this needs to be removed on unMount. This triggers a ton of warnings that setState is called on an unmounted component.

default_settings.py
```diff
@@ -537,21 +537,21 @@
 HAYSTACK_CONNECTIONS = {
     'default': {
         'ENGINE': 'seo.search_backend.DESolrEngine',
-        'URL': 'http://127.0.0.1:8983/solr/seo',
+        'URL': 'http://solr:8983/solr/seo',
```
mklauber: I think these changes are what are causing tests to fail.  We can talk tomorrow, after deployment.

PR #2870, opened 2017-03-20
Apply and source metrics [PD-3026]

tdphillips:
## What's this PR do?
This PR adds columns to analytics reports denoting percentages for the number of intended applies and the number of users who reached the job source.
2017-03-20-094711_1040x828_scrot
myreports/datasources/job_views.py
Rerun `dkm make_data`
Verify that your analytics reports have these new columns.


tdphillips:
Why do I even try?

PR #2871, opened 2017-03-21
[PD-3025] Group ReportingTypes on report intentions page.

almathew:
## What's this PR do?
Allows for grouping ReportingTypes so that the user can understand what they're looking at and find the reporting type they're looking for more easily.

report_intentions

* The view in `myreports/views.py` has been updated to return a nested dictionary of groups and reporting types, where previously it was only returning the reporting types. The outer level is the group, and the inner level contains group information, and a list of matching report types.
* The javascript has been updated so that, if a group is present, the reporting types are grouped together. Otherwise the reporting types should appear to be their own separate entity (as it was before).
* New reporting types, from start to finish, should be manageable in the Django admin. Unfortunately, due to limitations of the django admin, there's quite a few separate pages required to make this happen.
** I didn't add history to the objects like I jokingly said I might. Yet.

1. Run both migrations
2. Navigate to https://secure.my.jobs/reports/view/overview/ and check out the awesome new reporting intentions page.
3. Go to the Django admin and create your own new reports.
4. Navigate back to the page in step (2) and see how much more awesome the reporting intentions page got.

No.


myreports/models.py
```diff
@@ -156,23 +181,35 @@ def active_for_reporting_type(self, reporting_type):
 
 
 class ReportType(models.Model):
+    DATASOURCES = ((key, key) for key in ds_json_drivers.keys())
```
almathew: Similar to this and `UI_CHOICES`, I'd like to start providing choices to users where they could otherwise put in a nonsensical choice that might mess things up.

I know for sure there are probably other fields I've missed, so let me know if you notice any others. Otherwise I'll probably just fix them as I encounter them.

mwarnsleyDE:
BEFORE MERGING. I might have to take a look and edit this here and there for the Reporting Intentions screen

mwarnsleyDE:
For this one I guess the way the data is coming back it's a bit harder to try and format this as far as the UI goes. I'm not sure if this is how @fezick wanted this to look for when it gets done but if he verifies it then we can go ahead and merge it. We will eventually have to change it up a bit as far as the html structure goes for browser compatibility purposes.

almathew:
I was working off the mockup @fezick gave me:

pasted image at 2017_03_21 08_27 am

mwarnsleyDE:
Oh okay, I guess I didn't see that mockup. I should have looked. We might have to take the hover effect off of the indented ones that make them pop and make it a bit different instead but other than that the rest of it shouldn't be a problem then.

PR #2855, opened 2017-03-10
[PD 3006] Saved search monitoring

tejaswinigutti31:
## What's this PR do?
Monitor saved searches and alert through email when there are failures
## Where should the reviewer start?
Check mysearches/tasks.py for method monitor_saved_search()
## How should this be manually tested?
Run the file mysearches/tests/test_tasks.py

PR #2872, opened 2017-03-21
Move seoxml imports to a separate queue, and monitor them individually.

mklauber:
Despite resolving the memory leaks, imports are still hanging.  Based on circumstantial evidence, we believe that seoxml may be causing the hangups.  This separates out the seoxml tasks to a separate queue.  A related PR will create a separate worker for that queue. 

## What's this PR do?
Move seoxml tasks to a separate queue.
Combine daily success and failure monitoring into one monitoring_imports task that runs per import task type.

N/A

tasks.py->monitor_imports


seo/tests/test_tasks.py
```diff
@@ -49,14 +43,14 @@ def test_check_total_throughput_exceeded(self, mock_send_mail):
...
             TaskState.objects.create(
                 task_id=i,
                 tstamp=self.midnight,
-                name=choice(tasks.SOLR_TASKS),
+                name='tasks.etl_to_solr',
                 state='SUCCESS',
             )
-        tasks.check_total_throughput()
+        tasks.monitor_imports()
         mock_send_mail.assert_not_called()
```
JLMcLaugh: @tejaswinigutti31 saw an issue with assert_not_called here: https://github.com/DirectEmployers/MyJobs/pull/2855/files#diff-6613b68e7c93b8b236e47c0cb6bdb8e6R46

This should be sending 4 emails if you're only creating tasks for 1 of the 5 monitored tasks.

PR #2884, opened 2017-03-27
Fix intended_applies calculation [PD-3026]

tdphillips:
## What's this PR do?
Reduces `intended_applies` from a sum of `applications_intended` to a count of that field where it is greater than 0.
`applications_intended` can be None or any integer, not None or 1 as I originally thought.


analytics/management/commands/make_data.py
```diff
@@ -74,7 +74,7 @@ def handle(self, *args, **options):
                 view_source=random.choice(view_sources),
                 zip_code=zip_code,
                 view_count=random.randint(2, 75),
-                applications_intended=random.choice([None, 1]),
```
tdphillips: Just a dumb little test value. It's unimportant in every way.

myreports/datasources/job_views.py
```diff
@@ -38,7 +38,7 @@ def run_unaggregated(self, company, filter_spec, values,
                                        partial=partial)
         query_set = query_set.annotate(
             job_views=Sum('view_count'),
-            intended_applies=Sum('applications_intended'),
+            intended_applies=Count(Case(When(applications_intended__gt=0, then=1), output_field=IntegerField())),
```
tdphillips: Woo!

myreports/datasources/job_views.py
```diff
@@ -51,7 +51,9 @@ def run_aggregated(self, company, filter_spec, values,
         query_set = query_set.values(*values)
         query_set = query_set.annotate(
             job_views=Sum('view_count'),
-            intended_applies=Sum('applications_intended'),
+            intended_applies=Count(Case(When(
```
tdphillips: Woo!

PR #2887, opened 2017-03-27
add test diagnostic

darrint:
## What's this PR do?

I'm trying to see why some unit tests are slow. Don't merge this.


tdphillips:
In general we shouldn't be testing basic Django functionality to begin with. If that test is slow, could we just remove it?

almathew:
I'm not sure that the tests are necessarily worth the time it takes, but our seo admin is far from basic Django, which is why I imagine the tests were necessary in the first place. 

We might be better testing some specific urls that we know we have had problems with?
@JLMcLaugh

tdphillips:
I don't consider "page returns a 200 or 403" to be testing our custom admin functionality.

darrint:
Here are the results. After adding the flush, no particular admin page is taking that much time.

https://gist.github.com/darrint/8ad5348aa29751d19bf641aace88e78c

PR #2888, opened 2017-03-29
[PD-2833] Parse user agents to provide end-user with more readable information.

almathew:
## What's this PR do?
Parses the user agent and adds specific parsed info to the AnalyticsEvent object.

- Made small modifications to the UserLocationBaseModel to be a bit more consistent
- Added the user agent parsing. Should ensure that user agent is always created before first save.

Create objects every way you can with any user agent you can find. Make sure the info is set as you expected after the object is saved.


tdphillips:
This now has conflicts and it's not mine so I'm hesitant to push buttons.

almathew:
@darrint @tejaswinigutti31 are changes being made to this anyway?

darrint:
This was incorporated into a different branch.

PR #2890, opened 2017-03-30
[PD-3030] AnalyticsEvent datasource

almathew:
## What's this PR do?
Adds a datasource for AnalyticsEvent.
In order to re-use more code when adding this datasource, I added a base datasource for flat models using the Django ORM straight.

The analytics event datasource and the analytics event datasource filter are the only two really new items.
The most important new changes are in the new generic django orm base datasource in `simple_django_orm`

You could probably build a report out of it and use it in analytics if you were really motivated, but the easiest way is just to make queries against it in a similar way that it's being done in `tests/test_analytics_event_datasource.py`.

The generic datasource had to be done, but it has a lot of design decisions I don't like. I just don't have any better ways of doing them. I'd love suggestions there.


almathew:
I thought I found a problem, but it turns out my report structure data is just extremely messed up.

PR #2897, opened 2017-04-03
Fixed console errors

aljordan82:
## What's this PR do?
Removed event listener on when components unmounted.
Removed dead code.
Navigate to CBO wizard second page and then go back to the main page.
Clicking anywhere on the page shouldn't cause console errors.


koonsies:
It looks like this issue still persists on the edit contact form.

PR #2907, opened 2017-04-05
[PD-3105] map help text shows zero

mwarnsleyDE:
## What's this PR do?
Map Text for analytics fixed for showing the amount of job views

N/A

Pull down the repository. Go into the analytics app and check the World and USA State Maps and hover over a specific Country or State and you will see the job views displaying the actual views instead of 0

Quick Fix Needed THIS WEEK


mwarnsleyDE:
For whoever takes a look at this. Jason Sole suggested we put a Unit Test in this to make sure the job views doesn't break again. I'm not sure how we would go about doing this unless we used Enzyme just based off of the fact that it's not using Redux and anything so I don't really know how we would Unit Test certain code inside of the actual component itself. It's Javascript code that you will see was changed inside of the render function.

PR #2925, opened 2017-04-10
Pd 3122 cbo realign forms in wizard 2

aljordan82:
## What's this PR do?
Realign forms
screen shot 2017-04-10 at 12 53 29 pm


mwarnsleyDE:
@aljordan82 This all looks good to me. The only thing I'm seeing would be making the forms on the manage contacts page, manage distribution locations, and manage rules aligned with the wizard. They are all forms so we will have to make them all uniformed. We can either do it on this PR or we can make it work on another created ticket.

mwarnsleyDE:
@aljordan82 also for the name field inside of the wizard we should probably make the input longer and closer to the name instead of having the large gap next to it but other than that I think the code and using already components and everything looks great

PR #2926, opened 2017-04-10
[PD 3107] meeting length open text NUO

mwarnsleyDE:
## What's this PR do?
PR changes the meeting length field in the communication record form of NUO from an input field to a multi select dropdown with hours and minutes like the one in PRM.

**PRM**
screen shot 2017-04-10 at 1 15 47 pm

**NUO**
screen shot 2017-04-10 at 1 13 53 pm

Start inside of the NUO application

Pull down repository. Go into the NUO app and create a new partner OR select an already existing partner. Select a contact and go to the communication record page. On that page if you set the select dropdown to meeting, it will show the meeting location and meeting length which will be the Native Select boxes like in PRM.

Jason Sole said to use the Native Select since we will have to go to that eventually anyways. This doesn't include the functionality of saving this as it might be some backend work but I'm not for sure as of yet but this is for replacing the input box with a select box.


gulp/src/common/ui/TimeLengthPicker.jsx
```diff
@@ -0,0 +1,65 @@
...
+      return (
+        <option key={i}>{minute}</option>
+      );
+    });
+    return <div className="length-container"><select id="minutes_length">{minuteList}</select></div>;
+  }
+  render() {
+    return (
+      <div id="date_length_container" className="date-container">
+        {this.renderHours()}
```
aljordan82: Wouldn't it be better to render the hours and minute JSX outside of render()?
I don't see the need to recreate the JSX with each render()

mypartners/forms.py
```diff
@@ -375,7 +375,7 @@ class Meta:
         form_name = "Communication Record"
         fields = ('contact_type', 'contact',
                   'contact_email', 'contact_phone', 'location',
-                  'length', 'subject', 'date_time', 'job_id',
+                  'length', 'subject', 'length', 'job_id',
```
koonsies: date_time was changed to length, so the new line has "length" listed twice.

gulp/src/common/ui/TimeLengthPicker.jsx
mwarnsleyDE: @aljordan82 if I am reading what you're saying right, the jsx always re-renders on a render. This is actually rendering it outside of the render but you have to place it inside of the render in order for the JSX to be seen

mypartners/forms.py
mwarnsleyDE: Updated this

koonsies:
If I am reading this correctly, this code only changes the front end, not how we send the data or how the backend processes it. If that's the case and this pretty much breaks this field for NUO, we'll not be able to merge this as-is. Someone who wants to fix the backend needs to pull this ticket down and make the needed changes.

koonsies:
Also, can we fix the font on the "hours" and "minutes" labels? They are bolder than almost any font on the page.

gulp/src/common/ui/RemoteFormField.jsx
```diff
@@ -34,6 +35,10 @@ export default class RemoteFormField extends Component {
 
     switch (inputType) {
     case 'time':
+    case 'date_time':
```
koonsies: This doesn't look right. "date_time" is duplicated and "time" is being ignored here.

PR #2931, opened 2017-04-12
[PD 3155] make category required

mwarnsleyDE:
## What's this PR do?
This PR makes the categories a required field when adding a contact inside of CBO

N/A

Pull down the repository. You can either go to a CBO and add an additional contact from the contact list or you can create a new cbo and add a contact that way as well. If you try and go to the next page or save without selecting a contact you will see the validation message popup prompting that a category is going to be required


aljordan82:
@mwarnsleyDE  Does the other form need to be changes? contact/ContactForm.js

mwarnsleyDE:
@aljordan82 we do need it on the contact form but we can create a new ticket for that as I don't see there is any validation on the contact form at all right now as far as updating and saving a contacts information but we do need it

koonsies:
Good call @aljordan82, I was thinking the same thing (and have for a while). I actually recently created https://directemployersfoundation.atlassian.net/browse/PD-3177 to deal with that very problem.

PR #2932, opened 2017-04-12
[PD 3153] implement validation rule type location

mwarnsleyDE:
## What's this PR do?
PR adds validation to the Rule Type Location

N/A

Pull down the repository. Go into a cbo and then into a contact to the manage contact page. If you go down and click add additional rule and attempt to add a rule with the type being location and not having the field filled out there will be a validation message before allowing you to proceed.

Assistance required with the regex for this one to make sure it's in line. Not the greatest at regex and just want to make sure it's in the correct form. EXAMPLE: Indianapolis, IN Chicago, IL New York, NY etc


mwarnsleyDE:
@koonsies this is the PR I was referring to with the Regex in it for the Rule Type of Location

koonsies:
Most of this looks good. We have an issue where validators aren't being used for both edit and add forms. I'm going to actually start a refactor PR that should fix that and make it a little cleaner.

PR #2936, opened 2017-04-13
User locations added to reports

tejaswinigutti31:
## What's this PR do?
Adds User Location to reports


gulp/src/analytics/components/Charts/ChartContainer.js
```diff
@@ -38,7 +38,8 @@ class ChartContainer extends Component {
       break;
     case 'map:nation':
       const countryFilter = chartData.activeFilters.filter((x) => x.type === 'country')[0] || {};
-      if (countryFilter.value === 'USA') {
+      const userCountryFilter = chartData.activeFilters.filter((x) => x.type === 'user_location_country_code')[0] || {};
```
darrint: @mwarnsleyDE @tejaswinigutti31 It would be better if we could figure out how do this this in a reducer.

mwarnsleyDE:
@darrint check this now. I moved that into Redux if that's what you were referring too for filtering that out.

PR #2938, opened 2017-04-13
Pd 3173 analytics tables are cropped in mobile

aljordan82:
## What's this PR do?
Fix the CSS for mobile devices


gulp/src/sass/analytics.scss
```diff
@@ -802,7 +800,7 @@ path:hover {
 }
 .search-container {
   position: relative;
-  float: right;
+  //float: right;
```
koonsies: rather than commit commented code, just remove the line.

PR #2939, opened 2017-04-13
Pd 3122 cbo realign forms in wizard 2

aljordan82:
## What's this PR do?
Align form labels to the right.
screen shot 2017-04-13 at 12 44 29 pm


mwarnsleyDE:
@aljordan82 for this one we might want to wrap a div around those and target them that way inside of the css using specific classes instead of targeting the bootstrap classes. We are doing that globally for some reason from a long time ago and it creates a hassle if you ever want to add the bootstrap grid to a div with a specific ID and don't want those styling for the row class and the col-md-4 class.

PR #2948, opened 2017-04-17
Ensure that we don't show another partner's contacts when creating a partner [PD-2634]

tdphillips:
## What's this PR do?
Stops a heroic filtering that ended up being silly
The three modified lines
Go to NUO.
Begin creating a new partner through the NUO wizard
Ensure that existing contacts don't appear in the contact autocomplete


koonsies:
When I type a single letter into the contact search box, I get an exception on the Python API

```Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/http.py", line 42, in inner
    return func(request, *args, **kwargs)
  File "/MyJobs/mypartners/views.py", line 2442, in api_get_contacts
    } for contact in sorted_contacts]))
AttributeError: 'NoneType' object has no attribute 'pk'```

**Comment by tdphillips** [comment]
I'm not getting a stack trace.

There is an issue when you start with a contact rather than partner, however.

**Comment by tdphillips** [comment]
Taking another look at it, would you happen to have a contact with no partner?

File: mypartners/views.py [hunk 1]
```diff
@@ -2414,10 +2414,12 @@ def api_get_contacts(request):
...
     partner_id = (request.GET.get('partner_id')
                   or request.POST.get('partner_id'))
-    if partner_id:
-        contact_args = {'partner__id': partner_id}
-    else:
+    if partner_id is None:
         contact_args = {'partner__owner': company}
+    else:
+        if partner_id == '':
+            partner_id = 0
```
almathew: Is search for partner_id == 0 meaningful in some way?  
MySQL AUTO_INCREMENT should start at 1, so that would be the same as doing a query that force-returned None if I'm reading this correctly. So wouldn't it be better to just not do the query at all?

tdphillips: I suppose so. This was a dumb attempt at not wrapping everything in gratuitous `if` statements.

PR #2949, opened 2017-04-17
Pd 3164 cbo cbolist filter does not persist

aljordan82:
## What's this PR do?
    Implement Redux to store the CBO filter state.
    Use bootstrap table to display CBO list


mwarnsleyDE:
@aljordan82 I put a DO NOT MERGE on this just so we can merge the rest of what we have and have you do a pull to make sure there aren't any conflicts which I don't think there should be. As well, I am guessing if it's easier for us to use the bootstrap table that we should persist it which would mean implementing it inside of the ContactList.js
DistributionLocationList.js and RulesList.js where there are tables as well.

mwarnsleyDE:
@aljordan82 so there are going to be some merge conflicts that you'll have to fix since a lot of that stuff got merged into Quality-Control

koonsies:
Can we change "City" in the filter menu to "Distribution City"? To match "Distribution State"

gulp/src/cbo/actions/cboActions.js
```diff
@@ -266,6 +271,105 @@ export function deleteDistributionLocation(zipperedCbo) {
...
+      if (a.isCustom > b.isCustom) {
+        return 1;
+      }
+      return 0;
+    };
+
+    cbos.sort(sortFunc);
+    function searchCategoryForValue(contact) {
+      return contact.category && contact.category.some((category) => {
+        return this(category);
```
koonsies: I'm not familiar with this use of "this". What does this line do?

aljordan82: 'this' is the matcher function "categoryMatcher"

koonsies: Interesting. I admit I haven't done a ton with context binding, so using "this" like that is pretty new to me haha.

PR #2952, opened 2017-04-17
[PD-3077][PD-3073] Site analytics report part 1.

almathew:
## What's this PR do?
Introduces a site analytics report.

Some of the queries that Hal wants added to the code and "working" but not implemented, because the implemented part might be a bit too hard on MemSQL at the moment. 
At this point the queries are optimized to run as well as I can possibly get them, and adding them will be trivial to implement (they just get added to the values list in analytics/utils.py), but I want to mess around with a few different options before I resign myself to the way they work. 
So with this PR all that's included is job views, intended applies %, and reached source % for both reports. Those are the columns already present for job analytics right now. 

It looks like job analytics.

Site analytics re-uses a lot of the base for job analytics. In fact, pretty much everything should be the same other than the base filters (`view_host` rather than `buid` and a different time field).
Also, CompositeForeignKeys, like the one in analytics/models.py, are cool and you should check out how that works.
Note that the migration added doesn't actually do anything other than make Django happy.

Re-run make data to make some valid data. Check out site views and make sure what you're seeing makes sense.

Still waiting on me to figure out how to get MemSQL running the insanely complicated annotations. It might just work out of the box, it might not, but I'm not ready to just throw that in quite yet.
I kept messing up the data in myreports, so I added backwards migrations so I could go back to zero and fix everythign.


almathew:
This is ready for review, and I'm expecting this round of tests to pass.

As noted in the summary - this does NOT complete PD-3077. I'm working on completion for that ticket right now.

One weird thing for me is that intended applies % is always 0, but that was true for my data even before changes were made. So someone might want to look a bit harder at that specifically.

myreports/datasources/simple_django_orm.py
```diff
@@ -61,10 +71,31 @@ def filter_query_set(self, query_set):
...
+            # https://code.djangoproject.com/ticket/18437
+            # https://docs.djangoproject.com/en/dev/topics/db/queries/#spanning-multi-valued-relationships
+            else:
+                q.append(as_q(filter_field, field))
+
+        # Turn all of the Q objects into a single filter and apply them.
+        q = filter(None, q)
+        if q:
+            query = reduce(__and__, q)
+            query_set = query_set.filter(query)
```
almathew: This section is all query optimization. 
The previous version was chaining `.filter()` for every part of the query, which forces multiple joins  on the same two tables and ended up producing a query that wasn't actually correct.
The links should give a pretty good overview of why this was done if you're interested (and the whole thing is pretty interesting). 
So instead I deviate from the way the other data source

almathew:
I've changed my mind. Everything works great until you run it against a non-local db.
Despite having optimized the queries, something needs to be done to make this actually useable. 

Leaving this PR open because I still want feedback.

almathew:
Ok, problems related to query optimization are resolved. This is ready again.

PR #2957, opened 2017-04-18
Pd 3164 cbo cbolist filter does not persist

aljordan82:
## What's this PR do?
Fix bad merge


koonsies:
Looks like there are also some extra files committed to the repo. CboList_BACKUP, _BASE.. etc, etc.

koonsies:
It seems like the presence of the "delete" button is backwards. It's hidden for isCustom=True and shows for isCustom=False.

PR #2959, opened 2017-04-19
[PD 3190] create placeholder for contact with no distribution location

mwarnsleyDE:
## What's this PR do?
This PR adds a message into the distribution location table stating that emails can't be sent until a contact has a distribution location

screen shot 2017-04-19 at 11 49 08 am

Pull down repository. Go into the CBO app and find a contact inside of a CBO that has no distribution locations associated with it. You should see the message in the screenshot


koonsies:
Looks good! I was wondering if you guys think it would be better to have it highlighted a bit more? Perhaps with a yellow box and bold? I'm a little worried the user won't often see this notice @fezick ?

PR #2963, opened 2017-04-19
Use permissions for reporting.

darrint:
## What's this PR do?

* Removes an old hack and replaces it with checking permissions.

Not changed.

* Note changes to `update_activities` which fixed a bug.
* `ReportingTypeManager` has a more sophisticated way of getting active report types.
* The unit test feature has activities assigned to reporting types.
* Some view permissions are removed. (I'm a little unsure if I'm ok removing the one on `dynamicoverview` but I think it's ok.

Add and remove permissions to a user's role and see how their report intention choices change.

none.


myreports/migrations/0016_add_permissions.py
```diff
@@ -0,0 +1,62 @@
...
+
+    add_activity(Activity, prm, 'read partner')
+    add_activity(Activity, prm, 'read contact')
+    add_activity(Activity, prm, 'read communication record')
+    add_activity(Activity, prm, 'read tag')
+    add_activity(Activity, prm, 'read partner saved search')
+
+
+@only_on_default
+def backwards(apps, schema_editor):
```
tdphillips: Should this return early if we're running tests? It does not currently but `forwards()` does.

almathew:
@tdphillips thanks for taking care of the merge

@tejaswinigutti31 the only issue I'm seeing is because I have the migrations for site-traffic report, but that's not included in this branch. The error looks like this:
```
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/http.py", line 42, in inner
    return func(request, *args, **kwargs)
  File "/MyJobs/myreports/views.py", line 570, in filters_api
    driver = ds_json_drivers[datasource]
KeyError: u'analytics_event'
```

Everything else works as expected.

PR #2971, opened 2017-04-21
Added final requested column. [PD-3077]

almathew:
## What's this PR do?
Adds the last required field for the site analytics and job analytics report. 

screen shot 2017-04-21 at 9 16 17 am

Logic was all already there, this just enables the column and fixes the tests that look for it. 

Navigate to the site or job analytics report. The Source/ATS Apply Rate column should be present and populated.

The % in that column makes no sense because our make_data isn't random and nonsensical. Percentages are much, much lower using production data.


almathew:
There might be a slight ordering issue for the report dimensions, or this might just be me. Can someone run all the migrations and double check?

screen shot 2017-04-21 at 4 05 04 pm

PR #2976, opened 2017-04-24
Handle InvalidDatabaseError thrown by mmdb.

almathew:
## What's this PR do?
There are cases where the file in MMDB_PATH is wrong and the file needs downloaded.
In these cases, an InvaidDatabaseError is thrown. 
Catch that error and take appropriate actions.


almathew:
We will need to update Analytics too.

PR #2985, opened 2017-04-25
[PD 2896] app access activities UI

mwarnsleyDE:
## What's this PR do?
PR will hide and show things depending on the app access and activities for a current user

N/A

Pull down the repository. If you have the ability to switch the activities for certain Users or yourself you can do that and switch back and forth to see certain buttons and such showing and hiding depending on the level of access that you have


gulp/src/common/ui/UserCan.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+UserCan.propTypes = {
+  /**
+   * Children wrapped inside of the user activity permissions
+   */
+  children: PropTypes.object.isRequired,
+  /**
+   * Array of the activities that the User Currently has
+   */
+  userActivities: PropTypes.array.isRequired,
```
aljordan82: This can be pick up by connecting to redux store instead of passing from parent component.

gulp/src/common/ui/UserCan.jsx
```diff
@@ -0,0 +1,43 @@
+/*
+ * This is the component that you would wrap around elements that you either show or hide depending on the User Activities
+ */
+import React, {PropTypes, Component} from 'react';
+
+class UserCan extends Component {
+  activitiesCheck() {
+    const {userActivities, neededActivities} = this.props;
+    let activitiesPresent = true;
+    for (let i = 0; i < neededActivities.length; i++) {
```
aljordan82: use lodash _.every instead of the loop.

gulp/src/common/ui/UserCan.jsx
```diff
@@ -0,0 +1,43 @@
...
+      if (userActivities.indexOf(neededActivities[i]) === -1) {
+        activitiesPresent = false;
+      }
+    }
+    return activitiesPresent;
+  }
+  render() {
+    const {children} = this.props;
+    const childrenWithActivities = this.activitiesCheck() ? children : '';
```
aljordan82: I would return a null or false as recommended by in React documentation.

gulp/src/common/ui/UserCan.jsx
mwarnsleyDE: @aljordan82 we are moving this around in a sense and trying to put it in the common folder. It would be better to not connect it to the redux store inside of CBO based off of us trying to make it reusable

gulp/src/common/ui/UserCan.jsx
mwarnsleyDE: @aljordan82 I don't believe you can return a null or false value when it comes to rendering inside of a div inside of JSX

gulp/src/cbo/components/CboApp.js
```diff
@@ -121,9 +122,11 @@ class CboApp extends Component {
         <Link to="/" className="btn" onClick={(e) => this.updateCboAppState(e)}>
           TJD List
         </Link>
-        <Link to="/newcbo" className="btn" onClick={(e) => this.renderCboWizard(e)}>
-          New TJD
-        </Link>
+        <UserCan userActivities={activities} neededActivities={['create cbo', 'read cbo', 'update cbo']}>
```
aljordan82: activities shouldn't be needed.

gulp/src/common/ui/UserCan.jsx
```diff
@@ -0,0 +1,45 @@
...
+ */
+import React, {PropTypes, Component} from 'react';
+import {connect} from 'react-redux';
+import {difference} from 'lodash-compat';
+
+class UserCan extends Component {
+  activitiesCheck() {
+    const {userActivities, neededActivities} = this.props;
+    let activitiesPresent = true;
+    if (difference(userActivities, neededActivities).length === 0) {
```
aljordan82: I would just return (difference(userActivities, neededActivities).length === 0);

gulp/src/common/ui/UserCan.jsx
aljordan82: I see. I guess we don't need to connect this component at line 43

gulp/src/cbo/reducers/initialState.js
```diff
@@ -9,7 +9,7 @@ export default {
   categories: [],
   states: [],
   rules: [],
-  activities: [],
+  // activities: [],
```
aljordan82: I think activities should be defined.

gulp/src/common/ui/UserCan.jsx
mwarnsleyDE: @aljordan82 it's connected to the in order to access the activities being passed through to it so the component is reusable

gulp/src/common/ui/UserCan.jsx
mwarnsleyDE: @aljordan82 I'm not too familiar with lodash but I think just returning that only returns an array and not a true or false value

gulp/src/cbo/reducers/initialState.js
mwarnsleyDE: @aljordan82 I will check but I'm not sure that we would need activities inside of the initial state anymore based off of the fact that it's being populated for the user on a global scale. I will check on this however

darrint: If you take it out, take it out. Don't leave it commented.

gulp/src/common/ui/UserCan.jsx
```diff
@@ -0,0 +1,45 @@
...
+    if (difference(neededActivities, userActivities).length !== 0) {
+      activitiesPresent = false;
+    }
+    return activitiesPresent;
+  }
+  render() {
+    const {children} = this.props;
+    const childrenWithActivities = this.activitiesCheck() ? children : null;
+    return (
+      <span>
```
aljordan82: Why a span is needed?

mwarnsleyDE: @aljordan82 you either have to wrap it inside of a span or inside of a div tag when you're returning JSX. I was going to use a DIV instead but it messes up the css and the format on the page so a span sufficed with keeping things in order

mwarnsleyDE: It would essentially return this.props.children and you can't just return that without an enclosing div or span tag. If the child is lets say an <a><span></span></a> or something alike then it would have to be enclosed in a div or a span since it's adjacent JSX tags.

aljordan82: what about this
```
    render(){
        return {this.activitiesCheck() ? React.Children.only(this.props.children) : null}
    }
```

mwarnsleyDE: @aljordan82 only issue with that is not you are only allowing for the component to have 1 child element when it comes to this.props.children. If there are more than one child component when using it then it will fail

aljordan82: and if it theres is more than one child, then it should be wrapped in the parent component. 
This component shouldn't introduce any extra elements.

aljordan82: The component should fail if it had more than one element in props.children, it should be up to the parent to provide a single child

koonsies: I'm inclined to agree with Aladeen on this one. Similar to how `render` requires that only one jsx object be returned, I'd say that putting the onus on the component using the UserCan component to wrap their JSX returns makes sense. Prevents "span spam".

gulp/src/common/ui/UserCan.jsx
```diff
@@ -0,0 +1,41 @@
...
+    const {userActivities, neededActivities} = this.props;
+    return difference(neededActivities, userActivities).length !== 0;
+  }
+  render() {
+    const {children} = this.props;
+    const canRender = this.activitiesCheck();
+    if (!canRender) {
+      return null;
+    }
+    return this.activitiesCheck() ? React.Children.only(children) : null;
```
koonsies: This seems redundant. The canRender check returns null if false, then we do a conditional to check the same thing again.

Also, what happens when more than one child is supplied to this object? If we don't raise an error explicitly, I feel like it may give an unhelpful error like "render must return one object" or something. We want it to say specifically that multiple objects must be wrapped.

koonsies:
This doesn't seem to work currently. With just the read_tjd permission, I see all update buttons and such.

gulp/src/common/ui/UserCan.jsx
aljordan82: canRender was committed by mistake.
The error message when trying to pass an array of children is 
```
Warning: Failed propType: Invalid prop `children` of type `array` supplied to `UserCan`, expected `object`. Check the render method of `Connect(UserCan)`.
```

aljordan82:
I added a custom warning for this component. feel free to edit the warning.

gulp/src/common/ui/UserCan.jsx
```diff
@@ -21,7 +21,20 @@ UserCan.propTypes = {
   /**
    * Children wrapped inside of the user activity permissions
    */
-  children: PropTypes.object.isRequired,
+  children: (props, propName, componentName) => {
+    if (propName === 'children' && componentName === 'UserCan' && Array.isArray(props.children)) {
```
aljordan82: I need to fix this, isArray is wrong.

gulp/src/cbo/components/cbo/CboList.js
```diff
@@ -43,7 +44,9 @@ class CboList extends Component {
       <div className="cbo-table-container" id="cbo-list-table">
         <div className="wrapper-header cbo-table-wrapper">
           <h2>TJD(s)</h2>
-          <span id="add-new-cbo" className="table-header-link" onClick={() => this.addNewCbo()}>New TJD</span>
+          <UserCan neededActivities={['create cbo']}>
```
koonsies: This activity is incorrect. It should be "create tjd" .. The activities are "create tjd", "update tjd", "create tjd", "read tjd"

koonsies: Furthermore, this doesn't seem to work.. This case should automatically fail since "create cbo" isn't an activity, but the button shows up for me.

koonsies:
We should hide or disable the "update" buttons on the various forms when they do not have the update permission. Right now, they show, but an "error: forbidden" is raised when they are clicked.

koonsies:
@fezick @JLMcLaugh   Do we have a permission we'd like to wrap about setting the optin/optout status for the entire app? We do force the user to have all permissions? Or perhaps make one for the overalll optin/opt out?

PR #2986, opened 2017-04-25
PD-3195 adding job view estimate

tejaswinigutti31:
## What's this PR do?
Revert changes of PD-3073 to have Job Views display Sum of Job Views rather than count of Job views.
Have also changes tests for data_source job_views and analytics api
Check the data source job_views for changes made to code


myreports/datasources/job_views.py
```diff
@@ -54,9 +53,11 @@ def annotate_query_set(query_set, values=None):
 
         """
         values = [] or values
-
+        if 'job_views_estimate' in values:
```
almathew: You're going to hate me, but this should actually be `job_views_exact`.

PR #2987, opened 2017-04-25
Pd 3210 rename reports

tejaswinigutti31:
## What's this PR do?
Rename reports appropriately
A new migration 0018 in myreports/migrations


almathew:
I want to make a merge migration for this before it's merged, since both this PR and the one for turning site analytics on have a migration 0018.

PR #2991, opened 2017-04-26
[PD-3213] Secure job views and analytics events datasources.

almathew:
## What's this PR do?
Previously attempts to access site-traffic and site-analytics reports for companies that didn't have any associated seo.JobViewSites would allow those companies to query all the data instead.

Now, any attempt to create a report for JobViews or AnalyticsEvent without at least a site or a buid passed in will return an empty results set instead.

Unless you added JobViewSites in your test environment, the site-traffic and site-analytics reports should now be blank for you. However, the job-analytics report probably still has results.


almathew:
This is waiting on PR #2995 which should change the tests in a way that both causes merge conflicts and makes the tests on this no longer fail.

almathew:
I'd rather wait for #2997 too, since it will be easier to resolve conflicts from that branch on this one than it will to do the reverse.

PR #2979, opened 2017-04-24
Replace CBO with TJD

aljordan82:
## What's this PR do?
Use the new app name "Targeted Job Distribution" and "TJD"
Move link to the Employers menu
## Screenshots (if appropriate)
screen shot 2017-04-24 at 4 52 53 pm

PR #2993, opened 2017-04-26
[PD-3205] Enforce Activities on CBO API

koonsies:
## What's this PR do?

Gets the CBO activities in line with activities from other apps and enforces their use in the API calls.

Manage Users
* Click 'Roles'
* Create a new role
* Assign all of the manage users activities to that role.
* Assign your user account to that role
* Click 'Users' on the side navigation.
* Edit your User account
* Remove "admin" from your roles
* Add/Remove CBO permissions to test what happens when they aren't present.

When finished, edit your user and readd the Admin role.

Tediously.

None that I want to, no.


myjobs/models.py
```diff
@@ -1363,6 +1363,12 @@ def end(sender, impersonator, impersonating, request, **kwargs):
 
 
 ACTIVITIES = {
+    'CBO' : [
+        ("create cbo", "Create data for CBO app, including cbo, contact, etc"),
```
aljordan82: Is the "CBO" in the description visible to users?

koonsies: Yeah, it's visible in Manage Users. We haven't settled on a replacement for CBO, officially, and when we do we'll have to do a big refactor to fix this, the urls, and other places it says "CBO".

tdphillips: I think we may have had the replacement chosen for us already.

https://github.com/DirectEmployers/MyJobs/pull/2979

koonsies: Ahhh. Didn't realize the refactor had gone through already. Fair enough.

PR #2996, opened 2017-04-27
[PD-3225] page titles displaying line breaks

mwarnsleyDE:
## What's this PR do?
This PR fixes the issue with the titles on the sidebar wrapping and also in the table as well in Analytics

https://directemployersfoundation.atlassian.net/browse/PD-3225

Start inside of site-analytics

Pull down repository. Go into site analytics and view the sidebar along with the table values

https://directemployersfoundation.atlassian.net/browse/PD-3225


gulp/src/sass/analytics.scss
```diff
@@ -1737,6 +1745,13 @@ overflow: visible;
        .tab-content-container {
          margin: -49px 0 0 0;
        }
+       .rwd-table {
```
aljordan82: What about just changing the white-space on higher screens? we would want it to wrap in smaller screens
 (min-width: 480px) #analytics_content .rwd-table th:not(:nth-child(1)) {
    white-space: nowrap;

mwarnsleyDE: @aljordan82 I edited that we will see when it's on QC what Jason thinks if he still wants it like that and everything but I do agree with you on that. I was going to do that first just thought he wanted it wrapping on bigger screens

PR #3000, opened 2017-04-27
Add rs parameter to redirects for use in job view processing [PD-3227]

tdphillips:
## What's this PR do?
Adds a new parameter to redirect log entries
Start with redirect.views.home. If you're super determined and thread your way through it, you should find that rs is only set in the paths that lead to the job source.
We don't really test analytics in MyJobs sadly, so there's not much to test.


darrint:
Seems to work pretty well. Based on some discussion we had we may want to update the debug page to show what would get logged to analytics so that we can see `rs`.

We may also want to unset the `X-JSON-Header` when in the debug page so we can't use the debug page to put in a bunch of false `rs: 1`'s.

tdphillips:
@darrint, is this too ugly?

2017-04-28-082341_993x128_scrot

tdphillips:
The debug page is ours so we should be able to do basically anything we want with it.

PR #3001, opened 2017-04-27
PD-3205 listing user activities

tejaswinigutti31:
## What's this PR do?
Lists all the user activities
myjobs/views.py has a new function to list the user activities


gulp/src/common/myjobs-api.js
```diff
@@ -94,6 +94,10 @@ export class MyJobsApi {
 
     return this.parseJSON(await this.checkStatus(response));
   }
+
```
aljordan82: I would move this to api.js

PR #3010, opened 2017-05-01
Pd 3188 cbo form improvements

aljordan82:
## What's this PR do?
Remove Category tags (Pills)
Consolidate the focus effect on form elements.
Switch the order of the Rules form input fields. The Type field is now first.


tdphillips:
I was lax with backporting one of my hotfixes all the way to quality-control, which is why some of my commits are showing up. I would be inclined to let this branch proceed, with the side effect of backporting my changes for me.

gulp/src/sass/cbo/cbo_main_styles.scss
```diff
@@ -42,20 +42,14 @@ $large-desktops: 1200px;
 @mixin borderRadius($radius) {
   -moz-border-radius: $radius;
```
aljordan82: most vendor prefixes are no longer needed, I'll clean it up later.

PR #3017, opened 2017-05-03
Disable AuthorizeSauce's debug setting when in production [PD-3235]

JLMcLaugh:
## What's this PR do?
Disables the debug setting for credit card authorization when in production. Credit card transactions are being rejected because we're posting to authorize.net's sandbox api.
The best we can do for now in staging is verify we haven't broken the sandbox api requests. We can create a payed product for apply.directemployers.org in the staging environment and test that we're able to purchase it from another company.


postajob/payment.py
```diff
@@ -1,9 +1,12 @@
 from authorize.client import AuthorizeClient, AuthorizeCreditCard
 from authorize.data import Address, CreditCard
 
+from django.conf import settings
 
 def get_client(authorize_id, authorize_key):
-    return AuthorizeClient(authorize_id, authorize_key)
+    # Use the sandbox api if we're not in production
+    debug = (ENVIRONMENT != "Production")
```
tdphillips: `settings.ENVIRONMENT`?

JLMcLaugh: Ugh, thanks.

postajob/payment.py
```diff
@@ -1,9 +1,12 @@
 from authorize.client import AuthorizeClient, AuthorizeCreditCard
 from authorize.data import Address, CreditCard
 
+from django.conf import settings
 
 def get_client(authorize_id, authorize_key):
-    return AuthorizeClient(authorize_id, authorize_key)
+    # Use the sandbox api if we're not in production
+    debug = (settings.ENVIRONMENT != "Production")
```
darrint: Having a single setting that has (potentially) more than one effect can be a problem. Can we make a setting specifically for this?

tdphillips: Why would it be a problem?

JLMcLaugh: That makes one vote from @almathew for using DEBUG and one vote for using something more specific. I'm inclined to call this a happy medium unless someone really objects.

PR #3018, opened 2017-05-03
branch test tasks

koonsies:
## What's this PR do?

Adds a new gulp task, testjson, and sets it as the default when using gulp build (or just running gulp directly). This allows gulp test and gulp watch to continue to print out to the console for local testing.


gulp/gulpfile.js
```diff
@@ -45,6 +45,14 @@ gulp.task('bundle', function(callback) {
 });
 
 gulp.task('test', function() {
+  // used for local tests, prints to console
+  var testConfig = require('./jasmine.json');
+  return gulp.src(['./src/**/spec/*.js'])
+    .pipe(jasmine(testConfig));
+});
+
+gulp.task('testjson', function() {
```
aljordan82: Why is this needed?

aljordan82:
why not have two reporters?

`var Reporter = require('jasmine-terminal-reporter');`

``` gulp.task('test', function() {
  var testConfig = require('./jasmine.json');
  testConfig['reporter'] = [new reporters.JUnitXmlReporter({
    savePath: '../reports/',
    filePrefix: 'gulp_tests'
  }),
  new Reporter({
    isVerbose: true,
    showColors: true,
    includeStackTrace: true,
  })]

  return gulp.src(['./src/**/spec/*.js'])
    .pipe(jasmine(testConfig));
}); 
```

koonsies:
I wasn't aware you could have two reporters. I assumed that since the attribute was named 'reporter' that only one reporter could be active at once. You've confirmed that works?

PR #3025, opened 2017-05-04
Master -> Staging

almathew:


almathew:
I made some changes directly to master that need pulled back to staging

almathew:
I hate you, github.

PR #3026, opened 2017-05-04
Fix issues that have come up with job expiration. 

almathew:
## What does this PR do?
- Adjusts crontab to be correct
- Temporarily fixes an encoding issue while we research a better fix in the long term.


import_jobs/expire.py
```diff
@@ -70,6 +70,14 @@ def notify_expired():
     }))
     job_urls = ExpiredJobURL.objects.filter(claimed_by=unique_identifier)
 
+    # Some job urls have special characters in them. The python ssl library
+    # doesn't like those characters.
+    # This means we are technically not expiring the correct url, but
+    # it's close enough for now (until we've had a chance to sort out a
+    # better solution).
+    for url in job_urls:
+        url.url = url.url.encode('ascii', 'ignore')
```
mklauber: What's the long term fix for this?

almathew: Third party will need to tell us what they want.

PR #3034, opened 2017-05-05
Persist CBO table state

aljordan82:


aljordan82:
This PR is to persist the tables states, Page index and Pagination size.
I still need to implement the change for the Contacts and Distribution Locations tables.

gulp/src/cbo/components/cbo/ManageCboPage.js
```diff
@@ -32,6 +32,7 @@ export class ManageCboPage extends Component {
     this.setState({zipperedCbo: new CboZipper(this.state.zipperedCbo)});
   }
 
+
```
koonsies: looks like an unintentional newline here.

aljordan82:
@koonsies I could store the pagination values directly on 'this' or 'this.state' . I don't see the difference, we are not going to set those properties from an external resource(i.e. no need to call setState/and render)
I could use the redux store, but I don't see the advantage for this particular case.

PR #3035, opened 2017-05-08
Pd 3221 cbo persist pagination state

aljordan82:
## What's this PR do?
Persist table page index and rows count


gulp/src/cbo/components/CboApp.js
```diff
@@ -28,6 +28,17 @@ class CboApp extends Component {
       statusChecked: nextProps.cboOptStatus,
     });
   }
+  getAndSetTableState(table, page, paginationSize) {
+    if (table) {
+      this[table + 'Page'] = page || this[table + 'Page'];
```
koonsies: What decision led to these attributes being placed directly on the top level CboApp object, rather than in CboApp.state?

aljordan82: I didn't see the difference/advantage for on the CboApp.state

koonsies: The difference is that changing a value on this.<variableName> does not trigger a re-render of the page. Changing the value on this.state.<variableName> would force a rerender. As it stands, this doesn't seem to cause a problem, but if you introduced a flow that had this function called without something else causing a rerender, you could wind up with changes not displaying.

As long as the lack of an explicit re-render is understood and accounted for, this works.

PR #3036, opened 2017-05-08
[PD-3249] date range

mwarnsleyDE:
## What's this PR do?
Changes the Date Range Selection in Analytics to not go back any farther than 5/4/2017

N/A

Start inside any of the analytics applications

Pull down the repository. Go into Analytics applications and select a date from the date range calendar. You will see that it automatically wont' let you go back any farther than May 4th and if one of the quick selections is selected it still goes from May 4th as the start

Most of this is hard coded right now and needs to be for DEAM until we figure out how to make it dynamic


aljordan82:
Why are the dates hardcoded?

mwarnsleyDE:
@aljordan82 Any background context you want to provide? In there it tells why we had to hardcode it for DEAM. It's going to be changed to be dynamic once DEAM is over with anyways so it's getting overhauled

PR #3052, opened 2017-05-15
PD-3239 cbo consistency check

tejaswinigutti31:
## What's this PR do?
A script to check if there is a valid Customer ID for a user with company role
management sub-directory in CBO
Run the command cbo_consistency_check


cbo/management/commands/cbo_consistency_check.py
```diff
@@ -0,0 +1,27 @@
...
+from cbo.api import get_cbo_customer_id
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        users = User.objects.filter(roles__isnull=False).distinct()
+        cust_id = []
+        company_wo_custid = []
+        company_list = []
+        for user in users:
+            company_list = company_list + [r.company for r in user.roles.all()]
```
tdphillips: If `company_list` was a set, you could do `company_list.update([r.company for r in user.roles.all()])` and do away with the conversion to a set and back.

cbo/management/commands/cbo_consistency_check.py
```diff
@@ -0,0 +1,27 @@
...
+        company_list = list(set(company_list))
+        for company in company_list:
```
JLMcLaugh: You can iterate directly over sets. There's no need to have it start and end as a list.

cbo/management/commands/cbo_consistency_check.py
```diff
@@ -0,0 +1,27 @@
...
+            try:
+                cust_id = get_cbo_customer_id(company)
```
JLMcLaugh: Are you expecting exceptions here? A catch-all exception is usually a bad idea. If you put this in because you're seeing exceptions, we should look into that and handle it in the helper function.

cbo/management/commands/cbo_consistency_check.py
```diff
@@ -0,0 +1,27 @@
...
+            company_list = company_list + [r.company for r in user.roles.all()]
+        company_list = set(company_list)
+        for company in company_list:
+            try:
+                cust_id = get_cbo_customer_id(company)
+            except (AttributeError):
+                None
+            if not cust_id:
+                company_wo_custid.append(company)
+        if company_wo_custid: return company_wo_custid
```
almathew: Would we be better off doing something like a `print` here?

cbo/management/commands/cbo_consistency_check.py
almathew: I might be wrong, but something like `Company.objects.filter(role__user__isnull=False).distinct()` should work too? And eliminate the need for a loop or set.

cbo/management/commands/cbo_consistency_check.py
```diff
@@ -0,0 +1,27 @@
...
+        cust_id = []
+        company_wo_custid = []
+        company_list = []
+        for user in users:
+            company_list = company_list + [r.company for r in user.roles.all()]
+        company_list = set(company_list)
+        for company in company_list:
+            try:
+                cust_id = get_cbo_customer_id(company)
+            except (AttributeError):
```
JLMcLaugh: What is this catching here? If you're seeing an AttributeError, I'd like to know what's causing it. It may be from something we're doing wrong when we build company_list.

PR #3053, opened 2017-05-15
Upgrade js stuff 3

darrint:
## What's this PR do?

Updates us to node 7 and replaces npm with yarn. Webpack dev server appears to work.

This is all Dockerfile changes.

We should make sure all existing apps build and appear to work. We don't expect subtle breakage.

Run `dk rebuilddev`.

From now on we will not use `npm install`. Instead we will run `yarn`.


tdphillips:
If this is what we're doing going forward, I'd like to see our documentation updated.

darrint:
@tdphillips Thanks. I updated `README.md`.

PR #3057, opened 2017-05-15
Duplicate params in partner saved search

tejaswinigutti31:
## What's this PR do?
Ran into an scenario in which previous fix to remove duplicate parameters from partner saved search urls doesn't work. Fixed it
add_extra_params in mypartners/helpers.py


mypartners/helpers.py
```diff
@@ -65,7 +65,11 @@ def add_extra_params(url, extra_urls):
         new_queries['z'] = '1'
     if set(query.keys()) != set(new_queries.keys()):
         query.update(new_queries)
-    parts[4] = query.urlencode()
+        query_list = QueryDict('', mutable=True)
+        for key, value in query.items():
+            if value not in query_list.values():
```
tdphillips: Should this be `if key not in query_list.keys()`? I thought we were deduplicating keys, not values.

tejaswinigutti31: I've been noticing duplicate values for a key and therefore used the above condition

PR #3060, opened 2017-05-17
[PD-3261] export current tab

mwarnsleyDE:
## What's this PR do?
PR fixes the exporting for the current tab that the user is on inside of analytics

N/A

Start inside of any of the analytics reports

Pull down repository. Go into any one of the analytics and create more tabs by drilling down. Switch the tabs and export the data which should be the data for the current tab


gulp/src/analytics/actions/header-actions.js
```diff
@@ -4,12 +4,18 @@ export function doExportTab(presentationType) {
...
       const link = document.createElement('a');
+      const tabs = getState().pageLoadData.navigation;
       const currentReport = getState().pageLoadData.activeReport;
       const storedDates = {
         startDate: getState().pageLoadData.globalStartDate,
         endDate: getState().pageLoadData.globalEndDate,
       };
-      const storedFilters = getState().pageLoadData.activeFilters;
+      let storedFilters = [];
+      for (let f = 0; f < tabs.length; f++) {
```
aljordan82: It is not clear what this loop does.

mwarnsleyDE: @aljordan82 this loops through the tabs themselves and finds the active one that you are currently on and sets it. Before we were just doing the last tab that was created but now it does the currently active tab that you are on

aljordan82: Ideally, I would use 'find'
`var storedFilters = tabs.find((tab) => tab.active).activeFilters || []`

tdphillips:
Have the requested changes been made? I suspect so but I'd rather not merge without being certain.

@aljordan82, @mwarnsleyDE

mwarnsleyDE:
@tdphillips @aljordan82 I've made the changes requested

PR #3076, opened 2017-05-24
Remove race conditions, fix datetime format (again)

almathew:
## What's this PR do?
Updates the datetime format to a format that I've now had multiple people sign off on.
Removes a race condition where count() returns a number greater than zero, but things get claimed while the count is being done.


myreports/migrations/0016_add_permissions.py
```diff
@@ -23,6 +25,8 @@ def forwards(apps, schema_editor):
     ReportingType = apps.get_model('myreports', 'ReportingType')
     Activity = apps.get_model('myjobs', 'Activity')
 
+    update_activities()
```
almathew: I added this because there's a frustrating chance that the Apps/Activities/etc don't exist when this migration is run (from a fresh db instance).

PR #3078, opened 2017-05-26
Add criteria sheet to Excel exports [PD-3271]

tdphillips:
## What's this PR do?
Adds a new worksheet to Excel exports
2017-05-26-083033_324x173_scrot
Most of the relevant code is in `/myreports/presentation/xlsx.py` It's also magic and perfect.
Find a report. Export it. Verify there's a second worksheet.


myreports/presentation/xlsx.py
```diff
@@ -9,13 +11,54 @@ class Xlsx(Presentation):
 
     filename_extension = 'xlsx'
 
-    def write_presentation(self, values, records, output):
+    def write_presentation(self, values, records, output, report=None,
+                           report_configuration=None):
```
almathew: I have an incredibly stupid question. Above you made the function definition `write_presentation(self, values, records, output, *args)`, but here you're treating them as kwargs. Is there any reason for the inconsistency between args and kwargs? 
It shouldn't really matter either way since `presentation.write_presentation(values, sorted_records, output, report, report_configuration)` is correct. Mostly just curious.

tdphillips: The other presentation types don't care about these extra arguments. Due to our generic code, they had to accept them anyway. Outside of that, I didn't care enough.

almathew:
Merge conflicts :tada:

tdphillips:
Resolved. For the record, this one should have been merged first.

PR #3081, opened 2017-05-29
Add jobs

almathew:
I'm not sure my default query -> facets conversion is 100% correct, so I'd appreciate someone digging into that more to confirm what I've written (@JLMcLaugh)
@mklauber This will require us to do some deployment changes

## What's this PR do?
Adds new jobs based on whether or not they actually show up on a site.
While I'm making changes, also adds the zipcode to the job detail schema.

`task_etl_to_solr` and `task_update_solr` both eventually end up queuing jobs for addition via `import_jobs.init.determine_and_handle_adds`.
From there, most of the logic is handled in `import_jobs/update.py`
There's a lot of duplication for how we do default queries on sites done in `import_jobs.update.query_sites_for_guid`. This section could use quite a bit of extra attention. As noted above, I'm not sure I accounted for everything that can possibly be done. Also, this function runs the risk of being pretty horrible to Solr during updates.

For adding new jobs
- Make sure you have at least one seo site with `track_job_changes=True`
- Clear out your solr instance if it's not already clear.
- Import jobs via `task_etl_to_solr`. If you don't know how, ask me and I'll show you. There should be some `NewJobURL` objects created.
- Import jobs via `task_update_solr`. If you don't know how, ask me and I'll show you. There should be even more `NewJobURL` objects created.

For the zip code changes:
- Confirm that, if available, zip codes are showing up in view source
- Ask me directly for better clarification/confirmation/testing instructions


templates/job_detail.html
```diff
@@ -55,6 +55,9 @@
                           <span itemprop="addressCountry">{{ the_job.country }}</span>
                       {% endif %}
                   {% endif %}
+                  {% if the_job.zipcode %}
+                    <span style="display:none;" itemprop="postalCode">{{ the_job.zipcode }}</span>
```
almathew: @koonsies right now this styling is inlined because I couldn't find anywhere else to pull it from. Does something already exist that just has `display: none` that I can reuse.

import_jobs/expire.py
```diff
@@ -40,7 +40,8 @@ def queue_for_expiration(expired_jobs):
         for job in expired_jobs
     ]
     urls = [joblist_url({}, job) for job in expired_jobs]
-    [schedule_to_expire.delay("https://www.my.jobs%s" % url) for url in urls]
+    for site in SeoSite.objects.filter(track_job_changes=True):
```
almathew: @JLMcLaugh I made a choice here to have expire follow the same logic as add. I'm not even sure I agree with that choice, so I figured I'd ask you what you thought.

templates/job_detail.html
koonsies: I was thinking "class=hidden" as well, but I hear you and Aladeen talking about it over there. If that is not working, I don't see any issues with an inline style.

PR #3094, opened 2017-06-01
PD-3315 CBO-Customer ID validation

tejaswinigutti31:
## What's this PR do?
Adds customer Id validation to all API calls


cbo/api.py
```diff
@@ -46,54 +46,54 @@ def delete_required_data(send_url, _):
...
-              'insert_cbo': {'method': 'POST', 'URL': 'cbo/addcbo',
+              'insert_cbo': {'method': 'POST', 'URL': 'cbo/add',
                              'permissions': ['read tjd', 'create tjd']
                              },
-              'update_cbo': {'method': 'PUT', 'URL': 'cbo',
+              'update_cbo': {'method': 'POST', 'URL': 'cbo/update',
                              'permissions': ['read tjd', 'update tjd']
                              },
-              'delete_cbo': {'method': 'DELETE', 'URL': 'cbo',
+              'delete_cbo': {'method': 'POST', 'URL': 'cbo/delete',
```
aljordan82: Why the DELETE requests were changed to a POST?

koonsies:
The CA API is changing due to customer ID validation being added. We can talk more at the office here in a few.

cbo/api.py
```diff
@@ -9,7 +9,7 @@
 from django.views.defaults import bad_request
 from django.http import JsonResponse
 
-URL_ROOT = 'http://reports.directemployers.com/cboadmin/'
+URL_ROOT = 'http://reports.directemployers.com/qc/cboadmin/'
```
koonsies: @almathew is there a preferred way to do environment checking/implement environment specific variables? It's been a little while since I've done that.

cbo/api.py
```diff
@@ -46,54 +46,54 @@ def delete_required_data(send_url, _):
 
 @requires('read tjd')
 def proxy_request(request, function_name, query_arg=None):
```
koonsies: You can remove query_arg=None here

gulp/src/cbo/api.js
```diff
@@ -28,110 +28,143 @@ export default class Api {
...
+      cboId: cboId,
+    };
+    const response = await this.postToCboApi('/delete_cbo/', cboObj);
     return response;
   }
 
   async getContact(contact) {
     const contactId = contact.contactId;
-    const response = await this.getFromCboApi('/get_contact/' + contactId);
+    const response = await this.postToCboApi('/get_contact/', contactId);
```
koonsies: We don't actually use this function, but I don't think this would work if we did. We need to create an object like "cboObj" above in deleteCbo.

gulp/src/cbo/api.js
```diff
@@ -142,12 +175,12 @@ export default class Api {
   }
 
   async saveCategories(cbo) {
```
koonsies: I believe we can delete these functions. I may have said otherwise before, but we don't use them.. so.. might as well delete them rather than maintain them. Same goes for the ruleTypes deletes

default_settings.py
```diff
@@ -800,3 +804,4 @@
 }
 UPDATE_ENDPOINT = "https://www.my.jobs"
 UPDATE_SCOPES = []
+>>>>>>> 7353239a9d5a587ed0b0f822c011a2f93ef32b46
```
almathew: Left over merge conflict stuff

PR #3095, opened 2017-06-01
[PD-3272] Move DB reporting records to code

darrint:
If someone has time and can start looking over this code I'd appreciate it.

Tasks I have yet to do:

- [x] Data migration for existing dynamic reports.
- [ ] ~~Remove old tables.~~ Don't do this.
- [ ] Find and fix more regressions.
- [x] Rearrange the fields originally requested in the PR.
- [ ] `split(..., 3)` should have been `split(..., 2)`
- [ ] Fix JS unit tests.

## What's this PR do?

Moves the db records defining reports to code in the hopes that this will be easier to manage in the future.

Should be no changes.

There's a lot. But a few key areas are:

* There will be a data migration which affects the 1300-ish existing `DynamicReports`.
* `schema.py` has a `resolve` function that is interesting.
  * The `resolve` function is often called more than once in a single request. Is this ok?

* Go deep looking at Reporting and Analytics. There may be regressions.

na


vigowei:
Changes in this PR were pulled into [PR#3110](https://github.com/DirectEmployers/MyJobs/pull/3110), we did some basic work on top of these changes, and added a small fix to make reporting work with this new setup. The sooner we review and merge this PR and/or PR#3110 the sooner we can avoid doing migrations for updating reporting types and such.  Thoughts? @almathew @tdphillips @darrint

PR #3099, opened 2017-06-05
[PD-3275] Editing the table presentation

mwarnsleyDE:
## What's this PR do?
Edits the table presentation

N/A

Start inside of any of the analytics reports

Pull Down repository. Check the bottom of the table presentation at all of the analytic reports to make sure it isn't overflowing

Needed to fix the presentation at the bottom of site traffic report


gulp/src/analytics/components/Table/Table.js
```diff
@@ -11,13 +11,20 @@ class Table extends Component {
...
+
     if (currentPage === 1) {
       showingStart = 1;
     } else if (rowsThisPage !== pageSize) {
       showingStart = (totalRows - rowsThisPage) + 1;
     } else {
       showingStart = (pageSize * currentPage) - (pageSize - 1);
     }
+    // Loops through the amount of columns that there are and subtracts 2 from it so for the NBSP <td>'s that need to be created'
+    for (let i = 0; i < columnAmounts - 2; i++) {
```
aljordan82: I think [].fill is better for this case.
```
var arr = []
arr.length = 3; // any value
arr.fill(<td>&nbsp;</td>)
```

mwarnsleyDE: @aljordan82 is there a reason behind this? Also with fill is there a way to grab the specific index of the element because that has to go in as the key in JSX.

mwarnsleyDE: @aljordan82 I'm almost positive as well that fill is not supported in IE11 or earlier I think

aljordan82: Then this wont work if the index is needed, so the is fine as is.
We have the polyfill for IE11

PR #3102, opened 2017-06-05
PD-3239 cbo check

tejaswinigutti31:
## What's this PR do?
Fixed errors on cbo consistency check management command


cbo/management/commands/cbo_consistency_check.py
```diff
@@ -20,8 +20,12 @@ def handle(self, *args, **options):
                 cust_id = get_cbo_customer_id(company)
             except (AttributeError):
                 None
-            if not cust_id:
+            if cust_id is None:
```
JLMcLaugh: This should be verifying that you're getting an integer back.

cbo/management/commands/cbo_consistency_check.py
```diff
@@ -20,8 +20,12 @@ def handle(self, *args, **options):
                 cust_id = get_cbo_customer_id(company)
             except (AttributeError):
                 None
```
JLMcLaugh: 'pass' would be more common here, but we should be outputting errors anyway.

koonsies:
How is this doing? @JLMcLaugh are your concerns addressed?

PR #3107, opened 2017-06-06
[PD-3280] Upgrade React Router 4

koonsies:
## What's this PR do?

Ugh..

Well... The intent was to simply upgrade react router to the newest version. This is the last of our third party libraries that we needed to bring in line to be "up to date" in React.

The issue with this upgrade is that react router changed significantly from 1.0 -> 3.0, and again from 3.0 -> 4.0 .. To make this a bit more complicated, this upgrade touches all of our apps and it's not possible to upgrade them one-at-a-time. So.. You get a nasty PR full of all types of changes. It's worth noting that some apps also didn't use Redux (looking at you, manage users) properly, so I corrected a bit of that along the way.

With yarn. This PR will involve a lot of changes in your yarn.lock file, so be sure to run 
`dkg yarn`

Afterward, close out of webpack (if it is running) and relaunch it.

This PR will require a lot of testing. Please test any/all apps you feel comfortable testing and try to identify any bugs. We should be thorough here. Routing is the backbone to many of our apps, so these changes have potential to break a lot of stuff. 

When you review, please post a comment detailing WHICH apps you tested. That way, reviewers who come afterward can identify anything that is under-tested


analytics/management/commands/make_data.py
```diff
@@ -108,8 +108,10 @@ def handle(self, *args, **options):
                 browser_version=browser_version,
                 os_family=os
             ))
-        AnalyticsEvent.objects.bulk_create(analytics_events)
-        JobView.objects.bulk_create(job_views)
+        for analytics_event in analytics_events:
```
koonsies: I'm not sure if this should be committed. I had to add this bit of code (at the behest of @darrint ) to solve an issue where I couldn't run make_data without an exception. I know @mwarnsleyDE had the same issue at some point. Should I leave this in until we have a better workaround?

gulp/package.json
```diff
@@ -45,7 +45,6 @@
     "gulp": "^3.9.0",
     "gulp-jasmine": "^2.2.1",
     "gulp-util": "^3.0.7",
-    "history": "^1.17.0",
```
koonsies: implicit requirement now

gulp/src/cbo/components/CboApp.js
```diff
@@ -57,98 +66,70 @@ class CboApp extends Component {
     this.setState({
       modalShown: false,
     });
-    history.pushState(null, router);
+    history.push(router);
     dispatch(doResetApp(targetLocation));
   }
-  updateCboAppState(e) {
```
koonsies: Sidebar moved to separate component

analytics/management/commands/make_data.py
almathew: I'm not against it if it doesn't take incredibly long to run, although use `.save()` instead of `bulk_create` then.

gulp/src/reporting/components/DynamicReportApp.jsx
```diff
@@ -16,25 +19,27 @@ import {
 
 class DynamicReportApp extends Component {
   componentDidMount() {
-    const {history} = this.props;
-    this.unsubscribeToHistory = history.listen(
-      (...args) => this.handleNewLocation(...args));
+    const {history, dispatch} = this.props;
+    dispatch(markPageLoadingAction(false));
   }
+  //   this.unsubscribeToHistory = history.listen(
```
aljordan82: Remove the commented out code in this file.

gulp/src/cbo/api.js
```diff
@@ -28,110 +28,137 @@ export default class Api {
...
     const childlessContact = omit(zipperedCbo.contact, ['rules']);
-    const response = await this.postToCboApi('/update_contact/' + childlessContact.contactId,
-        childlessContact);
+    const response = await this.postToCboApi('/update_contact/', childlessContact);
     return response;
   }
 
   async deleteContact(zipperedCbo) {
-    const response = await this.deleteFromCboApi('/delete_contact/' + zipperedCbo.contact.contactId);
+    const contactObj = {
```
aljordan82: contactObj is not used

PR #3112, opened 2017-06-08
PD-3328 Analytics report order as in the interface

tejaswinigutti31:
## What's this PR do?
The exported data now has the order same as the one in interface.
It is in sorted order based on  'Job Views'  or 'Page Views' depending on the report selected.
Exporting data in either .xlsx or .csv format should reconcile with the order on the interface


tdphillips:
@vigowei, do you think that should be a part of this ticket or was that a suggestion for future work?

vigowei:
probably future work, I think it deserves its own ticket. @tdphillips

PR #3113, opened 2017-06-08
Pd 3342 analytics closing timeline chart sends ajax

aljordan82:
## What's this PR do?
Implement shouldCOmponentUpdate
Update CSS


gulp/src/sass/analytics.scss
```diff
@@ -40,6 +40,15 @@ footer {
   display: none;
 }
 
+/* Bootstrap overwrites*/
```
mwarnsleyDE: @aljordan82 we should find an alternative to overwriting bootstrap. Overwriting them could cause some problems in the future if we needed to use them somewhere else in the files.

gulp/src/analytics/components/Charts/ChartContainer.js
```diff
@@ -13,66 +13,122 @@ import {doGetSelectedFilterData} from '../../actions/table-filter-actions';
...
+          'day': removeLeadingZero,
+        };
+      }
+      return {
+        'Job Views': month['Job Views'],
+        'month': removeLeadingZero,
+      };
+    });
+  }
   constructor(props, context) {
```
mwarnsleyDE: @aljordan82 we should move this to the top of the class just to keep the structure the same as the rest of the components.

aljordan82: @mwarnsleyDE , the linter marked it as an error, static vars/methods needs to be on top.

mwarnsleyDE: @aljordan82 okay gotcha makes sense then we can leave this one as is

PR #3114, opened 2017-06-08
New shared settings and general improvements.

almathew:
## What's this PR do?
A bunch of things. A lot more than it should.
- Adds settings.py to the repository. This replaces the dev_settings.py file most people are using. If you want custom stuff for yourself, they now go in local_settings.py. But there should always be enough in settings.py to run any dev environment. 
- Adds a command, `dkd`. Same as `dkm` but runs microsites by default instead of myjobs.
- Fixes a few issues in tests that either make the tests fail (really, how was that test in seo/tests/test_views.py ever passing?) or prevents them from running with the `--keepdb` flag. So now `--keepdb` should work without fail. 
- Moves some migration-y logic out of the test setup and into the migrations. Because it's migration-y logic.
- A small improvement for tasks, where bad database connections gets closed before the tests are run. We've been seeing issues with bad MySQL connections causing task failures. This should fix some/most of them.

This now means that we should only need to give new devs solr configs and a secrets file on their first day.

settings.py is the biggest of the changes. Are the changes correct, agreeable, and positive?

Run tests. They should work for you with no effort (or, well, maybe a bundles/custom.css file because I can't figure that one out).

## Things I learned while doing this
- Sometimes we really do need to be using `TransactionalTestCase` - whenever there's multiple db interactions specifically - but it's really, really slow.
- Solr is really, really slow in dev, which makes running things like the seo tests awful. We might discuss getting away from just blanket inheriting off of base test classes that do things like add to/remove from Solr when the tests aren't even using Solr.


universal/decorators.py
```diff
@@ -44,25 +65,6 @@ def wrap(request, *args, **kwargs):
         return wraps(view_func)(wrap)
     return decorator
 
-def activate_user(view_func):
```
almathew: More unused code.

dk2/env.sh
```diff
@@ -3,11 +3,11 @@ dk() {
 }
 
 dkm() {
-    dk -P settings_myjobs manage "$@"
+    dk manage "$@"
 }
 
 dkr() {
-    dk -P settings_myjobs:. dorun "$@"
+    dk -e myjobs:. dorun "$@"
```
tdphillips: `dkd` has `-e ENVIRONMENT=` - does this command need it as well?

api/models.py
```diff
@@ -1,7 +0,0 @@
-from seo.models import User
-from django.db import models
-
-from tastypie.models import create_api_key
-
-models.signals.post_save.connect(create_api_key, sender=User)
```
almathew: Moved this to myjobs/models.py in case we still need it for saved search emails.

dk2/env.sh
almathew: It shouldn't. myjobs is the default.

almathew:
Unfortunately api.py is only nearly unused. There's one stupid admin page that uses it, despite the user being logged in. Not sure it even works for new users, but I guess we'll find out. Not really going to fix it since removing the API endpoints was already outside the scope of what I was doing.

PR #3119, opened 2017-06-10
[PD-3351] refactor analytics css

mwarnsleyDE:
## What's this PR do?
PR refactors and restructures the css for analytics

## Screenshots (if appropriate)
N/A

## Where should the reviewer start?
Start inside of the analytics App

## How should this be manually tested?
Pull down repository. Check and make sure inside of analytics that the application still looks the same

PR #3122, opened 2017-06-13
Revert "[PD-3351] refactor analytics css"

aljordan82:
Reverts DirectEmployers/MyJobs#3119


aljordan82:
Something is broken in Analytics, Page refresh doesn't work properly.

aljordan82:
It is better to revert all JS changes, then we can reopen it with just the CSS changes

PR #3126, opened 2017-06-13
Imports to multiple sources

almathew:
Just so no one is afraid of the impact on production, this points to a feature branch.

## What's this PR do?
Re-writes imports with the goal of allowing us to add/remove sources we want to put jobs in at will.
Breaks things into sensible, logical increments so that things are more testable and we don't have mega-files. 
Bases everything off a "source of truth" so that they're all as in-sync with the source of truth as possible. Because of this, catching task failures is going to be very important!
Introduces imports into the database and Redis. Redis may or may not stay around.

I'm not actually sure it works, so don't bother at this point. Those instructions will come with the testing PR.

The real goal of this was adding redis and the db. 

To Do (in different PRs because this one is already un-reviewably huuuuuge):
- [x] Break adds, updated, and deletes into separate steps
- [x] Testing
- [x] Logging
- [x] Handle postajob
- [ ] Update documentation
- [ ] Re-write task re-runner


almathew:
Ignore until I'm done messing with tests

almathew:
Annnnd finished.

almathew:
@tejaswinigutti31 @JLMcLaugh @mklauber @vigowei ... and anyone else

almathew:
Imports appear to work in at least some situations.

import_jobs/tasks/general_tasks.py
```diff
@@ -0,0 +1,292 @@
...
+        if is_priority:
+            steps.extend([
+                priority_do_add.s(add_chunk), priority_do_update.s(update_chunk),
+                priority_do_delete.s(del_chunk),
+            ])
+        else:
+            steps.extend([
+                do_add.s(add_chunk), do_update.s(update_chunk),
+                do_delete.s(del_chunk),
+            ])
```
almathew: @mklauber the final implementation I decided on.

almathew:
For the record, tests aren't going to pass on this until we get redis working.

almathew:
This isn't actionable until I get tests passing, but it is open because (minus things I am randomly finding) the only thing remaining is getting tests passing. 
As far as I am aware, Redis on jenkins continues to be the only roadblock.

almathew:
@mklauber claims it works

almathew:
Tests are passing now... I was just stupid and updated the CELERY_ROUTES and now they're not. Not sure if it's the AMI or what, but this is 100% ready to review.

darrint:
I'm happy with what I see for the tasks and the sources code. I don't have much experience with imports so I'd like someone else like @mklauber or @JLMcLaugh to take a look and push the merge button.

PR #3129, opened 2017-06-13
attempt to integrate redis

vigowei:
## What's this PR do?
Adding logic to get guids from solr first then pull more information from Redis if needed. 

N/A

seo/search_backend.py line#302

 job search on microsites should still work, but it relies on other unfinished work.

first draft - work in progress


default_settings.py
```diff
@@ -800,3 +800,5 @@
 }
 UPDATE_ENDPOINT = "https://www.my.jobs"
 UPDATE_SCOPES = []
+
+FIELDS_IN_REDIS = []
```
darrint: Can we scope this name better? Perhaps prefix it with something like `SEARCH_` or `SEO_`?

vigowei: thanks for the feedback, I'll reconsider the naming.
Closing the PR for now until we have a more testable & reviewable version.

vigowei:
this one displays too many changed files as a result of rebasing... reopened as: https://github.com/DirectEmployers/MyJobs/pull/3169,

PR #3133, opened 2017-06-15
Display reporting link in topbar for users with PRM permissions [PD-3372]

tdphillips:
## What's this PR do?
Adds an or.
I made a booboo.


myjobs/templatetags/common_tags.py
```diff
@@ -337,7 +337,7 @@ def get_menus(context):
             },
         ]
 
-    if employer_menu and can_view_analytics_info:
+    if employer_menu and can_view_analytics_info or can_read_partner:
```
JLMcLaugh: You want parenthesis around the 2nd two terms. This will be evaluated like (employer_menu and can_view_analytics) or can_read_partner

PR #3134, opened 2017-06-15
[PD 3303] early status change dropdown nuo

koonsies:
## What's this PR do?

Adds the workflow state dropdown to the initial process record page so that a user can quickly categorize a record without creating a contact record.

First, run migrations

Afterward, head to NUO

Click "Review" on one of the records. 
Change the workflow state for that record.

The workflow state "reviewed" is being merged with "complete", as per a discussion with Jen. A new workflow state "unqualified" is being added in "reviewed"s place.

The "incomplete" filter now only shows "new" and "needs more info" records.


mypartners/views.py
```diff
@@ -2216,19 +2219,24 @@ def all_forms_iter():
...
+    if process_type == 'workflowOnly':
+        outreach_form.save()
+        return HttpResponse('"success"')
 
     if api_errors:
         return HttpResponse(json.dumps({
             'api_errors': api_errors}), status=400)
-
-    if has_errors or validate_only:
+    import ipdb; ipdb.set_trace
```
almathew: pdb

koonsies: Ah.. Well then... Just making sure people were reading my PRs >.>

gulp/src/nonuseroutreach/actions/navigation-actions.js
```diff
@@ -51,7 +52,9 @@ export function doFilterRecords() {
 
     if (workflowFilter === 'Incomplete') {
       filtersActive = true;
-      filteredRecords = filteredRecords.filter(record => record.currentWorkflowState !== 'Complete');
+      filteredRecords = filteredRecords.filter(
+        record => contains(['New', 'More Info Needed'], record.currentWorkflowState)
```
aljordan82: 'Incomplete' statues is missing from the list.

almathew:
if @aljordan82 is fine with the JS, python all looks good to me.

PR #3141, opened 2017-06-16
Pd 3375 Fix loading spinner

aljordan82:
## What's this PR do?
Fix loading spinner for all apps


gulp/src/sass/custom.scss
```diff
@@ -43,7 +43,6 @@ footer {
 
 
 /*PAGE LOADING SPINNER CSS*/
-.analytics-loader {
```
tdphillips: Was this intentional?

gulp/src/common/ui/LoadingSpinner.jsx
```diff
@@ -4,8 +4,8 @@ import {Component} from 'react';
 class LoadingSpinner extends Component {
   render() {
     return (
-      <div id="loading">
-        <div className="analytics-loader"></div>
+      <div id="loading-container">
+        <div className="loading-indicator"></div>
```
koonsies: Can we switch the order that these are displayed? I don't like the spinner behind the overlay.

aljordan82: @koonsies , good catch!

PR #3152, opened 2017-06-21
removed error message after export format has been selected

strempe:
## What's this PR do?
Removes error message from export popover after a user has selected an export format.


aljordan82:
The issue still exists, and assigning a onExited callback that clears the state should solve this issue.

gulp/src/analytics/components/Header/ExportOptions.js
```diff
@@ -33,28 +44,30 @@ class ExportOptions extends Component {
       dispatch(doExportTab('csv'));
     }
     this.setState({
```
aljordan82: Is this needed?

PR #3157, opened 2017-06-22
Fix get activities behavior

koonsies:
## What's this PR do?
Changes user.get_activities(company) to only return activities that the company provided has app access to.
Anywhere user.get_activities is used

Have more than one role for a couple companies
Have those companies have different app access levels
Call user.get_activities and provide each company
Notice the difference in activities returned

This is currently a guidance requested PR. Most of the activity-related functions that use this function also do a separate call to make sure that the activities returned also belong to the company provided. This is not the case in the Javascript User.Can, however, and it made me wonder why we would return activities in get_activities that don't belong to the company provided. While we -could- change the javascript to filter out activities that don't relate to the company provided.. Is there any reason to provide these activities in the first place?


myjobs/tests/test_models.py
```diff
@@ -436,6 +436,33 @@ def test_activities(self):
...
+        new_activity_name = 'do_something_fun'
+        new_activity = ActivityFactory(name=new_activity_name,
+                                       app_access=new_app_access)
+
+        # mimic admin role behavior of having -all- activities
+        self.role.activities.add(new_activity)
+
+        # the newly created activity isn't in the list since the
+        # company doesn't have the proper access
+        self.assertIn(new_activity_name,
```
tdphillips: The comments preceding `assertIn` and `assertNotIn` don't agree with what the code is doing.

koonsies: oh.. Haha.

I'll correct that.

tdphillips:
> Most of the activity-related functions that use this function also do a separate call to make sure that the activities returned also belong to the company provided.

Where does this happen? I didn't see it.

koonsies:
https://github.com/DirectEmployers/MyJobs/blob/94f252e56b7dc7b2594c8c509980fc7964bde541/myjobs/decorators.py#L203-L203 

It is checked is the "requires" decorator, for example.

PR #3164, opened 2017-06-23
We're using the wrong column names [ST-4222]

tdphillips:
## What's this PR do?
Changes the names of columns that we use for criteria spreadsheet creation, among other things
Export a PRM report. It should fail. Check this out. It should succeed.


myreports/presentation/xlsx.py
```diff
@@ -43,9 +43,12 @@ def write_presentation(self, values, records, output, report=None,
                         break
                 # Magic happens here. Configuration columns are ordered so
                 # shenanigans like this are okay.
-                criteria_rows = [
-                    c.alias for c in report_configuration.columns][
```
tdphillips: `alias` and `column` were the same thing for analytics reports. They are not for PRM.

myreports/presentation/xlsx.py
```diff
@@ -43,9 +43,12 @@ def write_presentation(self, values, records, output, report=None,
...
-                        :len(report_filters)]
+                criteria_start = 0
+                column_names = [c.column for c in report_configuration.columns]
+                if column_names[0] == 'date_time':
```
tdphillips: This is ugly.

tdphillips:
Maxim 43 applies. I thought the original solution was clever.

PR #3166, opened 2017-06-23
Attempt at TJD data source [PD-3396]

tdphillips:
## What's this PR do?
Adds a data source for TJD
It can't. The api it's using doesn't yet contain any data to return.


tdphillips:
I just made a data source that returns data. Tell me what it needs to return and I'll make it so.

Not sure if the format is proprietary so check Slack for the input/output formats.

tdphillips:
Taking this down while we do more api shenanigans.

tdphillips:
I'm drawing a blank - what was left to do on this?

PR #3169, opened 2017-06-23
Pd 3363 get guids from solr and data from Redis

vigowei:
## What's this PR do?
Adding logic to get guids from solr first then pull more information from Redis or MySQL if needed. 

N/A

seo/search_backend.py line#302

 job search on microsites should still work,

To have it pull data from outside of solr, add fields to SEO_REDIS_FIELDS.
Still fixing some tests.


seo/search_backend.py
```diff
@@ -299,33 +301,41 @@ def search(self, query_string, sort_by=None, start_offset=0, end_offset=None,
...
+            # fields come in as list, set or string
+            if not isinstance(fields, (list, set)):
+                fields = fields.split()
+            requested_fields_in_redis = set(fields) & set(settings.SEO_REDIS_FIELDS)
+            fields = set(fields) - requested_fields_in_redis
+            if requested_fields_in_redis:
+                fields.add('guid')
+            fields = ' '.join(fields)
             kwargs['fl'] = fields
-       # This code was causing sort_by to break, but we're keeping it as a
```
almathew: @JLMcLaugh's old code can be deleted.

seo/search_backend.py
```diff
@@ -462,6 +472,12 @@ def search(self, query_string, sort_by=None, start_offset=0, end_offset=None,
             self.log.error("Failed to query Solr using '%s': %s", query_string, e)
             raw_results = EmptyResults()
 
+        if requested_fields_in_redis:
```
almathew: Is there a way we can have this rely on a flag too? There are some situations (mostly via CLI) in which we're going to bypass Redis regardless of settings.

vigowei: how about we allow something like `multi_src=True` to be passed as a kwarg to search?

almathew: Sure, that's exactly what I was hoping for.

almathew:
Closing while @vigowei implements the flag.

vigowei:
@almathew reopening with some updates including the flag option to allow bypassing Redis.  I added an`SEO_LOG_ES_HOSTS` option to default settings, it won't break anything if no valid ES url(s) are supplied, it'll just fall back to python logger which is probably stderr or stdout.  
I'm thinking that part (seo/utils/log.py) can be improved and generalized more, so all modules, not just seo can leverage it for logging... probably worth a new ticket.

almathew:
Now you have merge conflicts - probably because of some of the changes I made with logging.

vigowei:
@almathew just saw your comment, it was something in the settings file and I took care of it

almathew:
I'm starting a much more in-depth review of this, but @mklauber and @JLMcLaugh your input would also be welcomed.

almathew:
PR is not playing nicely with new solr schema.

PR #3176, opened 2017-06-26
Backport Impersonate to QC

koonsies:


dk2/dockerdev.sh
```diff
@@ -240,7 +240,7 @@ runserver() {
 }
 
 runmicrosites() {
-    pythonpatharg="-e PYTHONPATH=settings_dseo" \
+    pythonpatharg="-e ENVIRONMENT=dseo" \
```
koonsies: Looks like this change came from staging. Is this legit??

almathew: Yes, legit.

PR #3181, opened 2017-06-26
Add an asterisk to two labels [ST-4199]

tdphillips:
## What's this PR do?
Adds an asterisk to two labels

2017-06-26-164846_16x16_scrot

Begin creating a new organization in the TJD app. Note that the category label now contains an asterisk.


tdphillips:
CboReview.js doesn't include asterisks on any other fields so I left it alone.

PR #3191, opened 2017-06-27
Fix issue with PRM exports that should have been detected earlier

tdphillips:
## What's this PR do?
Removes some cleverness that was clearly broken
Start creating a new PRM report. Enter something in the partner or contact field. Save and try exporting as Excel.
If you're not on this branch, you've probably got a yellow screen.
The original hotfix was too clever for its own good.


myreports/presentation/xlsx.py
```diff
@@ -41,15 +41,12 @@ def write_presentation(self, values, records, output, report=None,
                         criteria.append(('date range', '-'.join(
                             report_filters.pop(field))))
                         break
-                # Magic happens here. Configuration columns are ordered so
-                # shenanigans like this are okay.
-                criteria_start = 0
                 column_names = [c.column for c in report_configuration.columns]
-                if column_names[0] == 'date_time':
-                    criteria_start = 1
-                criteria_rows = column_names[criteria_start:criteria_start+len(report_filters)]
```
tdphillips: This works for analytics and the basic PRM report with no filters. It's not appropriate for complexity of any kind.

PR #3194, opened 2017-06-28
PD-3400: Remove slow admin tests

mklauber:


mklauber:
Let's see how much of a speed up we get from this.

tdphillips:
So much speed. It finished in 46 seconds.

mklauber:
lol

PR #3198, opened 2017-06-29
[PD-3448] cleanup sidebar component code

mwarnsleyDE:
## What's this PR do?
PR cleans up the code inside of the sidebar component

N/A

Pull down repository. Go into any analytics application and make sure the sidebar is still functioning properly


mwarnsleyDE:
@aljordan82 I think we have to look and judge whether or not to use shouldComponentUpdate in every component. It's only suppose to be used if absolutely necessary for performance issues and are you sure that the sidebar component is one that has that many performance issues attached to it?

koonsies:
I think adding "shouldComponentUpdate" is unnecessary for more simple components. Much of the same benefit can be gained through being careful which props are being passed into our components. Only passing those explicitly needed (and not parent objects) will go a long way in preventing superfluous refreshes.

aljordan82:
Without shouldComponentUpdate, the SideBar render will get called if anything changed in the parent component. this will cause the r`eports.map` in the renderer to get executed.

koonsies:
I can see what you mean. I'm not completely convinced as to the wastefulness of this call, though. Calling a map on such a short list (there will only ever be 5-10 reports) isn't a resource-intensive action. An issue with throwing in shouldComponentUpdate all over the place, is that we could potentially get into trouble where components aren't updating when we expect them to, due to conditions not being met in the SCU function.

Just playing devil's advocate. I'm fine with either way.

aljordan82:
I am seeing the Sidebar render called 5 times with each click.

aljordan82:
https://facebook.github.io/react/docs/react-api.html#react.purecomponent

koonsies:
That's a good compromise. We should look into that for most of our components that are tied directly to AnalyticsApp, as that component will be re-rendered a lot.

mwarnsleyDE:
@aljordan82 @koonsies 
https://twitter.com/dan_abramov/status/759383530120110080?lang=en
This is coming from Dan Abramov himself. I really think Reacts Rendering speed is not being given enough credit here. Even some places have said that a component that isn't that heavy would have barely enough performance enhancement to warrant implementing PureComponent or shouldComponentUpdate

aljordan82:
@mwarnsleyDE I am not for using pureComponents everywhere, I am for placing them "strategically'"at the high level components.

mwarnsleyDE:
@aljordan82 then wouldn't it make more sense for us to use it in the AnalyticsApp.jsx file which is the highest level component. For the sidebar component it holds other components but it's not heavy at all and everything passed to it is from the AnalyticsApp.jsx

mwarnsleyDE:
I'm not saying we shouldn't implement it, I'm just saying we have to measure performance and see if it's worth implementing it in certain components. Tons of people say since they use Redux they don't really ever have to implement it and it's a matter of measuring the performance first before implementing it. We should check and see if we are hurting performance wise first before going ahead and just putting it in.

PR #3200, opened 2017-06-29
Analytics sidebar unit test

aljordan82:


aljordan82:
I don't see anything that can be removed

PR #3202, opened 2017-06-29
Data source for TJD

tdphillips:
## What's this PR do?
Adds a data source for TJD. Some of the necessary bits (like multiple contact or location selection) should be handled by schema_columns and/or column_formats.
¯\\\_(ツ)\_/¯
If it doesn't have some necessary functionality, let me know.


tdphillips:
Probably not exactly what we need but something is better than nothing.

PR #3204, opened 2017-06-30
Fix hanging Spinner

aljordan82:
Prevent storing the redux state if there are any pending Ajax requests.

This is difficult to test or to reproduce.


koonsies:
Adding DNM flag until I get a chance to talk with @aljordan82 about this

PR #3213, opened 2017-07-05
Pd 3452 my.jobs upgrade to webpack3

aljordan82:
## What's this PR do?
Upgrade to webpack 3
The production bundles should be minified.


tdphillips:
What guidance is being requested?

aljordan82:
I am not sure if I am missing anything with this upgrade.
@darrint DO you spot anything wrong with this PR?

koonsies:
What are the thoughts on the speed of webpack 3? If it's prohibitively slow, this update may not be worth it.

aljordan82:
@koonsies It's slightly slower, but I think will need this upgrade eventually.

koonsies:
I'm noticing that as well. Simple changes seem to recompile pretty quickly. I also agree that staying current on our libraries is the way to go.. Just wanted to touch base about speed and see what you thought.

PR #3217, opened 2017-07-05
Pd 3406 improve error handling

aljordan82:
## What's this PR do?
Adding an error handler for error-action
dispatch an error action.


koonsies:
I added a line that will cause the dynamic API report to return a 403.. Not sure why, but the first time it gets the 403, the error is not logged to the console (at least, not by this component). Any ideas?

PR #3229, opened 2017-07-07
PD-3451 Fix discrepancies in timeline

vigowei:
## What's this PR do?
Fixing glitches mentioned in PD-3451

**Before**: data were aggregated based on UTC, so a _day_ is UTC 00:00 - 23:59. That means for someone in New York, it's 19:00 to 18:59  (or 20:00 to 19:59 during daylight savings), that's why there is a "tail" at the end (extra 4-5 hours till Eastern midnight but next day for UTC)

image

**After**: each data point now represents total views of a day _in client timezone_, e.g. for someone in New York, it aggregates UTC 04:00am to 03:59am the next day.

image

`myreports/models.py`, line#456 , logic for timeline data formatting (previously `format_timeline_data`in analytics/api.py) has been moved here
`myreports/datasources` there is a small change in `job_views` and `analytics_event`, to shift UTC to default timezone before aggregating

Visit any report page with timeline, select a custom date range, you should expect the same behavior in the screenshots (timeline ending at right date and all days without data represented with zeroes). Timeline graphs exported in Excel also reflect this change.

image

No front end changes required.
Right now, USE_TZ is enabled for our app, which means naïve time objects will default to time zone: `America/New_York`. 
We should probably start using offset-aware timestamps everywhere. Include client time zone information in requests, this will enable localizing time series data for users in different timezones. To do that, we need to make sure front end always passes client timezone, and we do not lose that information during date re-formatting (where time gets truncated): everything that gets passed into `DateRangeFilter` (myreports/datasources/util.py) should be timezone aware (right now it's not the case).


vigowei:
closing for now, fixing tests

PR #3233, opened 2017-07-07
Final fixes for solr improvements.

almathew:
## What's this PR do?
Makes the final necessary fixes of issues discovered during load testing of new code.

PR #3234, opened 2017-07-07
Solr improvements

almathew:
## What's this PR do?
All the solr improvements.
To be merged after https://github.com/DirectEmployers/MyJobs/pull/3233 has been merged.

Wherever you want. It's a lot of changes. All of them should've been individually reviewed, but that doesn't mean it's not worth a second glance.

Index some jobs and test microsites, pretty much. These are some far-reaching changes.

This is going to be deployed to production on Sunday.
Seth has signed off on the overall output, and it's doing relatively well in load testing.


import_jobs/sources/redisjobsource.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def add_to_storage(self, jobs):
+        add_dict = {job['guid']: json.dumps(job, default=json_util.default) for job in jobs}
+        self.conn.mset(add_dict)
+
+    def delete_from_storage(self, guids):
+        self.conn.delete(*guids)
+
+    def to_storage_format(self, job):
+        return {k: v for k, v in job.iteritems() if k in self.fields}
```
vigowei: This is equivalent to `return {k: job[k] for k in fields if k in job}` but performance is different, running with timeit locally:

`{k: job[k] for k in self.fields if k in job}` :  averages 2.63s, 

`{k:v for k, v in job.iteritems() if k in self.fields}`: averages at 14.6s

almathew: Fixed in other PR

vigowei: I just noticed the same thing in `dbjobsource.py`, L#42

seo/tests/setup.py
```diff
@@ -89,7 +89,7 @@ class DirectSEOTestCase(DirectSEOBase):
     def setUp(self):
         super(DirectSEOTestCase, self).setUp()
         self.solr_docs = solr_settings.SOLR_FIXTURE
-        self.conn.add(self.solr_docs)
+        handle_imports(0, self.solr_docs)
```
vigowei: thanks for fixing these seo tests, I had fixed them in a slightly different way but I like your solution better.

almathew: Yeah I'm still going though patching up some smaller things. My work isn't going to be"done" with this PR, but this code is (obviously) production ready with this PR.

PR #3235, opened 2017-07-07
Changes to "Distribution Status" switch [PD-3430]

tdphillips:
## What's this PR do?
- Hides switch if the user is viewing a single organization
- Adds a confirm when disabling the switch
2017-07-07-145104_819x432_scrot
Select an organization. In any of that organization's subscreens, the switch should not be present.
Head back to your main TJD screen. Switch off distribution. You should see the pictured dialog.
The text is not my own. If anyone has suggestions, I'm all ears.


aljordan82:
The code looks fine, however, I couldn't test it locally.

gulp/src/cbo/components/commonCbo/Sidebar.js
```diff
@@ -84,16 +96,20 @@ class Sidebar extends Component {
...
         <a className="btn leave-feedback" href="http://www.directemployers.org/beta-feedback" target="_blank">Leave Us Feedback</a>
-          <h2 className="top">Organization Status</h2>
-          <div id="toggle_container">
-            <p className="distribution-status-title">Distribution Status</p>
-            <div className="toggle-inner-container">
-              <span className="toggle-main-container">
-                <input type="checkbox" className="toggle-input" id="toggle_input" name="check" checked={checked} onChange={(e) => this.updateCboStatus(e)} />
-                <label data-on="ON" data-off="OFF" htmlFor="toggle_input" className="toggle-label"></label>
-              </span>
+        {viewTJD ? '' :
```
koonsies: It looks like we want to restrict this control to display only on the main page. That is similar to the check on line 70 (82) .. perhaps we could use the same logic in both places?

tdphillips: Thanks. The way I did it was leaving the switch on the "new organization" form, which probably wasn't ideal.

tdphillips:
The page no longer reloads if you click cancel or otherwise dismiss the modal.

gulp/src/cbo/components/commonCbo/Sidebar.js
```diff
@@ -26,10 +27,18 @@ class Sidebar extends Component {
...
   }
-  updateCboStatus(e) {
+  async updateCboStatus(e) {
+    e.preventDefault();
     const {dispatch} = this.props;
     const checked = e.target.checked;
+    const message = 'Turning off distribution means your jobs will no longer be delivered to the listed organizations. Are you sure that is the action you want to take?';
+    if (!checked) {
+      if (!(await runConfirmInPlace(dispatch, message))) {
+        return false;
```
aljordan82: just `return;` is better

tdphillips: Why?

aljordan82: Nothing uses the return values, so it is better to keep the return value consistent

tdphillips: What about the `return true` a few lines later? That seems inconsistent with all other return values.

PR #3237, opened 2017-07-10
Pd 3406 improve error handling

aljordan82:
## What's this PR do?
Better error handling


aljordan82:
@koonsies The error didn't show up because it happened on an early stage before any component got mounted.

koonsies:
Alright. I can see why that would happen. If we have issues in the future regarding errors on mount, we may need to look into a solution for it. Until then, this is a good start.

vigowei:
I was debugging impersonation earlier and ran into a 404 status that caused the spinner to hang. Might wanna take a look. To reproduce just select a company to impersonate since that part is currently broken in QC.

PR #3255, opened 2017-07-12
[PD-3445] Update Report Generation Logic to Honor Impersonation

koonsies:
## What's this PR do?
Darrin's updates to the way we create/handle reports did not honor the impersonation logic for analytics. This PR add a specialized user.can call to filter out activities that are to be ignored for staff.

* Change your company to a company with the Impersonate app access.
* Ensure you are an admin/are assigned to a role with the "impersonate analytics" activity
* Head to Analytics
* First, ensure that the data displayed is correct for the company you are signed in as.
* Second, change the company in the textbox in the upper left corner.
* Finally, test the analytics platform (including exports) and ensure the data matches the impersonated company.
This ticket is a tricky one, because we have to be careful not to grant permission to staff to view protected information (such as PRM reports). This approach is designed to avoid that type of unintended access.

**Note, DNM Label added until I can add proper tests**


analytics/api.py
```diff
@@ -52,8 +52,6 @@ def validate_analytics_query_request(query_data, user, company):
                 field, 'Invalid date start: %s' % query_data[field]
             )
 
-    all_reports.resolve(query_data['report'], user, company)
```
koonsies: @darrint you'll need to verify this change. I saw that every time this function was called, it was proceeded by a call to `report_for_request` which did this exact call once again.

tdphillips:
We still use `company` here and two lines later.

https://github.com/DirectEmployers/MyJobs/pull/3255/files#diff-9512d7b95f52cc9d909f950219e6e7c8R180

PR #3257, opened 2017-07-12
Django says this is missing

tdphillips:
## What's this PR do?
Adds a migration that Django complains about
While ensuring my database was up to date, Django helpfully informed me that there are model changes that are not yet reflected in a migration. I didn't want this change to get caught up in an unrelated migration so here we are.


tdphillips:
@almathew, any problem with this? The original change is yours, I believe, and I wanted to make sure we don't expect this to cause any issues.

almathew:
No I'm fine with the change. I'm mostly just confused on how this only just got picked up now.

tdphillips:
Jenkins doesn't like this migration. I'll play with it a bit and see what happens.

tdphillips:
Tried to execute this migration myself. Was notified that there were more model changes that didn't have a migration. Generating them resulted in a duplicate analytics migration.

¯\\\_(ツ)\_/¯

PR #3259, opened 2017-07-13
Modify models to facilitate partner assessments [PD-3473]

tdphillips:
Also adds initial feature flag support

## What's this PR do?
Allows us to start work on partner assessments
Feature flag disabled (or nonexistent)
secure my jobs-prm-view-records-edit-id 2 partner 2 page 1 1

Feature flag enabled - this actually features horizontal radio buttons but the base bootstrap style we load makes them display:block
secure my jobs-prm-view-records-edit-id 2 partner 2 page 1
Check out the "edit communication record" form and verify that it doesn't have inputs for score or review. The fields currently don't show up on the NUO version, just the one in PRM.

Go to your admin and find the new section named "Waffle" and click on the "add" button next to "flags". Name your new flag "partner assessment" and enable it for superusers, staff, or authenticated. Your dev user is probably all of these so which you pick doesn't matter.

Go back to the "edit communication record" form and, like magic, there are new fields for you to fill out.


mypartners/forms.py
```diff
@@ -500,6 +501,39 @@ def save(self, request, partner, commit=True):
...
+class HorizontalRadioRenderer(forms.RadioSelect.renderer):
+    def render(self):
+        return mark_safe(u'\n'.join([u'%s\n' % w for w in self]))
+
+
+class ContactRecordFormWithAssessment(ContactRecordForm):
+    score = forms.ChoiceField(
+        widget=forms.RadioSelect(renderer=HorizontalRadioRenderer),
+        required=False,
+        choices=[(x, x) for x in range(0, 6)])
```
tdphillips: The first radio is for zero stars and will probably need to be hidden. This was an annoying fix. :(

mypartners/forms.py
```diff
@@ -500,6 +501,39 @@ def save(self, request, partner, commit=True):
...
+    review = forms.Textarea()
+
+    def __init__(self, *args, **kwargs):
+        if 'initial' not in kwargs:
+            kwargs['initial'] = {'score': 0}
+
+        super(ContactRecordFormWithAssessment, self).__init__(*args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        if self.instance.score == 0:
```
tdphillips: Again, more annoyances. We don't want to store a zero.

mypartners/views.py
```diff
@@ -804,6 +806,10 @@ def prm_edit_records(request):
     except ContactRecord.DoesNotExist:
         instance = None
 
+    Form = ContactRecordForm
+    if flag_is_active(request, 'partner assessment'):
```
tdphillips: If you haven't gone into your admin and created the flag that you're checking for, it defaults to disabled. I was worried about that but that's exactly what we need.

aljordan82:
Do we really want to merge this? 
or can we have a feature branch?

tdphillips:
I have no preference. The idea behind feature flags, however, is that you can merge partial work without affecting the users.

almathew:
I like the idea of it still. It gives us a lot more control over when it's truly released, that isn't necessarily tied to a deployment.  And I know this is pretty "worst case scenario," but faster rollback too.

tdphillips:
There we go.

PR #3264, opened 2017-07-14
Add jenkins script.

mklauber:


jenkins.sh
```diff
@@ -0,0 +1,48 @@
...
+    PATH=$WORKSPACE/venv/bin:$PATH
+    if [ ! -f "venv/bin/activate" ]; then
+            virtualenv venv
+    fi
+    . venv/bin/activate
+
+    # Update the virtualenv with the latest requirements.
+    pip install setuptools==24.0.3
+    pip install -r MyJobs/requirements.txt --download-cache=/tmp/$JOB_NAME
+    pip install MySQL-python
```
tdphillips: MySQL-python 1.2.5 is in requirements.txt already. Do we need a more up to date version?

mklauber: No.  That line can/should be removed.

aljordan82:
The /static directory also needs to be watched.

tdphillips:
Do we have any tests for files residing in the /static directory?

aljordan82:
@tdphillips It looks like the js files in /static are not tested at all. I thought they were tested.

PR #3265, opened 2017-07-14
Scaffolding for making radio buttons into stars

tdphillips:


mypartners/forms.py
```diff
@@ -502,8 +502,26 @@ def save(self, request, partner, commit=True):
 
 
 class HorizontalRadioRenderer(forms.RadioSelect.renderer):
+    radio_text = ('<input {checked} type="radio" id="{id}" name="{name}" '
+                  'value="{value}" /><label class="full" for="{id}" '
+                  'title="{title}"></label>')
+    radio_tooltips = {1: ':\'(',
```
tdphillips: These emoji should be replaced with what we actually want to show to the user.

PR #3266, opened 2017-07-14
Backport Staging to QC

koonsies:
*Note* The default settings changes are not mine. Are they legit?


default_settings.py
```diff
@@ -218,12 +218,12 @@
         'routing_key': 'solr.expire_jobs'
     },
     'tasks.monitor_imports': {
-        'queue': 'priority',
-        'routing_key': 'priority.monitor_imports'
+        'queue': 'priority_import',
```
koonsies: Legit??

mklauber: Yes.

PR #3269, opened 2017-07-17
Modify nuo backend to support assessment [PD-3488]

tdphillips:
## What's this PR do?
More flag shenanigans
2017-07-17-082631_663x847_scrot
The score field will need to be updated with the star widget eventually. Our remote form library turned it into a text input as well.


koonsies:
Do we want this pointed to QC, rather than the partner-reviews branch? I may have missed some conversation.

tdphillips:
No. I thought it was pointed at partner-reviews. Fixed.

koonsies:
For the NUO backend changes, when we get the field for score, the field.widget is returned as "textarea". I haven't looked into it yet, but I'm assuming that's a value returned from Django? To prevent hacky stuff as seen below, we'd need that widget type to be something unique to star.. Like... widget=StarWidget or something.

https://github.com/DirectEmployers/MyJobs/pull/3273/files#diff-3a8871df0b15f4586a16f51a0e06af7aR38

tdphillips:
It's the default widget for an int field.

I made it a radio button locally. As we don't currently have one, that means I get to create my own React radio button widget.

koonsies:
Ivan already created the widget itself. He simply needs the field.widget value to be returned as something that signifies which widget to use.

tdphillips:
We're using django-remote-forms to pass this form from Django land to React land. The widget name we're passing around is an actual Django thing. I guess I can make a dummy Django widget that's basically a copy of Radio.

koonsies:
That's pretty much what I mean. As it stands, the React code checks the widget name to determine what to display.

tdphillips:
Sorry, I modified my comment. Cloning Django's radio input doesn't work.

tdphillips:
Let's start over. This is going to require "hacky stuff" no matter where we do it. The question is where it's going to happen.

tdphillips:
We now pass "StarWidget". It's not going to display in the ui until @iartemiev's code is modified as well.

iartemiev:
Got it, I'll modify momentarily

tdphillips:
I also made it a radio select. I don't know if that matters for the React widget at all.

iartemiev:
Changes made. Works now

iartemiev:
Just tested it and it's not... It's returning `null` again...

iartemiev:
Every other form field has a label... Why can't this?

tdphillips:
Great question. It was a decision by the Django devs.

iartemiev:
Gotcha. That prop is how we set the label that's adjacent to the input in the form. So we need to figure out some kind of workaround then.

Currently, it looks like this:

image

tdphillips:
Any more feedback? Merging this would make other PRs look nicer.

iartemiev:
None from me. Everything seems to be working well now

PR #3270, opened 2017-07-17
Pd 3473 partner assessments create star select widget

strempe:
## What's this PR do?
New star rating widget for communication record.

screen shot 2017-07-17 at 12 22 18 pm
Communication record with new widget 

Reference PD-2868 for setting up waffle to review.


aljordan82:
This shouldn't go into quality-control. We need to push it to partner-reviews

mypartners/forms.py
```diff
@@ -513,7 +513,7 @@ class HorizontalRadioRenderer(forms.RadioSelect.renderer):
 
     def render(self):
         radios = []
-        for choice in self.choices:
+        for choice in self.choices[::-1]:
```
aljordan82: Can you add a comment here explaining why the list needs to be reversed?

static/star_rating.css
```diff
@@ -0,0 +1,61 @@
+@import url(//netdna.bootstrapcdn.com/font-awesome/3.2.1/css/font-awesome.css);
+
+.profile-form-input.score-field label {
```
aljordan82: `.profile-form-input` is too specific, can you the star styles more generic?

static/star_rating.css
```diff
@@ -0,0 +1,61 @@
...
+  text-align: center;
+  margin: 0 0 0.2em;
+}
+
+.profile-form-input.score-field label input[type="radio"] {
+  display: block;
+  margin: 0.5em auto;
+}
+
+fieldset, label { margin: 0; padding: 0; }
```
aljordan82: I don't think those are used
```
+fieldset, label { margin: 0; padding: 0; }
 +body{ margin: 20px; }
 +h1 { font-size: 1.5em; margin: 10px; }
```

mypartners/forms.py
strempe: Yes

static/star_rating.css
```diff
@@ -0,0 +1,61 @@
...
+
+.score-field > label:last-of-type:before {
+  content: "\f05c";
+    color: #a5a5a5;
+}
+.score-field > label:last-of-type:hover:before {
+color: #808080;
+}
+
+.score-field > .half:before {
```
iartemiev: This isn't needed, since we're only using full stars

templates/mypartners/edit_record.html
```diff
@@ -10,6 +10,7 @@
 
 {% block meta-extra %}
 <link href="{{ STATIC_URL }}jquery.tagit.css" rel="stylesheet" type="text/css">
+<link href="{{ STATIC_URL }}star_rating.css" rel="stylesheet" type="text/css">
```
aljordan82: Is this needed? it looks like adding the css file to base.html should do the job. not sure though!
Can you try removing the css link from this file?

koonsies: Aladeen is right that it is not needed in both files. It may be better to just include it where needed, rather than at the root (base.html).

templates/base.html
```diff
@@ -23,6 +23,7 @@
     {% compress css %}
     <link href="{{ STATIC_URL }}def.ui.css" rel="stylesheet" type="text/css">
     <link href="{{ STATIC_URL }}my.jobs.css" rel="stylesheet" type="text/css">
+    <link href="{{ STATIC_URL }}star_rating.css" rel="stylesheet" type="text/css">
```
koonsies: I see that the star_rating was moved to base.html - did you guys decide it was better to have this CSS loaded on almost every single page, rather than on a page-by-page basis?

aljordan82: I prefer not including the css in base.html, I think It's better if css was included as needed.

PR #3273, opened 2017-07-18
Pd 3473 nuo add stars to nuo forms

iartemiev:
## What's this PR do?
Adds star widget to NUO Communcation Record 

screen shot 2017-07-18 at 3 17 33 pm

StarWidget.jsx

1. NUO
2. Outreach Records 
3. Review
4. Fill out **Contact Search** -> Select Contact
5. The **Score** field with the corresponding Star Widget should be visible and should function as expected on click and on hover.
6. Data should also be getting passed successfully in the API request upon clicking **Add Record**

In `StarWidget.jsx`, I hate having to write out each `<input><label>` pair manually. Normally, I'd, of course, just do it inside of a loop. But, as I've mentioned in the comment, that would necessitate wrapping the adjacent elements in a single parent, consequently breaking the CSS...


gulp/src/nonuseroutreach/components/StarWidget.jsx
```diff
@@ -0,0 +1,36 @@
...
+import PropTypes from 'prop-types';
+
+const StarWidget = ({onChange}) => {
+  // Unfortunately, have to do it this way, instead of just mapping over an array of [0..5]
+  // Because if we map, we have to enclose each <input><label> pair in an enclosing tag, i.e. <span>
+  // And that will make the existing CSS (static/star_rating.css) not work
+  // (no easy way to get around that without a completely different CSS file for the component)
+  return (
+    <fieldset className="score-field">
+      <input type="radio" name="score" id="star5" value="5" onChange={onChange}/>
```
aljordan82: Is it safe to use a non unique id for input fields? we could have multiple StarWidgets in page

iartemiev: Technically, we don't need `id`'s at all. I'll take them out

gulp/src/nonuseroutreach/components/StarWidget.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
iartemiev: Sounds good, I'll move it

gulp/src/nonuseroutreach/components/StarWidget.jsx
aljordan82: what about the htmlFor? doesn't it need an id?

iartemiev: You're right... That was short-sighted of me. Since we're hiding the radio buttons, we need the `id` and `for`...

gulp/src/sass/custom.scss
```diff
@@ -2,6 +2,7 @@
 
 @import "variables";
 @import "media_queries";
+@import "../../../static/star_rating.css";
```
aljordan82: This file is importing Font-Awesome. but we don't really need it in React land because we already have font-awesome there.
We can discuss offline.

static/star_rating.css
```diff
@@ -0,0 +1,52 @@
...
+/****** Style Star Rating Widget *****/
+
+.score-field {
+  border: none;
+  float: left;
+}
+
+.score-field > input { display: none !important; }
+.score-field > label:before {
+  margin: 5px;
```
aljordan82: 5px is a bit too much, I would make just 2px.

iartemiev:
Requested changes made

gulp/src/common/ui/StarWidget.jsx
```diff
@@ -0,0 +1,53 @@
...
+      ids: Array.from(Array(6), () => uniqueId('star'))
+    });
+  }
+
+  render() {
+    const {onChange} = this.props;
+    const {name, ids} = this.state;
+    return (
+      <fieldset className="score-field">
+        <input type="radio" name={name} id={ids[5]} value="5" onChange={onChange}/>
```
aljordan82: A loop would be nice instead of having the almost identical lines 5 times. but it is up to you.

aljordan82: also create the unique ids inside the loop.

iartemiev: That won't work because you have to have a single root element in a loop. Therefore, we'd have to wrap them in a span, i.e.: 
```html
<span>
  <input/>
  <label>
</span>
```
And that will screw up the CSS.

aljordan82:
```
  constructor() {
    super();
    this.uniqueId = uniqueId('score');
  }

  render() {
    const {onChange} = this.props;
    // const {name} = this.state;
    const arr = Array.from(Array(6), (_, i) => i);
    const stars = Array.map(arr, (_, i) => {
      const id = this.uniqueId + i;
      return ([
        <input type="radio" name={this.uniqueId} id={id} value={i} onChange={onChange}/>,
        <label className="full" htmlFor={id}></label>
      ]);
    }).reverse();
    return (
      <fieldset className="score-field">
        {stars}
      </fieldset>
    );
  }
```

iartemiev: You're right. I guess what I wrote above is only true for code that's inside of the actual `return` of the render function.

koonsies:
This PR should point to the partner-reviews branch, not QC

PR #3280, opened 2017-07-19
Partner assessment api [PD-2869]

tdphillips:
## What's this PR do?
More backend work to support partner assessments

Communication record endpoint
2017-07-21-095416_1337x894_scrot

List endpoint (includes create)
2017-07-19-163246_1351x806_scrot

Detail endpoint
2017-07-19-163344_1359x536_scrot


tdphillips:
Remaining question: How much of the drf work do we think is built-in functionality?

To put it another way, I don't want to write tests for functionality I didn't write. If anyone thinks something is sufficiently custom, say so and I'll write a test.

mypartners/views.py
```diff
@@ -2513,3 +2531,69 @@ def api_get_workflow_states(request):
...
+        company = get_company_or_404(request)
+        partner_id = request.GET['partner_id']
+        if company.partner_set.filter(pk=partner_id).exists():
+            assessments = PartnerAssessment.objects.filter(partner=partner_id)
+            serializer = PartnerAssessmentSerializer(assessments, many=True)
+            return Response(serializer.data)
+        else:
+            return Response({'partner': [u'Invalid partner']})
+
+    @method_decorator(requires('read assessment'))
```
koonsies: This should probably be 'create assessment'.

koonsies:
I think the tests here are sufficient.

koonsies:
This looks good for the retrieval of existing partner assessments. We also need an API endpoint for communication records during a particular time period. These will be used to populate the RecordList in the mockups.

tdphillips:
Are start/end dates required or can one/both be left out?

koonsies:
Both will be provided.

tdphillips:
Could/should that be combined with the aggregate endpoint? We're basically doing the same filter generating the same list of records twice.

koonsies:
Oh, yeah. I edited my comment above to ask that same question. I was torn a bit on the issue of keeping the API clean (i.e., doing only one thing) and multiplying the call. Let's go with a single API endpoint in interest of speed.

tdphillips:
Updated the original post. The old aggregate endpoint now also returns a list of records.

tdphillips:
I just took a second glance and realized how ugly that return value is. I'll work on it.

tdphillips:
Made the communication record endpoint more pretty. It no longer resembles the structure returned by other endpoints, however.

koonsies:
Excellent! Thank you. Could you point this PR to the partner-review-stage-2 branch? These APIs aren't needed until that feature branch.

tdphillips:
This is an ugly diff.

koonsies:
A small nitpick, but DRF puts a link at the top of the API that, when clicked, throws a 500 (it trims off the partner_id). I'm not super, super freaked out by this bug, just wanted to call attention to it in case it's a quick fix.

PR #3284, opened 2017-07-21
do that waffle

koonsies:
## What's this PR do?
Add waffle flags to the star display (where they hadn't been previously)

BONUS! I also moved where the assessment box displays on the commRec detail view and added the review text.
Go to your waffle flag in admin.. Uncheck all user groups. Verify that you don't see stars anywhere. Or any other part of assessments for that matter.


templates/mypartners/includes/record_card.html
```diff
@@ -23,11 +23,14 @@
         <span class="tag-name small" style="background-color: #{{tag.hex_color}}">{{ tag.name }}</span>
         {% endfor %}
     </div>
+    {% load waffle_tags %}
```
koonsies: I put the "load" tag down here so when we remove it, it's easy to spot and doesn't accidentally get left at the top.

PR #3287, opened 2017-07-24
Pd 3510 prm mock redux state

iartemiev:
## What's this PR do?
Creates initial redux state and mocks communication records

`initialState.js`

Please let me know if I missed anything or if you think it should be structured differently.


gulp/src/partnerassessment/reducers/initialState.js
```diff
@@ -1,2 +1,525 @@
...
+    filters: [],
+    sortBy: '',
+    docs: []
+  }
 };
+
+// mock communication records
+// based on GET /prm/api/communicationrecords/?partner_id=1
+state.records.docs = [
+  {
```
koonsies: Sweet. Wanted to make sure it matched Troy's API and this seems to do so perfectly.

iartemiev:
The only thing I left out from @koonsies suggestions from the JIRA ticket is Communication Record Count. I thought that instead of having another prop to maintain in the state, we can just get `.length` on `state.records.docs`... But I'm open to other opinions on this matter.

koonsies:
We get returned a value for "scored_records" and "all_records". We could use the "scored_records" return to display The "Rating based on X reviews" or whatever. All_Records may not be as useful, as we apply filters, the count of the records displayed will change.

Speaking of filters. We probably want to design a way to store active filters and sort methods. One way is to store that in the state and have the ReportList display based on the active filters/sorts. What are your thoughts on that?

iartemiev:
I agree and will throw in a `scoredRecords` prop.

Also agree re: filters. Could be something like:
```javascript
filters: [
  {
    name: 'communicationType',
    value: 'phone'  
  },
  {
    name: 'assessmentRating',
    value: 5
  }
]
```

koonsies:
That sounds good. How about sorting? I believe there will be sorting? I could see something similar to your filter design.. {name: <ColumnName>, value:<Direction>}

iartemiev:
I'd probably do:
```javascript
sorting: {
  key: columnName,
  direction: 'desc' // or 'asc'
}
```

iartemiev:
I'd probably do:
```javascript
sortBy: {
  key: columnName,
  direction: 'desc' // or 'asc'
}
```

koonsies:
That's fine by me.

iartemiev:
Or, could keep it slightly more consistent by doing: 
```javascript
filters: [
  {
    key: 'communicationType',
    value: 'phone'  
  },
  {
    key: 'assessmentRating',
    value: 5
  }
],
sortBy: {
  key: columnName,
  direction: 'desc' // or 'asc'
}
```

PR #3293, opened 2017-07-25
PD-3514 fixing forwarded emails not creating PRM contact records

vigowei:
## What's this PR do?
de-duplicate contact emails in email processing

Remember to add `QC_EMAIL_KEY`


deploy/settings.myjobs_qc.py
```diff
@@ -97,6 +97,8 @@
 EMAIL_HOST_USER = QC_EMAIL_HOST_USER
 EMAIL_HOST_PASSWORD = QC_EMAIL_HOST_PASSWORD
 
+EMAIL_KEY = QC_EMAIL_KEY
```
mklauber: Where is this coming from?  Should be imported or defined somewhere?

vigowei: I think this should be added to the secrets file, @almathew can confirm

vigowei: I previously thought it would come from default settings but secrets makes more sense

almathew: I added it to the secrets file yesterday. It is a secret.

PR #3299, opened 2017-07-27
[PD-3465] Stop making new companies

almathew:
## What's this PR do?
Prevents company creation just because a company with multiple business units has a business unit with a different name associated with it.

The docstring for add_company is pretty descriptive about what the expected outcome should be. Does the function match the docstring? Are the assumptions in the docstring sensible?

Set up a test scenario similar to `import_jobs.tests.test_utils.AddCompanyTest.test_no_new_company_created` (which is actually annoyingly complicated to do - I can't actually come up with a way that doesn't involve faking job files).


import_jobs/utils.py
```diff
@@ -1,88 +1,76 @@
...
+            'action': 'switching company',
+            'buid': bu.id,
+            'company_id': existing_company.id,
+            'company_name': existing_company.name,
+        }))
         bu.company_set = [existing_company]
-        company = existing_company
+        return existing_company
+
+    company = bu.company_set.first()
```
mklauber: typo

import_jobs/utils.py
```diff
@@ -1,88 +1,76 @@
...
+
+    logger.info(json.dumps({
+        'action': 'creating company',
+        'buid': bu.id,
+        'company_name': bu.title,
+    }))
+    slug = slugify(bu.title)
+    logo_url = '//d2e48ltfsb5exy.cloudfront.net/100x50/seo/%s.gif' % slug
+    try:
+        company = Company.objects.create(name=bu.title, company_slug=slugify(bu.title), logo_url=logo_url)
```
mklauber: Use `slug` instead of re slugifying bu.title?

PR #3304, opened 2017-07-28
[PD-3385] tjd-distribution-location-and-rule-summary

mwarnsleyDE:
## What's this PR do?
This PR checks for the rules applied and depending on that basis determines what message to show the user in the Distribution Details box concerning where the emails are going to be sent

screen shot 2017-07-28 at 11 37 38 am

Start inside of the TJD Application

Pull down the repository. Go into TJD and select one of the Organizations. It would be easier if you selected one that was a custom CBO so you can edit the rule, locations, etc if need be. Go into edit or view a contact. On the Manage contact page you will see the Distribution Details on the right hand side. If you mess with the Rules or Locations and add or delete them, depending on the criteria settings, the message should change depending on what is selected.

N/A


mwarnsleyDE:
@aljordan82 Need some collaboration on this a bit to get any ideas from you as far as the way this looks or feels, etc.

aljordan82:
This looks good, the only issue I am seeing is the new div is showing up in sub-routes like:
https://secure.my.jobs/cbo/view/cbo/231/contact/305/loc/682
But I don't think is a major issue.

PR #3306, opened 2017-07-28
[PD-3539] Saved search sending improvements

almathew:
## What's this PR do?
Switches saved searches and saved search digests to send via id instead of the object itself.
Also adds some kwargs that should help in debugging issues that come up, discovering potential issues failed partner saved search sending, and detecting saved searches that are failing often.

tasks.py is where the new logic is at, although mysearches/models.py needed updated in a few places to use id instead of objects.

- Create a new saved search (after 10 AM) and confirm that an email pops up in your console.
- Create a saved search that should send today and attempt sending it using the tasks (after 10 AM). An email should pop up in the console then too.


tasks.py
```diff
@@ -161,24 +160,41 @@ def assign_ticket_to_request(key, access_request):
...
+            'search_id': search_id,
+        }.update(kwargs)))
+        return
+
     try:
-        newrelic.agent.add_custom_parameter("search", search)
         search.send_email()
     except Exception as e:
-        logger.error("Unable to send saved search for Saved Search ID: %s", search.pk)
+        logger.error(json.dumps({
```
tdphillips: `.update()` updates in place - this results in `json.dumps(None)`.

almathew: Good catch. Fixed.

PR #3312, opened 2017-07-31
[PD-3542] New Analytics Models for Traffic

darrint:
## What's this PR do?

Adds three new tables for Analytics and tests that they work. The tests are trivial but I added them because of the use of raw SQL in the migration.

Make sure the SQL and models are self-consistent.

* `dkm makemigrations` doesn't try to do new work on the tables.
* The tests pass.
* The sql for each field matches the model definition. (type, nullness, etc)

Also:

* ~~Should I make time and hostname not null?~~ Yes.

Run the tests.

na.


analytics/migrations/0020_traffic_tables.py
```diff
@@ -0,0 +1,431 @@
...
+  user_location_city varchar(255),
+  user_location_country_code varchar(255),
+  user_location_zip_code varchar(255),
+  user_location_state varchar(255),
+  user_location_lat float,
+  user_location_long float,
+  user_location as geography_point(user_location_long, user_location_lat)
+    persisted geographypoint,
+  time datetime(6) not null,
+  hostname varchar(255) not null,
```
vigowei: this field has `null=True` in models.py: Line#552

PR #3320, opened 2017-08-02
Conditional JS inclusion in PA

koonsies:
## What's this PR do?
Change base-react to conditionally include the core CSS files


templates/base-react.html
```diff
@@ -39,11 +39,13 @@
...
-        {% endcompress %}
+        {% block core-js %}
+            {% compress js %}
+                <script type="text/javascript" src="{% static "utils.184-4.js" %}"></script>
+                <script type="text/javascript" src="{% static "custom.166-06.js" %}"></script>
+                <script type="text/javascript" src="{% static "svg4everybody.min.js" %}"></script>
+            {% endcompress %}
+        {% endblock %}
         <script>
             svg4everybody();
```
aljordan82: Should we add the script tag inside the block core-js?
Also, I am not sure if we need it in PA.

koonsies: The svg4everybody(); call? Do you mean so that it can be overwritten by PA?

aljordan82: so, if we don't include it in PA, it won't undefined.
I don't see why we need svg4everybody.js in PA

koonsies: Good call.

One second.

PR #3324, opened 2017-08-03
Pd 3540 assessments css clean up

aljordan82:


gulp/src/partnerassessment/components/Records/RecordListFilter.jsx
```diff
@@ -4,16 +4,18 @@ import {connect} from 'react-redux';
...
-  {id: 2, display: '★★' },
-  {id: 3, display: '★★★' },
-  {id: 4, display: '★★★★' },
-  {id: 5, display: '★★★★★' },
+  {id: null, display: <StarDisplay score={0}/> },
+  {id: 1, display: <StarDisplay score={1}/> },
+  {id: 2, display: <StarDisplay score={2}/> },
+  {id: 3, display: <StarDisplay score={3}/> },
+  {id: 4, display: <StarDisplay score={4}/> },
+  {id: 5, display: <StarDisplay score={5}/> },
```
aljordan82: Would it be better to reverse the order and append `and up` to starDisplay?

gulp/src/sass/_base/components/filterBar.scss
```diff
@@ -1,20 +1,27 @@
...
   min-height: 35px;
-  padding-top: 5px;
+  padding: 5px 15px;
   clear: both;
-
-  .btn-clear-all {
-    margin-top: 5px;
-    margin-right: 8px;
-  }
+  border-top: 1px solid #d8d8d8;
```
aljordan82: Make #d8d8d8 a variable, it is used a number of times

gulp/src/partnerassessment/components/Form/AssessmentForm.jsx
```diff
@@ -60,7 +60,7 @@ class AssessmentForm extends Component {
       </FormGroup>
 
       <FormGroup className="assessment-rating" validationState={this.state.ratingError}>
-        <ControlLabel>Rate Effectivness</ControlLabel>
+        <ControlLabel>Rate Effectivness *</ControlLabel>
```
mwarnsleyDE: Rate Effectivness should be Rate Effectiveness

gulp/src/common/ui/StarDisplay.jsx
```diff
@@ -21,7 +21,28 @@ const StarDisplay = ({score}) => {
...
+    </div>
+  );
+};
+
+const StarDisplay = ({score, menuItem}) => {
+  return menuItem ? renderMenuItem(score) : renderDisplay(score);
+};
+
 StarDisplay.propTypes = {
+  menuItem: PropTypes.bool,
```
aljordan82: I don't think this belongs to the starDisplay, the plus sign could be appended to the <starDispay/> jsx

iartemiev: It gets complicated getting the spacing right for the + next to the star otherwise... 
If it's appended to `<StarDispaly/>`, the + will be outside of `div.star-ratings > div.star-ratings-top`

Also, for the `menuItem` context, we need to not render `star-ratings-bottom` at all (since we'er not showing the grayed out stars).

So I can't think of a way to do that without just passing a param or without creating a whole new component. 

If you have a solution, I'll be happy to consider it

PR #3328, opened 2017-08-04
[PD 3532] improved job

mwarnsleyDE:


aljordan82:
Typing in the ReqID is really slow. I don't think this will pass QC.

aljordan82:
I've noticed that text field values are not persisted. try the 'city' field.

myreports/datasources/tjd_report.py
```diff
@@ -316,7 +240,7 @@ def adorn_filter_items(self, company, found_items):
 
             adorned[u'Location']['state'] = {
                 found['value']: found
-                for found in [ result for result in self.help_state(company, data_source_filter, '')]
+                for found in [result for result in self.help_state(company, data_source_filter, '')]
```
vigowei: ````found['value']: found for found in [result for result in self.help_state(company, data_source_filter, '')]```
is the same as:
```found['value']: found for found in self.help_state(company, data_source_filter, '')```

and Line#248 @tejaswinigutti31

PR #3330, opened 2017-08-04
[PD-3558] query redirect db calls

tejaswinigutti31:
## What's this PR do?
Reduce number of DB calls when Query Redirect is called
test for number of DB counts in seo/tests/test_middleware


tejaswinigutti31:
Using select_related() will relatively reduce the number of DB calls. But it is in a state where addition DB calls are made probably by the middleware in general

seo/tests/test_middleware.py
```diff
@@ -163,3 +166,18 @@ def test_query_redirect_with_queries(self):
         )
         self.assertEqual(response.status_code, 200,
                          "Query redirect took place on the wrong site")
+
+    def test_num_of_db_calls_for_redirect(self):
+        from seo.models import SeoSite
```
almathew: Any reason you're doing the import here?
`self.test_site` might be the value you're looking for anyway.

almathew:
:+1: Looks good to me other than the one import in the test.

PR #3338, opened 2017-08-08
Use https resources on static error page

tdphillips:
## What's this PR do?
Uses an https link to css on our error page. js.nlx.org doesn't have a cert so I went with the long form url.

@fezick, would this be appropriate?


tdphillips:
js.nlx.org doesn't have a certificate.

PR #3339, opened 2017-08-08
Migration for adding is_bot to site traffic tables

tejaswinigutti31:
## What's this PR do?
Creates a migration to alter tables to include a new column is_bot
Migration analytics/0021


analytics/migrations/0021_add_is_bot.py
```diff
@@ -0,0 +1,58 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('analytics', '0020_traffic_tables'),
+    ]
+
+    operations = [
+
+        migrations.RunSQL(
+            sql="ALTER TABLE analytics_goalevent ADD COLUMN is_bot boolean;",
```
darrint: `is_bot boolean not null`

analytics/migrations/0021_add_is_bot.py
```diff
@@ -0,0 +1,58 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('analytics', '0020_traffic_tables'),
+    ]
+
+    operations = [
+
+        migrations.RunSQL(
+            sql="ALTER TABLE analytics_goalevent ADD COLUMN is_bot boolean NOT NULL DEFAULT FALSE;",
```
almathew: If we are going to have a default, can we set that default in the model too? You might need to re-create the model - Django likes to know when things have changed.

PR #3342, opened 2017-08-09
Pd 3384 analytics remove custom.css

aljordan82:
Layout shouldn't change in secure.my.jobs. especially Analytics,


strempe:
So much better!

mwarnsleyDE:
HIGH FIVE

PR #3346, opened 2017-08-10
Remove last visited network site

aljordan82:
## What's this PR do?
Remove "last Network site"


almathew:
The python test that's failing can be deleted.

tdphillips:
I think I'm done. There wasn't as much to remove as I thought - some of it is necessary for other features, which I thought we were also removing.

PR #3351, opened 2017-08-11
[PD-2783] Put defer attribute back on script tags.

almathew:
## What's this PR do?
What it says on the box.

The work is not my own. I am merely the resolver of merge conflicts.


templates/seo_base_v2.html
```diff
@@ -164,12 +164,12 @@
...
 {% else %}
   {% compress js %}
-      <script src="{% static "bundle/seo_base_scripts.js" %}"></script>
+      <script defer src="{% static "bundle/seo_base_scripts.js" %}"></script>
   {% endcompress %}
 {% endif %}
 
 {% if site_config.use_secure_blocks %}
-    <script>
+    <script defer>
```
aljordan82: I don't think this will have any effect because there isn't an `src` attribute

almathew: @darrint any comments?
I have literally no idea.
This is not my code @JLMcLaugh just wanted to finally be making some progress here so I unearthed the branch.

almathew: @darrint has taught me some js and now it's gone.

PR #3352, opened 2017-08-11
[PD-3506] Add display_name for AppAccess

almathew:
## What's this PR do?
Adds display name to AppAccess.
Switches all the user-facing AppAccess names (that I could find) to display_name.


myjobs/migrations/0003_appaccess_display_name.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.10 on 2017-08-11 19:19
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def forward(apps, schema_editor):
```
tdphillips: This might need the `@only_on_default` decorator.

myjobs/models.py
```diff
@@ -1048,9 +1048,11 @@ class AppAccess(models.Model):
     may belong to many roles, it may only be assigned one app access.
     """
     name = models.CharField(max_length=50, unique=True)
+    display_name = models.CharField(max_length=50)
 
     def __unicode__(self):
-        return self.name
+        print self.display_name
```
tdphillips: Left over from development?

almathew:
Anyone have any tips on the migration ordering issue? I have a feeling this is going to be more difficult to fix than adding the field was.

PR #3358, opened 2017-08-15
Only update what is needed.

almathew:
## What's this PR do?
Diffs updates against jobs already in Redis, in hopes that we're only updating what's needed.


import_jobs/utils.py
```diff
@@ -222,3 +223,63 @@ def update_business_unit_modified_dates(business_unit, crawled_date,
     if updated:
         business_unit.date_updated = datetime.datetime.utcnow()
     return business_unit
+
+
+def should_update(jobs):
```
mklauber: I feel like this is basically a pointless method, given all the work is being done in diff_jobs.

almathew: I don't disagree, but it was an easy "this does exactly what I need it to" access method.

mklauber: Is the return of `diff_jobs` a tuple for testing purposes?

almathew: ....I get your point. Fine. I'll change it.

almathew: Your opinions have been accounted for.

PR #3375, opened 2017-08-22
[PD-3487] Add PartnerAssessment handling to reporting

koonsies:
## What's this PR do?
This PR is the combination of a few different tickets. It does the following:

1) Adds Score, Review to the Communication Reports
2) Adds a new report category for PRM, "Partner Assessments"
3) Adds a datasource for exporting Partner Assessment Reports
4) Adds a simple mechanism to waffle-gate these new reporting changes.

Django Admin
* Navigate to the bottom of the admin page, to "Waffle"
* Click the "Add" button next to "Switches" (NOT flags)
* Add a new switch called "partner assessment"

Reporting App
* Select PRM
* Select Communication Record (default selection)
* Add score filter.
* Export a report
* Check that it contains score/review fields
* Select "Partner Assessment" from Report Category
* Ensure the form/export works properly.

Due to difficulties passing the request object into certain functions, I spoke with Jason Sole and we decided that we did not need user-specific waffle functionality for this app. Instead of using flags (user specific), we are using switches (on or off for all).

Also, there is a conspicuous lack of tests for the new datasource. I have created a spin-off ticket [PD-3633](https://directemployersfoundation.atlassian.net/browse/PD-3633) for that and am currently working on those tests.


myreports/schema.py
```diff
@@ -542,8 +551,27 @@ def ds_aggregated(columns):
 )
 
 
-all_reports = AllReports(
-    report_intentions=[rit_prm, ritg_analytics_group, rit_tjd])
+def all_reports():
```
koonsies: @darrint what are your thoughts on this? It's a pretty straight forward way to add/remove stuff from the all_reports object. Not a clever approach, but it seems to work.

aljordan82:
It doesn't look like the report is including records with null rating.

gulp/src/common/ui/ScoreReportFilter.jsx
```diff
@@ -0,0 +1,30 @@
...
+import Select from 'common/ui/Select';
+
+const defaultChoice = {
+  value: '',
+  display: 'Select a Score',
+};
+
+const choices = flatten([
+  defaultChoice,
+  map(['1', '2', '3', '4', '5'], i => ({
```
aljordan82: What about null rating?

myreports/schema.py
darrint: Looks about right to me. @tejaswinigutti31?

tejaswinigutti31: This looks good

tdphillips:
It would have been nice to have been included in the flag/switch conversation. Flags aren't necessarily per-user. They would allow us to enable a feature for all staff accounts with the press of a button. Then staff could get used to the new feature in preparation for questions from members.

koonsies:
That's my fault. I discussed it with Jason a few weeks back and should have run it by you. Should we need to revisit this issue and come to a different conclusion, I'm more than willing to do so.

PR #3378, opened 2017-08-23
Add new nav item to all PRM templates [PD-3629]

tdphillips:
## What's this PR do?
Add a new navigation element to PRM. We decided to hard-code it because we don't do Django anymore.
2017-08-23-084928_984x786_scrot


templates/mypartners/nonuseroutreach/outreach_inbox_management.html
```diff
@@ -1,98 +0,0 @@
-{% extends "base.html" %}
```
tdphillips: This isn't in use. We did this then decided that everything should be React.

PR #3386, opened 2017-08-24
Pd 3616 user mgmt navigation

strempe:
## What's this PR do?
- Moves navigation for DE 'Manage Users' Application to a top Sub Nav.
- Moves task actions to the top of the page. (ie: Add User)

screen shot 2017-08-24 at 11 55 17 am

screen shot 2017-08-24 at 12 18 39 pm

- Apps > Manage Users

- Each task in the Sub Nav should change the view accordingly when clicked 
- The tasks 'Users' and 'Role' should toggle to the actions, 'Add User' and 'Add Role'
- Actions should also appear in top right corner of the contents container 

Mobile
- Tasks are shifted to a dropdown
- Clicking a task in the Sub Nav dropdown should change the view accordingly


mwarnsleyDE:
Nice Job Sara

koonsies:
When in mobile view with a longer company name, I get some text overlap. 
screen shot 2017-08-25 at 10 05 43 am

koonsies:
I know it's a bit late-in-the-game to comment on the design, but I find it a bit disconcerting that when we select a link (say "Users" in this example), that the link to User's is immediately replaced by "Add User". It's pretty subtle and I'm not sure many users will understand why it happened. Not to mention, you lose a bit of the feeling of knowing "where you are".

Could we, perhaps, do a submenu that "slides" out of the clicked link? I made a really, really bad mockup of what I mean. User would remain visible, but clearly selected, and the subnav would appear beside it, clearly differentiated from the main nav entries.

sillymockup

Perhaps this has already been discussed and discarded. Just checking if it's a possibility.

PR #3389, opened 2017-08-25
ES Job Source, Optional facet by tag rather than query.

almathew:
I'm putting this branch up to be merged into a feature branch now that I have others working with me and can't just try out ideas on my own.
Plus this branch is already ridiculously large. 

## What's this PR do?
- Introduces ES job source. ES is required because it will be used to handle tagging. 
- A beginners attempt at tagging and faceting on tags. I'd prefer if these were both seriously reviewed, but also accept that it's no where near the final version of code. Neither of them are impacting and they're both subject to significant change. At this point I'm looking for someone to tell me I'm doing something very, very wrong, or going in a terrible direction. 

Hopefully it looks the same way it always did.

Focus on elasticsearchjobsource.py as a final product
Am I going in the right direction with tagging and faceting?

Toy with importing things using elasticsearchjobsource


seo/views/search_views.py
```diff
@@ -632,62 +632,6 @@ def posting_stylesheet(request, cid=None, css_file="posting-stylesheet.173-18.cs
 
 
 @custom_cache_page
-def job_listing_nav_redirect(request, home=None, cc3=None):
```
almathew: There were some changes I needed to make here, but I opted to delete it instead, since it is completely unused.

PR #3391, opened 2017-08-25
Use our other method of activity determination [ST-4335]

tdphillips:
## What's this PR do?
Migrates our permission management from `@requires` to `user.can` as `@requires` is an all-or-nothing deal.
"read tjd" permissions were required when previewing PRM reports.


myreports/views.py
```diff
@@ -779,12 +779,15 @@ def get_default_report_name(request):
...
 
 
-@requires('read partner', 'read contact', 'read communication record', 'read tjd')
 @require_http_methods(['GET'])
 def old_report_preview(request):
     company = get_company_or_404(request)
     report_id = request.GET.get('id', 0)
     report = get_object_or_404(DynamicReport, pk=report_id)
+    if category == 'tjd_reports':
+        request.user.can(company, 'read tjd')
```
tdphillips: `user.can` raises a 404, just like `@requires`. If it instead returned a 404, this wouldn't work.

PR #3400, opened 2017-08-29
Some multifilter analytics API

darrint:
## What's this PR do?

Modifies the `dynamic` api in Analytics and the front end to provide better sense of where we are in the dimension list when querying. It also adds a `multifilter` when one is supposed to be displayed.

The new report schema changes I made are worth questioning. These are the bulk of the changes.

`analytics/api.py` contains most of the back end code for the new multifilter control. That control also requires the "group_by" column to be a plural entity.

Most of the changes are a result of the two items above.

Set the "new-site-traffic" switch.

`make_data` generates site traffic test data so you can use that to see working charts.

Make sure that the analytics for traffic in particular aren't broken.

Check that exports still work.

Check the "Source" and "Technology" 

Turn off the "new-site-analytics" switch and make sure that the old site traffic analytics dimensions still work.


aljordan82:
Please add some padding for the buttons,
IE specific: Search field is not displayed correctly, Adding "display: flex" to the .table-search container should fix the issue.

screen shot 2017-09-06 at 1 57 21 pm

gulp/src/sass/analytics/_table.scss
```diff
@@ -1,368 +1,381 @@
...
-       .active-selected {
-         background-image: none;
-         background-color: #337ab7;
-         border-color: #2e6da4;
-         color: $white;
-         letter-spacing: 0.6px;
-         outline: 0;
+        &:hover {
+          background-image: none;
+          background-color: $red;
```
aljordan82: Can you change the hover to a more subtle color? also, the hover is only on the selected buttons and not all button states
Is "background-image" needed?

gulp/src/sass/analytics/_table.scss
```diff
@@ -1,368 +1,381 @@
...
+      .non-active {
+        letter-spacing: 0.6px;
+      }
+    }
+  }
+  .table-search {
+    margin-top: 0;
+  }
+  .filter-container {
+    border-top: 1px solid #ddd;
```
aljordan82: Please use a variable instead of #ddd

gulp/src/analytics/actions/sidebar-actions.js
```diff
@@ -34,19 +43,11 @@ export function doSwitchMainDimension(mainDimension) {
...
-        getSources = await api.getAvailableCombos(mainDimension);
-        dispatch(generateTrafficSources(getSources));
-      }
       dispatch(switchMainDimension(allLoadData));
-      dispatch(setMainDimension(mainDimension));
       dispatch(markNavLoadingAction(false));
       dispatch(doBumpHistory());
     } catch (error) {
-      dispatch(errorAction(error.message));
+      throw error;
```
aljordan82: We probably should keep dispatching the errorAction for now.

gulp/src/analytics/actions/spec/sidebar-actions.test.js
```diff
@@ -7,45 +7,80 @@ import createReduxStore from '../../../common/create-redux-store';
 import {combineReducers} from 'redux';
 
 class FakeApi {
-  getMainDimensionData() { }
+  getMainDimensionData() {}
 }
 
-describe('headerReducer', () => {
+fdescribe('headerReducer', () => {
```
aljordan82: Remove the "f"

vigowei:
clicking on bar graph doesn't trigger a drilldown.

myreports/datasources/goal_events.py
```diff
@@ -283,10 +283,17 @@ def run_gather_timeline_data(self, company, filter_spec, field=None, partial=Non
 
     def apply_filters(self, company, filter_spec, field=None, partial=None):
         if field and partial:
-            query_field = self.get_partial_field(field)
-            modified_filter_spec = filter_spec.clone_without(field)()
-            query_set = self.filtered_query_set(company, modified_filter_spec)
-            query_set = query_set.filter(**{query_field: partial})
+            if isinstance(field, list):
+                search_condition = reduce(__or__, [{self.get_partial_field(f.column_name): partial} for f in field])
```
vigowei: wouldn't this throw an exception?

vigowei: maybe: 
```from django.db.models import Q
reduce(__or__, [Q({self.get_partial_field(f.column_name): partial}) for f in field])```

gulp/src/sass/analytics/_table.scss
tejaswinigutti31: discussed offline. Just going to leave in

darrint:
Added https://directemployersfoundation.atlassian.net/browse/PD-3663 to address the padding on IE problem.

PR #3401, opened 2017-08-29
Add "Partners" link to all PRM navigation, add nav to partner saved search

tdphillips:


strempe:
The sub navigation should stay consistent on the NUO, Partner, Other

tdphillips:
I think I addressed the bits that don't deal with NUO.

PR #3407, opened 2017-08-30
Modify topbar to check truthiness of company rather than user.roles.exists

tdphillips:
## What's this PR do?
Hopefully fixes topbar. I was unable to reproduce, so ¯\\\_(ツ)\_/¯


myjobs/views.py
```diff
@@ -526,21 +526,9 @@ def cas(request):
 def topbar(request):
     callback = request.GET.get('callback', request.POST.get("callback"))
     use_v2 = request.GET.get('v2', request.POST.get('v2', 0))
-    user = request.user
     impersonating = request.GET.get('impersonating', request.POST.get('impersonating'))
 
-    if not user or user.is_anonymous():
```
tdphillips: This provides no value whatsoever. `topbar.html` only renders topbar items if the user is authenticated.

PR #3409, opened 2017-08-30
[PD- 3642] excluding standard facets in calculating total jobs count

tejaswinigutti31:
## What's this PR do?
Standard facets are not included in total jobs count query
total job counts calculation in seo/cache.py


seo/helpers.py
```diff
@@ -1383,3 +1383,27 @@ def get_company_thumbnail(filters):
...
+    sqs = sqs_apply_custom_facets(custom_facets, sqs, exclude_facets)
+    sqs = _sqs_narrow_by_buid_and_site_package(sqs, buids=jsids)
+
+    sqs = sqs.order_by(sort_order_mapper.get(sort_order, '-score'))
+
+    # The boost function added to this search query set scales relevancy scores
+    # by a factor of 1/2 at ~6 months (1.8e-11 ms) in all future queries
+    sqs = sqs.bf('recip(ms(NOW/HOUR,salted_date),1.8e-9,1,1)')
+
+    if facet_limit > 0:
```
almathew: These checks and parameters shouldn't be needed since we're explicitly eliminating facets.

seo/helpers.py
```diff
@@ -1383,3 +1383,27 @@ def get_company_thumbnail(filters):
...
+                   sort_order='relevance'):
+    """
+    Used to calculate total job counts without applying standard facets
+
+    """
+    sqs = DESearchQuerySet()
+    sqs = sqs_apply_custom_facets(custom_facets, sqs, exclude_facets)
+    sqs = _sqs_narrow_by_buid_and_site_package(sqs, buids=jsids)
+
+    sqs = sqs.order_by(sort_order_mapper.get(sort_order, '-score'))
```
almathew: No need to sort. We're only getting a count.

seo/helpers.py
```diff
@@ -1383,3 +1383,27 @@ def get_company_thumbnail(filters):
             return None
     else:
         return None
+
+def get_jobs_count(custom_facets=None, exclude_facets=None, jsids=None,
+                   facet_limit=250, facet_sort="count", mc=1,
+                   sort_order='relevance'):
```
almathew: You probably only need custom_facets, exclude_facets, and jsids passed into the function.

seo/helpers.py
```diff
@@ -1383,3 +1383,27 @@ def get_company_thumbnail(filters):
...
+
+    # The boost function added to this search query set scales relevancy scores
+    # by a factor of 1/2 at ~6 months (1.8e-11 ms) in all future queries
+    sqs = sqs.bf('recip(ms(NOW/HOUR,salted_date),1.8e-9,1,1)')
+
+    if facet_limit > 0:
+        sqs = sqs.facet_limit(facet_limit)
+    sqs = sqs.facet_sort(facet_sort).facet_mincount(mc)
+
+    return sqs
```
almathew: You can switch this to return sqs.count() and then the function name might be a bit more accurate.

I say this because, it's a bit of a trivial optimization, but if you add sqs[0:0] before it will make the solr query a slight bit more efficient. [0:0] would be telling it not to get any rows.

so 
```
sqs[0:0]
return sqs.count()
```

and then fix the call in cache.py to not return a count. Maybe? This one is a bit opinionated and would have to be tested.

PR #3411, opened 2017-08-30
Semantic satiation

tdphillips:
## What's this PR do?
Hotfixes our hotfixes to fix bugs introduced by said hotfixes


templates/includes/topbar.html
```diff
@@ -6,6 +6,16 @@
 {% compress css %}
   <link href="{% static "topbar.myjobs.css" %}" rel="stylesheet" type="text/css">
 {% endcompress %}
+{% if user.is_authenticated and request.session.keys|length %}
+    {% is_a_group_member company user "Employer" as group_member %}
```
tdphillips: Moved up from the bottom as we're using the `companies` variable.

templates/includes/topbar.html
```diff
@@ -6,6 +6,16 @@
 {% compress css %}
   <link href="{% static "topbar.myjobs.css" %}" rel="stylesheet" type="text/css">
 {% endcompress %}
+{% if user.is_authenticated and request.session.keys|length %}
```
tdphillips: Added this. It worked without this before one of our hotfixes. Why does it not work now? Because we're not doing annoying shenanigans relating to replacing users in the topbar view, so `get_company_name` has the potential of receiving an AnonymousUser.

templates/includes/topbar.html
```diff
@@ -21,7 +31,7 @@ <h3 id="topbar-logo">My.jobs</h3>
                 <div class="menu-container employer-menu" role="menu">
                     <ul>
                     {% for menu in menus.employer %}
-                        <li {% if menu.submenus %}class="has-drop"{% endif %}>
```
tdphillips: We don't have submenus anymore.

templates/includes/topbar.html
```diff
@@ -21,7 +31,7 @@ <h3 id="topbar-logo">My.jobs</h3>
...
+                        <li {% if companies|length > 1 %}class="has-drop"{% endif %}>
```
tdphillips: We add the companies in topbar.js. Don't ask me why we don't just do so here.

myjobs/templatetags/common_tags.py
```diff
@@ -259,36 +259,14 @@ def get_menus(context):
         return []
 
     employer_menu = {}
-    beta_submenu = []
```
tdphillips: With the introduction of waffle, we no longer require the beta menu.

myjobs/templatetags/common_tags.py
```diff
@@ -259,36 +259,14 @@ def get_menus(context):
...
-    try:
-        can_view_analytics_info = user.can(
```
tdphillips: Moved this closer to its point of use now that it is no longer required for the beta menu.

myjobs/templatetags/common_tags.py
```diff
@@ -259,36 +259,14 @@ def get_menus(context):
...
-            company, "view analytics")
-    except MissingAppLevelAccess:
-        can_view_analytics_info = False
 
-    if user.is_superuser and 'apps.directemployers.org' in user.email:
```
tdphillips: Removed the if. Deindented its contents.

templates/includes/topbar.html
koonsies: Unrelated, but that is_group_member function is strange. Takes company, user, and group but only used company.. *shrug*

tdphillips: It's a relic from the past and should probably just be removed.

PR #3418, opened 2017-09-01
Add common react chunks

aljordan82:
## What's this PR do?
Extract the common js file used in react apps into a react-common.js
file.
There is no need to include common js libraries into each app.

File sizes before and after.
screen shot 2017-09-01 at 3 17 59 pm

screen shot 2017-09-01 at 4 19 12 pm


iartemiev:
Will Django add a hash to the filename (for cache-busting purposes)? 
If not, it can be done in the webpack config, by adding `[chunkhash]` to the file name.
This would only need to be done when building for production, so we'd need to add a check for that

PR #3432, opened 2017-09-06
Remove unused code.

almathew:
## What's this PR do?
Remove unused code.

It's removing functionality, so ensure that I didn't remove anything important?


seo/admin.py
```diff
@@ -1257,40 +1258,6 @@ def check_inline_instance(obj, req):
         return obj.inline_instances
 
 
-def copy_to_qc(modeladmin, request, queryset):
```
tdphillips: Is this unused? It's in the dropdown on various admin pages.

almathew: It's in the dropdowns, but no one actually uses it.

PR #3433, opened 2017-09-07
PR Checklist

darrint:
## What's this PR do?

Adds a code review checklist to the GitHub PR form.

Look at the items on the list. Are they good? Are there too many?

Should we merge this, we'll also have some sense of what we expect each other to do with it. Putting it on the PR probably means we at least expect someone reviewing and/or merging code to look at it. 

Some examples of things to think about:

* I used check boxes. Check boxes might be a problem since we might add changes that require reexamining an item. Maybe no one will bother to look since the box is already checked. Maybe that's not a problem.
* There are 16 items on the list and no front end items there yet. Is that so many items that we are likely to not look at the list at all?
* It's a retrospective and we think adding an item to this list might be helpful for awhile. Which item would you remove to make room for the new item?

na

na


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,32 @@
 
 ## Any background context you want to provide?
 
+## PR Checklist
+
+### API
+
+- [ ] API is  consistent about sending structured values to the front end. -- If a simple value isn't enough, follow the value, display, description convention. -- Example: {value: 3, display: "Bob Vance", description: "Bob Vance of Vance Refrigeration is civic leader and stand up guy."} Anti-example: {key: 3, value: "Bob Vance", text: "Bob Vance of..."}
```
almathew: I know that this is based on the Django forms thing, but I'm going to kind of disagree with this...

For instance, if I'm trying to supply a list of user permissions the sensible decision would be to just turn the user model into a dictionary and go with: 
```
{'id': 1, 'email': 'dev@apps.directemployers.org', 'permissions': ['write sonnets', 'read sci-fi']}
```
which goes against this convention.

Even if the permissions are more verbose, the following format is more in line with Django's model system:
```
{
    'id': 1, 
    'email': 'dev@apps.directemployers.org', 
    'permissions': [
        {'id': 1, 'key': 'write sonnets', 'description': 'Can write amazing sonnets.'}
        {'id': 3, 'key': 'read sci-fi', 'description': 'Can read basic sci-fi novels.'}
]
```
I guess the more that I'm typing... this method seems really, really specific? It only accounts for the key, value, description patterns, but not ones more complex than that, and directly contradicts Django's suggested object formatting in the process.

It just seems like a very, very limited checkbox.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,32 @@
...
+- [ ] API doesn't invent new conventions unnecessarily. -- Example: View uses existing utilities like ApiValidator for communicating errors to the client.
```
almathew: How will I know when I'm doing this?
I'd say the biggest issue here is lack of knowledge about existing conventions, so if both the implementer and reviewer don't know about a convention, then this is ignored. I don't know that there are many instances that we are knowingly ignoring conventions.

Also, at this point what even is the convention? I know we discussed moving away from ApiValidator, for a convenient example.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,32 @@
...
+
+### Separation of Concerns
+
+- [ ] Implement back end "business logic" in plain python, Django model classes, and Django forms, not Django views or other frameworks. -- Example: A few Django model classes have several methods named after the business problem they solve. View methods for the API and Celery tasks are short. Anti-example: A few Django model classes exist which only declare fields. Extensive code is written in Django views, Django signal handlers, and Celery tasks.
```
almathew: I'm not sure I agree with this necessarily. More discussion is required.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,32 @@
...
+- [ ] Separate concerns of UI, api, and model are in separate code units. -- Example, model method returns a query set. View code converts query set to relevant json document. Anti-example: model method takes request as a parameter and returns a python dict ready for sending to the client as a json doc.
```
almathew: Possibly pedantic: model methods should ~~not~~ rarely return querysets.
methods on django.db.models.Managers and django.db.models.QuerySets, sure, but not the models themselves.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,32 @@
...
+- [ ] Try to be very DRY about view code. When encoding for the front end be greedy about extracting plain python "view support" functions. -- Example: Reuse ApiValidator. Anti-example: Make up dict on the spot.
+
+### Tests
+
+- [ ] Test raw python business logic extensively.
+- [ ] Test Django models and business logic in them extensively.
+- [ ] Test Django views/integration lightly.
+- [ ] Test Django plain python view support functions extensively.
+- [ ] Use unittest.TestCase for speed and access to the testfast command when tested code does not reference Django.
```
almathew: As long as you're verrrrrry careful about it not referencing Django.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,32 @@
...
+- [ ] Use a Django TestCase or derivative when tests or set up reference Django models.
```
almathew: Use TransactionTestCase only when absolutely necessary.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,32 @@
...
+- [ ] API doesn't invent new conventions unnecessarily. -- Example: View uses existing utilities like ApiValidator for communicating errors to the client.
+
+### Separation of Concerns
+
+- [ ] Implement back end "business logic" in plain python, Django model classes, and Django forms, not Django views or other frameworks. -- Example: A few Django model classes have several methods named after the business problem they solve. View methods for the API and Celery tasks are short. Anti-example: A few Django model classes exist which only declare fields. Extensive code is written in Django views, Django signal handlers, and Celery tasks.
+- [ ] Separate concerns of UI, api, and model are in separate code units. -- Example, model method returns a query set. View code converts query set to relevant json document. Anti-example: model method takes request as a parameter and returns a python dict ready for sending to the client as a json doc.
+- [ ] Try to be very DRY about view code. When encoding for the front end be greedy about extracting plain python "view support" functions. -- Example: Reuse ApiValidator. Anti-example: Make up dict on the spot.
+
+### Tests
```
almathew: Two things I've been thinking about with regards to tests:

1. Less important since I think this is pretty decided, but fixtures vs. factories. More important, because it varies so much in our code, but when to use factories vs. using the model directly. I don't have an opinion on this. My biggest observation is that we use both pretty much randomly.
2. Being careful about the TestCase type that's being used as the base. So much of our test time is spent doing random or useless stuff during setup because when tests were written they just copied from somewhere else without really thinking about if they needed everything the base TestCase did.

tdphillips:
Added the "do not merge" label - we're discussing this tomorrow.

tdphillips:
I'm so glad we have to wait for tests to finish on this.

aljordan82:
It will be great if we decide what "Approved" means. was the code tested or just glanced over? 
and at what point a PR can be merged. I see some PR getting merged without a proper review (after hours PRs).

almathew:
Proposed solution - we stop approving pull requests. I think we all agree that it has so many meanings that it's meaningless at this point.

If you want to say it looks good to you and you're okay merging after someone else has looked at it, or after test pass, or you have no idea what it does but the code is very pretty, or whatever else, then just write a comment saying that.

PR #3439, opened 2017-09-07
[PD-3659] use waffle flag to control access to new site-traffic reporting

vigowei:
## What's this PR do?
Switch to using waffle flag (instead of switch) to give our staff access to the new site-traffic reporting features before rolling out to members.

image

Go to reports-> site-traffic, you should only see three reporting categories (Browser, job-seeker locations and OS) and they should still work as before. 
Create a waffle flag in Django admin, set values according to screenshot above
Click on the search icon next to `Users`, then locate the local user for testing, the user ID will be added. Save the flag. 
Go to reports -> site-traffic, you should see all new report categories on the left side. 

The `Everyone` option at the top overrides other settings, if set to `no`, even whitelisted users won't have access to the flagged feature, so make sure to set it to "Unkown" and then give access to specific users, or user groups.


myreports/schema.py
```diff
@@ -709,7 +710,7 @@ def all_reports():
         if rc_partner_assessment in rit_prm.report_categories:
             rit_prm.report_categories.remove(rc_partner_assessment)
     del rit_site_traffic.report_categories[:]
-    if switch_is_active('new-site-traffic'):
+    if flag_is_active(current_request(), 'new-site-traffic'):
```
tejaswinigutti31: Is there any possibility of passing request to all_reports()??

vigowei: yes, we will pass request to another function, but probably a good idea since these are async requests, I'll update

vigowei: I think we may have talked about this... request needs to be passed wherever all_reports() is used... except for reports that are not part of analytics. Since we are relying on request to determine if feature flag is active, I decided to only change the parts pertaining to analytics/site-traffic, that means making request an optional arg to `all_reports()`, and also passing request to `TempDynamicReport` objects (request is now used in `regenerate_data` calls and `search`)

myreports/schema.py
```diff
@@ -692,7 +692,7 @@ def ds_page_data(dimensions):
 )
 
 
-def all_reports():
+def all_reports(request=None):
```
darrint: I think we chose the greater of two evils instead of the lesser. This is "model" code and it shouldn't ever have to deal with request objects. Otherwise this code can't every be used for celery, for instance, which for reporting code is a big deal. We have to keep that door open.

See how the `resolve` method in this module deals with this. It asks for a user and company object since that's what it needs. This interface should also ask for just what it needs. What it needs the _result_ of a waffle flag. Really here we don't need to know about waffle even. IMHO this is an "option" and we could pass in a dict or a named tuple.

The downside of this is that all callers to `all_reports` need to create this options object and pass it here. We can have a method that takes a request and makes that options object and use it in lots of places in the API/views.

vigowei: This is my problem with waffle. Even though it allows us to control feature access, there is no easy way to not refactor, or get messy in some parts of our code.  This was a hard decision to make, for all the reasons you described. It's a bad thing that we have to pass request to such low levels.. some of the callers to all_reports live in even lower levels - models. Instead of refactoring that much, I just make request an optional argument, and we can just clean up after this feature is released.

vigowei:
Closing for now, we are looking for a better solution because dotjobs deserves it.

myreports/schema.py
koonsies: I'm inclined to agree with Darrin on this. It would be fairly trivial to create a function/method that takes the request object, tests the given request.user object against the various flags, and returns a list of the flags that are active for the user. A list of flags is something easily mockable, perhaps even with the addition of an object containing an overwrite value, so celery can overwrite all flags in one go.

`{overwite: False, flags: ['partner_assessment', 'otherthings', 'evenotherthings']}`

vigowei: @koonsies thanks for your input. I agree that passing a dict around is definitely better than passing the entire request object. I also wish waffle only required the user object to determine if a flag is active. After further discussion, we are thinking maybe it's better to use AppAccess for this reporting feature. Right now we are trying to get Sole's opinion on this, would you like to be included in the conversation when it happens?

vigowei: Update: we are going to create a new app called site-traffic-v2-preview, so members who will have access to the new site traffic report will temporarily have access to both old and new. Once we roll it out to everyone, replace site-traffic with the new version, delete the v2 app.  🎩 🐰

PR #3445, opened 2017-09-08
[PD-3579] dropdown combos when switching pages

mwarnsleyDE:
## What's this PR do?
PR changes the state to not show the multifilter dropdown if it is being shown when switching between dimensions

N/A

Start inside the site traffic analytics page with the multifilter dropdowns

Pull down the repository. Go into the analytics application and go to a multifilter page. When you open it, switch dimensions to another multifilter page and it should be closed now do to the state change


iartemiev:
Just curious, why are we storing `multifilterShown` in redux? It looks to me that it is an example of UI state, which is generally always better to store in the component state, as opposed to redux...

mwarnsleyDE:
@iartemiev I was storing in inside of redux in order to reset the state of the component once a dimension is changed. I typically even try and store as much in redux in generate just because it holds most of the state in the first place. I was actually thinking of trying to have most of the components be functional stateless components compared to having individual state since it's better to have redux do most of the work when it comes to managing the state.

iartemiev:
https://github.com/reactjs/redux/issues/1287 
2nd comment (gaeron is Dan Abramov)

UI state is best stored in the component, app state is best stored in redux. Trying to store everything in Redux gets very messy and needlessly complex very quickly.

mwarnsleyDE:
@iartemiev I think for the most part it doesn't matter too much to be honest. I think the main reason I did it there was because of it not toggling back to off when switching dimensions in the application. I more so have a preference of just embracing the single state tree. I think UI state is ok there as long as it's not too tedious to do all the time. I think the ONLY reason I switched it to redux was because of how the app is built with switching dimensions when it's actively open it should shut when switching that dimension.

mwarnsleyDE:
@iartemiev is there a way to reset the state back to false when switching dimensions? That's the main reason I switched it over to the redux state.

iartemiev:
If it's relevant to components outside of the direct ancestor chain, I think then it makes sense to throw it in Redux.

iartemiev:
Otherwise, just lifting the state up seems to be the best practice.

mwarnsleyDE:
@iartemiev and @aljordan82 for this one we have to be able to close or reset the state when a dimension is switched. That's why it is in redux. If done using the state I'm guessing the only way is through a lifecycle method which will also close it when navigating and filtering down on the table so that is why I stored it in redux unless there is another option.

koonsies:
Conflicts

mwarnsleyDE:
@iartemiev @aljordan82 you guys can check this now

mwarnsleyDE:
@aljordan82 fixed

PR #3449, opened 2017-09-08
[PD-3623] Preview Solr searches when making custom facets.

almathew:
## What's this PR do?
Allows admins to preview Solr queries when creating custom facets.

screen shot 2017-09-08 at 2 49 52 pm
 
API is in seo/api/v2/
JavaScript is in the only .js file added.

Import a few jobs.
Go to /admin/seo/customfacet/1/change/
Click the Advanced options (Show) toggle.
Click search and watch the results section magically populate. 
Now mess with a combination of search terms and business units. Changing those should impact the search results. Changing anything else should not. 

This was broken for a while, so the current workflow for DS is to create a facet, save, and then check it out on the site.
Because facet tagging will slow that process down quite a bit (since it requires a reimport), we opted to fix this as a happy medium between tagging on the fly and forcing DS to deal with how slow the process will be.


almathew:
Since I was changing related files, I pulled in my other PR to avoid merge conflicts with myself.
Sorry.
Primary changes are https://github.com/DirectEmployers/MyJobs/compare/8fbabcf...6fd5ff6

dseo_urls.py
```diff
@@ -1,32 +1,19 @@
...
 from myjobs.urls import impersonate_patterns
-from seo.views.search_views import (BusinessUnitAdminFilter, SeoSiteAdminFilter,
-                                    Dseo404)
+from seo.views.search_views import BusinessUnitAdminFilter, SeoSiteAdminFilter, Dseo404
 from seo.views.settings_views import secure_redirect
 from registration import views as registration_views
-from ajax_select import urls as ajax_select_urls
 from myjobs.forms import MyJobsAdminAuthenticationForm
 
-# This is a bit of code pulled from a Django TRAC ticket describing a problem
```
tdphillips: :1st_place_medal:

PR #3450, opened 2017-09-11
PD-3653 - Analytics - Initial time range is hardcoded at May 4th

iartemiev:
## What's this PR do?
Changes the initial date range in analytics to the following:
from: 31 days ago
to: yesterday
thereby, setting the default range to 30 days

Converted the date format `MM/DD/YYYY` that is deprecated by MomentJS to the supported `YYYY-MM-DD`

Also: Prettier made some formatting changes in header-reducer.js and Header.js. The reviewers can ignore these, as the logic has remained unchanged.


mwarnsleyDE:
@iartemiev I haven't taken too deep of a dive into the code as of yet but this still doesn't permit people to go back any farther than May 4th does it? There shouldn't be any data farther back than that so it should still be set as the farthest date back. Haven't looked at the code too much but just asking in advance so if so just excuse this comment based on the fact of me being lazy LOL

iartemiev:
@mwarnsleyDE there was a request to change the default date to 30 days. From yesterday - 30 to: yesterday. That's all this PR does

PR #3451, opened 2017-09-11
[PD-3664] fix header in export

vigowei:
## What's this PR do?
Start using `alias` in excel/csv exports, this includes headers and filter criteria.

Before:

image

After:

image

Export for different types, both excel and csv should have their headers set properly. For excel, the criteria sheet should also have correct labels.

@tejaswinigutti31 and I decided to refactor a bit, and renamed `label` to `alias` for columns defined in schema_dimensions.py - `alias` was introduced in an old commit, we'll let it live, and let's agree on its meaning and purpose: "label for a column, something easier to read than snake case, used for front end representation"


myreports/schema_dimensions.py
```diff
@@ -8,14 +8,14 @@
...
-DataColumn = namedtuple('DataColumn', ['column_name', 'output_format', 'label', 'description'])
+DataColumn = namedtuple('DataColumn', ['column_name', 'output_format', 'alias', 'description'])
 
 
 data_cols_job_views = [
     DataColumn(
         column_name=u'job_views',
         output_format=u'text',
-        label=u'Job Views',
+        alias=u'Job Views',
```
darrint: How about `display`?

myreports/presentation/csv.py
```diff
@@ -4,22 +4,21 @@
...
+
         group_by_overwrite = kwargs.get('group_by_overwrite')
         if group_by_overwrite:
-            if len(group_by_overwrite) > 1:
-                group_by_key = ' / '.join(group_by_overwrite)
-                values = [group_by_key] + [v for v in values if v not in group_by_overwrite]
-                for i, r in enumerate(records):
-                    group_by_value = '/'.join(r.get(g) or 'Unavailable' for g in group_by_overwrite)
-                    r[group_by_key] = group_by_value
+            group_by_key = ' / '.join(group_by_overwrite)
```
darrint: This code probably wants to be somewhere else close to where we generate the report data, not in every Presentation type.

myreports/schema_dimensions.py
vigowei: `alias` is introduced in this commit: https://github.com/DirectEmployers/MyJobs/commit/a027e97029670fa6215f2d6d5e09bf1ff1798624#diff-16b165fb63b242298ecb4afd96d917d1R65

if we rename it, it will probably a bigger refactor, a new ticket?

myreports/presentation/csv.py
vigowei: updated, thanks.

tejaswinigutti31:
Conficts

myreports/schema_dimensions.py
darrint: Let's leave it as you have it.

PR #3452, opened 2017-09-11
PD-3656 analytics predefined date range doesn't work

iartemiev:
## What's this PR do?
- Fixed date range selection issues (namely 'Last Month' and 'Custom Range')
- Refactored Calendar.jsx, added `shouldComponentUpdate`
- Moved UI State into Calendar.jsx
- Changed all Analytics Component file extensions from `.js` to `.jsx`


aljordan82:
I am seeing issues that might not be related to this PR.
Selecting a custom range doesn't allow selecting today's date.
Selecting "yesterday" will cause the charts to have data from January

iartemiev:
@aljordan82 good catch! Should be fixed now

PR #3456, opened 2017-09-12
[PD-3569] contact and location clarification on creation

mwarnsleyDE:
## What's this PR do?
PR adds a clarification message to the NEW Cbo page and Additional Contact page

N/A

Start inside of the the TJD application

Pull down the repository. Go into the CBO application and go to create a NEW CBO. You will see the message on the right hand side with the facets. As well go into a CBO to edit that is already created. Go and add an additional contact and you will also see the message on the right hand facet side.

https://directemployersfoundation.atlassian.net/browse/PD-3569


mwarnsleyDE:
Anyone who reviews this, do me a favor. Maybe I am losing my mind but we are calling setState in some of the functions yet we don't have state declared anyone inside of the class. Maybe I'm losing my mind or just confused at how that's working or maybe it's just dead code but I'm not sure lol.

iartemiev:
@mwarnsleyDE It looks like dead code to me. I see the call to `setState` but I don't see anything using `this.state`...

koonsies:
Some of the dead setState logic is likely due to the refactor a did a while back in order to introduce react-router-4. At one point in time, a great deal of logic was held within the CboApp.js. When I moved the code out of CboApp, I must have missed some of these references. That was a messy PR...

gulp/src/cbo/components/commonCbo/Sidebar.js
```diff
@@ -13,37 +13,34 @@ import {updateCboStatus} from '../../actions/cboActions';
...
-    this.setState({
-      targetLocation: clickedButton,
-    });
-  }
-  renderCboInformation() {
+  };
+  contactLocationMessage = () => {
+    const {location} = this.props;
+    const infoText = location.pathname.includes('additionalcontact')
+      ? 'Additional contacts can be added after creation of the contact'
```
aljordan82: This should be "additional Location" according to the ticket.

mwarnsleyDE:
@aljordan82 Fixed

PR #3458, opened 2017-09-12
[PD-3678] Fix None and timedeltas.

darrint:
## What's this PR do?

Fixes a problem where a datasource was broken due to all None values in a timedelta column.

`utils.py` has a new utility for converting milliseconds to a timedelta and there is a test.

I can't actually reproduce it locally. It was found on QC.

na.


myreports/tests/test_datasource_utils.py
```diff
@@ -728,4 +728,11 @@ def test_build_or(self):
...
-        self.assertEquals(expected, result)
\ No newline at end of file
+        self.assertEquals(expected, result)
+
+
+class TestToTimeDelta(TestCase):
+    def test_milliseconds_to_timedelta(self):
+        self.assertEqual(timedelta(minutes=1), milliseconds_to_timedelta(60000))
+        self.assertIsNone(milliseconds_to_timedelta(None))
+        pass
```
tdphillips: pass?

darrint: fail

PR #3462, opened 2017-09-12
[PD-3680] add admin page for analytics channels

vigowei:
## What's this PR do?
Add a simple admin page for channels so users an add entries; 
Can no longer add duplicate entries

Go to Django admin -> Analytics -> Channels, and add/remove entries from the interface.
We will not be using company_id now, until we figure out a solution to infer company id during the traffic processing phase.


darrint:
Since we're doing migrations, can we remove the not null constraint and the default value from the company_id column?

PR #3461, opened 2017-09-12
PD-3671 - UX - Replace Jasmine with Jest

iartemiev:
## What's this PR do?
**Please re-source `dk2/env.sh` and run `dky` when testing this PR**
* Migration to Jest for our front end unit testing.
* Removal of gulp (and related dependencies) 
* Building and deployment is done through Yarn scripts

I standardized the naming of our test files across all of our React apps.
Here's the convention: `someDir/__tests__/aBeautifulUnitTest.test.js`
This is the standard naming convention for Jest, but if you guys don't like it, I can rename the test dir to `spec`.
I created separate `__tests__` dirs for actions and reducers and broke up the existing tests into the relevant dirs.

Off the top of my head, I'd say that roughly 95% of the tests were copied as-is, unchanged. Some required minor alterations. Some were cleaned up (unused dependencies removed) and since we use the Prettier formatting tool, some of the formatting has been cleaned up automatically as well.

One test in particular: `analytics/actions/__tests__/sidebar-actions.test.js`, had some errors in it that Jasmine must have been silently failing on, if you look at the diff, you'll see what I'm talking about.

## Any background context you want to provide?

I've created a new function in `dk2/env.sh` for calling yarn: `dky`
So the following scripts will be called i.e.: `dky test:watch`

I've added 5 new yarn scripts for testing:
* `yarn test` - will run Jest once, then quit
* `yarn test:watch` - (my favorite) starts Jest in watch mode and only test newly changed files (those which haven't yet been committed to the git branch). 
* `yarn test:watchAll` - same as above except it will run all tests
* `yarn test:coverage` - Istanbul code coverage
* `yarn test:jenkins` - run tests in --ci mode and create XML report with JUnit (for consumption by Jenkins)

And 4 new scripts for building and deployment:
* `yarn build` - production webpack build. (Sets NODE_ENV to production)
* `yarn build:dev` - dev webpack build. (Sets NODE_ENV to development)
* `yarn deploy` - Everything that's needed for deployment: yarn install --frozen-lockfile, production build and `test:jenkins`
* `yarn deploy:npm` - Same as above, but works without Yarn (npm fallback)

PR #3463, opened 2017-09-12
PD-3670 - UX - remove Gulp

iartemiev:
## What's this PR do?
Step 2 of https://github.com/DirectEmployers/MyJobs/pull/3461

I renamed the **gulp** dir to **secure**. Just the first thing that came to my mind. I'm open to suggestions though


iartemiev:
Seems that only some python tests are failing

darrint:
Any reason not to just move everything to the root? I don't think the secure name is going to hold up. I may have missed some discussion though.

iartemiev:
@darrint The only issue I see with that is that there are already a few dozen (mostly python) files in the root, so things like our `package.json`, `webpack.config`, `.babelrc`, etc. might become a bit more difficult to find. I personally like having the front end in a separate dir from an organizational perspective... I could likely be swayed though (:

@aljordan82, @mwarnsleyDE - what do you guys think?

darrint:
Back in the day I got a few complaints because of the fact that it was in a subdir.

mwarnsleyDE:
@iartemiev and @darrint I'm not too familiar with how the secure or other domains relate to the folder structure but I can say I'm all for having the front end in a different directory compared to the backend. I do think it's easier to find and honestly for me I think finding some of the front end stuff even in django right now hard to find because of how many files we currently have. Right now I think it's easier to find all of the react projects because we know it's the only thing that goes into the gulp folder and nothing else so I think that structure makes it easier to navigate the code base. I don't know how that relates to anything else as far as domains or anything like that but if we could I think it would be easier keeping them separated.

PR #3464, opened 2017-09-12
Pd 3664 fix header in export again

vigowei:
## What's this PR do?
Fixes a bug where combos are still not represented correctly

before: 

image

after:

image

Download different reports and make sure this fix is reflected.


tejaswinigutti31:
The PR looks good. If we plan on filling empty values with something(e.g. 'Unavailable') rather than leaving the cells blank, could we do it as part? @darrint thoughts on this?

darrint:
Let's make a new subtask for additional work

PR #3470, opened 2017-09-14
[PD-3689] Fix server errors.

almathew:
## What's this PR do?
Fixes some errors popping up in new relic.

All the changes are focused around search_views.
The first, and biggest, problem was that guid wasn't properly getting added to the job if the job is retrieved from MySQL because it was popped off the results, so I prevented that. I also converted the guid to uppercase for search, since I really don't care about guid case when determining if the URL is the canonical URL.
The second is that people are searching for multiple spaces via autocomplete, which messes with haystack, causing errors. Less important, but might as well fix it while I'm there.

Ask me for a business unit/url that was causing this problem. Confirm that it was broken before my changes, and works after. You can also mess around with confirming things still actually work.
For autocomplete, you can test searching for spaces. It currently errors. On the new code it just short-circuits and does nothing instead (which is better than an error). 

An increase error rate is causing issues with the load balancers marking servers as down. This fixes the cause of the increased error rate.


almathew:
Important change 1: https://github.com/DirectEmployers/MyJobs/pull/3470/files#diff-b1cb1c063c796b68c0ee52fb93fdfcacR120
Important change 2: https://github.com/DirectEmployers/MyJobs/pull/3470/files#diff-59a5dfba4c33738addad3440fa8d4efdR297
Important change 3: https://github.com/DirectEmployers/MyJobs/pull/3470/files#diff-59a5dfba4c33738addad3440fa8d4efdR1094

almathew:
Just added fix for the issue with imports too:
```  File "/home/web/MyJobs/MyJobs/import_jobs/utils.py", line 284, in job_is_same
    if set(job_a.get('assets').items()) != set(job_b.get('assets').items()):
AttributeError: 'NoneType' object has no attribute 'items'
```

PR #3474, opened 2017-09-14
PD-3688 Fixing columns in Apply Clicks

tejaswinigutti31:
## What's this PR do?
Apply Clicks dimension of new site-traffic needs to include Job data along with Goal Data. Fields corresponding to Job Data have been added
a new dimension for apply clicks has been added to schema_dimensions.py
Check the Apply Clicks dimension of new site traffic report to find Job data(Title, Req ID, Location)


myreports/schema_dimensions.py
```diff
@@ -482,6 +516,25 @@
...
+            GroupColumn(
+                column_name='goal',
+                filter_interface_display=u'Goal',
+                description=None,
+                active_only_with=None,
+                output_format='text',
+            ),
+        ],
+    ),
+    data_columns=data_cols_goals + data_cols_job_data,
```
darrint: This is a lot of columns. We may need @haliscooper to let us leave off the goal columns and just do the job columns.

PR #3482, opened 2017-09-15
Fix error where request is coming in without an HTTP_HOST

almathew:
## What's this PR do?
Sometimes we get requests without an HTTP_HOST.
While they're probably [mostly] a waste of time, there's no reason we can't serve them.

I'd recommend against running new relic locally unless you're actively doing extensive work on it, so instead just copy the line to some other middleware section (like multihost) and confirm that it works.


vigowei:
Does this mean we only track POSTs with new relic? 
what about GET requests?

PR #3487, opened 2017-09-18
[PD-3703] Remove Null on Goal Path

tejaswinigutti31:
## What's this PR do?
Goals labelled as "Unavailable" are no longer displayed in Goal Flow
Changes in data source of goal_events to not include the rows with "Unavailable" goal
In Goal flow dimension of Site Traffic v2 Preview, observe that there are no rows with Goal value being "Unavailable" in Goal Minus One and Goal Minus Two


myreports/datasources/goal_events.py
```diff
@@ -303,7 +303,15 @@ def calculate_fields(self, company, result, filter_spec, values):
                     filter_list[k] = f.value
         total_visits = Visit.objects.filter(**filter_list).count()
 
-        for row in result:
+        for row in reversed(result):
```
darrint: Did we try filtering the nulls from the queryset?

PR #3488, opened 2017-09-18
PD-3701 Renaming Page Types

tejaswinigutti31:
## What's this PR do?
Rename the Page Types to make it more readable
A small change in site_traffic.py to map the notations to readable format
In Site Traffic v2 Preview, check the dimension


vigowei:
There is a bug that prevents drilldowns from working correctly in this dimension, 
because display values are used in filters. 
image

example bad request: 
```
{
  "date_start":"2017-08-19 00:00:00",
  "date_end":"2017-09-17 23:59:59",
  "active_filters":[
    {"type":"page_type",*"value":"Job Apply"*}
  ],
  "report":"site-traffic-v2.page_type.aggregated",
  "reporting_type":"site-traffic-v2",
  "timeline":false,"dimension":"path"
}
```

myreports/datasources/site_traffic.py
```diff
@@ -313,3 +325,24 @@ def filtered_query_set(self, company, filter_spec):
...
+        """
+        page_type_reverse_lookup = {
+            "Job List": "joblist",
+            "Search": "jobsearch",
+            "Job Apply": "jobapply",
+            "Job View": "jobview",
+            "Content": "generic",
+            "Unknown": "unknown"
+        }
+        if 'path' in values:
```
vigowei: it's probably better to check if page_type is part of the filter,  i.e. `hasattr(filter_spec, 'page_type')`,  instead of relying on the presence of 'path' in query values,

myreports/datasources/site_traffic.py
```diff
@@ -202,6 +211,9 @@ def run_aggregated(self, company, filter_spec, values,
                     row['engagement_rate'] = 0
                 del row['engagement_count']
             row['time_on_page'] = milliseconds_to_timedelta(row['time_on_page'])
+            if 'page_type' in values and row['page_type'] is not None:
+                page_type = page_types.get(row['page_type'])
+                row['page_type'] = page_type
```
vigowei: This will ignore all page views that doesn't have 'page_type' set, which I believe can happen in real data, so I would recommend removing `row['page_type'] is not None` from the if condition.
`page_type` could end up being None if it's not defined in page_types, which shouldn't happen, but maybe it's better to rewrite line 215-216 as: 
`row['page_type'] = page_types.get(row['page_type'], row['page_type'])`

vigowei:
timeline appears to be broken for page types.. probably because the same reverse lookup should be done for `run_gather_timeline_data`. That makes me wonder if we should add reverse lookup to init

PR #3491, opened 2017-09-18
The default value is a list, which is unhandled.

almathew:
## What's this PR do?
Handles the condition where the default value (a list) is still the default value for assets.

There are task failures happening because of this, so if you're wondering what this is fixing it's easy to look up.


import_jobs/utils.py
```diff
@@ -281,8 +281,17 @@ def job_is_same(job_a, job_b, include_facets=False):
     for field in diff_fields:
         if job_a.get(field) != job_b.get(field):
             return False
-    if set(job_a.get('assets', {}).items()) != set(job_b.get('assets', {}).items()):
+
+    job_a_assets = job_a.get('assets', {})
+    job_b_assets = job_b.get('assets', {})
+    # The default value is a list. Because of that, we have to coerce it to a dictionary.
```
tdphillips: I'm trying to understand this.

If `assets` is present but blank, it's an empty list.
If it's present and populated, it's a dictionary.

It doesn't look like we're coercing anything either, just assigning an unrelated value.

almathew: Yeah, this is a huge pain, as shown by the fact that this is the second time I'm fixing it. There are four possible conditions:
- It's blank, and has never been assigned. Therefore it will be assigned the default value of `[]`
- It's blank, but has been coerced into the correct format by the code that checks for assets. In this case it will be `{}`
- It's blank because someone did something manually. In this case it will be `None`
- It has a valid value. It will be a dictionary with contents.

It's a TextField that gets coerced to JSON, so technically it could have any other number of values, but those are the explainable ones.

There are about 20 different ways that this could be handled. 
The best solution would be changing the default behavior, but the fallout of another migration is not worth it. 
The second best involves ensuring that the value gets set to the correct datatype, but since it's not properly handled on the database, there's still a chance that something might get around that (through manual changes or non-traditional addition/editing of jobs).
I went with the third best option I could come up with, which is just handling all the values that aren't completely wrong.

tdphillips: I merged this before fully processing what you said.

`{'assets': None}.get('assets', {})` still results in `None`.

almathew: Good catch. I throw up another fix against QC when this one gets merged back, since I think I'm the only one who is playing around with the actual data in production to cause that case.

PR #3495, opened 2017-09-18
PD-3682 - PA/Analytics - Footer Style

iartemiev:
## What's this PR do?
Fixes footer style in PA and Analytics

image

image


gulp/src/sass/partner_assessment/main.scss
```diff
@@ -18,8 +18,10 @@
 @import "./components/_record_list_filter";
 @import "./components/_assessment-form";
 
-
 .partner-assessment-app {
+    background-color: white;
+    border: 1px solid #333;
+    padding-bottom: 15px;
```
aljordan82: Is this padding needed? I think it should be fine without it.

aljordan82: Also, I would probably remove border left and right.

iartemiev: With padding:
image

Without padding:
image

iartemiev: If you like the way that looks, I'm happy to remove it

aljordan82: hmm? it looked like it has excessive padding on my machine?

PR #3496, opened 2017-09-18
Tests for things on microsites that have had issues recently.

almathew:
## What's this PR do?
Adds tests for issues that have been discovered throughout the past few deploys.
Many of these things were either previously untested, or not tested in a meaningful way.

Since these are recent issues that were mostly hotfixed, these tests won't pass until I have all the fixes pulled back to QC and then into this branch.


almathew:
Tests are passing in case anyone was kind of sort of keeping an eye on this.
Each commit has a comment about why a test wasn't previously passing.

PR #3504, opened 2017-09-19
[PD-3722] table page size broken

mwarnsleyDE:
## What's this PR do?
PR fixes the page sizing that was broken on the analytics tables

N/A

Start inside of any analytics app

Pull down the repository and go into the analytics application. Job Titles might be the best dimension to test but in the table, change the page size and you should see it work now instead of throwing an error.


mwarnsleyDE:
Some of the other code besides fixing the page size was not necessary for this PR but i was already working on it since it needed to be done and just included it in here for the future. The only real code change was inside of the PageSize.jsx file

tejaswinigutti31:
Conflicts

PR #3506, opened 2017-09-19
[PD-3611] Implement Confirm handling for TJD wizard navigation

koonsies:
## What's this PR do?
This PR does a few things

1) Expands the Confirm.jsx module to allow multi-response handling.
2) Adds a special case to clicking the "New TJD" navigation that verifies the user's intent
3) Makes resetting of the TJD wizard more explicit
4) Begins removing the "CBOModal" component
5) Moves wizard-specific files to a new wizard directory, instead of under the cbo directory

Bonus) Adds a "restartwebpack" dk command.. Cuz I got sick of doing app/stopapp when devserver alone failed.

TJD. Verify that the "New TJD" navigation opens a prompt when you've begun creating a new TJD, navigate away, and attempt to navigate back.

1) Start a new TJD in the wizard (at LEAST enter a name and hit "Next")
2) Click another link, or press Back
3) Click "New TJD" again. You should be prompted with the new Confirm box.

This is the beginning of a refactor of the TJD wizard. Going forward, there will be tickets for
1) Detecting partial form inputs by either just putting the data in redux as users enter it, or prompting before they navigate away
2) Add history handling to the wizard (replacing the current switch statement method)
3) Remove uses of the CboModal component and replace with Confirm.


koonsies:
Good call on the multiple "Add Organization" buttons. I will modify the handling to work with both.

About the form fields.. As far as I know, they've never been detected/prevented. I didn't want to cram everything into one PR, so it is slated for the future. The work related to that is covered under bullet point (1) under the "Background" description.

Thanks for the review!

dk2/dockerdev.sh
```diff
@@ -184,6 +184,12 @@ restartmicrosites() {
     runmicrosites || true
 }
 
+restartdevserver() {
```
koonsies: In case someone doesn't want to do dk appstop/start just to restart dev server

gulp/src/cbo/actions/cboWizardActions.js
```diff
@@ -378,6 +379,7 @@ export function doSaveCbo() {
     const cboData = getState().newCbo.newCbo;
     return api.saveCbo(cboData).then((cboWithId) => {
       dispatch(saveCboSuccess(cboWithId));
+      dispatch(resetWizard());
```
koonsies: Abstracted the logic to reset the wizard, created a new action for it.

gulp/src/cbo/components/CboApp.js
```diff
@@ -2,82 +2,36 @@ import React, {Component} from 'react';
...
 class CboApp extends Component {
   constructor(props) {
     super(props);
 
     this.state = {
-      modalShown: false,
-      targetLocation: '',
       statusChecked: props.cboOptStatus || null,
     };
+    this.confirmWizardNavigation = confirmWizardNavigation.bind(this);
```
koonsies: Pulled this function out into it's own file in order for it to be used in more than one place. Due to this, it has to be re-bound to any class using it. Not sure if there's a more elegant solution to this type of issue, so input is welcome.

gulp/src/cbo/components/wizard/CboWizard.js
```diff
@@ -28,9 +27,6 @@ class CboWizard extends Component {
     case 'NEW_CBO_CONTACT_LOCATION':
       wizardContent = <AddDistribution/>;
       break;
-    case 'NEW_CBO_CONTACT_RULE':
```
koonsies: Wizard doesn't currently allow addition of rules. This was scrapped a while ago.

gulp/src/cbo/components/wizard/WizardConfirm.js
```diff
@@ -0,0 +1,38 @@
+/**
```
koonsies: This function is the abstracted-wizard-navigation thingie. It must be bound to a class that has access to certain props. This approach makes me a little queezy, but I better idea hasn't occurred to me.

gulp/src/cbo/components/wizard/CboWizard.js
aljordan82: I ma not sure why it adding a rules isn't in the wizard.
@mwarnsleyDE any ideas?

mwarnsleyDE: @aljordan82 when we first built it, we were told to not put the rules inside of the wizard instead. I agree that it should be inside of the wizard as well however.

koonsies:
When it was first built, the add rule functionality was bugged. We didn't have time to correct the issue fully, and it was scrapped due to timing. Furthermore, I do believe a contact/cbo can be added without rules applied.

aljordan82:
@koonsies There is still a non-standard modal that appears when pressing the "previous" button in the Contact form page. is this a part of this ticket?

koonsies:
Ah, yeah, I didn't explicitly say it, but that's part of item 2 in the Background Information (next steps) section. When I revisit the way we handle routing (as well as storage of the information in Redux as it's input) then I will remove that modal. For now, it serves a minor purpose and I decided to leave it in.

koonsies:
I didn't want to do all of the work in a single, huge PR, so I found a good stopping point and created this one.

PR #3511, opened 2017-09-20
Setting queryset limit

tejaswinigutti31:
## What's this PR do?
Set a limit on querysets in datasources of new site traffic report
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


vigowei:
duplicate of #3512

PR #3513, opened 2017-09-20
Pd 3751 extract common api calls

aljordan82:
## What's this PR do?
Extract common api call that are shared across apps
partnerassessment/main.jsx


gulp/src/nonuseroutreach/main.jsx
```diff
@@ -12,7 +12,7 @@ import {combineReducers} from 'redux';
 import {installPolyfills} from '../common/polyfills';
```
aljordan82: Ignore this file

koonsies:
This is cleaner. Good refactor. I'll manually check some of the APIs for each app and get this merged.

PR #3517, opened 2017-09-21
Pd 3773 replace TJD acronym

tejaswinigutti31:
## What's this PR do?
Change the acronyms of TJD and PRM in report screen and App access list to full names
screen shot 2017-09-21 at 10 10 46 am

Changes in schema.py and myjobs/models
Check the pages on report screen and app access list to find the full names
### Separation of Concerns

### Tests

### Migrations


myreports/schema.py
```diff
@@ -506,7 +506,7 @@ def ds_page_data(dimensions):
 )
 rc_tjd_logs = ReportCategory(
     datasource=u'tjd_report',
-    display=u'TJD reports',
+    display=u'Targeted Job Distribution',
```
aljordan82: Do you want to keep "reports"?

myjobs/helpers.py
```diff
@@ -125,3 +125,17 @@ def get_title_template(next_url):
 
 def impersonate_access_function(request):
     return request.user.is_staff
+
+def get_app_access_name(app_access_name):
```
koonsies: This function does something that we intend to fix with the ticket here: https://directemployersfoundation.atlassian.net/browse/PD-3506

I wonder if we want to go ahead and tackle the ticket above instead of putting in a function to replace the missing description field.

@fezick , thoughts? I see that the AppAccess ticket in question isn't priority, so this may be fine in the interim.

almathew: For the record I'm like 99% done with that ticket, there's just some stupid stuff with migrations that might actually be resolved thanks to work from @darrint 

However, I kind of de-prioritized it given all the other work I was doing, and the fact that we've often talked about taking a different approach anyway. I'm happy to finish it up if that would make anything easier for anyone else.

aljordan82:
@almathew @koonsies Can this be merged?

aljordan82:
I've tested this and it looks fine, I am not sure the python stuff.

myjobs/helpers.py
almathew: https://github.com/DirectEmployers/MyJobs/pull/3527

almathew:
@aljordan82 @koonsies asked me about my PR to see if we can avoid adding some unnecessary code

myjobs/helpers.py
koonsies: The ticket above is now merged. We can go ahead and remove this logic and create a follow-up ticket to migrate the acronym-named app accesses accordingly.

tejaswinigutti31:
I've created a sub-task PD-3817 to create a migration to update full names in activities page

PR #3519, opened 2017-09-21
Update code to pull from the new jobsfs url.

almathew:
## What's this PR do?
Pulls info from the new jobsfs server.

Run a `from_jobsfs` import.

Tested and confirmed working with help from @RonCraig. They should be 100% prepared for us to make the move.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


almathew:
Changes are being made to the endpoint. This will return momentarily.

almathew:
Changes have been finalized and made. This code works with those changes. This is ready (from the perspective of the person who wrote it) to be merged.

PR #3521, opened 2017-09-21
PD-3739 fixing timeline for new Site traffic report

tejaswinigutti31:
## What's this PR do?
Fix the time line data for new site traffic report
Check the updated python_tld() in myreports/models.py
In new site traffic report, check if values in timeline and table below reconcile
Check if the labels displayed are correct on timeline
### Front-End

- [x] Works on IE 11 and Edge
- [x] Works on Mobile

### Separation of Concerns

### Migrations
N/A


gulp/src/analytics/components/Charts/Line/TimelineChart.jsx
```diff
@@ -59,6 +59,8 @@ class SimpleLineChart extends Component {
       this.viewType = 'Page Views';
     } else if ( getDataKeys.indexOf('Goals') > -1) {
       this.viewType = 'Goals';
+    } else if ( getDataKeys.indexOf('Apply Clicks') > -1) {
+    this.viewType = 'Apply Clicks';
```
darrint: indent

darrint: I am not saying we have to fix this now. But:

How can we eliminate this decision from the API? Where in schema dimensions would this "viewType" choice live?

myreports/models.py
```diff
@@ -448,7 +448,10 @@ def regenerate_timeline_data(self, user, company):
         else:
             datasource = self.data_source
         driver = ds_json_drivers[datasource]
-        data_type_name = 'gather_timeline_data'
+        if data_set.value in 'apply_clicks':
```
darrint: I am not saying we have to fix this now. But:

Where in schema dimensions should this go?

myreports/models.py
```diff
@@ -478,38 +481,51 @@ def python_tld(self):
         report_intention = self.report_schema.split('.', 1)[0]
         if report_intention == 'site-traffic' or report_intention == 'site-traffic-v2':
             if self.data_source == 'goal_events':
-                site_traffic = {
-                   'view_type': 'Goals',
-                   'time_filter': 'time'
-                }
+                if self.report_schema.split('.', 2)[1] in 'apply_clicks':
+                    site_traffic = {
```
darrint: I am not saying we have to fix this now. But:

Where in schema dimensions should this go?

myreports/models.py
vigowei: is using `in` intended or should this have been an equality comparison?

myreports/models.py
vigowei: one way I can think of is to separate timeline from the regular run_* methods, then set up a similar dispatch for `run_timeline`, or have one `run_timeline_aggregation` method per data_source, that requires metric type(s), for example:
`run_timeline_aggregation(metrics=['visits', 'page_views'], filter_spec=...)` will gather time series data for both visit and page views. 
Now I think it'll be a good opportunity to refactor when we work on the new story of adding multi-dimensions to timeline.

myreports/datasources/visits_data.py
```diff
@@ -257,7 +257,7 @@ def run_gather_timeline_data(self, company, filter_spec, field=None, partial=Non
                                        partial=partial)
         timeline_data = query_set.annotate(
             day=TruncDay('time')
-        ).values('day').order_by('day').annotate(views=Count('site_visit_id'))
+        ).values('day').order_by('day').annotate(visits=Count('site_visit_id'))
```
vigowei: this is returning a higher number for timeline in Goal related dimensions, the difference a bit hard to notice, you may not observe it locally if you have good fake data. The delta is caused by the exclusions (`None`/Unavailable) we do for the "regular" goal queries, which is not done for timeline.

aljordan82:
I've tested this PR and I don't see any issues.

PR #3522, opened 2017-09-22
Pd 3746 prm common action button

aljordan82:
## What's this PR do?
Fix Navigation Bar behavior in mobile. 
Navigation dropdown shouldn't push the page down when it is opened

### Front-End

- [x] Works on IE 11 and Edge
- [x] Works on Mobile


templates/mypartners/prm_base.html
```diff
@@ -46,9 +46,9 @@
                         {% if can_read_tag %}<div><a {% if current_page == 'partner_tagging' %} class="active"{% endif %}href="{% url 'partner_tagging' %}">Tag Management</a></div>{% endif %}
                         {% if has_nuo %}<div><a href="{% url 'nuo_main' %}">Non-User Outreach</a></div>{% endif %}
                     </div>
-                    <div class="nav-common-action">
-                        <!-- <div class="btn"> Common Action </div> -->
-                    </div>
+                    <!-- <div class="nav-common-action">
```
tdphillips: Why is this here? We've tried to not keep code that's been commented out.

PR #3523, opened 2017-09-22
[PD-3687] Allow microsites to specify locked css path

vigowei:
## What's this PR do?
Add `locked_css_path` (an S3 folder) to site configurations admin page, to be used with `lock_css`.

- lock_css is enabled, and locked_css_path is provided, template will load CSS from that path
- lock_css is enabled, locked_css_path is not provided, template will load site specific CSS
- lock_css is disabled, locked_css_path is ignored, template will load default CSS

image

please look at the migration first, let me know if the description text should be updated.

run microsites locally,
go to admin page, Configurations, update css path for my.jobs 
reload my.jobs, style should change/break

### Separation of Concerns

### Tests

### Migrations


seo/models.py
```diff
@@ -1226,6 +1229,8 @@ def get_template(self, template_string):
 
     lock_css = models.BooleanField("Lock CSS", default=False,
                                    help_text='Use hard-coded CSS instead of dynamic')
+    locked_css_path = models.CharField(max_length=255, blank=True,
+                                       help_text='Specify the S3 bucket and folder for your CSS files')
```
almathew: Path and file name, not bucket. They don't get a choice about which bucket.

vigowei: updated.

seo/models.py
```diff
@@ -1059,7 +1059,10 @@ def save(self, *args, **kwargs):
         # Increment the revision number so a new cache key will be used for urls
         # of the seosites linked to this configuration.
         self.revision += 1
-        if self.group: self.group.save()
+        if self.group:
+            self.group.save()
+        if not self.lock_css:
+            self.locked_css_path = ''
```
koonsies: So, if I understand this.. If they set a path, but forget to check the box, it clears the path without an error or any other indicator? I think this is a bit heavy handed. I feel like it's a better approach to let the path be set, but simply check the flag as to whether or not to display it. Someone may want to turn on/off the CSS as they are testing, without having to reset the path each time.

almathew:
We need to talk through some of the implementation details with DS.

seo/models.py
vigowei: Thanks for the feedback, this has been addressed. Hopefully it's more intuitive now.

tdphillips:
Where did the migration go?

vigowei:
@tdphillips my bad, I forgot to add it after deleting the old version, just pushed it up.

almathew:
I'm ok with this, although I'll leave it open for @JLMcLaugh or @tdphillips (or @koonsies I suppose) to comment on more.

tdphillips:
Did the DS conversation happen? I don't want to merge this if they haven't signed off on whatever was going to be asked.

almathew:
Yes, sorry, it happened before @vigowei re-opened the PR.
DS has confirmed that they're ok with the specifying only folder name(s).

PR #3530, opened 2017-09-25
PD-3817 updating full names on activities page

tejaswinigutti31:
## What's this PR do?
A data migration to update full names on Activity page
screen shot 2017-09-25 at 11 12 22 am

Migration 0004 in myjobs
In https://secure.my.jobs/manage-users/view/activities check that full names are being displayed for TJD, PRM and NUO
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


myjobs/migrations/0004_rename_displays.py
```diff
@@ -0,0 +1,52 @@
...
+        'TJD': 'Targeted Job Distribution',
+        'PRM': 'Partner Relationship Manager',
+        'NUO': 'Non-User Outreach'
+    }
+
+    app_accesses = AppAccess.objects.all()
+
+    for app_access in app_accesses:
+        if app_access.display_name in display_name_lookup:
+            aa = AppAccess.objects.get_or_create(name=app_access.name)[0]
```
tdphillips: You've already got this AppAccess, right? Could `app_access` be modified instead of retrieving it from the database again?

tejaswinigutti31: Updated it. Thanks for pointing that out

PR #3531, opened 2017-09-25
[PD-3776] App for API key management. 

almathew:
## What's this PR do?
Introduces API key management.
Right now it runs as an app on secure, but it is designed to run as it's own app (hence the new database). 

This adds two types of keys for use with our APIs:
- Tokens. Right now we have no way of issuing these to users (this is temporarily a feature, not a bug). They belong to organizations, not specific users. 
- Temporary tokens. Good for 1 hour, eventually we will have permissions associated with these. This exists so organizations can hand out temporary, limited API access.

I didn't write endpoint documentation yet because I don't know where @darrint and @tejaswinigutti31 are going with it, and I think I'm better conforming to their standards rather than making my own.

The custom authentication backend for temporary keys is in `api_management/auth_backends.py`. Otherwise, most of the magic happens in `api_management/api/`.

Check out the new endpoints for:
- Creating temporary API keys
- Listing existing, un-expired API keys
- Deleting API keys (even expired ones) 

You'll have to make your own endpoint for testing the TempTokenAuth, which is the whole point of creating/listing/destroying these keys.

### ~~Front-End~~
[removed by darrint]

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations

- [x] Data migrations' forward/reverse functions should be decorated with `@only_on_default`.


api_management/models.py
```diff
@@ -0,0 +1,32 @@
...
+            self.key = self.generate_key()
+        if not self.expiration:
+            self.set_expiration()
+        return super(TempToken, self).save(*args, **kwargs)
+
+    def generate_key(self):
+        return binascii.hexlify(os.urandom(20)).decode()
+
+    def set_expiration(self, hours=0, minutes=0, seconds=0):
+        self.expiration = datetime.now() + timedelta(hours=hours, minutes=minutes, seconds=seconds)
```
darrint: Consider `django.utils.timezone.now()`

api_management/api/views.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class CreateTempTokenView(views.APIView):
+    authentication_classes = (TokenAuthentication, SessionAuthentication)
+
+    def post(self, request, *args, **kwargs):
+        if not hasattr(request, 'user') or request.user.is_anonymous():
+            raise exceptions.NotAuthenticated()
+        token = TempToken(user=request.user)
+        token.set_expiration(seconds=3600)
```
darrint: Make a setting?

api_management/models.py
almathew: Consider it considered.

api_management/api/views.py
almathew: I considered that, but I was thinking that we probably wanted to make it dynamic? So I left that as more of a placeholder while we figure that out.

So consider the discussion about whether or not to make that setting dynamic open for discussion.

Another discussion point is whether or not use of the key pushes out the expiration period.

I don't have much experience working with keys that expire in general, so I'm just making things up as I go along.

darrint: I think of these tokens are more or less "Access Tokens" and the usual expectation is for them to have lifetimes only a few hours long. I don't think it's common to have them renew themselves.

https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/

In my limited experience we never needed to vary this setting for different customers but I could turn out to be wrong about that. I can't think of a good specific reason to make this dynamic. "So you want to be less secure?"

almathew: Ok. I'm still not going to make it a setting, but I will make it a default value on the function.

almathew:
Sorting out the db stuff first, even though it's going to make this PR bigger.

almathew:
DB stuff has been sorted. PR description has been updated to account for this sorting.

almathew:
Also, I think it's important to note that this will require you to add a new database. The documentation has been updated to reflect this.

Specifically you'll have to do this:
```
dk maint mysql:5.5 
mysql -u root -p -h $myip
CREATE DATABASE api;
exit
dkm migrate --database api
```

api_management/tests/api/test_views.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+@override_settings(ROOT_URLCONF="myjobs_urls", PROJECT="myjobs", MEMOIZE=False,
+                   HAYSTACK_CONNECTIONS=settings.TEST_HAYSTACK_CONNECTIONS,
+                   TEMPLATES=settings.TEMPLATES,
+                   CELERY_ALWAYS_EAGER=True)
+class APIViewTest(TestCase):
+    def setUp(self):
+        super(APIViewTest, self).setUp()
+        SeoSiteFactory(id=1)
```
darrint: What is this for?

almathew: Getting through MultiHostMiddleware. 
It will be less necessary once we're testing this as a stand-alone app, and not a member of secure/dseo.

darrint:
Code looks good. Have not tested yet.

darrint:
I tested the happy paths and a few accidental error paths locally with curl.

PR #3532, opened 2017-09-25
Pd 3819 pa filters

aljordan82:
## What's this PR do?
Make the filters div collapsible in mobile view
In Partner Assessment list page, the filters should be a dropdown i mobile view.
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


koonsies:
This probably applies to both filter dropdowns, but do we want to add collapse-on-blur logic? Or always have them have to tap the arrow to hide the menu? @strempe ?

strempe:
I'm a big fan of on-blur, but with the current UI we risk the user accidentally tapping another action. That's why we are using more intentional 'escape' - I could see it closing after a filter is selected... @koonsies

PR #3536, opened 2017-09-26
Pd 3567 remove tips from column

strempe:
## What's this PR do?
Removes trips from the right column for: 
- PRM
- Manage Users
- TJD

**With Tips** 
screen shot 2017-09-25 at 4 00 26 pm

**Tips Removed**
screen shot 2017-09-26 at 11 05 03 am

NUO tips will be removed in a separate PR due to changes made with mobile filters. 

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


templates/mypartners/edit_item.html
```diff
@@ -92,29 +92,6 @@
...
-                but you need to be sure the following fields are completed:
-                <ul>
-                    <li>- City</li>
-                    <li>- State/Region</li>
-                </ul>
-                {% endif %}
-                {% if contact and can_update_contact %}
-                    <div class="navigation">
-                        <h2>Actions</h2>
-                        <a class="btn" href="{% url 'edit_location' %}?partner={{ partner.id }}&id={{ contact.id }}">Add Location</a>
```
tdphillips: We need this. This is how we add locations to contacts.

gulp/src/cbo/components/commonCbo/Sidebar.js
```diff
@@ -30,15 +30,8 @@ class Sidebar extends Component {
     }
     dispatch(updateCboStatus(checked));
   };
-  contactLocationMessage = () => {
```
aljordan82: This was added last sprint, Do we really want to remove it?
@mwarnsleyDE didn't just add this?

strempe:
@aljordan82 
**DONE** – added span and location button. 
"Location button and removing the span4"

screen shot 2017-09-26 at 12 22 36 pm



**DONE** – removed changes from gulp/src/cbo/components/cboRoutes.jsx

gulp/src/cbo/components/commonCbo/Sidebar.js
mwarnsleyDE: @aljordan82 yes. @strempe did you mean to delete this for some reason because i noticed it looking at the code as well? Was this placed somewhere else or are we getting rid of the message?

mwarnsleyDE:
@strempe other than the merge conflicts to fix and possibly the comment from above with deleting the contactLocationMessage, I don't see a problem with the rest of the code. NICE JOB

templates/mypartners/edit_item.html
```diff
@@ -116,6 +103,7 @@ <h2 class="top">Tips</h2>
             </div>
         </div>
     </div>
+    </div>
```
tdphillips: Are these divs balanced?

gulp/src/cbo/components/CboApp.js
```diff
@@ -1,16 +1,16 @@
-import React, {Component} from 'react';
-import {Route} from 'react-router-dom';
+import React, { Component } from 'react';
```
koonsies: Was there a conversation about switching to spacing around object imports? I'm not particularly for or against it, but barring good reason for the refactor, I'm not sure it's worth the change. As of right now, it just breaks consistency.

strempe:
@koonsies It's a formatting extension in VS Code called prettier, it automatically formats when saved.

koonsies:
Understood. That said, auto-linters can be pretty intrusive. Unless as front-end devs you've all decided you'd like to switch to that style, I'd disable it. The problem with auto-formatters is they create a lot of "fluff" in your PRs and make it more difficult to figure out where the actual changes are. Do they have a place? Perhaps, but most of the benefits require universal adoption.

strempe:
@koonsies I'm pretty sure I was the last out of the four of us to install it....so we all probably have some fluffy PR's. I can disable it, it wasn't my call.

koonsies:
Haha. Fair enough. It's not my call either. I'm not even a front-end dev! That said, I hadn't noticed it in anyone else's PRs. Aladeen has a PR up now that still has the no-spaced objects.

tdphillips:
Tacking `?w=1` to the end of the url when looking at the list of files changed will hide most whitespace changes.

koonsies:
Ah! Thanks, Troy. I knew there was a way to do it, but I hadn't remembered how.

strempe:
@tdphillips that's really cool. Thanks for the tip!

strempe:
@koonsies hmm, you're right. I'll look into it. :+1:

templates/mypartners/edit_item.html
```diff
@@ -94,19 +94,6 @@
         </div>{# /delete-confirm #}
         <div class="span4">
             <div class="sidebar">
```
aljordan82: The sidebar can be wrapped in `{% if contact and can_update_contact %}` so it doesn't show the grey background.
screen shot 2017-09-27 at 7 47 51 am

strempe:
@aljordan82 
**DONE** –  "The sidebar can be wrapped in {% if contact and can_update_contact %} so it doesn't show the grey background."
screen shot 2017-09-27 at 11 25 12 am

PR #3537, opened 2017-09-26
[PD-3737] history not working multi selects

mwarnsleyDE:
## What's this PR do?
PR fixes the history back and forward buttons for the multi-select combos

N/A

Pull down repository. Go into site traffic reports and select from the multi-select combos. Use the back and forward buttons to make sure history is working

### Front-End

- [x] Works on IE 11 and Edge
- [x] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


gulp/src/analytics/components/AnalyticsApp.jsx
```diff
@@ -69,7 +69,7 @@ class AnalyticsApp extends Component {
       tabsMenuActive: false,
     });
   };
-  handleNewLocation(location, action) {
+  handleNewLocation = (location, action) => {
```
koonsies: I am not up to speed as to the significance of this change.

iartemiev: This is done so that we don't have to explicitly bind the functions to `this` in the constructor since it behaves like any other arrow function wrt `this`

mwarnsleyDE: OH yea @koonsies it's basically a Class Property that's used in Babel. So we implemented it maybe last sprint or so I think. So now instead of doing this.someFunction = this.someFunction.bind(this); inside of the constructor we just use it like this someFunction = () {} and we don't have to bind it anymore. We also don't have to put the state inside of the constructor anymore either

koonsies: Fair enough. I'll keep that in mind moving forward.

gulp/src/analytics/actions/table-filter-actions.js
```diff
@@ -231,10 +231,10 @@ export function doSetMultiFilterOptions(filterValue, activeReport, reset) {
       };
       const filterData = await api.setMultiFilter(allFilterData, activeReport, dimension, multifilter);
       dispatch(setMultiFilter(filterData));
+      dispatch(markNavLoadingAction(false));
+      dispatch(doBumpHistory());
```
koonsies: This looks sufficient. Normally when something fails to be stored in history it's because an action triggered a change to redux without calling the "doBumpHistory" function. Perhaps we can find a more automatic way of handling this in the future.

mwarnsleyDE: @koonsies Honestly when I first looked at it, I didn't even know it was that simple as to why the filters weren't loading in history lol. I had to look at the code forever in the Analytics App and see what you were passing and everything else but I did get a better understanding of it even though I'm not too equipped with history stuff lol.

koonsies: It's pretty simplistic. It just backs up data from redux into the history store. When a refresh happens, or a navigation occurs, it loads the data back from redux. If the call isn't made to backup to history, the call to refresh from history won't have any data that was changed since the last backup.

PR #3538, opened 2017-09-26
[PD-3686] update some resource urls for microsites

vigowei:
## What's this PR do?
Switch the url for some frequently requested resources from the old (`d2e48ltfsb5exy`) to new (`d1g1peq7ldieal`), this time around we are updating these 6 libs/files:

- /framework/v2/js/code/jquery-1.7.1.js
- /framework/v2/js/code/jquery-ui-1.8.17.custom.min.js
- /dseo_static/def.ui.microsite.141-20.js
- /dseo_static/underscore-1.3.1.min.js
- /dseo_static/jquery.placeholder.js
- /dseo_static/pager.164-21.js

Before on microsites: 

image

Desired behavior:

image

There are 3 files that had reference to files listed above:
static/filter_carousel.js
templates/myblocks/myblocks_base.html
templates/seo_base.html
For some urls, it's direct substitution, for the ones that relied on `STATIC_URL`, I added a new settings option called `STATIC_URL_PRIMARY`, and exposed it to templates through a context processor. This required updating some settings files, are those acceptable?

run microsites locally, inspect page source after adding and removing `STATIC_URL_PRIMARY` from settings. 

For example, without changing the settings file, you should see this:
image

### Front-End

- [x] Works on IE 11 and Edge
- [x] Works on Mobile

### Separation of Concerns


almathew:
I'm going to be completely honest. I didn't test this in Edge, IE 11, or mobile.
It works in Chrome just fine. Pretty sure the others don't matter.

PR #3542, opened 2017-09-27
Add partner autocomplete to communication record form [PD-3810]

tdphillips:
## What's this PR do?
Adds a Partner autocomplete to the communication record form if you reach it from the common action button.

We also remove waffle switches as the switched feature has been released.

2017-09-27-152452_801x257_scrot

Most of the code is in `mypartners/forms.py`. A bit more is in `mypartners/views.py`.

Go to PRM. Click the button. Open a partner. Click the button. Go to NUO and Partner Assessment. Click the button on each.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


gulp/src/partnerassessment/components/PartnerAssessmentApp.jsx
```diff
@@ -52,6 +42,16 @@ class PartnerAssessmentApp extends Component {
   }
   render() {
     const {partnerName, partnerId, primaryContact} = this.props;
+    const commonAction = (
```
tdphillips: We do this in `render()` in `PATertiaryNavigation.jsx` so I assume that's okay here as well.

mypartners/forms.py
```diff
@@ -364,33 +366,81 @@ def save(self, request, commit=True):
         return instance
 
 
+class HorizontalRadioRenderer(forms.RadioSelect.renderer):
```
tdphillips: This has moved. Nothing was changed.

mypartners/forms.py
```diff
@@ -364,33 +366,81 @@ def save(self, request, commit=True):
...
+                id='%s_%s' % (self.attrs['id'], choice[0]),
+                name=self.name,
+                value=choice[0],
+                title=self.radio_tooltips.get(choice[0], '')
+            ))
+        return mark_safe(u'\n'.join([u'%s\n' % w for w in radios]))
+
+
+@register('partner')
+class PartnerLookup(LookupChannel):
```
tdphillips: Basically straight from the django-ajax-selects docs.

mypartners/forms.py
```diff
@@ -364,33 +366,81 @@ def save(self, request, commit=True):
...
+        return item.name
+
+
 class ContactRecordForm(NormalizedModelForm):
     date_time = SplitDateTimeDropDownField(label='Date & Time')
     length = TimeDropDownField()
     attachment = MultipleFileField(required=False,
                                    help_text="Max file size %sMB" %
                                              MAX_ATTACHMENT_MB)
+    score = forms.ChoiceField(
```
tdphillips: This looks like an addition but really it's replacing `ContactRecordForm` with `ContactRecordFormWithAssessment`

templates/mypartners/edit_record.html
```diff
@@ -47,7 +53,6 @@
                 {% include "includes/form-error-highlight.html" %}
                 {% csrf_token %}
                 <input type="hidden" name="company" value="{{ company.id }}">
-                <input type="hidden" name="partner" value="{{ partner.id }}">
```
tdphillips: We're including `partner` in the form now. We don't need to fake it here.

templates/mypartners/edit_record.html
```diff
@@ -85,6 +90,8 @@
 {% endblock %}
 
 {% block extra-js %}
+    {{ form.media }}
```
tdphillips: django-ajax-selects requires some JS.

tdphillips:
This needs some CSS work.

tdphillips:
I need to hide some fields and do ajax shenanigans.

tdphillips:
There's a bunch of jQuery. Read it and weep.

Proof of concept. If we don't want to do this, no problem.

static/partner_edit_record.192-28.js
```diff
@@ -1,10 +1,50 @@
...
+
 $(function() {
+    $("#prm-application-body").on("click", "[id^='kill_'][id$='id_partner']", function() {
+        $("input#id_partner_text").show();
+        hide_form();
+    });
+
+    $("#prm-application-body").on("click", ".ui-autocomplete li", function() {
+        $("input#id_partner_text").hide();
+        populate_contacts($('#id_partner').val());
```
tdphillips: This should probably be memoized somewhere so we don't retrieve the same list of contacts multiple times. I'll get to that soon.

koonsies:
Great job with the look and feel of the controls! During my review, I found that the following workflow throws an error

1) Select a partner from the partner list
2) Click the common action button
3) Click the "trash" to remove the partner as the selected partner
4) Select another partner
5) Fill out the form as usual
6) Save and see the failed validation on contact.

tdphillips:
Whoops. Thanks. I should check more than just the happy path.

tdphillips:
Sorry, that change was stashed. I pushed it. There are still some tests to work on, however.

koonsies:
This is weird behavior and I don't know what to make of it. If you create a new CommRec via the common action button, then immediately click "Edit" on the CommRec summary screen shown, you can change the partner on the form still. If you click back to Partners, select the same partner and go to that communication record summary, edit does not allow changing of the partner.

Not sure why this is, but I assume we don't want those two to be inconsistent.

koonsies:
Oh, it's because "common" remains in the url after saving the form.

myreports/tests/test_views.py
```diff
@@ -548,6 +553,7 @@ def test_select_data_type_api(self):
             u'selected_report_category': u'communication-records',
             u'selected_report_intention': u'prm',
         }
+        import ipdb; ipdb.set_trace()
```
koonsies: ipdb statement

tdphillips: It's been fixed.

PR #3547, opened 2017-09-28
Move the Edit and Copy buttons

aljordan82:
Move the buttons and added the Saved Search title.

## What's this PR do?
In Saved Search and PRM Saved Search, the Copy and the Edit buttons were moved.
Added Saved Search title in PRM.
Truncated long Saved Search titles 
screen shot 2017-09-28 at 1 53 42 pm
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


strempe:
Looks good, there are a few things that need editing to match the style of some of our other tables. 

**Icons:**  
- Edit – We need to stick with one edit icon 'fa-pencil'
- Copy – It's small and possibly picky, but could we switch the copy icon to 'fa-clone'. These images are small with little pixel space to show detail. If we can always use simple icons that still communicate purpose, we are doing it right.  

**Typography:** (for now)
-  'edit' and 'copy' should not be bold. 
- Subtitle – if you can also remove the bold style from 'Matching Jobs....." that would really help with hierarchy.
 
**Color:** 
- Change the font and icon color to #444444
- Add a responsive hover to the print and edit icons. Color: #797979

PR #3551, opened 2017-09-28
[PD-3712] 404 handling and correcting Non-referral Apply Rate value format

tejaswinigutti31:
## What's this PR do?
1. First Viewed On is being 'my.jobs' when it actually is a 404. Filtered out the records doing this
2. When non-referral apply rate is None, 'Unavailable' is being displayed in the reports. Corrected it to display 0.00
check the datasource file job_views
Test Job/Site analytics report to identify the above changes.
### Separation of Concerns


tejaswinigutti31:
I've combined PD-3712 and PD-3839 into this PR as they are very small changes in jobviews datasource

PR #3555, opened 2017-09-29
[PD-3850] hotfix comm rec models

koonsies:
## What's this PR do?
Moves the validation logic from a form to the model, since multiple forms reference this model and the lack of validation can cause errors.


aljordan82:
This PR is going against QC! should it?

koonsies:
Haha whoops..

PR #3556, opened 2017-09-29
[PD-3849] Timeline does not align with Table data

tejaswinigutti31:
## What's this PR do?
Timeline data doesn't seem to reconcile with table data in Job/Site analytics. 
Check the code change for run_gather_timeline_data in job_views datasource
In Job/Site analytics, check that sum of Job views on timeline matches with that in the table
demote_joins turned out to be necessary for timeline data as well
### Separation of Concerns


almathew:
I am happy with this PR, but I am also the person who suggested this as one of the valid ways of fixing the problem, so I don't want to be the one who merges it.

PR #3557, opened 2017-09-29
remove overview page

strempe:
## What's this PR do?
Removes 'Overview' page from manage users.

screen shot 2017-09-29 at 2 23 20 pm

Manage Users, the first view should be the User list. The tab should be active. 

Click through tabs, make sure routes work. 
* I did not rename the routes, but would happy to do so

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


aljordan82:
This looks good!

PR #3558, opened 2017-09-29
Pd 3673 nuo table unreadable

aljordan82:
## What's this PR do?
Fix NUO table in mobile
Remove css file that is not really used.

Test NUO and TJD tables in different browsers and screen sizes.

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


strempe:
This looks good to me! Will merge at the end of the day if there are no objections.

koonsies:
Can we force the button to not shorten the text? I think it looks a bit strange to have the text truncated.

screen shot 2017-09-29 at 3 48 51 pm

koonsies:
Tested in IE11, works even better there (the button text doesn't truncate)

PR #3565, opened 2017-10-02
[PD-3854] Filtering bots from Site Traffic data

tejaswinigutti31:
## What's this PR do?
Filters out bots from Site Traffic reports(new and old ones)
The datasources for site traffic are updated to filter out bots
Ensure that no bots are displayed in the data
### Separation of Concerns


darrint:
`AnalyticsEvent` in `make_data` needs `is_bot` too.

Do we need to filter `is_bot` for timelines as well?

Is it worth writing a single function in each datasource to sanitize all three queries (assuming timeline matters)?

tejaswinigutti31:
@darrint Yeah we can just write it into a single function, I'm currently making the change...will be pushing up another commit shortly

myreports/datasources/analytics_event.py
```diff
@@ -87,11 +88,16 @@ def run_unaggregated(self, company, filter_spec, values,
     def run_gather_timeline_data(self, company, filter_spec, field=None, partial=None):
         if not filter_spec.is_valid():
             return []
+        timeline_data = self.timeline_queryset(company, filter_spec, field, partial)
+        return list(timeline_data)
+
+    def timeline_queryset(self, company, filter_spec, field=None, partial=None):
```
vigowei: can you explain the reason for separating `run_gather_timeline_data` into 2 methods?

myreports/datasources/goal_events.py
```diff
@@ -184,13 +184,17 @@ def run_aggregated(self, company, filter_spec, values,
...
+
+    def filter_queryset(self, query_set, values):
         if 'goal_minus_one' in values:
             query_set = query_set.filter(~Q(goal_minus_one=''), ~Q(goal_minus_one=None))
         if 'goal_minus_two' in values:
             query_set = query_set.filter(~Q(goal_minus_two=''), ~Q(goal_minus_two=None))
-        result = [self.extract_record(r, values, ['num_pageviews', 'visits_count']) for r in query_set[0:200]]
-        return self.calculate_fields(company, result, filter_spec, values)
-
+        query_set = query_set.exclude(is_bot=True)
```
vigowei: same as other instances, in my opinion, this exclude can be pulled into `apply_filters` which is called by most methods, that way you don't have to worry about missing excluding bots for all queries. 
For example `run_gather_apply_clicks_timeline_data`

myreports/datasources/analytics_event.py
```diff
@@ -64,6 +64,7 @@ def aggregated_queryset(self, company, filter_spec, values,
         query_set = self.annotate_query_set(query_set, values)
         if 'num_pageviews' in values:
             query_set = query_set.order_by('-num_pageviews')
+        query_set = query_set.exclude(is_bot=True)
```
vigowei: can go into apply_filters

myreports/datasources/site_traffic.py
```diff
@@ -262,12 +263,18 @@ def get_partial_field(self, field):
     def run_gather_timeline_data(self, company, filter_spec, field=None, partial=None):
         if not filter_spec.is_valid():
             return []
+        timeline_data = self.timeline_queryset(company, filter_spec, field, partial)
+        return list(timeline_data)
+
+    def timeline_queryset(self, company, filter_spec, field=None, partial=None):
```
vigowei: not sure if this refactor is really needed

myreports/datasources/analytics_event.py
tejaswinigutti31: This datasource uses SimpleDjangoORM, so change in apply_filter for is_bot will turn being an error for job_views. It will be simpler if its handled here

myreports/datasources/analytics_event.py
tejaswinigutti31: Increase in number of filters, I felt it would be tidier with a separate method. but as its not really huge and adding a separate function is increasing number of lines of code, I've taken it off. Thank you!!

myreports/datasources/goal_events.py
tejaswinigutti31: Done!!

PR #3566, opened 2017-10-02
Pd 3811 tjd common action

strempe:
## What's this PR do?
Creates common action button for TJD

Styles have not been added. 

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


aljordan82:
@strempe this has extra files. is it based off of the feature branch?

strempe:
@aljordan82 yes, it is based off the feature branch

gulp/src/cbo/components/CboApp.js
```diff
@@ -33,7 +33,7 @@ class CboApp extends Component {
     return (
       <div>
         <Confirm />
-        {navigationRoutes(this.confirmWizardNavigation)}
+        {navigationRoutes}
```
aljordan82: This is reverting QC changes.

PR #3568, opened 2017-10-02
[PD-3778] Demo: First API view for data.

darrint:
## What's this PR do?

Adds an endpoint with some documentation for a theoretical 3rd party developer interested in querying our data.

There is a basic UI for assisting with prototyping queries. If there are browser or other problems with it we won't fix it at this time.

2017-10-02-112956_443x821_scrot

* The view work is done in `views.py` and `serializers.py`.
* Also look at the new endpoint schema for datasources. 

Go to https://secure.my.jobs/datasources/v1/page/ . Try to get some data.

na

### Separation of Concerns

### Tests


tejaswinigutti31:
I've tested this locally. Data outputted by the end point looks pretty good!

myreports/datasources/api/serializers.py
```diff
@@ -0,0 +1,68 @@
...
+    timestamp_start = serializers.DateTimeField(default=lambda: utcnow() - timedelta(days=30))
+    timestamp_end = serializers.DateTimeField(default=lambda: utcnow())
+
+    def validate_filter(self, data):
+        if not isinstance(data, dict):
+            raise serializers.ValidationError('Must be a dictionary.')
+        return data
+
+    def validate_include_columns(self, data):
+        if not isinstance(data, list) or len(data) < 0:
```
vigowei: don't think len() can ever be negative, did you mean `len(data) < 1` or `not data`? I did notice that when you post with include_columns: [], results contain all columns, maybe that behavior is intended.

darrint: Good catch. I added some tests.

PR #3573, opened 2017-10-03
[Pd 3834] fix topbar issues

aljordan82:
## What's this PR do?
Show ellipses for long company names
Prevent “My Apps” from wrapping
Show dropdown arrow if there were more than company
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


koonsies:
The ellipses are good when a user has multiple companies. When they have only one, though, the text gets all wonky.

screen shot 2017-10-03 at 11 23 40 am

aljordan82:
@koonsies Good catch!

PR #3574, opened 2017-10-03
tdj common action button

strempe:
## What's this PR do?
Add TJD common action button.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


gulp/src/cbo/components/cboRoutes.jsx
```diff
@@ -79,9 +72,26 @@ export const routes = [
   },
 ];
 
+const commonAction = wizardConfirm => (
+  <UserCan neededActivities={[activities.create_tjd]} key="addOrganization">
+    <NavLink
+      exact
+      to="/newcbo"
+      className="nav-action-btn"
+      activeClassName={'active'}
```
aljordan82: Those are not needed
```
 activeClassName={'active'}
 key="item-2"
```

gulp/src/cbo/components/cboRoutes.jsx
```diff
@@ -79,9 +72,26 @@ export const routes = [
   },
 ];
 
+const commonAction = wizardConfirm => (
+  <UserCan neededActivities={[activities.create_tjd]} key="addOrganization">
+    <NavLink
```
aljordan82: This can be a <Link  instead on <NavLink

PR #3562, opened 2017-09-29
[PD-3852] Analytics-metrics on incorrect reports

tejaswinigutti31:
## What's this PR do?
Fixed incorrect metrics on analytics reports
## Where should the reviewer start?
Verify schema and schema dimension
## How should this be manually tested?
Intended to Apply and Reached Job Source should not be on site owner syndication (site analytics) 

Referral Apply Rate and Non-referral Apply Rate should not be on job owner syndication (job analytics)

### Separation of Concerns

- [x] Back end "business logic" is implemented in plain python, Django model classes, and Django forms. (not Django views or other frameworks)
- [x] Concerns of UI, api, and model are in separate code units.

PR #3577, opened 2017-10-04
Master to Staging

JLMcLaugh:
We haven't backported this hotfix yet: https://github.com/DirectEmployers/MyJobs/pull/3562


vigowei:
looks like this hot fix split `data_cols_job_views` into two parts and got rid of the original definition, while it's imported by the new API feature for myreports.

PR #3585, opened 2017-10-05
[PD-3721] Block transitions on PA creation form

koonsies:
**NOTE** This PR includes the changes from https://github.com/DirectEmployers/MyJobs/pull/3597
## What's this PR do?
Prevents transitioning while editing the partner assessment form.

Partner Assessment

Start a new assessment or edit an existing one.
Change a value in one of the fields.
Try reloading, navigating away, etc.

Notice there are two distinct dialogue boxes if you hit "Reload" or click the "Back" link from the assessment editor. This is due to React-Router not handling navigation outside of React. Navigating to a link in Django, or another webpage cannot be handled in the same way internal React navigation is. Instead, it must be handled via window.onbeforeunload, which has a different dialogue window in Chrome.

### Front-End

- [x] Works on IE 11 and Edge
- [ ] Works on Mobile


gulp/src/partnerassessment/components/Form/AssessmentForm.jsx
```diff
@@ -28,23 +30,39 @@ class AssessmentForm extends Component {
...
+
   onDropDownChange(value) {
     const {dispatch} = this.props;
     dispatch(updateRating(value));
   }
 
+  formTouched() {
+    if (!window.onbeforeunload) {
+      window.onbeforeunload = () => {
+        return 'Your changes are unsaved. Continue?';
```
aljordan82: Can you make this string a variable so it isn't duplicated.

gulp/src/partnerassessment/components/Form/AssessmentForm.jsx
```diff
@@ -28,23 +30,39 @@ class AssessmentForm extends Component {
...
+    window.onbeforeunload = () => {};
+  }
+
   onDropDownChange(value) {
     const {dispatch} = this.props;
     dispatch(updateRating(value));
   }
 
+  formTouched() {
+    if (!window.onbeforeunload) {
```
aljordan82: This wont work if the `onbeforeunload` is added to the window object somewhere else in the code. but I don't see a way around it.
So this is fine as is!

vigowei:
after making changes on the page, leaving it there for a while, the "Do you want to leave" window keeps popping up. I accidentally ran into this when I got distracted from testing the branch.
image

koonsies:
Ahh. Yeah, that's an interesting situation. We have an auto-logout timer that runs through javascript. Did you try hitting "Stay" and see what happens? Probably keeps you in the broken React app (since the APIs would fail). We probably should update the auto-logout script to clear any function in the window.onbeforeunload placeholder.

koonsies:
Removing DNM tag, added handling in the auto-logout to clear the navigation block.

PR #3587, opened 2017-10-06
Fix a bug, city field not refreshing

aljordan82:
## What's this PR do?
Fix bug in SearchInput.jsx
The bug happens because the SearchInput constructor I called only once
any subsequent updates to the SearchInput “value” prop were ignored.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


aljordan82:
@koonsies last attempt to fix this caused a regression and it needed a hotfix.

koonsies:
Code looks sound. I'll review this after I finish the common-action PR

tejaswinigutti31:
The fix actually make the clone that was broken work properly for city field

koonsies:
I don't remember the regression you mentioned. However, this seems to work fine.

PR #3588, opened 2017-10-06
Merge staging to qc

darrint:
## What's this PR do?

This merges fixes back from staging.


darrint:
I tested that the service starts and analytics still works.

PR #3589, opened 2017-10-06
Use base-react.html on root url if user is authenticated [PD-3864]

tdphillips:
## What's this PR do?
Toggles what is displayed on `/` if the user is logged in. This does not touch the login action so logging in still redirects to the profile page.
Unauthenticated:
2017-10-06-083335_955x745_scrot

Super wonderful authenticated styles:
2017-10-06-083419_1440x81_scrot
The `home` view, I suppose.
Log out. You see a form. Log in and go back to `/`. You see my amazing CSS skills.
This is the beginning of our dashboard work. It was decided that the unauthenticated portion could remain as-is.
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


tdphillips:
I'm sure tests will fail. We have tests checking the content of the page I mucked with.

aljordan82:
Looks good! We can merge this even if the tests are failing

PR #3594, opened 2017-10-06
 [PD-3866] Redirect click on logo to secure

tejaswinigutti31:
## What's this PR do?
On clicking the logo My.jobs on the topbar, we are now redirected to secure.my.jobs
Its a small change in topbar.html template
Create a waffle flag with name "dashboard" and on setting it, clicking on logo should be redirected to secure.my.jobs
If the flag is turned off, it is redirected to www.my.jobs
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


templates/includes/topbar.html
```diff
@@ -21,9 +22,15 @@
     <div class="topbar-inner">
         <div id="de-topbar-content" class="wrapper">
             <div id="logo-block" class="de-topbar-item">
-                <a id="anchor-logo" href="http://www.my.jobs/">
+                {% flag "dashboard" %}
+                    <a id="anchor-logo" href="http://secure.my.jobs/">
```
aljordan82: We can use `https` instead of http

PR #3598, opened 2017-10-09
[PD-3847] Add Delete to Partner Assessments

koonsies:
## What's this PR do?
Adds delete functionality to partner assessments

Partner Assessments

Create an assessment
Edit that assessment
Click "delete"
Confirm
See that it doesn't show up in the list (it's actually archived, not deleted)

This relies on the ubuntu:xenial changes in another PR. 

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


koonsies:
Closing until I can add tests

gulp/src/partnerassessment/components/Form/AssessmentForm.jsx
```diff
@@ -157,6 +180,11 @@ class AssessmentForm extends Component {
         <Button onClick={this.handleSave} bsStyle="primary" bsSize="small">
           Save
         </Button>
+        {match.params.assessmentId ?
```
aljordan82: Should this be wrapped in UserCan?

aljordan82:
@koonsies what's the advantage of confirm over the custom modal?

koonsies:
I am a bit torn on that myself. For me, I was wondering what the advantage of a custom modal was over the built-in (as we have to maintain a custom, but built may be more reliable)

koonsies:
I kind of lean toward the window.confirm, since that is what React-Router uses for Prompt, and we can't coerce another window type as of right now.

I don't have a major commitment to either. Kind of a toss up for me. Window.confirm was easier, less code overhead.

aljordan82:
I think it would be better to keep the Alerts consistent across apps.

gulp/src/sass/partner_assessment/components/_assessment-form.scss
```diff
@@ -30,17 +30,23 @@
...
+
   .form-group {
     label {
       margin-left: 5px;
       margin-bottom: 8px;
     }
   }
 
+  .btn {
+    margin-right: 5px;
```
aljordan82: The margin-right is affecting other buttons.

aljordan82:
Front-end code is fine, this can be merged when the python code get reviewed

PR #3601, opened 2017-10-10
Fancy logging decorator.

almathew:
## What's this PR do?
Introduces a fancy logging decorator, which can hopefully be used to clean up the logging in imports as a starting place.

Also cleans up log files, since the only justification for the reason we had two log files was that I didn't bother fixing it when I merged the projects. Is the chosen log file the best one for the job? I don't know, but it's one that already exists, so the risk in making a new one that makes more sense isn't really worth it. 

## Usage example
screen shot 2017-10-10 at 10 52 30 am

results in the log line:
```
INFO 2017-10-10 10:51:09,429 decorators 13 140496719558400 {'duration': 0.6159131526947021, 'nexturl': 'None', 'func': 'home'}
```

The decorator is in universal/decorators.py.
This includes a manager to keep track of the current frame so that variables can be accessed after the function has already returned.
This is all inspired by the post here: http://scottlobdell.me/2015/08/taking-logging-next-level-python/

See the screenshot above. Try it out for yourself.

Logging for imports is starting to get out of control and make the code unreadable. After brainstorming a few ways to clean it up, this was the best solution I was able to find.

### Separation of Concerns


darrint:
Do you have any sense of the overhead that this introduces?

almathew:
In terms of? 

Memory is my biggest concern, because it will probably mess with garbage collection a little bit, but with everything else the overhead is roughly equal to regular logging. 
I'm choosing to be ok with the memory issue right now because I'm fairly confident that it won't change anything in imports, which is where this will have the biggest immediate impact. 

It's a potential issue we'll probably have to be aware of. On the plus side it should be super easy to diagnose when it happens.

almathew:
Actually, I take that back. 

It will be slower too. But I think how slow is a property of the variables declared and all that. My claim that it's not much of one is based on the fact that, when testing with imports. it wasn't noticeably slower. However, I'm also not capable of running large imports, so actual results may vary.

almathew:
TL;DR: Possible memory issues, also possibly makes things slower. Actual results will vary based on how it's used. I didn't run into any problems using it myself. 

Ok, I'm done spamming everyone with comments for now.

PR #3613, opened 2017-10-12
Exclude ga in dev environment

aljordan82:
## What's this PR do?
Exclude Google analytics in dev environment


tdphillips:
Should we do the same for quantcast?

almathew:
I'm not really sure if anyone has a strong opinion about this, but I didn't bother having @aljordan82 write a whole new context processor just for passing a setting in. 
Maybe the webpack one should be more generally named or something if we need to add more in the future ¯\_(ツ)_/¯

PR #3615, opened 2017-10-12
Attempt to fix disappearing spinner

aljordan82:
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


aljordan82:
@koonsies 
reloadFromProps()  as getting called twice, this cause the api requests to be sent twice. and caused the spinner state to be set to false when the first batch pf api calls are resolved, and when the spinner is hidden, user can select another category from the dropdown before the second batch of the api calls are resolved.

koonsies:
Ah, good catch. I was trying to figure out why this was happening, then I remembered that there is a redirect component that is mounted each time the category changes, thereby unmounting SetUpReport, and remounting (calling componentDidMount on SetUpReport) each time there is a history.push to that location.

koonsies:
Weird. This is showing the refresh fixes as being in this PR, even though that code is already in QC.

PR #3618, opened 2017-10-13
Remove reference to nonexistent template

tdphillips:
## What's this PR do?
Removes a reference to a template that was removed some time ago.
It's one line.
Hit the activation page on your dev server. If debug is enabled, you should not see a yellow screen. If debug is disabled, you should not see an error in your runserver log.
We removed a template. We did not remove references to it.
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


almathew:
:tada:

PR #3622, opened 2017-10-13
[PD-3882] fix refresh report

koonsies:
## What's this PR do?
Adds an action to clear the "refreshing" state from reports after they have been selected to refresh.

* Create a report
* Export that report, check the values.
* Modify something that would cause that report to be different (adding a new comm record, for instance)
* Refresh the report
* Notice the spinner goes away (this is the actual change)
* Verify that running the refreshed report has the change made above.

Not sure how the spinner ever cleared on "Refresh"..

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


aljordan82:
@koonsies Is there a ticket for this?

koonsies:
Done

PR #3623, opened 2017-10-16
Switch to TestCase for faster tests.

almathew:
## What's this PR do?
Use TestCase instead of TransactionTestCase because speed.

It shouldn't.

There used to be 2 tests that required TransactionTestCase. Somewhere along the way I apparently either fixed or deleted those tests. Either way, I wasn't going to keep all the tests on TransactionTest Case unless I have to.

### Tests


almathew:
Frustratingly, failures have nothing to do with actual tests.

almathew:
The improvement is small, but meaningful. 8ish minutes less?

myjobs/tests/setup.py
```diff
@@ -93,9 +95,13 @@ def login_user(self, user):
...
 
 
+def default_site():
+    return SeoSite.objects.first()
+
+
 @override_settings(ROOT_URLCONF="myjobs_urls", PROJECT="myjobs", MEMOIZE=False,
                    HAYSTACK_CONNECTIONS=settings.TEST_HAYSTACK_CONNECTIONS,
-                   TEMPLATES=settings.TEMPLATES,
+                   TEMPLATES=settings.TEMPLATES, SITE=SeoSiteFactory.build(),
```
tdphillips: `.build()` doesn't save the object. Are you okay with that?

almathew: I used .build() for that exact reason.
I personally am ok with it since myjobs-related apps rarely ever depend on seo stuff (postajob is the only notable exception, really). In cases where it doesn't work being a fake object, it should be pretty obvious. I can see why someone would disagree with that approach, though.

There's actually some more stupid stuff that's going on with context processors that caused this, which I've been messing around with, but this is a cleaner solution while I figure out exactly what's needed and what isn't.

PR #3624, opened 2017-10-16
Pd 3871 clean up custom css block

aljordan82:
## What's this PR do?
Clean up "custom-css" block
Reduced the number of generated hashed css blocks by ({% compress css %}) 
Extracted common css file into one block in base-react.html.
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


koonsies:
I ran compress locally and did not see any issues with CSS. Not an expert on Django compressor, but this looks good.

PR #3626, opened 2017-10-17
Order and limit communication records [PD-3887]

tdphillips:
## What's this PR do?
Orders returned communication records by `date_time` and allows limiting the returned records with `&limit=&offset=`.
N/A
mypartners/views.py
Open `/prm/api/communicationrecords/?partner_id=<partner_id>` and ensure things are ordered and nothing is limited.
Tack on some combination of `&limit=&offset=` and ensure things are limited in the manner you would expect.
This will be used for the dashboard system we're working on. @aljordan82 thought others might want it.

This is builtin rest_framework functionality. I'd prefer not writing tests for external code, hence why you don't see any here.
### Separation of Concerns

### Tests


mypartners/views.py
```diff
@@ -2539,15 +2540,19 @@ def get(self, request, format=None):
...
         scored_records = all_records.filter(
             score__isnull=False)
         if scored_records:
             avg = sum([record.score
                        for record in scored_records]) / float(scored_records.count())
         else:
             avg = None
-        serializer = CommunicationRecordSerializer(all_records, many=True)
+        paginated = all_records
+        if 'limit' in request.GET:
```
tdphillips: I wanted to keep our previous functionality (a request without limit or offset returns everything) so I didn't provide a value for the default limit in default_settings. There's no great solution - providing a value limits all uses of LimitOffsetPagination in that manner. Not providing a value forces a default limit of None, which rest_framework interprets as an empty list.

Nice middle ground. Don't provide a value and only use the pagination class if a limit parameter was provided.

mypartners/views.py
```diff
@@ -2539,15 +2540,19 @@ def get(self, request, format=None):
...
+            paginator = api_settings.DEFAULT_PAGINATION_CLASS()
+            paginated = paginator.paginate_queryset(all_records, self.request, view=self)
```
tdphillips: We use APIView, not GenericAPIView. Per the docs, we have to manually invoke pagination.

aljordan82:
This works as expected.

PR #3631, opened 2017-10-19
Allow archiving of OutreachEmailAddresses [PD-3735]

tdphillips:
## What's this PR do?
Adds ArchivedModel superclass to OutreachEmailAddress
N/A
The various api views would be a good place.
Have an outreach address that has records associated with it. Delete it. Verify that you still see your records. Create another identical address.
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


gulp/src/nonuseroutreach/components/OutreachRecordTable.jsx
```diff
@@ -68,8 +68,6 @@ OutreachRecordTable.propTypes = {
     PropTypes.shape({
       // date the record was created
       dateAdded: PropTypes.string.isRequired,
-      // the inbox the record went to
-      outreachEmail: PropTypes.string.isRequired,
```
tdphillips: We don't display this anywhere. Removed from here and the api.

mypartners/views.py
```diff
@@ -1747,9 +1747,17 @@ def api_add_nuo_inbox(request):
         raise Http404("This view is only accessible via POST method, not %s" %
                       request.method)
 
-    inbox = OutreachEmailAddress.objects.create(
-        company=company,
+    inbox = OutreachEmailAddress.all_objects.filter(
         email=request.POST.get("email"))
+    # This looks awkward. Blame ArchivedModel - changing the previous line to `all_objects.get_or_create`
+    # was reasonable but would not allow an archived instance to be saved after retrieval.
+    if inbox and inbox[0].company == company:
```
tdphillips: I'm not super happy with this whole section. Thoughts?

tdphillips:
There's some discussion on the ticket so we may not have to do something weird like this.

tdphillips:
Discussion on the ticket appears to have determined that we should do what I did here.

PR #3638, opened 2017-10-25
First cut at scopes

darrint:
## What's this PR do?

Note that this PR is based on a feature branch.

Converts most Analytics datasources to a new scope model from companies.

* `UserAndCompanyGrant`
* Note that all converted datasources have lost their `is_valid` filter method. All queries in those datasources are based on the scope and timeframe provided from the API consumer.
* `analytics.api.py` Uses the new grant system instead of `get_company`. Some special cases that were scattered around the API are gone.
* The unit tests from the datasources should fail if you comment out the scope limit in the corresponding datasource, at least for `run_with_meta_aggregated`.

* The tables and charts appear to still work.

* I and @almathew did most of the design on this scope system. My hope is that within a year we will be using a token authentication system in place of the user/company cookie that the current login creates. If/when we get there we will no longer have a use for `UserAndCompanyGrant` and will instead have other implementations of `Grant` based purely on tokens.

### Separation of Concerns

### Tests


analytics/api.py
```diff
@@ -172,37 +173,44 @@ def dynamic_chart(request):
 
 
     """
-    company = get_company_or_impersonated(request)
+    grant = get_grant_from_request(request)
+    grant = transfer_grant_if_needed(request, grant)
```
tdphillips: I wasn't fond of `get_company_or_impersonated` but it hinted at what it might be doing.

Months or years down the line, are you going to know that `transfer_grant_if_needed` deals with impersonation and nothing else?

api_management/models.py
```diff
@@ -7,6 +8,122 @@
...
+
+def disallow_empty_scope_constructor(self, arg, error_msg):
+    if not arg:
+        raise ScopeError(error_msg)
+    super(self.__class__, self).__init__(arg)
+
+
+class BusinessUnitScope(BusinessUnitScopeTuple):
+    def __init__(self, business_units):
+        disallow_empty_scope_constructor(self, business_units, "No business units allowed.")
```
tdphillips: None of these statements make sense in combination. If the scope constructor is empty, we call `disallow_empty_scope constructor`. Makes sense so far.

But then we check if an arg is not provided, and if it's not provided say that that arg is not allowed.

api_management/models.py
```diff
@@ -7,6 +8,122 @@
 from django.utils.timezone import now
 
 
+BusinessUnitScopeTuple = namedtuple('BusinessUnitScope', ['business_units'])
```
tdphillips: Couldn't a tuple with a constructor be called a class?

analytics/api.py
darrint: Thank you for paying attention to this.

The name did need to change. When I ran some of this by @fezick he told me that he didn't like calling analytics company switching "impersonation" because it isn't that. It's granting someone  access to data they wouldn't otherwise be able to see. But there is no other user being impersonated. "Transfer" is a verb I made up that is distinct from impersonate and the mental image I get from it seemed appropriate to me. I didn't bother changing the activity name though, as that already exists and is visible to at least some users. Hopefully only staff but I don't know that for sure.

I looked at how I'm affecting the real kind of impersonation. Best I can tell Django changes the user in some middleware before the view is invoked so `UserAndCompanyGrant` should just work as before. If someone knows better than me about that I'd like to fix this.

I'm viewing `Grant` as a superset of activities. Activities regulate what actions you are allowed to do at all. Grant provides a list of allowed activies and also other lists of limits, like what data you are allowed to see.

api_management/models.py
darrint: Would you rather see classes with __eq__, __str__ etc. methods? I'm ok with doing that.

api_management/models.py
darrint: Thanks again. These are helping me write out some important things.

So here's an assumption I'm making that I think is important: We don't like empty scopes because we're likely to just feed empty things to other internal apis and queries. Those api's/query_sets might interpret an empty constraint as no constraint and give someone access to everything. Therefore make it hard for a Grant to construct an empty scope.

So the error message...

The caller asked for allowed business units but there are none. The caller gets back an error meaning there are no business units allowed in this grant. That's not legal so I'm raising an error.

However, I did confuse myself at least once looking at one of those error messages. Maybe I should make the error a complete sentence?

analytics/api.py
almathew: For the record, I think @darrint and I disagree on this, but I can see this expanding out beyond impersonation in the future.

I'm ok with a better name, but having impersonation in it, when really it is a modification of privileges/permission/whatevers that may or may not be related to impersonation, seems short-sighted.

darrint:
https://github.com/DirectEmployers/MyJobs/pull/3638/files#diff-350b799e4396b2873fb512bdf6666a29R2

Helpful?

analytics/api.py
darrint: @almathew The only thing I'm leaving alone is the name of the activity. The method `get_company_or_impersonated` is replaced with a Grant transfer.

darrint: Ok. So we're going to stick with this for now. I added some docs at the module level in `api_management/models.py` to explain some of this.

PR #3639, opened 2017-10-25
[PD-3252] Allow for deleting reports.

almathew:
## What's this PR do?
Allows users to delete reports.

This is currently a delete forever action, not an archive. I asked about that in the ticket because it's a bit of a deviation from how we currently do things, but it's also not like we're permanently deleting unrecoverable information.

@aljordan82 I think I addressed your request to use listReports properly?

screen shot 2017-10-25 at 2 59 39 pm

Make a report.
Delete a report.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests


aljordan82:
Can the TJD reports be deleted as well?

almathew:
This doesn't add TJD. Just PRM.

aljordan82:
@almathew should TJD be added into this PR? or a new ticket?

almathew:
I will create a new ticket. Now that I know what I'm doing it shouldn't be that hard, but this PR is a complete unit of work, so no reason to stop one just to get the other in.

aljordan82:
Front-end code is fine.

tdphillips:
If you're going to tack your work onto someone else's, you should probably tag your commit with your ticket.

PR #3640, opened 2017-10-26
[PD-3876] Changes needed for Forward Goal Path and a lot of cleanup

darrint:
## What's this PR do?

**Note:** ~Do not merge is in place so we can see where we stand with the unit tests on Jenkins.~ Tests are passing now.

The business requirement driving most of this change is "forward goal path" which drove the need for us to prefab query filters and give them to UI to give back to us.

Our desire was to not layer on yet more hacks but take things apart and put them back together with a more flexible design. As big as the PR is, we could have done a lot more and stopped at the point of adding the scope system.

There are several significant changes in this PR.

### Grant/Scope System

When we started on enabling external API's we envisioned some kind of scope system. This is a way to say that this user or API consumer has access to these exact things. Currently in our API's and template views we usually write code to limit data access by filtering by seo Company. The scope system aims to replace that mechanism (there must be a company) with an explicit interface which can be queried for lists of things. Currently we have lists of activities, buids, and hostnames where a user's employer might have traffic to query (traffic sites).

When we did the initial cut of external API's we more or less left that undefined. Now it's defined. The biggest change there is that when using this system you don't need get_company_or_404. You ask for a grant and possibly ask for a "transfer" (like the analytics misnamed impersonation feature).

The only changes I can easily anticipate are that we may change the way functions obtain a grant and instead attach this to request.user.grant. And we will add more types of scopes.

### UserAndCompanyGrant

This is an adapter so that our existing views which rely on cookie sessions for `secure` will work with the Grant system. This work includes the resolver which builds the grant from the request, as well as the grant itself. The `transfer` method and the `get_company_grant_from_request` methods could use a close look.

Also, we're renaming the analytics-company-impersonate-thing as `transfer` and this is on purpose. Impersonate to us means there is a *person* being impersonated, including all their activities. It requires explicit approval from the company of the person being impersonated. We should not throw that term around lightly.

The analytics feature is just a person who has access to a lot of data looking at different data. I've plugged a possible security hole (maybe?) because I've hard coded a very limited set of activities for the user once they have made the transfer. I can't imagine how one might exploit the old way of doing it but that's not saying much.

### Run with Meta

All datasources used by Analytics were "upgraded" to support a new interface called `run_with_meta`. The old interface return value was a list of dictionaries with "values" in it. The new return value used by the new methods returns some more structured object that contain more metadata.

The most important new metadata for this task was a prefabricated filter for each row returned. When the user clicks on that row in analytics, the new drilled down filter should be the one we provide. This is because in a few cases, the field we're filtering on isn't the same as the one being clicked on.

While we were there we went ahead and give the datasource the responsibility for formatting the values for display, since that isn't a simple proposition anymore, and it's clearer to just do what we want in the datasource.

We're finding that we have uses for metadata at pretty much every level of the query results.

### TempDynamicReport

We wrote `TempDynamicReport` as a way to make datasource queries without creating records. This is turning out to be overkill and it results in an unnecessary serialization/dserialization step, which caused problems. So we're working on removing it in this PR. Instead we're going straight to the datasource json driver and calling it directly. This takes care of what TempDynamicReprort did without the additional overhead.

We could have done more such as convert all the PRM datasource interfaces. This would have been nice even since analytics `*_with_meta ...` calls now now have inconsistent interfaces from the old calls PRM makes. We expect to finish that when we do reporting for screenshots.

We had hoped to eliminate all `search_*_aggregated methods` but there are still a few around with some unit tests on them.

### resolve_scope

Upgraded datasources have a resolve_scope method. This replaces a bunch of code scattered around which modified query filters based on which datasource you were about to run. The newer interface accepts a timeframe and scope which are separate from the filter parameters. The datasource then decides what to do internally with timeframe and scope.

Since each datasource has it's own idea of what scope or scopes it should support we have the `resolve_scope` method where it can attempt to grab the scope it wants used. Job views is a special case because it works with more than one kind of scope. It is the only data source which uses the other parameter to `resolve_scope`.

### apply_clicks.py

`goal_events.py` used to do double duty as the apply clicks datasource since the queries were similar. We split it into it's own datasource because it makes the interface more consistent. Some special cases were removed from API code.

### UI: API calls

We use the `dynamic` api call as a workhorse for getting a page of analytics data. We're relatively happy with that but we had grown several different ways of calling it. We refactored this so that there is only one way to call various API's.

### UI: Component Changes

Some components were written to consume their props directly from data which looks exactly like the return value of `dynamic`. For instance we want maps components to be very generic and just take data as props in a generic way. We didn't get all the way there but we improved the situation.

Many components had special cases in them where they would behave differently based on hard coded strings known to be returned from `dynamic` at different times. We've tried to eliminate as much of this code as possible and make `dynamic` just tell the UI what to do in a clear way.  Again, we're not all the way there but we improved this situation and the behavior is more robust in the face of unusual user actions.

### UI: Unit Tests

Some unit tests have been disabled. There are large swaths of tests that are just check that "reducer updates thing in action". They don't provide any value. **Should we just delete them?**

### Date formats

The API was being inconsistent about date formats. Sometimes they were `mm/dd/yyyy` and sometimes ISO format. We tried to use ISO format consistently. The UI can format dates.

### Report Schema

The schema for all reports is based on scopes now. We left adapters in place for user/company so we could leave old reporting and TJD alone. The unit tests for schema resolution etc. are heavily rewritten and based on grants/scopes now. The changes for schema itself weren't all that big. I would pay the most attention to `resolve_grant` and `report_intentions_for_grant`

### External API

We made the demo external API not have it's own schema and function more like dynamic. It should not be usable in production because there is no way to get a valid token for in the production api management database. (am I right??) 

We removed a separate schema for the external API and just had it use the analytics schema for now.

### SimpleDjangoORM

`SimpleDjangoORM` hasn't quite lived up to it's promise (yet) but we moved it closer. We've made it so that dynamic `clone_*` methods are class attributes. Unit test failures comparing filters are more readable now.

The next pass of reporting improvements we expect to make for screenshots reporting should result in more work done in `SimpleDjangoORM` and fewer methods needed in datasources derived from it.

Go over Analytics as thoroughly as you know how. Including downloading exported reports and randomly bouncing around the UI. The UI should be much more solid vs this kind of behavior now.

Go over PRM and TJD reports. The backend code for these was lightly touched and could break in subtle ways.

### Future

Things we considering doing or hope to do as part of upcoming work:

* Eliminate the entire `search_*` family of methods from datasources. We think that with some thought we can do everything needed with `run_with_meta`.
* Use SimpleDjangoORM more.
* Use grant/scope in PRM and TJD reporting.
* Make grant a property of AbstractBaseUser.
* Simplify the Analytics UI code more.
* Eliminate the need to instantiate column format objects when exporting reports.
* Eliminate DynamicReport completely; stop saving snapshots of report data. Apparently no one wants that.
* Move datasources into the apps which contain the data being queried. Consider making a datasources app which contains just the base classes and utility code. This should help break some unnecessary  dependencies.
* and many more

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


koonsies:
On analytics:

Click the chart or table a a few times until you get a couple of tabs.
Change the date
Click on one of the open tabs (not the active one)
Error on server console

Tested and this does not occur on QC.

koonsies:
I can't track down the exact parameters of this, but the company impersonation isn't working as well as it had.

I have a company has no app access/users. When I try to impersonate that user, I get ```Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/MyJobs/analytics/api.py", line 140, in dynamic_chart
    data_source, values, group_by, intention, data_set = report_for_request(request)
  File "/MyJobs/analytics/utils.py", line 71, in report_for_request
    report_schema, request.user, company)
  File "/MyJobs/myreports/schema.py", line 352, in resolve
    activity_scope = grant.allowed_activities()
  File "/MyJobs/seo/models.py", line 732, in allowed_activities
    role__user=self.user, role__company=self.company, app_access__in=self.company.app_access.all())])
  File "/MyJobs/api_management/models.py", line 91, in __init__
    raise ScopeError("Cannot create empty activities scope.")
ScopeError: Cannot create empty activities scope.```

This error does not occur when operating on quality_control branch.

koonsies:
I do not have site-traffic data (in fact, I don't know how to populate it) but when I attempt to go to the site-traffic report, I get the following error

```
  File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 58, in wrapped_view
    return view_func(*args, **kwargs)
  File "/MyJobs/analytics/api.py", line 149, in dynamic_chart
    scope = driver.resolve_scope(grant, intention.scope_data)
  File "/MyJobs/myreports/datasources/jsondriver.py", line 49, in resolve_scope
    return self.ds.resolve_scope(grant, scope_data)
  File "/MyJobs/myreports/datasources/job_views.py", line 59, in resolve_scope
    return grant.allowed_traffic_sites()
  File "/MyJobs/seo/models.py", line 723, in allowed_traffic_sites
    return TrafficSiteScope([jvs.site for jvs in self.company.jobviewsite_set.all()])
  File "/MyJobs/api_management/models.py", line 84, in __init__
    raise ScopeError("Cannot create empty traffic site scope.")
ScopeError: Cannot create empty traffic site scope.
```

In the QC branch, it just shows the empty analytics screen when I try to access these reports.

aljordan82:
@koonsies good catch! fix pushed. 
There is still a lot of refactoring needed for Analytics. We'll do it gradually.

koonsies:
Test failures

almathew:
Executive decision. I'm closing this until at least tomorrow.

almathew:
Now that it's going to have longer in QC, I'm fine with this being open again.

myreports/datasources/goal_events.py
```diff
@@ -64,12 +86,12 @@ def filter_query_set(self, query_set):
 
             # If the search is explicitly for None, the search
             # needs to be for null values.
-            if hasattr(filter_field, 'value'):
-                filter_value = getattr(filter_field, 'value')
-                field_is_no_filter = isinstance(filter_field, NoFilter)
-                if filter_value is None and not field_is_no_filter:
-                    field = '%s__isnull' % field
-                    filter_field.value = True
+            # if hasattr(filter_field, 'value'):
```
darrint: Can we remove this?

vigowei: @darrint yes, I think so, and this may not be the only data source this exact block is commented out. The reason to get rid of `__isnull` is because it's equivalent to MatchFilter with `value=None`. I'm going to delete those now

vigowei: created a branch off of pd-3876 called (pd-3876-remove-obsolete-code) .. will open a PR once the tests are fixed on the feature branch

PR #3641, opened 2017-10-27
[PD-3718] Radius search.

almathew:
## What's this PR do?
Allows for some locations to be search for via radius from a latitude/longitude rather than exact match.

Whether or not a site does this is managed on the configuration.

It's just a flat term -> latitude/longitude mapping, managed by us. I have a dump of the ones that the US jobs API is currently using to do an initial seed. The term mapping gives us more options to do fun things in the future (military bases? nicknames for cities?).

Radius is determined by a hidden input field.

Also includes zip code in the search terms for location. 

Bulk of the logic was added in the new function in seo/helpers.py. The docstring on that function should pretty much describe my thought process.

Add some term->location mappings to your database. 
Add some jobs that do/don't match those terms. 
Search away.

### Separation of Concerns

### Tests

### Migrations


vigowei:
This is how I tested it manually in case anyone is also looking at this PR:
- apply seo migrations 0018 and 0019
- spin up runsecure and runmicrosites
- in admin, check allow radius search for my.jobs
- in dkm shell, import jobs for a small company. Note their locations and add a nearby centroid, for example, my new jobs a in Richland, WA, I added a geo point (that's within 25mi radius): `TermToCentroidMapping.objects.create(term='Kennewick, WA', latitude=46.2002991, longitude=-119.2470993)`
- in local myjobs, searching with `Kennewick, WA` will also bring up jobs in `Richland, WA`

seo/admin.py
```diff
@@ -90,7 +90,7 @@ class ConfigurationAdmin (admin.ModelAdmin):
                 ('title', 'group', 'status', 'percent_featured'),
                 ('doc_type', 'language_code'),
                 ('view_all_jobs_detail', 'show_social_footer', 'show_saved_search_widget', 'use_secure_blocks',
-                 'template_version'),
+                 'template_version', 'allow_radius_search'),
```
vigowei: doubt anyone would care .. but this row in the admin UI looks crowded. might wanna make them 2 rows or something, totally up to you

seo/helpers.py
```diff
@@ -963,13 +1003,15 @@ def prepare_sqs_from_search_params(params, sqs=None):
...
-    # in each location-y field in the index. If the `exact` parameter is
-    # `true` in the querystring, search locations for EXACT matches only;
-    # the rationale being that if a user clicks on "San Diego" he probably
-    # doesn't give a crap about "San Francisco" or "San Jose" results.
     if loc_val:
-        sqs = sqs.filter(full_loc=loc_val)
+        try:
+            # The multiplication by 1.60934 is to convert miles to km, since Solr
+            # searches in km.
+            radius = float(radius) * 1.60934
```
vigowei: don't think float() is necessary

seo/models.py
```diff
@@ -1119,6 +1119,7 @@ def get_template(self, template_string):
     # navigation section
     defaultBlurb = models.TextField('Blurb Text', blank=True, null=True)
     defaultBlurbTitle = models.CharField('Blurb Title', max_length=100, blank=True, null=True)
+    allow_radius_search = models.BooleanField(default=False)
```
vigowei: what if we also added `radius` as a config option to be used with this (similar to locked css path), so they can specify the radius value as well. Right now the default radius is a constant and there doesn't seem to be an easy way to control that on the front end. Maybe that restriction is intended?

seo/models.py
```diff
@@ -1513,6 +1514,12 @@ def __unicode__(self):
         return self.url
 
 
+class TermToCentroidMapping(models.Model):
+    term = models.CharField(max_length=255, blank=False, unique=True)
+    latitude = models.FloatField(default=0)
+    longitude = models.FloatField(default=0)
```
vigowei: Do we need max, min constraints on these 2 fields? I guess it's fine if these are added by us.

almathew: They're added by us, so I'm totally ok with us causing ourselves problems.

seo/admin.py
almathew: Agreed. The config admin as a whole is getting unusably confusing at this point.
I put it there because that's where all the other random one-off booleans are at. If someone has an opinion on where else it might make sense, I'm happy to move it.

seo/models.py
almathew: The customizable radius is as it was requested. From the descriptions I've been provided so far, I think it's going to be a bit more dynamic and changed more frequently than maybe you're picturing it.

seo/helpers.py
almathew: If provided as a query parameter, the radius will come in as a string, so it needs to be casted to something that can actually be used in multiplication. 
Float might be a bit excessive. I'm not sure DS is going to be that creative with their input. But why not?

vigowei: ah, makes sense.

seo/models.py
vigowei: I fine with way it is, and other ways to handle this if there is a need for radius to be customizable, otherwise ignore me.

seo/helpers.py
```diff
@@ -904,7 +908,41 @@ def create_sq(facets):
...
+
+    :return: The SQS object with the relevant location search applied.
+
+    """
+    if not location:
+        return sqs
+
+    if allow_radius_search:
+        try:
+            centroid = TermToCentroidMapping.objects.get(term__iexact=location)
```
JLMcLaugh: You enforce unique on the model, but this might return more than one result if you're using iexact.

tdphillips: Assuming `TermToCentroidMapping` lives on MySQL, the `__iexact` isn't doing much - it's already case-insensitive.

almathew: Good point.

almathew: At @JLMcLaugh's suggestion, I switched it to `.first()` despite the fact that it doesn't really matter.

PR #3642, opened 2017-10-31
[PD-3884] remove explicit report blocking if no contacts/partners exist

koonsies:
## What's this PR do?
Removes an explicit check on contact to verify that contacts exist before running an export. Since contact records can still exist after their partner/contact is deleted, this rule was moved to instead rely on the recount count for the report.

Reporting.

Verify that reports will run when no contact exists (i.e., have been archived)

This change is made in response to [This](https://directemployersfoundation.atlassian.net/browse/PD-3884?focusedCommentId=320241&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-320241)
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


aljordan82:
I've tested this and js code looks good.
This can be merged if backend code is fine.

PR #3644, opened 2017-11-01
QC to Staging

mklauber:
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


fezick:
@koonsies needs to patch something on QC before this is merged. It should only take a moment.

koonsies:
I'm adding a DNM since I had one small PR to squeak in. https://github.com/DirectEmployers/MyJobs/pull/3645

PR #3646, opened 2017-11-02
Handle 403 status code

aljordan82:
## What's this PR do?
Show a custom message when a 403 is returned

### Front-End

- [x] Works on IE 11 and Edge
- [x] Works on Mobile

### Separation of Concerns

### Tests

### Migrations
N/A


gulp/src/analytics/components/Tabs/TabsContainer.jsx
```diff
@@ -51,6 +51,9 @@ class TabsContainer extends Component {
     const activeTab = navigation.find(nav => nav.active);
     let chartTitleByDisplay = '';
     let chartTitleDisplay = '';
+    if (activeTab.PageLoadData.forbidden ) {
+      return <div ref="contentContainer" className="alert alert-warning text-center">There are no charts available due to security configuration.</div>;
```
tdphillips: In general, we don't tell users stuff like this.

aljordan82: @darrint would this be visible to none staff?

darrint: It would take some work. Either doing URL wrangling or having your permissions or other scope changing out from under you. @tdphillips is it the word security that's the problem? Right now we're telling them that they have no data. That's incorrect. They don't have access to any data.

tdphillips: Until now, we've tried not to leak information of any kind. "You have no data" is a non-leaky way of saying "You don't have access to data."

If we think only staff will see this message, why not make that assumption clear in the code?

darrint: The API is returning 403 due to a permissions problem which we already do for many existing API calls. @fezick @JLMcLaugh ?

tdphillips:
That last comment was not a dig at anyone. You apparently have to provide a comment to dismiss your own review.

gulp/src/analytics/components/Tabs/TabsContainer.jsx
JLMcLaugh: If we consistently return a 403 for any unauthorized request, I think it's fine. There's no harm in telling them this is a chart page that they can't see. It becomes a problem if we respond differently to unauthorized requests for items that do or don't exist. e.g. we give a 404 if someone requests charts for a partner that doesn't exist and a 403 for an unauthorized request that would have data if it was authorized.

We could argue that a consistent 404 would help keep the endpoints a bit more secure by obscuring them a bit, but as long as we're open source I don't see a problem with 403.

PR #3651, opened 2017-11-06
Clean up feedback

darrint:
## What's this PR do?

Addresses some feedback on the pd-3876 feature branch and cleans up some remaining items.

* `TempDynamicReport` is gone.
* A few unit tests are fixed.
* Some docs and an interface are updated.

Walk through PRM/TJD reporting and Analytics apps.

na

### Separation of Concerns

### Tests


vigowei:
exporting in analytic causes an exception:
```AttributeError: 'ActivityScope' object has no attribute 'allowed_activities'```

tejaswinigutti31:
@vigowei We were previously passing activity scope to report_for_request but now it requires grant. I've made the change, so export should work fine now

PR #3652, opened 2017-11-06
Remove unused code in datasource filters

vigowei:
## What does this PR do?
This PR removes some previously commented out code.

We commented out code that handles `MatchFilter`s with a None value by converting them to `__isnull` filters,  because `field=None` and `field__isnull=True` are equivalent. 

A more important reason to remove this code block in these filter classes is that a field name like `xyz__isnull` is unrecognized by the filter class since only `xyz` is listed in the models filter_fields either manually or through model._meta.fields, so we ended up passing `xyz: true` back to the front end, when passing either `xyz__isnull: true` or `xyz: null` would be correct. 

### Front-End
N/A

### Separation of Concerns
N/A

### Tests
N/A

### Migrations
N/A


vigowei:
wait for 3651 to merge first.

vigowei:
reopening after merging with base branch.

PR #3653, opened 2017-11-07
[PD-3942] list recent comm records

vigowei:
## What's this PR do?
Adds a new API endpoint (/prm/api/samplecommrecs) to list recent communication records for a company.

mypartners/view.py L#2518

visit https://secure.my.jobs/prm/api/samplecommrecs and try it out there.

We are creating a new API instead of tweaking the existing `CommunicationRecordList` because the existing API does a lot of filtering and it's partner-centric. It would essentially require a rewrite if we incorporated this feature.

### Front-End
N/A

### Separation of Concerns

### Tests

### Migrations
N/A


tejaswinigutti31:
I've tested the API endpoint and the output looks correct.

mypartners/views.py
```diff
@@ -2515,6 +2515,40 @@ def api_get_workflow_states(request):
...
+        company = get_company_or_404(request)
+        try:
+            sample_size = int(request.data.get('limit'))
+        except (TypeError, ValueError):
+            sample_size = 5
+        contact_types = request.data.get('contact_types')
+
+        sample_records_qs = ContactRecord.objects.filter(partner__owner=company).order_by('-created_on')
+        if contact_types and isinstance(contact_types, list):
+            sample_records_qs = sample_records_qs.filter(contact_type__in=contact_types)
```
koonsies: We should probably return a "Bad Request" with an error message when the contact_types is provided, but in an incorrect format. Silent ignores can result in irritating debugging.

vigowei: thanks, updated.

PR #3655, opened 2017-11-07
[PD-3943] Model, APIs for Dashboard Tiles

koonsies:
## What's this PR do?
Adds the model for dashboard tiles, as well as the APIs to access them.

REST Client, note: you need to set the csrf token header `X-CSRFToken` for these to work
The payloads are as follows:

URL: `https://secure.my.jobs/dashboard/api/tiles/`

**Get list of all tiles for a user**
Method 
`GET`
Payload 
`None`

**Create a new tile for a user**
Method:
`POST`
Payload
```
    {
        "id": 1,
        "title": "Won",
        "params": "{}",
        "data_grid": "{}",
        "tile_type": "brokenTile"
    }
```

**Update tile(s)**
Method:
`PUT`
Payload:
```
[
    {
        "id": 1,
        "title": "Won",
        "params": "{}",
        "data_grid": "{}",
        "tile_type": "brokenTile"
    }
]
```

**Delete tile**
Method:
`DELETE`
Querystring:
`?tile_id=<id>`

Hit the various APIs. Ensure they are creating, updating, and retrieving as expected

There's some foo with getting bulk update to work. This is required by the front end, as moving tiles will cause the position of all tiles to be updated, and we didn't want to fire off an API call for each tile, every time.

In review, you may notice the multiple serializers. I spend a good amount of time trying to get the multi update to play nice, but found that previously accepted solutions to this issue were not working in our version of DRF.

If anyone knows of a better way for me to tackle the list update, please let me know.

## TODO

Add tests. I want to get this merged before writing tests so that Aladeen can begin integrating it. I have added a ticket for tests [here](https://directemployersfoundation.atlassian.net/browse/PD-3951) 

### Front-End
N/A

### Separation of Concerns

### Tests

~- [ ] Base classes used for python tests don't do too much unnecessary work.~ (ticket to add tests exists)

### Migrations


user_dashboard/api.py
```diff
@@ -0,0 +1,99 @@
...
+from user_dashboard.models import DashboardTile
+from universal.helpers import get_company_or_404
+
+
+class DashboardTileListSerializer(serializers.ListSerializer):
+    """
+    Special list serializer that allows for list (bulk) updates of dashboard tiles.
+
+    Requires that the dashboard id be set as an editable field and therefore must protect data integrity itself.
+    """
```
koonsies: Here's the weirdness. Handling of list updates in DRF seems to have changed many times over the course of the project's life. I have had trouble tracking down a solution that works. If someone who has a bit more Google foo can find a more up-to-date solution, please let me know.

user_dashboard/api.py
```diff
@@ -0,0 +1,99 @@
...
+        if serializer.is_valid():
+            serializer.save(user=request.user, company=company)
+            return Response(serializer.data,
+                            status=status.HTTP_201_CREATED)
+        return Response(serializer.errors,
+                        status=status.HTTP_400_BAD_REQUEST)
+
+    def put(self, request, format=None):
+        company = get_company_or_404(request)
+        tile_ids = [tile.get('id') for tile in request.data if tile.get('id')]
```
vigowei: ah, the payload for updates will have to be a list even if you are only updating one tile. I'm okay with that as long as we make it clear that's the expectation. I would add a check here since this throws an exception when you pass in a different type other than list.

user_dashboard/api.py
```diff
@@ -0,0 +1,99 @@
...
+        tile_instances = DashboardTile.objects.filter(id__in=tile_ids, user=request.user, company=company)
+        # Note the special serializer used here for bulk updates
+        serializer = DashboardMultiTileSerializer(instance=tile_instances, data=request.data, many=True)
+        if serializer.is_valid():
```
vigowei: looks like one bad tile in a bulk will cause entire bulk request to fail, which may be an okay thing to do. are we also concerned with duplicate tile ids?

user_dashboard/api.py
```diff
@@ -0,0 +1,99 @@
...
+            serializer.save()
+            return Response(serializer.data,
+                            status=status.HTTP_200_OK)
+        return Response(serializer.errors,
+                        status=status.HTTP_400_BAD_REQUEST)
+
+    def delete(self, request, format=None):
+        tile = DashboardTile.objects.filter(pk=request.GET.get('tile_id')).first()
```
vigowei: is there a reason for the primary key to not be unique? is this related to the side effect of allowing overwriting tile_ids from bulk updates?

user_dashboard/api.py
vigowei: but of course I'm fine with things like this being improved when you start writing tests.

user_dashboard/api.py
tdphillips: `primary_key=True` implies both `null=False` and `unique=True`. Jon is just doing a `.filter().first()` here to avoid having to catch a DoesNotExist.

vigowei: @tdphillips I see. I would have used get() + catch, but I'm fine with this.

user_dashboard/models.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+
+
+class DashboardTile(models.Model):
+    id = models.AutoField(primary_key=True)
+    user = models.ForeignKey('myjobs.User', editable=False, on_delete=models.CASCADE)
+    company = models.ForeignKey('seo.Company', editable=False, on_delete=models.CASCADE)
+    title = models.CharField(max_length=120, blank=True, null=True)
+    params = models.TextField(blank=True, null=True)
```
tdphillips: When I was on the team, I was campaigning against the easy path of storing json in a text field. I'm tired of having to argue that frontend devs should not dictate anything connected to the database so consider this an approval under duress.

user_dashboard/api.py
vigowei: just read the [DRF doc](http://www.django-rest-framework.org/api-guide/serializers/#customizing-multiple-update), so I understand your solution better.. I think you are doing everything the recommended way except for having to declare `id` editable. Their example does not have read_only=True but weirdly it doesn't work for us if you leave it out. If we acknowledge and document clearly that this is just a hack to make `id` available in the list serializer then it's okay? I would recommend changing L#19 to:         
`data_mapping = {data.pop('id', None): data for data in validated_data}`
to explicitly state we are not overwriting the `id` field.

koonsies: That makes sense. That would serve as a double-guard against overwriting IDs, but since I am matching the ID to an existing instance tied to that user (the IDs being the same being the only way they would match), the presence of ID isn't a major concern. However, I can see how verifying ID isn't present could prevent an edit on the code from easily introducing a bug.

user_dashboard/models.py
koonsies: Hmm. I understand your issue. Perhaps it's a lack of cleverness of my part, but a better solution didn't present itself to me. What params are supplied are based on what the various APIs/components will require in order to do their rendering. If I tried to define fields, I figured I'd have to have a model for each component. While this is possible, I liked that solution even less than storing JSON.

aljordan82:
@koonsies I am getting a js error when using the `DELETE` because it returns an empty response. can it return at least a `{}` ?

PR #3657, opened 2017-11-08
[PD-3900] Add a debounce for TJD

vigowei:
## What's this PR do?
Adds a debounce to run_trial_report, so it won't be triggered every user edit 

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns
n/a

### Tests
n/a

### Migrations
n/a


vigowei:
closing to fix tests

gulp/src/reporting/components/SetUpReport.jsx
```diff
@@ -111,10 +114,10 @@ class SetUpReport extends Component {
     this.dispatchUpdateFilterWithDependencies();
   }
 
-  dispatchUpdateFilterWithDependencies() {
+  dispatchUpdateFilterWithDependencies = debounce(() => {
     const {dispatch, filterInterface, reportDataId} = this.props;
     return dispatch(doUpdateFilterWithDependencies(filterInterface, reportDataId));
-  }
+  }, 500);
```
aljordan82: I've tested and it works.
There is a `typingDebounce` function that we use, but it's hardcoded to 300ms. We can you that if you are fine with 300ms.

gulp/src/reporting/components/SetUpReport.jsx
```diff
@@ -402,7 +412,7 @@ class SetUpReport extends Component {
                 <TextField
                   name={currentFilter[col.filter] || ''}
                   value={currentFilter[col.filter] || ''}
-                  onChange={v => this.dispatchFilterAction(setSimpleFilterAction(col.filter, v.target.value))}
+                  onChange={v => this.dispatchFilterAction(setSimpleFilterAction(col.filter, v.target.value), 'debounced')}
```
aljordan82: I would use true and you can add a comment like .. , true /* debounce */)

vigowei: @aljordan82 either way would be readable right? if we allow anything truthy in the function, how does this make a difference?

aljordan82: True, there is not difference, it is just not expected to compare against a string.

koonsies:
I'm getting some weird results with the debounce. None of the fields are causing a trial report to be run instantly. When I change a multi-select box, a trial report isn't fired until after a looonngg delay. Debounce seems to work for the text field, but it's also way too long.

PR #3658, opened 2017-11-08
Pd 3568 add dashboard sub nav

aljordan82:
## What's this PR do?
Add Navigation ribbon
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


koonsies:
Clicking the text on the PRM button does the same thing as clicking on the bottom. Both go to "Add Record"

koonsies:
The finger hover effect occurs whenever the user hovers over one of the buttons, but the click event only happens on the text. I'm thinking the entire area should be clickable.

aljordan82:
@koonsies PRM link is now fixed.

PR #3659, opened 2017-11-09
Stop running a query for all business units on thread startup.

almathew:
## What's this PR do?
Stops running a query for all business units on thread startup.

Same as https://github.com/DirectEmployers/MyJobs/pull/3449

While messing around with Datadog we came across this really strange 5+ second query for all business units with no where clause. It looks like this is the cause.


seo/api/v2/serializers.py
```diff
@@ -23,10 +23,14 @@ def update(self, instance, validated_data):
 
 
 class QueryPreviewSerializer(serializers.Serializer):
-    business_units = serializers.MultipleChoiceField(allow_blank=True, choices=lazy(business_unit_choices, list)())
```
tdphillips: Would this work if you left off the last set of parens? An initial reading leads me to believe that would force the execution of the thing that was intended to be lazy.

almathew: You get this error when you try it that way:
```  File "/MyJobs/dseo_urls.py", line 16, in <module>
    from seo.api.v2 import urls as v2_urls
  File "/MyJobs/seo/api/v2/urls.py", line 3, in <module>
    from seo.api.v2.views import QueryPreviewView
  File "/MyJobs/seo/api/v2/views.py", line 4, in <module>
    from seo.api.v2 import serializers
  File "/MyJobs/seo/api/v2/serializers.py", line 25, in <module>
    class QueryPreviewSerializer(serializers.Serializer):
  File "/MyJobs/seo/api/v2/serializers.py", line 26, in QueryPreviewSerializer
    business_units = serializers.MultipleChoiceField(allow_blank=True, choices=lazy(business_unit_choices, list))
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/fields.py", line 1376, in __init__
    super(MultipleChoiceField, self).__init__(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/fields.py", line 1325, in __init__
    self.grouped_choices = to_choices_dict(choices)
  File "/usr/local/lib/python2.7/dist-packages/rest_framework/fields.py", line 153, in to_choices_dict
    for choice in choices:
TypeError: 'function' object is not iterable```

I think the implementation of `lazy` there is correct, I just think it's being used/evaluated earlier than I wanted it to be anyway.

almathew: Although I will admit lack of knowledge here too - `lazy` is not the best documented function.

darrint: Wouldn't the fix be: `for choice in choices():` ?

darrint: Ok. Disregard what I said earlier. Based on reading here:

https://stackoverflow.com/questions/28357646/how-to-understand-lazy-function-in-django-utils-functional-module

The problem is that you call the function right away. You get a proxy list, but the initializer for the serializer field immediately starts using the list so the benefit of lazy is lost.

You may be better off using a less interesting field and writing a validator method on the serializer.

PR #3662, opened 2017-11-13
Remove time from dates

aljordan82:
## What's this PR do?
Remove time from dates
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

- [ ] ~~Back end "business logic" is implemented in plain python, Django model classes, and Django forms. (not Django views or other frameworks)~~

### Tests

- [ ] ~~Base classes used for python tests don't do too much unnecessary work.~~

### Migrations

removed


darrint:
Deferring mobile/IE testing to feature branch.

PR #3663, opened 2017-11-13
Fix timelines the hard way

darrint:
## What's this PR do?

The visible bug was: Analytics Timeline charts had an extra day at the end. This appeared to be because the chart spanned a DST transition.

The only robust solution I could find was to stop considering times at all during these queries and to only query with the resolution of a day.

I also added a python function to fill in missing days.

In the future if we want to do hourly or to-the-minute queries we will need a separate API for that. Based on the work I did here I think this is reasonable. The weird time considerations for an hourly query vs. a query based on calendar dates are different.

Poke around the analytics UI.

The best solution I could come up with involved calling my congressperson. I decided to do this instead.

### Front-End

### Separation of Concerns

### Tests


darrint:
@tejaswinigutti31 Noticed this problem: https://directemployersfoundation.atlassian.net/browse/PD-3955 But since it is present on QC we're going to just create the ticket.

myreports/datasources/analytics_event.py
```diff
@@ -1,14 +1,17 @@
...
 )
 from myreports.datasources.base import (
     QueryResult, ResultPage, ResultRow, ResultCell, ResultColumn, DataSourceClientError)
-from myreports.datasources.util import OrGroupFilter, MatchFilter, DateRangeFilter, apply_filter_to_queryset
+from myreports.datasources.util import (
+    OrGroupFilter, MatchFilter, DateRangeFilter, apply_filter_to_queryset, fill_time_series)
+from myreports.column_formats import UsDate
 from universal.helpers import dict_identity
-from BPTrunc import TruncDay  # this API is available in Django 1.10, remove this when upgrade
+from BPTrunc import TruncDate  # this API is available in Django 1.10, remove this when upgrade
```
vigowei: Can you please help me understand the difference between this and TruncDay, and when we would prefer one over the other?

darrint: @vigowei TruncDay returns a timestamp at midnight of the day which it turns out might not be what you thought you were getting when you have a series of days including a DST transition. TruncDate casts the value to a simple date with no time information.

How the bucketing works in the database isn't 100% clear to me but at least the damage is limited to the totals. It doesn't cause additional strange behaviors like hour long days in the fall or who knows what in the spring.

gulp/src/analytics/reducers/initial-state.js
```diff
@@ -2,11 +2,11 @@ import moment from 'moment';
 
 const startDate = moment()
   .subtract(30, 'days')
-  .format('YYYY-MM-DD 00:00:00');
+  .format('YYYY-MM-DD');
 
 const endDate = moment()
   .subtract(1, 'day')
-  .format('YYYY-MM-DD 23:59:59');
+  .format('YYYY-MM-DD');
```
vigowei: not sure if we still want to subtract 1 day after getting rid of the time part?

darrint: Turns out no. The `DateRangeFilter` in reporting takes care of making date ranges inclusive when querying.

PR #3664, opened 2017-11-14
Datadog

almathew:
darrint: We're going to go ahead and merge this as the review looks good.

### Separation of Concerns


settings.py
```diff
@@ -203,3 +203,9 @@
     from local_settings import *
 except ImportError:
     pass
+
+DATADOG_TRACE = {
```
darrint: Any ideas for distinguishing between developers in `local_settings.py` ?

tasks.py
```diff
@@ -153,6 +152,7 @@ def assign_ticket_to_request(key, access_request):
 @task(name='tasks.send_search_digest', ignore_result=True,
       default_retry_delay=180, max_retries=2, bind=True,
       time_limit=30)
+@tracer.wrap()
```
darrint: Do we want to set a service name on these?

universal/helpers.py
```diff
@@ -598,3 +599,12 @@ def jobs_from_redis(guids):
     redis = StrictRedis(host=settings.REDIS_HOST)
     jobs = redis.mget(guids)
     return [json.loads(job, object_hook=json_util.object_hook) for job in jobs if job]
+
+
+def log_dict(**kwargs):
```
darrint: Would it be overkill to re-export tracer.wrap or put an adapter here?

settings.py
almathew: I just added one. It's very clever.
My original plan was to give instructions to override it, since it's going to require some slight secrets-type customization anyway.

darrint: I think that's just going to be `root` in Docker, no?

almathew: Oh, you're right. I *AM* too clever. 
Guess I'll do the secrets-y way, and just provide some setup instructions. I think I'm going to have to do a team walk-through sometime in the next few days anyway.

darrint: Putting `local_trace_user = 'darrint'` in `local_settings.py` might be ok. Having the trace inactive isn't out of the question. Or set it to `dev-nobody`.

darrint:
I've removed the DNM tag. It was just there to try and get traces of unit tests. That isn't worth holding this up over.

PR #3665, opened 2017-11-14
Pd 3111 systems notifications not displayed analytics

strempe:
## What's this PR do?
Fix notifications in analytics 

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


dseo.log.2017-11-07
```diff
@@ -0,0 +1,35 @@
+ERROR 2017-11-07 12:43:11,554 base 1902 140541723404032 Internal Server Error: /prm/api/nonuseroutreach/inbox/list
```
tdphillips: We don't want these logs to be committed.

myjobs/views.py
```diff
@@ -1225,6 +1225,7 @@ def process_access_request(request, access_id, accepted):
     The target of a second party access request will hit this view to
     approve or reject a given request.
     """
+    print "I'm here"
```
tdphillips: print

mymessages/context_processors.py
```diff
@@ -2,11 +2,12 @@ def message_lists(request):
     """
     Ensures lists of messages, if any, are always in template contexts.
     """
+    # import ipdb; ipdb.set_trace();
```
tdphillips: commented debugger

settings.py
```diff
@@ -147,6 +147,10 @@
     'django.template.loaders.filesystem.Loader',
     'django.template.loaders.app_directories.Loader',
     'django.template.loaders.eggs.Loader',
+    
+)
+TEMPLATES[0]['OPTIONS']['context_processors'] += (
```
tdphillips: Using this context processor was causing some issues when setting up a new database.

almathew: I'm kind of confused. Is it intentional that this context processor is only available for local development, then?

tdphillips: No. It's added in the settings files in the deploy directory.

It may not be an issue now that we're on a Django version higher than 1.6. Back then, this context processor would, in development, execute before the database was set up. Don't ask me how. The end result was the inability to run migrations.

PR #3669, opened 2017-11-15
Allow for using Datadog locally (web requests only).

almathew:
## What's this PR do?
Allows you to push your local results to Datadog. This can be used for tracking things, or for setting up and configuring tags or traces.

Run runserver. Look for your app.

This is against staging (because everything else was too). I'll backport as soon as all the Datadog changes are complete.


manage.py
```diff
@@ -1,10 +1,7 @@
 #!/usr/bin/env python
-from ddtrace import patch_all, tracer
 import os
 import sys
 
-patch_all()
```
almathew: Turns out we need to let Django handle all this.

settings.py
```diff
@@ -204,13 +204,12 @@
...
-except Exception:
-    current_user = 'dev'
-
+MIDDLEWARE_CLASSES = list(MIDDLEWARE_CLASSES)
+MIDDLEWARE_CLASSES.insert(1, 'middleware.ConfigureTracerForDevelopment')
+MIDDLEWARE_CLASSES = tuple(MIDDLEWARE_CLASSES)
 DATADOG_TRACE = {
     'DEFAULT_SERVICE': 'development',
-    'TAGS': {'env': 'dev-%s' % current_user},
+    'TAGS': {'env': 'dev'},
```
almathew: If anyone is actually doing testing locally, they should probably override the env here so we are not interfering with one another.

settings.py
```diff
@@ -204,13 +204,12 @@
...
     'ENABLED': False,
```
almathew: Anyone who wants to actually use this will need to set enabled to `True`. That, and being on the VPN, should be the only barrier to entry.

almathew:
Mental note to myself: after this is merged I will likely go ahead and do the first staging -> QC merge.

PR #3670, opened 2017-11-16
[PD-3970] fix frontend paging error

vigowei:
## What's this PR do?
Updates a function that handles pagination size change. We are already getting rows correctly out of query_result everywhere else, this function got missed somehow.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


darrint:
Deferring proper front end testing until we review the real feature branch.

PR #3672, opened 2017-11-16
Implementing tiles api

aljordan82:
## What's this PR do?
Add a Layout grid library and implement a basic Tile.
run `dky` and restart the devserver
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


gulp/src/dashboard/actions/mainActions.js
```diff
@@ -1,12 +1,57 @@
...
 
+export const setUserTiles = createAction('SET_USER_TILES');
+
+export function loadUserTiles() {
+  return async (dispatch, _, { api }) => {
+    try {
+      dispatch(markPageLoadingAction(true));
+      const tiles = await api.getUserTiles();
+      await dispatch(setUserTiles(tiles));
+    } catch (e) {
```
tdphillips: Granted this is my Python experience talking, but what is this giving us that simply not catching anything doesn't?

aljordan82: Nothing really.

aljordan82:
DNM, I found a bug while writing unit tests

gulp/src/common/CommonApi.js
```diff
@@ -1,7 +1,12 @@
...
 
 export default class CommonApi extends MyJobsApi {
   async getUserActivities() {
     return await this.get('/useractivities');
   }
+  async getSampleCommRecs({ limit, contact_types }) {
+    const sendUrl = '/prm/api/samplecommrecs';
+    const returnValue = await this.postJson(sendUrl, { limit, contact_types });
+    return returnValue;
+  }
```
koonsies: I'm not sure this should be in CommonApi. Do you envision other apps using this particular call?

gulp/src/dashboard/actions/mainActions.js
koonsies: I suppose it gets us the "finally".

gulp/src/dashboard/api.js
```diff
@@ -0,0 +1,54 @@
...
+  const formatedTiles = rawTiles.map(tile => {
+    try {
+      const newTile = {
+        ...tile,
+        dataGrid: tile.data_grid ? JSON.parse(tile.data_grid) : {},
+        tileType: tile.tile_type,
+        params: tile.params ? JSON.parse(tile.params) : {},
+      };
+      delete newTile.data_grid;
+      delete newTile.tile_type;
```
koonsies: I'm not following. Why do we parse/declare these items, only to delete them out of the dictionary?

gulp/src/common/CommonApi.js
aljordan82: I thought this would be used in other apps, I can put it in dashboard/api.js for now.

PR #3673, opened 2017-11-20
Pd 3637 ui clear spinner and show error

aljordan82:
## What's this PR do?
Create a wrapper function for the Spinner. 
Any action that should show a Spinner should be wrapped with `wrapSpinnerAndErrors` or `wrapSpinner`.
This is just for Reporting, other react apps will need this change before merging this PR.
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


gulp/src/reporting/actions/compound-actions.js
```diff
@@ -151,10 +150,9 @@ export function doReportDataRedirect(history, intention, category, dataSet) {
       dispatch(markOtherLoadingAction('dataSetMenu', false));
     } catch (e) {
       dispatch(markOtherLoadingAction('dataSetMenu', false));
-      dispatch(markPageLoadingAction(false));
-      dispatch(errorAction(e.message));
+      throw e;
```
aljordan82: If the wrapped action creator has a try/catch , then we must throw so the wrapper handles the error.

koonsies:
I like this idea. While we're refactoring the handling of spinner, etc, perhaps we could revisit how we "mark" a page as loading. One method I've seen was in CBO, and that was to store a counter of functions being waited on. So, if you dispatched three functions that were wrapped in the spinner wrapper, each would increment a global function counter by 1. When a function completed (or errored out), it would decrement the counter until the counter reached 0 and the spinner was removed.

We may have discussed this before, but I can't recall. Just wanted to bounce that idea around again.

vigowei:
maybe this is also an opportunity to agree on the format for error messages? Sometimes the error string is at the top level of the response object, so we can get it via response.data, sometimes, it's in an object, like `{'message': 'reason for failure...'}`, now we have to pass response.data.message to toastr since it doesn't display objects. 

we could stringify/format whatever is in response.data, or designate a field for error message, which means making sure all error responses from the backend follow the same convention.

koonsies:
Closing this until it's ready to review again.

PR #3674, opened 2017-11-20
Datadog for celery.

almathew:
## What's this PR do?
Sets up Datadog for celery. 
Removes a bunch of stuff that is irrelevant if Datadog becomes a thing, and needs replaced anyway if it doesn't.
I also have a better understanding of how traces and the AGENT_HOSTNAME setting works, so remove the middleware and use that instead.
Because of the above point, I'm going to take another stab at getting Jenkins working with APM.

Major changes are in imports. 
There should be no actual change to how the code functions.

A very, very limited version of the imports changes are already in production, so check out Datadog for that.
For the local changes, in settings.py set `DATADOG_TRACE['ENABLED'] = True`. Please also change the env tag to something specific to you (I've been using my name).

@darrint asked me to do some voodoo with the decorators, so we can hide a lot of the logic and make it less Datadog-y (like I tried to do with the `log_dict` function). 
This is messy and impractical at the moment, especially when, if we don't go with Datadog, the answer is going to be to pull it all out.

### Separation of Concerns


vigowei:
don't see a `tracer.configure()` call anymore, does it get run automagically?

almathew:
It does. The middleware they package handles a lot of it it already. I just messed up the first time I tried to set it up.

import_jobs/tasks/general_tasks.py
```diff
@@ -373,52 +261,14 @@ def handle_imports_without_buid(add_jobs, update_jobs, del_guids,
...
-    source_update_chunks = chunk(update_jobs)
-    for source_update_chunk in source_update_chunks:
-        source.update_jobs(source_update_chunk)
-    logger.info(json.dumps({
-        'buid': None,
-        'action': 'updated source of truth',
-        'num_jobs': len(update_guids),
-        'duration': time.time() - start
-    }))
+    add_to_source_of_truth(add_jobs)
```
vigowei: can you please double check this one?

deploy/settings.myjobs_staging.py
```diff
@@ -126,8 +126,10 @@
...
 }
 
+
+DD_ENV = 'staging'
 DATADOG_TRACE = {
     'DEFAULT_SERVICE': 'myjobs',
-    'TAGS': {'env': 'staging'},
+    'TAGS': {'env': DD_ENV},
     'ENABLED': True,
 }
```
vigowei: maybe I'm missing something but shouldn't `AGENT_HOSTNAME` be included as well?

almathew: No, the AGENT_HOSTNAME is only required in places where we won't be running the agent ourselves.
I can't be run inside the docker container, so nothing in dev. Same with Lambda and on the Jenkins slaves.

mklauber: `AGENT_HOSTNAME` only need to be set on the jenkins settings because we're not going to install the DataDog agent on every jenkins slave as it would mess with our server count.

mklauber: It defaults to localhost, which should work for everything else.

vigowei: makes sense, thanks guys.

PR #3676, opened 2017-11-20
[PD-3883]  calendar updates should not clear search

vigowei:
## What's this PR do?
Prevent search term from being cleared when the calendar is updated. 

compare search box behavior on QC and this branch, notice how it's affected by selecting new dates from the calendar.

Added an action to set the search term ('partial') for the currently active tab. This search value is then applied on the backend when the user updates the calendar.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


aljordan82:
Looks good, but there is minor issue. I think we need to bind the search field value to the stored `partial` in redux, so when the tab gets changed, the search value gets updated.

gulp/src/analytics/actions/calendar-actions.js
```diff
@@ -26,6 +26,8 @@ export function doSetSelectedRange(startDate, endDate) {
       const dimension = activeNavigation.PageLoadData.dimension;
       // Grab multi filter from the state navigation
       const rawMultiFilter = getState().main.navigation[0].PageLoadData.multifilter;
+      // Check if the search box has value
+      const partial = getState().main.navigation[0].partial || '';
```
aljordan82: `partial` default value can be added when the `navigation` item is created in reducers.

vigowei: I think that's `SET_PAGE_DATA`, I've added a default value for it and removed the OR.

vigowei:
@aljordan82 I agree we should be binding partial to tab, right now it's sort of a global state.. I have another branch for this exact issue, will open a new PR once this one is merged.

gulp/src/analytics/actions/calendar-actions.js
vigowei: reverted, will address this in the next PR

PR #3678, opened 2017-11-21
[ST-4456] TJD Reporting Fixes

koonsies:
## What's this PR do?
Fixes TJD reporting, which had the following issues:
1) Reports did not apply filters
2) Some filters were confusing in their display (saying "tags" instead of categories, etc)
3) State help results contained duplicates
4) CA's API requires that if location is sent, that city and state each have a value and not be omitted
5) When a select box is changed from "No Filter" to "Filter by..", a trail report would be run. This caused extra round trips to the server with no meaningful difference in filters.

TJD Reporting. Verify that all of the filters work properly.
I also removed a step in the search multiselect field. Make sure this still works in other reports (example: tags, category)

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests


myreports/datasources/tjd_report.py
```diff
@@ -249,9 +253,14 @@ def filtered_query(self, company, filter_spec):
...
+            formatted_filter = build_tjd_query(f)
+            filter_data[k] = formatted_filter
         del filter_data['SentDate']
-        response = post_required_data(settings.TJD_ROOT_URL + 'report/report_a',
+
+        if isinstance(filter_data['Location'], dict):
+            filter_data['Location']['city'] = filter_data['Location'].get('city') or ''
+            filter_data['Location']['state'] = filter_data['Location'].get('state') or ''
+
+        response = post_required_data(settings.TJD_ROOT_URL + 'report/report_b',
```
tejaswinigutti31: If for any reason the response is 404, we will end up in an error `ValueError: No JSON object could be decoded`. This case needs to be handled

koonsies: I agree. Since this issue wasn't introduced by this PR, I'd say that it isn't really a reason to stop this PR from being merged. On that note, I don't see anywhere that we gracefully handle runtime exceptions with reporting. We always return a 200, or, in the event of an error, a 500. So, we either return a default value when there's a non-JSON response (which hides the error more than handles it), or we create a workflow to return proper status codes from the run report fucntions. I'll create a ticket to handle different status codes coming out of reporting.

tejaswinigutti31: A ticket for that sounds good.

vigowei:
image
one minor thing, State is showing up as a number in exported excel, maybe it's always behaved that way?

vigowei:
filters allow you to see values that I'm not sure if the user should have access to, for example, filtering by contact shows a long list of names that may or may not belong to the same organization

tejaswinigutti31:
@vigowei a ticket has been created for the helper texts https://directemployersfoundation.atlassian.net/browse/PD-3979

vigowei:
@tejaswinigutti31 thanks, I just read the ticket and I understand the issue better now, it's more of a filtering issues than access issue. My previous comment can be ignored.

myreports/datasources/tjd_report.py
```diff
@@ -1,4 +1,5 @@
 import json
+from collections import OrderedDict
```
vigowei: not seeing this import used

koonsies: Fixed.

PR #3680, opened 2017-11-21
[PD-3940] fix search behavior in analytics

vigowei:
## What's this PR do?

Pull `searchValue`out of the AnalyticsApp state and manage it in TableSearch.

This fixes a couple of problems:

Any change in the search box will not trigger a chain of updates to propagate from the Analytics App down to table (the tabs container and table container used to re-render on every keystroke).

Search term is bound to each tab. Setting the value for search box will no longer set it for other tabs. The search term used to stay after drilling down in a dimension, and vice versa. This causes confusion since previous search term is nowhere to be found on the new tab.

Use search in analytics reports and verify the problems no longer exist.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile


aljordan82:
Looks good!
The only issue I am seeing is when changing the global date and then click on a hidden tab with a search term. but this is not a major issue and I think it's better just to clear the search field.

koonsies:
Once entered, how does one clear a search term?

vigowei:
@koonsies are you talking about unapplying a search? they simply need to clear whatever is in the search box and hit return. The only indicator that a search is applied is something is in the search box.

koonsies:
Strange. I tried that and had no results. I filtered to a single result by entering a search term, deleted the term and pressed enter, but the results in the table remained limited to 1.

koonsies:
I had an old version of the branch. Pulling new corrected the issue.

PR #3683, opened 2017-11-23
Pd 3637 ui clear spinner and show error

aljordan82:
## What's this PR do?
Create wrappers for Spinner and Error handlers for actionCreators
Fix a bug in NUO. The wrong spinner was used instead of a form field spinner.
gulp/src/common/actions/loading-actions.js

Loading spinner should work and error should be caught.
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


aljordan82:
DNM until dashboard feature branch is merged.

PR #3688, opened 2017-11-27
[PD-3327] suggested roles

tejaswinigutti31:
## What's this PR do?
Adds Suggested Roles to user management
screen shot 2017-11-27 at 4 24 38 pm

FrontEnd: A new file SuggestedRoles.jsx has been added
BackEnd: 2 new functions for suggested roles have been added in myjobs/views.py
Under Manage Users, check for a new tab Suggested Roles
A Suggested Role can be added in admin.
Click through buttons "Save Role"/"Reject"
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


koonsies:
Merge conflicts.

koonsies:
The suggested role admin view should have a few more items listed. Right now, it just shows the name of the "Suggested Role", not the company it's affiliated with, etc.
screen shot 2017-11-28 at 2 30 19 pm

koonsies:
Do we know if they want suggested roles to be deleted if they reject them? The wording is a bit off on the reject confirm. It currently reads "Are you sure you want to Reject?" .. Perhaps "Are you sure you want to reject this role?"

myjobs/views.py
```diff
@@ -1237,3 +1236,78 @@ def api_get_user_activities(request):
...
+    for role in roles:
+        sr_id = role.id
+        sr_name = role.name
+        sr_activities = []
+        activities = role.suggested_activities.values()
+        for activity in activities:
+            sr_activity = {}
+            sr_activity['name']=activity['name']
+            sr_activity['description']=activity['description']
+            # sr_activity['appAccess']=activity['app_access']
```
koonsies: Commented code

myjobs/views.py
```diff
@@ -1237,3 +1236,78 @@ def api_get_user_activities(request):
...
+    """
+    ctx = {}
+    if request.method != "POST":
+        ctx["success"] = "false"
+        ctx["message"] = "POST method required."
+        return HttpResponse(json.dumps(ctx),
+                            content_type="application/json")
+    else:
+        company = get_company_or_404(request)
+        role = SuggestedRole.objects.get(id=role_id)
```
vigowei: this may fail. Maybe use try - catch to handle unmatched role_ids

myjobs/views.py
```diff
@@ -1237,3 +1236,78 @@ def api_get_user_activities(request):
...
+            activities = Activity.objects.filter(name__in=[activity['name'] for activity in suggested_activities])
+
+            # Create the new role
+            new_role, _ = Role.objects.get_or_create(name=role.name, company_id=company.id)
+
+            # Update the above created role with activities
+            if activities:
+                new_role.activities.add(*activities)
+            ctx["success"] = "true"
+        SuggestedRoleStatus.objects.get_or_create(company=company, name=role.name, role_accepted=role_accepted)
```
vigowei: This may create duplicate entries for the same role, correct me if I'm wrong but you want to only update the role_accepted field here?

tejaswinigutti31:
@koonsies I made change to admin such that role and company will be displayed now. Thank you !!

myjobs/views.py
```diff
@@ -1237,3 +1236,78 @@ def api_get_user_activities(request):
...
+            role_id=int(sr_id),
+            role_name=sr_name,
+            activities=sr_activities,
+        )
+        sr_formatted.append(roles_data)
+
+    return HttpResponse(json.dumps(sr_formatted), content_type='application/json')
+
+@requires('create role')
+def api_create_suggested_roles(request, role_id=0):
```
koonsies: We switched recently to making all new Python APIs use DjangoRestFramework. It shouldn't change too much, but we should be consistent.

myjobs/views.py
```diff
@@ -1237,3 +1236,78 @@ def api_get_user_activities(request):
...
+    Based on the id of Suggested Roles, make an entry into Role table
+    :param request:
+    :param role_id:
+    :return:
+    """
+    ctx = {}
+    if request.method != "POST":
+        ctx["success"] = "false"
+        ctx["message"] = "POST method required."
+        return HttpResponse(json.dumps(ctx),
```
koonsies: This stuff will be handled by DjangoRestFramework, but an issue with this return would be that the status could be "200 (Success)", but there would have been an error.

myjobs/views.py
koonsies: Do we have a way of viewing these SuggestedRoleStatuses? What does the presence of this model offer us that isn't offered by simply having a few additional columns on SuggestedRole, i.e.,

Created: Boolean Field
DateAcknowledged: DateField

PR #3690, opened 2017-11-28
changed width to 100%

strempe:
## What's this PR do?
Removes topbar constraints. 

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


aljordan82:
This change will affect all React apps. I thought it is only for Dashboard and Analytics.

strempe:
@aljordan82 Yes, if it changes when you click on other applications we will lose consistency. The purpose of making the topbar full width is to provide more continuity for the different apps in my.jobs. :)

PR #3699, opened 2017-11-30
[PD-3992] Update Job analytics report with new table columns

tejaswinigutti31:
## What's this PR do?
Adds new column in Job Owner Syndication Report
screen shot 2017-11-30 at 4 44 45 pm

Check new queries in Job Views datasource
In Job Owner Syndication look for newly added table columns and check the values for correctness
### Separation of Concerns

### Tests


analytics/api.py
```diff
@@ -273,7 +273,7 @@ def format_meta_data(query_result):
         records = []
         for row in query_result['page']['rows']:
             cells = row['cells']
-            cell_list = {k: v['display'] for k, v in cells.iteritems()}
+            cell_list = {k: (v['display'] if k != 'drop_off_rate' else v['value']) for k, v in cells.iteritems()}
```
darrint: Why is this one special?

myreports/datasources/job_views.py
```diff
@@ -122,9 +122,18 @@ def annotate_query_set(query_set, values=None):
             ))
         if 'reached_job_source' in values:
             query_set = query_set.annotate(
-                reached_job_source=Sum('reached_source')
+                reached_job_source=Sum(Case(When(reached_source=True, then=1), default=0, output_field=IntegerField())
+            ))
+        if 'reached_job_source_rate' in values:
```
darrint: This looks redundant with `reached_job_source` above.

myreports/datasources/job_views.py
```diff
@@ -261,13 +269,24 @@ def extract_record_with_meta(self, record, values, filter_spec, cell_fields):
...
+                    else:
+                        display = Percentage().format(value)
                 else:
                     display = unicode(value)
                 cell_dict['display'] = display
+
+            trace_string = ''.join('{}: {} '.format(key, val) for key, val in record.items() if key not in
+                                   ['reached_job_source_rate', 'rm_platform_apply_rate', 'referral_apply_rate'])
+            if key == 'drop_off_rate' and value < 0:
+                span.set_tag('drop_off_rate', trace_string)
```
darrint: @almathew How should we be doing this? We want to be able to track these so we can look for data problems later.

myreports/schema_dimensions.py
```diff
@@ -21,17 +21,22 @@
...
     DataColumn(
         column_name=u'reached_job_source',
         alias=u'Reached Job Source',
         description=u'The percentage of job seekers who, after finding and possibly viewing your job elsewhere on the '
                     u'internet, arrived at the source of your job content, usually your ATS.',
     ),
+    DataColumn(
+        column_name=u'reached_job_source_rate',
+        alias=u'Reached Job Source Rate',
+        description=u'TBD',
```
darrint: The TBD should be in the `reached_job_source` column above. This description should be the one used in `reached_job_source`.

myreports/tests/test_job_views_datasource.py
```diff
@@ -456,6 +468,23 @@ def test_run_timeline_with_meta(self):
         self.assertEqual(0, query_result.page.rows[1].cells['job_views'].value)
         self.assertEqual(12, query_result.page.rows[2].cells['job_views'].value)
 
+    def test_run_aggragated_filter_drop_off_rate(self):
```
darrint: We need at least one test where drop off is a real number.

darrint: Ideally there a test where dropoff is negative as well, so we at least run the code for the tracking. I think that code would throw an exception right now.

myreports/datasources/job_views.py
darrint: After talking with Ashley we decided to handle it this way.

1. At the top of the file `import logging`
2. After imports `data_events_logger = logging.getLogger('analytics.data-events')`
3. Here: `data_events_logger.error("IA_RS inconsistent. Company: %d", company.name)`

tdphillips: Assuming that string uses standard formatting options, `%d` is for ints.

myreports/tests/test_job_views_datasource.py
tejaswinigutti31: As discussed, this wouldn't be a case that would arise

myreports/datasources/job_views.py
```diff
@@ -261,13 +266,23 @@ def extract_record_with_meta(self, record, values, filter_spec, cell_fields):
...
             if 'display' in cell_fields:
-                if key in ['intended_to_apply', 'reached_job_source', 'rm_platform_apply_rate', 'referral_apply_rate']:
+                if key in ['reached_job_source_rate',
+                           'rm_platform_apply_rate', 'referral_apply_rate']:
                     display = Percentage().format(value)
                 elif key == 'timeline_ts':
                     display = UsDate().format(value)
+                elif key == 'drop_off_rate':
+                    if value < 0.02 and destination == INTERACTIVE:
+                        display = '< 2%'
```
darrint: We should clamp the value here has well.

PR #3704, opened 2017-12-04
[PD-4003] Remove is_staff authorization requirement on PartnerLookup.

almathew:
## What's this PR do?
Removes is_staff requirement in PartnerLookup and instead allows get_company_or_404 to handle the authorization check.

Using a non-staff user, attempt to create a record using the "Add Record" quick button thing on the topbar. 
With the current behavior, when a non-staff user enters a letter the API will return a 403 and the user will get redirected back to the login page.
With the new behavior the API works as intended.

### Tests


darrint:
Looks good.

PR #3705, opened 2017-12-04
[PD-3994] Send email notification

tejaswinigutti31:
## What's this PR do?
A function to send email notification
Check the function send_email() in notifications/rules.py
### Separation of Concerns


notifications/rules.py
```diff
@@ -99,15 +99,19 @@ def send_email(self, template_context, html_template, plain_template):
...
-        email_domain = 'my.jobs'  # Need to figure out what else would that be(from email)
-        recipients = self.task.freeform_email
+        email_domain = 'notifications@my.jobs'  # Need to figure out what else would that be(from email)
+        recipients = self.task.freeform_email.split(',')
         if template is not None:
-            body = template_context.body
-            send_mail(subject, body, email_domain, recipients)
+            email_body = template.body
+            if html_template is not None:
+                html_message = html_template.body
```
koonsies: Few issues here. html_template, as passed into the function, is a Django template object. When you call html_template.render, it returns a string, or rendered template that was generated using the context. That string needs to be passed in as the html_message.

So..

```rendered_html_template=html_template.render(template_context)
html_message = rendered_html_template```

File: notifications/rules.py [hunk 2]
```diff
@@ -99,15 +99,19 @@ def send_email(self, template_context, html_template, plain_template):
 
         """
         print '===PRINTING HTML EMAIL==='
-        print html_template.render(template_context)
+        html_template = html_template.render(template_context)
         print '===PRINTING PLAIN EMAIL==='
         template = plain_template.render(template_context)
```
koonsies: This should be renamed to rendered_plain_template, or at least something that better represents what it is.

notifications/rules.py
```diff
@@ -99,15 +99,19 @@ def send_email(self, template_context, html_template, plain_template):
...
         subject = self.get_subject()
-        email_domain = 'my.jobs'  # Need to figure out what else would that be(from email)
-        recipients = self.task.freeform_email
+        email_domain = 'notifications@my.jobs'  # Need to figure out what else would that be(from email)
+        recipients = self.task.freeform_email.split(',')
         if template is not None:
-            body = template_context.body
-            send_mail(subject, body, email_domain, recipients)
+            email_body = template.body
```
koonsies: Same issue as below. Not template.body, but rather rendered_plain_template, or w/e you named it above.

notifications/rules.py
```diff
@@ -98,16 +98,12 @@ def send_email(self, template_context, html_template, plain_template):
...
+        html_rendered_template = html_template.render(template_context)
+        rendered_text_template = plain_template.render(template_context)
         subject = self.get_subject()
-        email_domain = 'my.jobs'  # Need to figure out what else would that be(from email)
-        recipients = self.task.freeform_email
-        if template is not None:
-            body = template_context.body
-            send_mail(subject, body, email_domain, recipients)
+        email_domain = 'notifications@my.jobs'  # Need to figure out what else would that be(from email)
+        recipients = self.task.freeform_email.split(',')
```
koonsies: This also needs to send emails to any of the users in tasks.users field that have emails.

notifications/rules.py
```diff
@@ -109,16 +109,19 @@ def send_email(self, template_context, html_template, plain_template):
...
+        users_email = self.task.users.all()
+        recipients += users_email
```
koonsies: I don't believe this will work. At this point, users_email is a queryset of User objects. At very least, this would need to be `recipients += [user.email for user in users_email]`

PR #3708, opened 2017-12-04
added transform hover effect

strempe:
## What's this PR do?
Added transform hover effect showing Rebecca she can 'click' a tab or link to take her back to the landing page. 

screen shot 2017-12-04 at 4 25 34 pm

PRM - file change in _navigationBar.scss

Select a tab then hover over that tab to see it grow! 
### Front-End

- [x] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


gulp/src/sass/_base/components/_navigationBar.scss
```diff
@@ -137,6 +137,10 @@ $link-color: #5a6d81;
                     &.active {
                         background: white;
                         color: black;
+                        &:hover {
+                            transform: scale(1.1);
+                            font-weight: 600px;
```
aljordan82: `px` is an invalid property for font-weight.

strempe:
@aljordan82 fixed

aljordan82:
Looks good!

PR #3709, opened 2017-12-05
Fix eslint rule: no-useless-return

aljordan82:
## What's this PR do?
Fix eslint rule: no-useless-return


gulp/.eslintrc.json
```diff
@@ -9,19 +9,14 @@
...
     "object-shorthand": 0,
     "comma-dangle": [1, "always-multiline"],
     "function-paren-newline": 0,
-    //  "block-spacing": 0,
-    //  "array-bracket-spacing": 0,
     "no-continue": 0, // this can stay off
-    //  "no-confusing-arrow": 0,
     "no-multi-assign": 0,
     "no-underscore-dangle": 0,
-    "no-useless-return": 0,
```
darrint: Why are we removing the rule if we're fixing it?

aljordan82: Rules are enabled by default.

darrint:
I tried to test the changed code but was blocked by this bug: https://directemployersfoundation.atlassian.net/browse/PD-4023

darrint:
Hm. I think I pressed the merge button by mistake. I'm going to leave it since this is such a low-risk PR. The above bug is noted though.

PR #3717, opened 2017-12-06
Monitoring error improvements.

almathew:
## What's this PR do?
1. Moves `log_dict` to monitoring where it belongs.
2. Sets the redis db, now that we're using multiple.
3. Logs encountered exceptions to redis (so they later can be consumed).
4. Changes 404 status codes for Datadog purposes (and only Datadog purposes) to 274, since they consider 404 an error and that's annoying.

Everything is in monitoring.py.
All the other stuff is just moving stuff around and adding settings.

Print statements. 
You can print what's in redis by using `print queue.get()` right after the `queue.put()` .
You can test the error code thing by printing `span.meta` after the response code is set.
You can try pushing to Datadog if you really want to, but that's kind of a pain.

### Separation of Concerns


seo/utils/search.py
```diff
@@ -73,9 +73,11 @@ def get_job_data_from_redis(guids):
     :param guids: set of guids to retrieve job data for
     :return: a dict with guid keys and job data values, plus a set of guids with no job data found
     """
+    if not guids:
+        dict(), set()
```
vigowei: I think you meant to return

almathew: I did, thanks. It's fixed.
Funny enough, I was only adding that because of the Datadog error on the mget...

PR #3719, opened 2017-12-07
[PD-4016][PD-4001][PD-3997] Priority, MyJobs Message, Dynamic Choices for Rules

koonsies:
## What's this PR do?
Adds a priority field to the notification task model
Adds dynamic choices and validation on rule_name to ensure a rule_name cannot be entered that does not exist.
Adds handling for MyJobs message sending
Squashes migrations into a single migration.. since we had a ton.

Shell.
Try creating various notification tasks to see if they let you use weird rule_names
Also see that priority accepts a small int value (0 to 32767)
Also set the send_myjobs_message to True and see if it creates a message in MyJobs.models.Message
### Front-End
N/A
- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


notifications/migrations/0003_auto_20171130_1609.py
```diff
@@ -1,20 +0,0 @@
...
-
-
-class Migration(migrations.Migration):
-
-    dependencies = [
-        ('notifications', '0002_notificationtask_subject'),
-    ]
-
-    operations = [
-        migrations.RenameField(
```
tdphillips: Did we decide to do away with this rename? It's not displayed in the new 0001 migration.

notifications/models.py
```diff
@@ -1,13 +1,33 @@
...
+    Check to verify that a class returned from getmembers actually belongs to the notification.rules module, and is
+    a class
+
+    :param member: member returned by inspect.getmembers
+    :return: true/false
+    """
+    return (inspect.isclass(member)
+            and member.__module__ == notifications.rules.__name__)
+
+# attempt to get all classes defined in the rules.py file, to prevent invalid rules from being saved to tasks.
```
tdphillips: Would it be easier to implement something mimicking JS exports?

`rules.py`:
```
class RuleA
class RuleB
class RuleC
class NotARule
def foo()
def bar()
rules = [RuleA, RuleB, RuleC]
```

notifications/rules.py
```diff
@@ -52,6 +56,20 @@ def _get_template_by_type(self, type):
...
             return get_template(template_format % ('default', type))
 
+    def get_expiration_for_myjobs_message(self):
+        """
+        Get the date which the MyJobs mail will stop displaying
+        :return: date - MyJobs mail expiration date
+        """
+        return datetime.now() + timedelta(days=30)
+
+    def get_users_from_task(self):
```
tdphillips: Is this supposed to save on characters or something? `self.get_users_from_task()` is longer than `self.task.users.all()`

notifications/rules.py
```diff
@@ -52,6 +56,20 @@ def _get_template_by_type(self, type):
...
+        """
+        Retrieve the users querydict from the task
```
tdphillips: This method deals with querysets.

notifications/migrations/0003_auto_20171130_1609.py
koonsies: Wow, sharp eye. So, I can definitely be accused of taking some liberties with the naming. I've waffled back and forth on what this should be called. I changed it to mail, thinking that was a good representation of what it is, then reconsidered since the project it's referencing is called "mymessages" and the model is "Message". I'm fine with either naming scheme, I just settled on the latter.

notifications/models.py
koonsies: Ah, so just have a global variable inside of rules.py that stores all of the classes, and can be imported itself? I hadn't thought of that. I like that it prevents any additionally declared items in rules.py from being considered. Good call!

notifications/rules.py
koonsies: I may have gone a bit overboard on stubbing out functions. As I was writing this, it crossed my mind that someone subclassing this object may want to include/remove certain users based on the rule type. On second thought, it's probably not very likely so I'm cool with removing it.

PR #3721, opened 2017-12-07
Switch from redis to kinesis so we can use lambda.

almathew:
## What's this PR do?
Switches from using redis for the error queues to Kinesis.

Add AWS credentials.
Cause errors.
Confirm that errors are getting added to the Kinesis stream.
Check out Kibana too. They should be there.


almathew:
I don't think anyone is itching to merge this, and I want to see some stuff get pushed to kibana...

PR #3723, opened 2017-12-08
[PD-4010] Validate freeform fields of notifications task

tejaswinigutti31:
## What's this PR do?
Validate freeform fields(email/sms) of notifications task for a valid pattern
Check the validation methods in notifications/models.py
A couple of unit tests are written to test the functions.
### Separation of Concerns

### Tests


koonsies:
It'll probably be worthwhile to make sure these validate functions are called in the task's save and/or clean functions.

notifications/tests/test_models.py
```diff
@@ -0,0 +1,20 @@
...
+class NotificationTaskModelTest(TestCase):
+    def setUp(self):
+        models.NotificationTask.objects.all().delete()
+
+    def test_validate_freeform_email(self):
+        """
+        Test for invalid email addresses in a given freeform form email address
+        """
+        with self.assertRaises(ValidationError):
+            models.NotificationTask.objects.create(freeform_email='abc@xyz.com, abc.jobs, abc123@xyz.com')
```
koonsies: We should probably also test the positive case. Enter an example that does not raise a validation error.

tejaswinigutti31: Done!

koonsies:
Merge conflicts from the other PR :(

PR #3724, opened 2017-12-11
[PD-4012]  new analytics dimensions and UI

vigowei:
## What does this PR do?
Offers a new way to look at analytics data. 

Go to New Analytics and play with the dimension groups, different column combos and make sure they all update the table correctly (selecting a new intention from the dropdown should change data columns, checking a new column on the left should add it to the table etc); make sure export still works. Note: timeline and bar charts are temporarily disabled in this PR. Talk to @aljordan82 if you have any concerns or questions about those.

New "intentions" are added to a special report intention called `analytics-pa` (pa = pick any).
The new intentions are based on data columns. When you look at the code, these intentions are actually defined as categories on the back end, I have a plan to refactor once we are ready to get rid of the old intentions. Doing this for now lowers the risk of breaking current reports.
Each new intention has multiple themed "dimension groups". A dimension group  is a group of multiple columns (for example, a location themed dimension group will have Country, State, City, ZIP). These definitions can be found in `pick_any_dimensions.py` they probably need to be revised as we move along in development for this feature.

### Front-End

- [x] Works on IE 11 and Edge
- [x] Works on Mobile

### Separation of Concerns

### Tests

### Migrations
N/A


gulp/src/analytics/reducers/__tests__/header-reducer.test.js
```diff
@@ -42,27 +42,27 @@ describe('headerReducer', () => {
...
-      const action = setSelectedRange(loadedRange);
-      const results = headerReducer(initialRange, action);
-      expect(results).toEqual({
-        startDate: '2016-05-04',
-        endDate: '2017-05-04',
-        stateCustomRange: '05/04/2016 - 05/04/2017',
-      });
-    });
-  });
+  // describe('setSelectedRange', () => {
```
darrint: Delete?

aljordan82: I would rather to keep it commented out for now.

darrint:
I'm deliberately not holding the PR or making tickets for 2 issues I saw:

 * On firefox I'm getting an error toast and an exception that appears to be history API related.
 * There could be unit tests on some new JS action creators but I'm not worrying about it at the moment.

PR #3730, opened 2017-12-12
[Ignore Me!] This PR doesn't exist.

koonsies:
## What's this PR do?
Shhhhhh.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


tdphillips:
There's a discussion about this in Slack from Dec 8. The get_template method appears to be forcing the caching of context processors.

koonsies:
Interesting. I missed the resolution of that conversation. Thanks! Talked to Ashley and trying out her idea.

PR #3733, opened 2017-12-12
[PD-4023] remove references that caused an error in search

koonsies:
## What's this PR do?
Removes references to outreachEmail, which we removed from the NUO records

The issue that brought this to attention was that the search stopped working in NUO.
Ensure that works.

The rest of the changes only involved automated tests, but clicking through the app never hurts.

### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


aljordan82:
This works as expected. I'll merge it today.

PR #3736, opened 2017-12-14
Create AnalyticsDateRangePicker

aljordan82:
## What's this PR do?
The previous version was getting rerendered needlessly
Create a connected AnalyticsDateRangePicker component. This component
is re rendered ONLY if the startDate or the endDate properties were
changed.

This part of an ongoing performance enhancements to Analytics App.


vigowei:
bug or feature? date range doesn't show before range is picked

PR #3737, opened 2017-12-14
[PD-4028] add activity for new analytics pick-any

vigowei:
## What's this PR do?
A very small change to add an app and activity to hide the new report from users until we are ready to show them.

you will notice that New Analytics (Pick Any) is no longer visible. 
Add the activity to your local company to see it again.

```
from myjobs.models import AppAccess, update_activities
from seo.models import Company

update_activities()
company = Company.objects.get(name="your_company_name")
company.app_access.add(AppAccess.objects.get(name="New Analytics"))
```

### Separation of Concerns


vigowei:
delaying, since we are replacing the old interface

PR #3740, opened 2017-12-18
[PD-4027] usage tracking

darrint:
## What's this PR do? (and why)

Integrates the analytics tracker into myjobs and uses it for tracking usage of analytics itself. Goals are sent whenever a user arrives at a new intention and when a user exports a report.

In development, goal achievement is logged to the console. In production goal achievement is logged to analytics.

Why?

We are trying a grand experiment.

We're finding that we have chronically poor engagement with analytics both from sales and customers who both would benefit from it. We've done a bunch of research and identified some ways analytics does not meet needs and are working on those.

However, we are also receiving a lot of conflicting feedback from various parties. Our own management wants to expose lots of power to users in a generic way which tends toward a few abstract controls. Our sales wants the user's able to obviously see how to get answers to their questions which tends towards lots of concrete controls. The various sales teams and individuals have differing ideas of what the users want to see.

To solve this problem, this PR makes some useful but primitive measurements available in our own analytics. These measurements are how we will measure engagement at least until we consider the engagement problem solved. While we will pay attention to subjective feedback from our own staff it won't be our primary focus. We will continue to observe users using ours and others' analytics products and question our staff on _why_ users need to know certain things. We will base our designs on this information.

Our team and our management will be eating our own dog food in our analytics.

As part of this experiment @haliscooper and @fezick  are giving us less low-level direction on analytics UI/UX design. Instead they are going to join us in better understanding the users and watching these measurements in our own analytics to see how some major changes impact engagement.

There's a new context processor which sets up a variable which templates can use to control the JS which does the analytics. If adding yet another context processor is not ok or you have a better idea let me know here.

The JS code for development and production handling of tracking events is in `track-usage.js`. I also built out action creators. Analytics' `main.jsx` shows an example of how to configure Redux so the action creators do the right thing in development vs. production.

Normally we shoot for not having anything logged to the console. I'm breaking that rule here. In production the console is still silent. In development we log every time we achieve a goal. I think this is a legit use of the console. For the analytics team, knowing that we are measuring these goals correctly is as important as the app behaving correctly. We need to always have the goal information easily available so I chose console logging. If this seems wrong to you let me know here.

Actually testing the JS for sending an event involves monitoring new events in the production database. The fact that your events will go to the database as `secure.my.jobs` is probably ok for now, since we aren't actually measuring yet. Once we go to production this won't be ok anymore. In the future if you want to test the event sending mechanism you'll need to configure your revproxy and settings to use non-production hostnames.

If you don't want to go to any of that trouble it is probably ok to just test that the console logging is working in various browsers. We can test further in QC which has appropriate hostnames. If we choose to defer the browser testing into QC we'll need to test all the supported browsers in QC to be sure this is working.

I can help with monitoring the production DB for new goal events if someone wants to browser test locally.

See what and why above.

### Front-End

- [x] Works on IE 11 and Edge
- [x] Works on Mobile
- [ ] ~~Unit tests added or ticket exists to add tests~~

### Separation of Concerns


gulp/src/analytics/components/Header/ExportOptions.jsx
```diff
@@ -35,6 +36,8 @@ class ExportOptions extends Component {
       return;
     }
 
+    dispatch(doRecordUsageExportGoal());
```
aljordan82: can this be moved to the action creator in header-actions.js?

gulp/src/analytics/track-usage.js
```diff
@@ -0,0 +1,117 @@
...
+  handleLoaded() {
+    for (const dispatch of this.deferredDispatches) {
+      dispatch();
+    }
+    this.deferredDispatches = [];
+    this.loaded = true;
+  }
+
+  dispatchClick(element) {
+    const clickEvent = new MouseEvent('click', {
```
aljordan82: Do we need to track on mobile?

templates/base-react.html
```diff
@@ -53,6 +53,12 @@
             {% endcompress %}
         {% endblock %}
             
+        <script>
```
aljordan82: Will this track actions on local dev environment?

aljordan82:
Do we need special settings for EU countries?

darrint:
No, we do not need special setting for EU countries here. The users are logged into our system and we very much positively identify them and keep logs of actions.

gulp/src/analytics/track-usage.js
```diff
@@ -0,0 +1,117 @@
...
+ * Make them invisible.
+ */
+export function buildTrackerDom(companyName, rootElement, onTrackerLoad) {
+  window.de_track_custom = {};
+  window.de_track_custom.g = [
+    { o: formatGoalName(companyName, 'analytics_intention'), r: ANALYTICS_INTENTION_GOAL_ID },
+    { o: formatGoalName(companyName, 'export'), r: ANALYTICS_EXPORT_ID },
+  ];
+  const div = document.createElement('div');
+  div.style = 'display: none';
```
aljordan82: This is throwing an error in IE11. 
this would work `div.style.display = 'none';`

aljordan82:
IE fix works!

darrint:
Unit tests weren't relevant.

PR #3744, opened 2017-12-20
[PD-3466] Update dev containers

darrint:
## What's this PR do?

Makes a couple of changes to the developer environment.

* Update dev to xenial.
* Update yarn and fix the version number so it's controlled.
* Tell memsql it's ok to not have 4GB of memory.
* Remove the old shrinkwrap file.

It's pretty short.

```
$ dk rebuilddev
$ dk appstop
$ # Start and test app
```

Make sure things function at all. I'd like to hear from a Mac user who bumps their docker memory down to the default instead of 6+ GB. Does analytics still function?

na.

### Some testing specific to this PR:

- [ ] The app works on a Mac.
- [ ] Analytics works on a Mac when docker memory is constrained below 4GB. Ideally this works at whatever the Docker for Mac default is.

### Separation of Concerns


almathew:
Tested and working on my mac.
Memsql acted up a little bit at first, but that's not really unusual.

PR #3748, opened 2017-12-20
Pd 4014 analytics filters ui

aljordan82:
## What's this PR do?
Major update to Analytics
screen shot 2017-12-20 at 1 10 38 pm
Test all report intentions
### Front-End

- [ ] Works on IE 11 and Edge
- [ ] Works on Mobile

### Separation of Concerns

### Tests

### Migrations


analytics/api.py
```diff
@@ -110,16 +168,33 @@ def dynamic_chart(request):
     scope = driver.resolve_scope(grant, intention.scope_data)
     time_frame = (date_start, date_end)
 
+    run_kwargs = {}
+    if query_data.get('page_query'):
+        get_pagination_params(query_data.get('page_query'), run_kwargs)
```
darrint: Consider moving `get_pagination_params` and `get_order_params` into the serializer itself in the future. That way `query_data.get('order_by')` will return exactly what's needed.

